PR #32, opened 2016-10-23
Feature/use json

MizukiSonoko:
### Remove 
- message pack
- yaml

### Future
- use json only

### Reason
- Message-Pack was too hard to develop. 🌵 
- I dislike yaml, sorry...


MizukiSonoko:
I mistake .

PR #160, opened 2017-02-15
Port numbers for grpc and http are configurable now.

luckychess:


core/infra/server/http_server_with_cappuccino.cpp
```diff
@@ -74,7 +75,26 @@ namespace http {
...
+
+        std::string param2 = "-p";
+        char *param2_c = new char[param2.length() + 1];
+        strcpy(param2_c, param2.c_str());
+
+        std::string port = std::to_string(config::IrohaConfigManager::getInstance().getHttpPortNumber(1204));
+        char *port_c = new char[port.length() + 1];
+        strcpy(port_c, port.c_str());
+
+        char *params[] = {param1_c, param2_c, port_c};
```
Warchant: how about 
```C++
std::string port = std::to_string(config::IrohaConfigManager::getInstance().getHttpPortNumber(1204));
char *params[] = {param1_c, param2_c, port.c_str()}; // ?
```
or http://en.cppreference.com/w/cpp/language/const_cast

Warchant:
@MizukiSonoko [please, review this line](https://github.com/hyperledger/iroha/blob/2d1b18854e70cde18960ddf25fa0182691718b7e/core/infra/server/http_server_with_cappuccino.cpp#L93)

That function has the following declaration:
```C++
void Cappuccino(int argc, char *argv[])
```

In the code inside, if you don't change `argv`, make it `const`, please.
So @luckychess will be able to write:

```C++
std::string port = std::to_string(config::IrohaConfigManager::getInstance().getHttpPortNumber(1204));
char *params[] = {"", "-p", port.c_str()};
```

instead of [this :-1: ](https://github.com/hyperledger/iroha/blob/2d1b18854e70cde18960ddf25fa0182691718b7e/core/infra/server/http_server_with_cappuccino.cpp#L79-L97)

core/infra/server/http_server_with_cappuccino.cpp
luckychess: As far as I know this way is not safe in the case if a constless string be modified.

core/infra/server/http_server_with_cappuccino.cpp
```diff
@@ -76,25 +76,12 @@ namespace http {
...
-        char *params[] = {param1_c, param2_c, port_c};
-        Cappuccino::Cappuccino( 3, params);
-
-        delete[] param1_c;
-        delete[] param2_c;
-        delete[] port_c;
+        std::vector<std::string> params = {"", "-p", std::to_string(config::IrohaConfigManager::getInstance().getHttpPortNumber(1204))};
+        std::vector<char*> argv;
+        for (const auto& arg : params)
+            argv.push_back((char*)arg.data());
```
luckychess: Exactly the same problem as with the const_cast with only difference you do it in a C-style way.

PR #203, opened 2017-03-07
Bugfixes and improvements.

Warchant:


Warchant:
For some reason both cappuccino (asio) and grpc are binded onto ipv6 socket:
```
root@e80412b5ad30:~# netstat -ntulp
(No info could be read for "-p": geteuid()=0 but you should be root.)
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp6       0      0 :::50051                :::*                    LISTEN      -               
tcp6       0      0 ::1:1204                :::*                    LISTEN      -
```

And any attempts to connect fail. This is why test on circle-ci does not work.

PR #222, opened 2017-03-14
Add my proposal

MizukiSonoko:
- Command should be enum. and separate command and object. I referred to @emilmelnikov 
- Domain should have ledgerId, Asset should not it.
- Add QueryFilter and Query has it.
- Add CommandOption for some command like Transfer

### DIFF
--- a/flatbuf/api.fbs
+++ b/flatbuf/api.fbs
@@ -6,7 +6,7 @@ file_identifier "IRH";
 file_extension "iroha";
 
 union Object { Currency, KeyValueListAsset, Domain, Account, Peer, Message }
-union Command { Add, Transfer, Update, Remove, Batch, Contract, ChaincodeBatch }
+enum Command : byte { Add = 1, Transfer, Update, Remove, Batch, Contract, ChaincodeBatch }
 
 table Permission {
     readable: bool(required); // queryable
@@ -50,7 +50,6 @@ table KeyValueListAsset {
   name: string;
   description: string;
   tag: string;
-  ledgerId: string;
   keyValueList: [Api.KeyValueObject];
   permissions: Api.Permission;
   isEditable: bool;
@@ -59,6 +58,7 @@ table KeyValueListAsset {
 table Domain {
   name: string(required);
   permissions: [Permission];
+  ledgerId: string;
 }
 
 table Account {
@@ -83,9 +83,14 @@ table Signature {
   signature:string;
 }
 
+table QueryFilter {
+  value: string;
+  condition: string;
+}
 table Query {
   type: string;
   senderPubkey: string;
+  filter: [QueryFilter]
 }
 
 table ConsensusEvent {
@@ -95,38 +100,19 @@ table ConsensusEvent {
   status:string;
 }
 
-// Commands
-
-table Add {
-    object: Object(required);
-}
-
-table Transfer {
-    object: Object(required);
-    receiver: string(required);
-}
-
-table Update {
-    object: Object(required);
-}
-
-table Remove {
-    object: Object(required);
-}
-
 table Transaction {
-  command: Command(required);
+  creatorPubKey: string(required);
+  command: string(required);
+  object:  Object;
 
   signatures: [Api.Signature];
   tag: string;
-  creatorPubKey: string;
   hash: string;
   timestamp: ulong;
   message: string;
 }
 
 // Communication
-
 table TransactionResponse {
   message: string;
   code: ushort;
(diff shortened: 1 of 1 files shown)


flatbuf/api.fbs
```diff
@@ -83,9 +83,14 @@ table Signature {
   signature:string;
 }
 
+table QueryFilter {
```
l4l: Could you explain what does it intend to be/do?

flatbuf/api.fbs
```diff
@@ -83,9 +83,14 @@ table Signature {
...
+  value: string;
+  condition: string;
+}
```
l4l: add new line between tables plz

flatbuf/api.fbs
```diff
@@ -95,38 +100,23 @@ table ConsensusEvent {
...
-}
-
-table Remove {
-    object: Object(required);
-}
-
 table Transaction {
-  command: Command(required);
+  creatorPubKey: string(required);
+  command: string(required);
```
l4l: Why this is string?

flatbuf/api.fbs
```diff
@@ -95,38 +100,23 @@ table ConsensusEvent {
...
-// Commands
-
-table Add {
-    object: Object(required);
-}
-
-table Transfer {
-    object: Object(required);
-    receiver: string(required);
+table CommandOption {
```
l4l: could you explain this as well?

emilmelnikov:
Unfortunately, when I've suggested replacing commands with enums, I didn't think of command parameters (e.g. the _transfer_ command should have the _receiver_ parameter).
The solution with the `CommandOption` table doesn't look nice and complicates the ledger storage.

Maybe revert back to command tables/structs, but define them separately for each object type?

Example API sketch:

```
table Transaction {
  object: Object;
}

union Object { Currency, Peer }

table Currency {
  value: ulong;
  command: CurrencyCommand;
}

union CurrencyCommand { CurrencyAdd, CurrencyRemove, CurrencyTransfer }

struct CurrencyTransfer {
  receiver: string;
}
```

The downside of this approach is a lot of object-command tables.

emilmelnikov:
The better option is to remove command/object nesting altogether:

```
table Transaction {
  command: Command;
}

union Command { CurrencyAdd, CurrencyRemove, CurrencyTransfer, PeerAdd, PeerRemove };

table CurrencyAdd {
  currency: Currency;
}

table CurrencyRemove {
  currency: Currency;
  // The alternative way is to list the "primary key" of the object:
  // name: string;
  // domain: Domain;
  // ledgerID: string;
}

table CurrencyTransfer {
  currency: Currency;
  // The "alternative way" also applies here.
  receiver: string;
}
```

In code, it will be handled with a one giant switch command, but one can hide it in some kind of factory:

```
class Command {
  virtual ~Command() {};
  // Common command interface...
}

class CurrencyAddCommand : Command {
  virtual ~CurrencyAddCommand() override {};
  // Implementation of the command interface...
}

class CurrencyRemoveCommand : Command {
  virtual ~CurrencyRemoveCommand() override {};
  // Implementation of the command interface...
}

class TransactionFactory {
  Command getCommand(const uint8_t* buffer);
}

Command TransactionFactory::getCommand(const uint8_t* buffer) {
  uint8_t* buf;
  auto transaction = GetTransaction(buffer);
  switch (transaction->command_type()) {
    case (Command_CurrencyAdd):
      // Handle CurrencyAdd
      return CurrencyAddCommand(params);
    case (Command_CurrencyRemove):
      // Handle CurrencyRemove
      return CurrencyRemoveCommand(params);
  }
}
```

MizukiSonoko:
Ok thanks, I understanded it
I think code increase on large scale...

flatbuf/api.fbs
MizukiSonoko: Ok thanks

flatbuf/api.fbs
MizukiSonoko: QueryFilter is used by query.
For example
```
table QueryFilter {
  value: string = "my=asset-name"
  condition = "equal"
}
```
```
table QueryFilter {
  value: string = "value"
  condition = "not-equal"
}
```

Warchant:
My proposal:
https://gist.github.com/Warchant/b7049e2f0653bd30c78f06b283c30bf0

Sorry, there are lots of files, probably some can be combined.
Also, don't worry about lots of includes:
> include automatically ensures each file is parsed just once, even when referred to more than once., -- Google

takemiyamakoto:
Meeting tonight at 19:00 JST to discuss the data format some more.

MizukiSonoko:
**Unnessecarry**

PR #224, opened 2017-03-15
Fix cmake file for MacOS. Add target link libraries.

cimadai:
Added following link libraries:
- create_objects_helper
- java_data_structure

Change libjvm from "so" to "dylib".

I tested on MacOS 10.11.6.


MizukiSonoko:
Ok thanks!

PR #235, opened 2017-03-17
Migrating to Circle 2.0

Warchant:
### Introducing `tiny` 

tiny - is a minimal docker image, which includes **only** iroha binaries compiled in Release mode, config files and libraries.
It can be built by calling `docker/build.sh` in one minute.

As a consequence, we do not create isolated environment to build iroha image (iroha-dev), so *to build tiny you should be able to build iroha in your OS*.

But this environment was saved as `warchantua/iroha-dev` and used in circle 2.0 to speed-up the build process.

### Config-discovery was removed

It means, that you **can not** create a network with docker-compose (it was also removed). 

BUT, with use of peer service, I can write new docker-compose file, which will be able to:
 - create arbitrary number of nodes connected in a network
 - deploy this stack on a docker swarm in one command

(docker dev team has released docker-compose v3.0 and this is possible now!)

### Circle 2.0

Now circle 2.0 builds and pushes to docker hub successful builds from *any branch*.
To access your image:
```
docker pull hyperledger/iroha-docker:branch_name
```

And then use. 
List of all releases is on docker hub: https://hub.docker.com/r/hyperledger/iroha-docker/tags/


Warchant:
@l4l 
```
1489771498 DEBUG [config] load json is {
  "group": {
    "ip": "172.17.0.3",
    "name": "82ed218c77f7",
    "publicKey": "BKges18NwuIi3OO3b0k+4aV6zfgW1dHkX+HuN9khLnc="
  },
  "me": {
    "ip": "172.17.0.3",
    "name": "82ed218c77f7",
    "privateKey": "EB1xqzIpXPD6z+3sNkbw5N/fOzXyv96zpctTTO0T+WT5b9UnepLcvq0Es/7jOHzBIUKs14LIiofw/XXwqcUsag==",
    "publicKey": "BKges18NwuIi3OO3b0k+4aV6zfgW1dHkX+HuN9khLnc="
  }
}

1489771498 WARNING [peer service with json] type must be 2, but is 1
1489771498 WARNING [peer service config] ParseFromStringException in config/sumeragi.json
1489771498 WARNING [peer service config] config/sumeragi.json is set to be default.
terminate called after throwing an instance of 'std::domain_error'
  what():  cannot use value() with null
Aborted (core dumped)
```

PR #265, opened 2017-04-14
Refactoring

kamilsa:
This is what is done so far:
 
Remove unused include
1. Apply clang-format
2. more clearly value name
3. Correct my wrong English
4. Remove magic numbers and hard coded values.
5. Gather if (getenv("IROHA_HOME") == nullptr) {...} into config. (done by introducing config::get_iroha_home())
6. Mark Deprecated to JVM and smart contracts (review if it is done properly)
7. Replace assertions with exceptions except in test codes (none of that found)

### DIFF
--- a/core/infra/virtual_machine/jvm/java_virtual_machine.cpp
+++ b/core/infra/virtual_machine/jvm/java_virtual_machine.cpp
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include <assert.h>
 #include <algorithm>
 #include <array>
 #include <util/logger.hpp>
@@ -40,7 +39,7 @@ namespace jvm {
     std::unique_ptr<JavaContext> initializeVM(const std::string& packageNameUnderInstances, const std::string& contractName) {
 
         const auto IrohaHome = []() {
-            const auto p = getenv("IROHA_HOME");
+            const auto p = config::get_iroha_home();
             if (p == nullptr) {
                 throw "[FATAL] Set environment $IROHA_HOME";
             }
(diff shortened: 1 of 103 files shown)


core/infra/virtual_machine/jvm/java_virtual_machine.cpp
```diff
@@ -40,7 +39,7 @@ namespace jvm {
     std::unique_ptr<JavaContext> initializeVM(const std::string& packageNameUnderInstances, const std::string& contractName) {
 
         const auto IrohaHome = []() {
-            const auto p = getenv("IROHA_HOME");
+            const auto p = config::get_iroha_home();
             if (p == nullptr) {
                 throw "[FATAL] Set environment $IROHA_HOME";
```
motxx: 1. This shows:
```
terminate called after throwing an instance of 'char const*'
Aborted
```
C++ cannot throw message directory. It needs some exception call. (Sorry, actually this is my mistake ;(

2. I think it is better to include this nullptr check into `config::get_iroha_home()`. If this change is applied, all variables such as `IrohaHome` can be removed and replaced with `config::get_iroha_home()`, and in that case, exception message `"[FATAL] Set environment $IROHA_HOME"` is better.

motxx: 1. This shows:
```
terminate called after throwing an instance of 'char const*'
Aborted
```
`throw` needs an exception class in order to show message correctly. (actually, this is my mistake. sorry ;(

2. I think it is better to include this nullptr check into `config::get_iroha_home()`. If this change is applied, all variables such as `IrohaHome` can be removed and replaced with `config::get_iroha_home()`, and in that case, exception message `"[FATAL] Set environment $IROHA_HOME"` is better.

motxx:
In my quick review, fixes of `6. Mark Deprecated to JVM and smart contracts` looks good to me.

luckychess:
Looks good for me.

PR #343, opened 2017-05-16
Feature/quick build

tkyonezu:
This is Makefile version for Iroha quick building. 
You can build Iroha simply with typing 'make'. And you can run Iroha with 'make run' to running single Iroha container. You can view logs by 'make logs' and so on...


motxx:
The github account is valid CLA?

PR #367, opened 2017-06-17
Feature/add crypto tests

motxx:


libs/crypto/keypair.hpp
```diff
@@ -118,6 +117,28 @@ class Keypair {
   explicit Keypair(const pubkey_t &pub)
       : pubkey(std::move(pub)), has_private(false) {}
 
+  /**
+   * Build a keypair with public and private key in base64 encoded format (string)
+   * @param pub
+   * @param priv
+   */
+  struct tag_base64_encoded {};
+  explicit Keypair(const std::string &pub_base64, const std::string &priv_base64, tag_base64_encoded)
```
motxx: Added Keypair generator from base64 encoded. (use tag dispatch)

libs/crypto/keypair.hpp
```diff
@@ -202,32 +234,30 @@ class Keypair {
                        : nonstd::nullopt;
   }
 
+  static Keypair generate_keypair() {
```
motxx: Add `generated_keypair()` as factory method.

PR #384, opened 2017-06-22
Ametsuchi api

grimadas:


MizukiSonoko:
Why we use `rx c++` instead of `observable`?

irohad/ametsuchi/storage.hpp
```diff
@@ -1,81 +0,0 @@
...
-  namespace ametsuchi {
-    class Storage {
-     public:
-      // factory
-      static std::unique_ptr<Storage> create();
-
-      // block_store
-      virtual void insert_block(uint64_t block_id,
-                                const std::vector<uint8_t> &blob) = 0;
-      virtual void erase_block(uint64_t block_id) = 0;
```
MizukiSonoko: When it used? erase_block, rollback?

grimadas: This code is already erased ;)

MizukiSonoko: Oh, sorry

irohad/ametsuchi/query_api.hpp
```diff
@@ -0,0 +1,106 @@
...
+       */
+      virtual std::vector<iroha::dao::Asset> get_domain_assets(
+          std::string domain_full_name) = 0;
+
+      /**
+       * Get all transactions of an account.
+       * @param pub_key - account's first public key
+       * @return observable of DAO Transaction
+       */
+      virtual rxcpp::observable<iroha::dao::Transaction>
```
motxx: Can Iroha filter transactions with arbitrary conditions by using this api?

irohad/ametsuchi/ametsuchi.hpp
```diff
@@ -0,0 +1,51 @@
...
+     * Storage class, which allows queries on current committed state, and
+     * creation of state which can be mutated with blocks and transactions
+     */
+    class Ametsuchi : public QueryApi {
+     public:
+      /**
+       * Creates a mutable state from the current state
+       * @return Created mutable state
+       */
+      virtual std::unique_ptr<MutableState> createMutableState() = 0;
```
motxx: Is it used before committing the block?

irohad/ametsuchi/query_api.hpp
grimadas: I think we will introduce filters a little bit later. Right now you will get for example all transactions of an asset, of a wallet or an account.

irohad/ametsuchi/ametsuchi.hpp
grimadas: Yes. This will be used as the only way to commit the block to Ametsuchi

lebdron: It is definitely used for stateful validation - this mutable state will be modified, validated, and discarded.
It's not exactly clear how should `MutableState` interact with `Ametsuchi` to write the block. Maybe it should write the block directly during `apply(dao::Block block)`.

irohad/ametsuchi/command_executor.hpp
```diff
@@ -0,0 +1,43 @@
...
+     * Applies command to the world state view
+     */
+    class CommandExecutor {
+      /**
+       * Executes a command in a temporary state
+       * @see TemporaryWsv, MutableStorage
+       * @param command Command to execute
+       * @return True if the command is successfully executed, false otherwise
+       */
+      virtual bool execute(dao::Command command) = 0;
```
motxx: Does Ametsuchi have executor?
I heard Ametsuchi is only for database and doesn't have logic and stateful validator.

irohad/ametsuchi/temporary_wsv.hpp
```diff
@@ -0,0 +1,61 @@
...
+       * Applies a transaction to current state
+       * using logic specified in function
+       * @param transaction Transaction to be applied
+       * @param function Function that specifies the logic used to apply the
+       * transaction
+       * Function parameters:
+       *  - Transaction @see transaction
+       *  - CommandExecutor
+       *  - WSVQuery
+       * Function returns true if the transaction is successfully applied, false
```
motxx: Is stateful validation inside Ametsuchi?
What does "successfully applied" mean? Which does it mean a success of stateful validation or not broken feature of DB? Both?

irohad/ametsuchi/mutable_storage.hpp
```diff
@@ -0,0 +1,54 @@
...
+       *  - CommandExecutor
+       *  - WsvQuery
+       * Function returns true if the block is successfully applied, false
+       * otherwise.
+       * @return True if block was successfully applied, false otherwise.
+       *
+       */
+      virtual bool apply(
+          dao::Block block,
+          std::function<bool(dao::Block, CommandExecutor, WsvQuery)>
```
motxx: What feature of this `std::function`?
I think this function does:
1. stateful validation. If validation is failed, returns false.
2. executor. Check the content of tx and apply new data to WSV.

irohad/ametsuchi/mutable_storage.hpp
```diff
@@ -0,0 +1,54 @@
...
+       * @param block Block to be applied
+       * @param function Function that specifies the logic used to apply the
+       * block
+       * Function parameters:
+       *  - Block @see block
+       *  - CommandExecutor
+       *  - WsvQuery
+       * Function returns true if the block is successfully applied, false
+       * otherwise.
+       * @return True if block was successfully applied, false otherwise.
```
motxx: Does "successfully applied" mean stateful validation is valid and DB has changed?

schema/commands.proto
```diff
@@ -96,153 +108,186 @@ message Domain {
...
   }
 
-  // transaction "creator" invites account with account_uuid to the domain
+  // transaction "creator" invites account with account_id to the domain
   message InviteAccount {
-    string account_uuid = 1; // account uuid
-    string domain       = 2; // invite to this domain
-    uint32 permissions  = 3; // logical or of Permissions enum
-    string username     = 4; // account will join with this username: <username>@soramitsu.co.jp
+    string account_id = 1; // invitee
```
motxx: Why does name have changed? For redundant?

irohad/ametsuchi/mutable_storage.hpp
grimadas: Block is not checked for stateful validation. Successfully applied means that it was checked that hash, merkle root in block is consistent with Ametsuchi. And the block is successfully written to block store and wsv (No database errors).

irohad/ametsuchi/mutable_storage.hpp
grimadas: In this case you use apply(Block), this is used when you need to write to Ametsuchi. Std::function in this case will contain logic for checking hashes, merkle root and how to apply
 transactions within this block to WSV. 
1. WSVQuery is used only for read-only get queries, it has no apply.
2. MutableStorage will not be used for statefull validation. Also check TemporaryWsv (it will be used for Transaction StateFull validation). We also need further Factory for Validation (It will show how to validate concrete command, for example for command Transfer - you need to have specific amount of Assets).
3. CommandExecutor knows how to apply command to Wsv.

PR #390, opened 2017-06-24
Feature/crypto solve odr (more strict array type definition)

motxx:
## What is this pull-request?
Promote to use more strict type array definition.
1. type.hpp: type definition which size is strictly defined (mainly for crypto).
2. Microsoft/GSL: it allows to use `array_view`, `span`

### type.hpp
- `blob_t<N>`: blob type that can be convertible to base64 or hex string
- `ed25519::sig_t`, `ed25519::pubkey_t`, `ed25519::privkey_t`
  - type aliases that have strict sizes.
- `ed25519::keypair_t`: pair of `pubkey_t` and `privkey_t`

### Microsoft/GSL
- For `array_view`, `span`
- (Please add comment here)

## P.S.
I propose anyone to edit comments to this PR description.


libs/crypto/keypair.hpp
```diff
@@ -86,9 +86,9 @@ namespace iroha {
       }
 
       /**
-        * Build a keypair with public key in binary format
-        * @param pub
-        */
+       * Build a keypair with public key in binary format
+       * @param pub
+       */
       explicit Keypair(const uint8_t *pub) : has_private(false) {
```
lebdron: Is it OK to construct a pair with only one object? Does not seem like a pair anymore to me

libs/crypto/keypair.hpp
```diff
@@ -29,7 +29,7 @@
 
 namespace iroha {
   namespace crypto {
-  using namespace ed25519;
+    using namespace ed25519;
```
lebdron: > Do not use using-directives (e.g. using namespace foo).

https://google.github.io/styleguide/cppguide.html#Namespaces

libs/crypto/keypair.hpp
motxx: Yes, I think so, too. Maybe another class name is recommended. @Warchant

libs/crypto/keypair.hpp
motxx: Yes. So, can I separate implementation of this class?

libs/crypto/hash.hpp
```diff
@@ -48,39 +48,39 @@ namespace iroha {
       return SHA3_512(output, input, in_size);
     }
 
-    nonstd::optional<std::vector<uint8_t>> sha3_224(uint8_t *input,
+    inline nonstd::optional<std::vector<uint8_t>> sha3_224(uint8_t *input,
```
lebdron: `SHA3_*` methods call `KeccakWidth1600_Sponge` methods with predefined parameters. It returns 1 if there is an error with predefined parameters, according to [this](https://github.com/gvanas/KeccakCodePackage/blob/master/Constructions/KeccakSponge.inc) file. Maybe we should guarantee that only correct `unsigned char *output, const unsigned char *input, size_t inputByteLen` are passed, and always return `vector<uint8_t>` or `array<uint8_t, >` of required length?

motxx: The functions, `nonstd::optional<> sha3_...()`, seem simply wrapping the features of `SHA3_...()`
Ofcourse, guranteeing correct input is always required. Hmm... in this situation, optional may be redundant.

Warchant:
We propose the following solution:
```
types.h:

template <size_t size>
using bytes_t = std::array<uint8_t, size>;

namespace ed25519{
  using pubkey_t = bytes_t <32>;
  using sig_t = bytes_t <64>; 
  using privkey_t = bytes_t <64>;
}

// because to_string is in std::
namespace iroha {
  template <size_t size>
  std::string to_string(const bytes_t<size> & b){
    return std::string{b.begin(), b.end()};
  }
}
```

And then, we write just *funcitons* which receive public key, private key, whatever.

Warchant:
I added `types.h` but unfortunately I  can't check if it compiles or not, I have some problems with `libcurl` :(

libs/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_subdirectory(crypto)
 add_subdirectory(datetime)
 add_subdirectory(timer)
-add_subdirectory(dao)
\ No newline at end of file
+add_subdirectory(dao)
+add_subdirectory(logger)
```
motxx: Should logger be in libs/?
The feature of logger will have not only current std::cout logging but also telling clients the logs. It seems so fat that logger is in libs/.

motxx:
What is the purpose of BOOST SML lib?

libs/CMakeLists.txt
Warchant: logger seems to me like a library, which can be reused by client / irohad / etc

Warchant:
@motxx I was playing with state machines and forgot to remove sml from libs, sorry. it is not necessary (at least now)

MizukiSonoko:
Maybe not important, This is memo
 I confused Microsoft/GSL and GNU/GSL...  😭 
https://www.gnu.org/software/gsl/

libs/common/types.hpp
```diff
@@ -0,0 +1,119 @@
...
+ */
+namespace iroha {
+
+  using byte_t = uint8_t;
+
+  /**
+   * Base type which represents blob of fixed size.
+   */
+  template <size_t size_>
+  class blob_t : public std::array<byte_t, size_> {
```
motxx: Is some comment needed here because `std::array` doesn't have destructor?

libs/common/byteutils.hpp
```diff
@@ -7,12 +7,41 @@
...
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+//
+// Created by bogdan on 24.06.17.
+//
```
Warchant: Sure.

libs/common/types.hpp
Warchant: Copying is ok at this case, because `blob_t` neither add new members, nor use pointers.
`blob_t` is just `std::array` but with extended interface, so it is safe.

motxx: Ok. It's nice.

motxx:
I don't know whether we should install Microsoft/GSL for now.
`array_view` and `span` seems convenient, but is it needed in current implementation?
It seems to increase libraries which are not used so much.
Actually, I think it is sufficient to use `blob_t`, `pubkey_t`, ... based of `std::array`.

I propose we accept following commit in new PR and close this PR for now:
https://github.com/hyperledger/iroha/pull/390/commits/06d2ddad45b25f10f905d539eeedbc8350328bd0

It seems better to introduce Microsoft/GSL if we really need it.

motxx:
Now, this PR can be merged.

Warchant:
:+1:

PR #391, opened 2017-06-25
Feature/client interfaces

muratovv:
This proposal contains useful interfaces for starting processing user intents(transactions and queries) in the system.

Interface depends on RxCpp.


muratovv:
@MizukiSonoko 
> Torii should invoke stateless-validator
Yes, you are totally correct.
But interfaces it is about usage, but not behaviour. Implementation of transaction processor will be invoke stateful validation. After validation it push in notify() observable response (transaction send to network or stateless validation broken).

>Torii should send valid tx to ordering service.
It is totally correct too. But also, it is about behaviour. After validation transaction processor implementation will be share it to instance of class called "PeerCommunicationService" for invoke method "propagate_transaction".

>so QueryProcessor and TransactionProcessor is not torii's knowledge.
As i know before torii was only grpc, but organization of pipeline processing is very important. Class TransactionProcessor is start point of transaction pipeline., and QueryProcessor is start point of query pipeline, accordingly.

MizukiSonoko:
Ok, I see.
but I think Torii receive tx as can change Iroha's database. so a query is not in Torii
and I propose this architecture/
```
torii/processor.cpp <= prev torii/impl/processor.cpp
pipeline/query_processor.cpp <= prev torii/query_processor.cpp
pipeline/client_processor.cpp <= prev torii/client_processor.cpp
pipeline/transaction_processor.cpp <= prev torii/transaction_processor.cpp
```

Torii's behaviour is in torii/processor.cpp.
pipeline processing component is in pipeline/

muratovv:
@MizukiSonoko 
For me not clear what is responsibility of torii in your case?

MizukiSonoko:
I think responsibility of Torii is 
- receive tx as can change Iroha's database
- if they received valid tx, they reply `ok`
- if they received invalid tx, they reply `reject`

Torii is consist of TransactionProcessor
but TransactionProcessor, QueryProcessor is not Torii's knowledge

muratovv:
@MizukiSonoko 
>receive tx as can change Iroha's database

Torii can't change database. It may say for system process transaction. What should do system not own responsibility.

>if they received valid tx, they reply ok
>if they received invalid tx, they reply reject

It is responsibility of transaction processor.

In my opinion torii (as grpc or client server) _use_ client processor for starting pipeline. How deserialize byte stream in dao it own deal. But organization pipeline and response on Transaction/Query interface

MizukiSonoko:
Torii is an interface.
TransactionProcessor changes Iroha database.

muratovv:
> TransactionProcessor changes Iroha database.

Tx processor hide what should system do. But in fact there is no directly affecting on database. Transaction processor use stateful validation and pass tx to PeerCommuniationService.

PR #396, opened 2017-06-27
Feature/application

lebdron:
## What is this pull request?
Concrete class that contains instances of components which are shared in the system
   
It will be used in main application during system initialization, used by Torii, and other components
  
Instantiate this class in main application, handle the errors and ensure the correctness of component initialization, and pass it to other components

## Details/Features
Application class which contains following instances:
* Ametsuchi storage
* Peer communication service
* Crypto provider
* Hash provider

### DIFF
--- a/irohad/main/application.hpp
+++ b/irohad/main/application.hpp
@@ -22,12 +22,12 @@
 #include <ametsuchi/ametsuchi.hpp>
 #include <dao/dao.hpp>
 
-namespace iroha {
+namespace main {
 
   /**
    * Contains instances of global services and shared providers
    */
-  class Application {
+  class Irohad {
    public:
 
     ametsuchi::Ametsuchi &ametsuchi;
--- a/irohad/main/application.hpp
+++ b/irohad/main/application.hpp
@@ -1,0 +1,43 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
+ * http://soramitsu.co.jp
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef IROHA_APPLICATION_HPP
+#define IROHA_APPLICATION_HPP
+
+#include <network/network_api.h>
+#include <ametsuchi/ametsuchi.hpp>
+#include <dao/dao.hpp>
+
+namespace iroha {
+
+  /**
+   * Contains instances of global services and shared providers
+   */
+  class Application {
+   public:
+
+    ametsuchi::Ametsuchi &ametsuchi;
+
+    network::PeerCommunicationService &peerService;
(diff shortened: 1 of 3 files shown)


motxx:
Will `Application` be expected to consist of other all modules?

irohad/main/application.hpp
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#ifndef IROHA_APPLICATION_HPP
+#define IROHA_APPLICATION_HPP
+
+#include <network/network_api.h>
+#include <ametsuchi/ametsuchi.hpp>
+#include <dao/dao.hpp>
+
+namespace iroha {
```
MizukiSonoko: Why Is `iroha` namespace used?

irohad/main/application.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Contains instances of global services and shared providers
+   */
+  class Application {
```
MizukiSonoko: Why Is Application class?
I think main contains `int main(){};` basically, 
In my opinion, Application class is a little fat...

muratovv: Application is bundle with all services. Torii module know about other in general. But parts of torii will consume part of application directly.

irohad/main/application.hpp
muratovv: What is your proposal for this?

muratovv:
@motxx idea is global modules knows about other modules. Each module by itself separate services by responsibilities  and provide only part of interface.

For example, stateful validation should know about ametsuchi?
`Proposal validate(Proposal, Ametsuchi)`
But this is bad interface. For performing validation you should use only temporary database state (without affecting on blockstore) - interface will be:
`Proposal validate(Proposal, TemporaryWsv)`

We reduce responsibility of validate interface.

For using it in business logic you should invoke like this:
`// peer communication class`
`auto verified_proposal = validate(proposal, ametsuchi.getTemporaryWsv())`

So, what happens? Global module PeerCommunicationService know about ametsuchi, but parts like stateful validation works only with small part of it.

motxx:
@muratovv
class PeerCommunicationService : public TransactionPropagator,
                                    public ConsensusListener {};

motxx:
I'm sorry, I pushed the close button by mistake. Looks nice for me.

irohad/main/application.hpp
motxx: I think the name of `Application` assumes to contain consensus and other modules.
How is `SharedServices`?

irohad/main/application.hpp
MizukiSonoko: I think this module is `iroha` obviously. and this module is not shared by another library. only iroha.
so I propose 
`namespace main{}`

irohad/main/application.hpp
MizukiSonoko: ok i see

MizukiSonoko:
In my opinion, and 
**-- In the future iroha, this is only memo --**
If there is application in hyperledger/iroha, There is `class Application` in libs/application. and it has
```
virtual void initialize();
virtual void commandParser(int argc,char* argv);
virtual void run();
...
```

and There is `class Irohad : private Application{}` in irohad/main/irohad.hpp.

I think Application is separated from irohad/main is better. 
but, Is it complex?
**-- --**
or 
I propose we use name `class Irohad` instead of `class Application`

muratovv:
@MizukiSonoko 
I don't mind for renaming. Pls, commit changes in this branch.

MizukiSonoko:
Ok 
Can I rename like this?
`class Irohad` instead of `class Application`

muratovv:
Yes, u are.

muratovv:
@motxx 
Could you review and merge it?

PR #402, opened 2017-06-28
Feature/ametsuchi stubs

lebdron:
## What is this pull request?
IR-177
Dummy implementations of interfaces provided by 'Ametsuchi'
  
Instantiate `AmetsuchiStub`

## Details/Features
Provides `AmetsuchiStub`, `CommandExecutorStub`, `TemporaryWsvStub`, `MutableStorageStub`


test/module/irohad/ametsuchi/command_executor_test.cpp
```diff
@@ -0,0 +1,32 @@
...
+using iroha::ametsuchi::CommandExecutorStub;
+using namespace iroha::dao;
+
+TEST(CommandExecutorTest, SampleTest) {
+  AmetsuchiStub ametsuchi;
+  CommandExecutorStub executor(ametsuchi);
+
+  ASSERT_TRUE(executor.execute(AddPeer{}));
+  ASSERT_FALSE(executor.execute(Command{}));
+}
```
motxx: Can we add following here?:
```
class DummyCommand : public Command {};
...
ASSERT_FALSE(executor.execute(DummyCommand{}));
```

motxx: Maybe redundant in this case but we can add:
```
struct DummyCommand : public Command {};
...
ASSERT_FALSE(executor.execute(DummyDommand{}));
```

motxx:
This is our discussion in our slack.
```
mizuki
Where will command's main logic be in ? (edited)

motxx
What is main logic? Maybe execute() or executeAddPeer() in CommandExecutor.

motxx
dao::Command can have method but, dao seems to have only data more like struct than class.

mizuki
I think main logic is in two place,
- variable declaration ( int z = 333 )
- executor()

motxx
If this should have only data (variables), executor() is ok. But If so, I want to change all dao::Hoge class -> struct.
https://github.com/hyperledger/iroha/blob/develop/libs/dao/commands/add_peer.hpp

But It may be confusing OOP or data oriented architecture.

mizuki
So I propose

class Command{
   virtual void executor() = 0;
}
```
Continue again in this PR.

MizukiSonoko:
I propose 
https://wandbox.org/permlink/nUrTnoNJSsWI6uHC
instead of @motxx 's propose

muratovv:
@motxx 
In current architecture, where dao is only structures without business logic. Dao _inpedenden_ from other parts. In your proposal connectedness increase because dao should know about business logic (wsv executions).

Think about internal changes in ametsuchi, for example internal scheme of indexes changed. In your proposal we should change two parts: indexes and dao. But in current architecture only indexes.

Also, unfortunately, in c++ not available parametric polymorphism. For make polymorphic behaviour on parameters Andrey write [library](https://github.com/hyperledger/iroha/pull/401) for it.

test/module/irohad/ametsuchi/command_executor_test.cpp
lebdron: The calls to `execute` should be made from `Command` reference like
```C++
AddPeer apcmd{};
Command &cmd = apcmd;
executor.execute(cmd);
```

PR #405, opened 2017-06-29
Fix abstract command in transaction

lebdron:
Vector should contain pointers to allow storing objects of derived classes


motxx:
Looks nice.

PR #417, opened 2017-07-02
Feature/application

lebdron:
## What is this pull request?
`main()` method, which initializes the system
   
Run `iroha-main` application

## Additional dependencies
Other pull request currently open


irohad/main/application.hpp
```diff
@@ -22,22 +22,28 @@
...
 #include <dao/dao.hpp>
 
-namespace main {
+namespace iroha {
 
   /**
    * Contains instances of global services and shared providers
    */
+  using namespace iroha;
   class Irohad {
```
motxx: What is the role of Irohad?
In other words, what is the difference between instances in this class and `main()`?

irohad/main/iroha-main.cpp
```diff
@@ -14,26 +14,62 @@ See the License for the specific language governing permissions and
...
+  iroha::ordering::OrderingServiceStub ordering_service;
+  iroha::consensus::ConsensusServiceStub consensus_service;
+  iroha::network::PeerCommunicationServiceStub peer_communication_service(
+      ametsuchi, stateful_validator, chain_validator, ordering_service,
+      consensus_service, crypto_provider);
+  iroha::torii::ClientProcessorStub client_processor(
+      stateless_validator, peer_communication_service, crypto_provider);
+  iroha::torii::ToriiStub torii(client_processor);
+  // shows required order of execution, since callbacks are called synchronously
+  peer_communication_service.subscribe_on_proposal();
```
MizukiSonoko: Why all service is in main?
I think irohad should have all components and manage it.
and main function doesn't know irohad's components.

irohad/network/peer_communication_stub.hpp
```diff
@@ -0,0 +1,61 @@
...
+  namespace network {
+    class PeerCommunicationServiceStub : public PeerCommunicationService {
+     public:
+      PeerCommunicationServiceStub(
+          ametsuchi::Ametsuchi &storage,
+          validation::StatefulValidator &stateful_validator,
+          validation::ChainValidator &chain_validator,
+          ordering::OrderingService &orderer,
+          consensus::ConsensusService &consensus,
+          dao::DaoCryptoProvider &crypto_provider);
```
MizukiSonoko: I think PeerCommunicationService has many many components

MizukiSonoko: Really they need **CryptoProvider and **Validator?

irohad/main/application.hpp
lebdron: Currently it is not used, since all objects are initialized in `main()`. It should contain shared objects (such as ametsuchi, crypto/hash providers) in the future, since their initialization might be complex.

irohad/main/iroha-main.cpp
muratovv: Correct, but in Irohad was broken initialization. And all services temporarily moved in main. But good news is there is no singleton in iroha! All dependencies are clear and passed only via constructor.

irohad/network/peer_communication_stub.hpp
muratovv: Good point. Yesterday/today we discuss responsibilities  of consensus service and remove from this module responsibility for stateful validation. Further we will discuss how to decrease number of dependencies in PCS. And also making more correct interfaces via virtual destructors, and set const qualifier.

PR #419, opened 2017-07-03
Fix/torii add wsv dependency

lebdron:
## What is this pull request?
Adds `WSVQuery` and `BlockQuery` to `QueryProcessor` of `Torii`.
Removes unnecessary `ClientProcessor`, which is composed of `QueryProcessor` and `TransactionProcessor`


irohad/torii/processor/impl/stub_query_processor.cpp
```diff
@@ -25,7 +25,9 @@ namespace iroha {
     using dao::QueryResponse;
     using dao::Client;
 
-    QueryProcessorStub::QueryProcessorStub() {
+    QueryProcessorStub::QueryProcessorStub(ametsuchi::WsvQuery &wsv,
+                                           ametsuchi::BlockQuery &block) :
```
motxx: Is it possible to separate the instance of Ametsuchi to WsvQuery and BlockQuery?
In other words, is `wsv_(wsv), block_(block)` better than `wsv_(ametsuchi), block_(ametsuchi)`?

And also, the interface of TxStore is now in BlockQuery because of downloading txs from blocks in flat files, is it right?

lebdron: Ametsuchi is a highly coupled component - it will be a problem to answer query such as `getWalletTransactions`, if BlockQuery and WsvQuery are separate, so Ametsuchi implements both interfaces.
Yes, it is because transactions are stored in blocks.

motxx: > Ametsuchi is a highly coupled component - it will be a problem to answer query such as getWalletTransactions, if BlockQuery and WsvQuery are separate, so Ametsuchi implements both interfaces.

Thanks, I understand that.
So, it can be expressed:
```
QueryProcessorStub::QueryProcessorStub(AmetsuchiStub &ametsuchi) :
    wsv_(ametsuchi), block_(ametsuchi) {
```

in `irohad/main/iroha-main.cpp`
```
iroha::torii::QueryProcessorStub qp(ametsuchi);
```
Using ametsuchi 2 times to 1 time.

lebdron: Yes, but in this case `QueryProcessorStub` will know about `AmetsuchiStub` methods (createTemporaryWsv and so on), which he doesn't need

motxx: Yes. Of course, AmetsuchiStub methods don't need.
I agree that QueryProcessorStub has `ametsuchi::WsvQuery wsv_` and `ametsuchi::BlockQuery block_`.

I mean:
`iroha::torii::QueryProcessorStub qp(ametsuchi, ametsuchi);`
can be changed to:
`iroha::torii::QueryProcessorStub qp(ametsuchi);`

but it's minor things, it can be ignored.

motxx: Ah, ok. I'm wrong. The constructor should not get Ametsuchi.

PR #429, opened 2017-07-11
Add Irohad class

MizukiSonoko:
## What is this pull request?
Add Irohad, context
- Add Irohad
- Add Context

make main function more simple


Warchant:
Thanks. I am going to extend this. Pull it ASAP

PR #431, opened 2017-07-11
Feature/add tx publisher

MizukiSonoko:
## What is this pull request?
Add TxPublisher module

TxPublisher task is sent confirm block information to API( another server without Iroha) 
In now, it uses grpc/protobuf.

Iroha will support this command
`AddNotify( address ) ` (tentative named)

```
+---------+
| client📱|
+--|------+
   | transaction
   V
+~~⛩~~~~~~~~~~~~+                              +--------+
|  Iroha network |-------(confirmed block)----> | server |
+~~~~~~~~~~~~~~~~+                              +--------+
```


irohad/network/block_publisher_with_grpc.cpp
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <network/block_publisher_with_grpc.hpp>
+
+namespace network {
+
+    using iroha::protocol::Block;
```
satellitex: I think it uses no `iroha::protoco::l`, yes `iroha::model::`.

irohad/block_publisher/block_publisher.cpp
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <block_publisher/block_publisher.hpp>
+
+namespace block_publisher {
+
+  void BlockPublisher::publish(const Block &block) {
+    // TODO: implement
```
takemiyamakoto: 途中なところでコメントを書いて貰うと他人に優しいです

irohad/network/block_publisher_with_grpc.cpp
```diff
@@ -0,0 +1,53 @@
...
+  ) {
+    ApiClient client(targetIp, 50051); // TODO: Get port from config
+    ApiBlock apiBlock;
+    return client.receiveBlock(apiBlock);
+  }
+
+  ApiClient::ApiClient(const std::string& ip, int port) {
+    // TODO(motxx): call validation of ip format and port.
+    auto channel = grpc::CreateChannel(ip + ":" + std::to_string(port), grpc::InsecureChannelCredentials());
+    stub_ = iroha::protocol::ApiService::NewStub(channel);
```
takemiyamakoto: This is somewhat ugly

irohad/network/block_publisher_with_grpc.cpp
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  ApiResponse ApiClient::receiveBlock(const ApiBlock& block) {
+    ApiResponse response;
+    auto status = stub_->receiveBlock(&context_, block, &response);
```
takemiyamakoto: I'm not sure I like this use of stub

irohad/network/block_publisher_with_grpc.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    if (status.ok()) {
```
takemiyamakoto: A cleaner way to do this error handling might be to use try to construct the ApiResponse object and throw and exception instead of returning status. Then if the exception is thrown, then the ApiResponse object won't be constructed, so you can just make your own response then notifying of an error. Does this make sense?

irohad/network/block_publisher_with_grpc.hpp
```diff
@@ -0,0 +1,51 @@
...
+    );
+  };
+
+  class ApiClient {
+    public:
+      ApiClient(const std::string& targetIp, int port);
+      iroha::protocol::ApiResponse receiveBlock(const iroha::protocol::ApiBlock&);
+    private:
+      grpc::ClientContext context_;
+      std::unique_ptr<iroha::protocol::ApiService::Stub> stub_;
```
takemiyamakoto: I'm not sure I like this. It is only temporary, right?

neewy:
In general this is not what should be defined as an interface between Iroha and external clients in my opinion. We should decide on something more concise to return, like a model object or a specific wrapper for a command. Why should we return block? It will involve parsing logic on the client

neewy:
screen shot 2017-07-12 at 10 33 38
Do we need a thin or somewhat thicker client?

neewy:
Another comment is that in open ledger blocks as a transport unit is acceptable, but we have ledger with permissions systems, so the client may receive block which potentially has transactions which are not allowed for the client to view, but the client can get the data in it no matter what its permissions are. What do you think about this?

MizukiSonoko:
In my opinion
In now, Iroha does not support transaction hidden by domain so all transaction send to API server
but Iroha will support it
- Server Send `AddNotify( address )` with my address, my publicKey and signatures
- After consensus formed, filtering transaction using tx creator publicKey and make filtered tx list

irohad/network/block_publisher_with_grpc.cpp
motxx: これはconfigを使うべきという意味ですか？
今のところportについて、PeerのModelとすべきか、configに置くべきか固定にすべきか立ち位置が決まっていません

irohad/network/block_publisher_with_grpc.cpp
motxx: gRPCはRPCの通信先のピアをstubという命名で扱っているようです。
https://github.com/grpc/grpc/blob/master/examples/cpp/helloworld/greeter_async_client.cc#L40

irohad/network/block_publisher_with_grpc.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include <grpc++/create_channel.h>
+#include <grpc++/security/credentials.h>
+
+namespace network {
+
+  /*
+   * Send confirmed block to other server without Iroha,
+   * This is only interface
+   */
+  class BlockPublisherClientWithGrpc: BlockPublisherClient{
```
takemiyamakoto: Publish := send to peer?

MizukiSonoko: Send confirmed block to other server without Iroha.
But what is best name?

PR #433, opened 2017-07-12
Update proto files

grimadas:
Updating obsolete proto files to be consistent with model

### DIFF
--- a/schema/block.proto
+++ b/schema/block.proto
@@ -32,7 +32,7 @@ message Block {
     uint32             tx_number      = 1;  // the number of transactions inside. Maximum 16384 or 2^14
     uint32             height         = 2;  // the current block number in a ledger
     bytes              merkle_root    = 3;  // global merkle root
-    bytes              prev_hash      = 4; // Previous block hash
+    bytes              prev_block_hash      = 4; // Previous block hash
   }
 
   message Body {
--- a/schema/block.proto
+++ b/schema/block.proto
@@ -3,35 +3,44 @@ package iroha.protocol;
 import "commands.proto";
 import "primitive.proto";
 
+message Header {
+  uint64             created_time    = 1;
+  repeated Signature signatures       = 2;
+}
+
 message Transaction {
-  message Header {
-    uint64             created_time    = 1;
-    repeated Signature signature       = 2;
-    uint32             version         = 3; // show scheme version like 1, 1.1, ... added by mizuki
+
+  message Meta {
+    bytes creator_pubkey = 1;
+    // used to prevent replay attacks.
+    uint64  tx_counter = 2;
   }
 
   message Body {
-    bytes            creator_pubkey  = 1;
-    repeated Command command          = 2;
+    repeated Command commands        = 1;
   }
 
   Header header = 1;
-  Body   body   = 2;
+  Meta meta = 2;
+  Body   body   = 3;
 }
 
 message Block {
-  message Header {
-    uint32             tx_n           = 1;  // the number of transactions inside. Maximum 16384 or 2^14
+
+
+  message Meta {
+    uint32             tx_number      = 1;  // the number of transactions inside. Maximum 16384 or 2^14
     uint32             height         = 2;  // the current block number in a ledger
     bytes              merkle_root    = 3;  // global merkle root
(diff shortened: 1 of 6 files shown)


schema/block.proto
```diff
@@ -3,35 +3,44 @@ package iroha.protocol;
 import "commands.proto";
 import "primitive.proto";
 
+message Header {
+  uint64             created_time    = 1;
```
MizukiSonoko: Can we use `google.protobuf.Timestamp`?
https://github.com/google/protobuf/issues/991

schema/commands.proto
```diff
@@ -1,293 +1,69 @@
...
-
-  // transaction "creator" invites account with account_id to the domain
-  message InviteAccount {
-    string account_id = 1; // invitee
-    string domain = 2; // invite to this domain
-    uint32 permissions = 3; // logical or of Permissions enum
-    string username = 4; // account will join with this username: <username>@soramitsu.co.jp
+    string account_name = 1;
+    string asset_name = 2;
+    string amount = 3;
```
motxx: Is this not `Asset` or `Currency`?

schema/commands.proto
```diff
@@ -1,293 +1,69 @@
...
+  message SetQuorum {
+    string account_name = 1;
+    int32 quorum = 2;
   }
 
-  // set peer role in the network
-  message SetRole {
-    string ip = 1;
-    Role role = 2;
+  message Transfer {
```
MizukiSonoko: Should it be in? I think it's in Asset.

schema/commands.proto
grimadas: We should decide how to make it better, it could be included in 'string amount'. In order to speed up development I use string.

schema/commands.proto
grimadas: This is just naming. But you actually Transfer from Account to Account

satellitex: I think `Asset{}`, `Account{}`, `Domain{}`, and so on are related by Permission model.
So, `Asset` has permission that are `Add`, `Subtract`, `Transfer`.

schema/commands.proto
```diff
@@ -1,293 +1,69 @@
...
+    string account_name = 1;
+    string asset_name = 2;
+    string amount = 3;
   }
 
-  // remove account from a domain (creator should have according permissions)
-  // this is for domain admins to kick accounts from a domain
-  message RemoveAccount {
-    string username = 1; // full username with full domain name: username@domain
+  message Create{
```
satellitex: `AssetCreate` should move to `Domain`.
The reason is below.

grimadas:
I updated schema of proto files, and removed confusing nesting.

schema/block.proto
lebdron: Does it support 64-bit timestamps? Since they are used here https://github.com/hyperledger/iroha/blob/develop/irohad/ordering/queue.cpp#L56

schema/block.proto
```diff
@@ -3,35 +3,44 @@ package iroha.protocol;
...
-    uint32             tx_n           = 1;  // the number of transactions inside. Maximum 16384 or 2^14
+
+
+  message Meta {
+    uint32             tx_number      = 1;  // the number of transactions inside. Maximum 16384 or 2^14
     uint32             height         = 2;  // the current block number in a ledger
     bytes              merkle_root    = 3;  // global merkle root
-    uint64             created_time   = 4;  // block creation time
-    repeated Signature peer_signature = 5;  // peers' signatures
+    bytes              prev_hash      = 4; // Previous block hash
```
takemiyamakoto: *prev_block_hash* is better

schema/commands.proto
```diff
@@ -1,293 +1,84 @@
...
-  }
-
-  // set peer role in the network
-  message SetRole {
-    string ip = 1;
-    Role role = 2;
-  }
+message TransferAsset {
+    string src_account_id = 1;
+    string dst_account_id = 2;
```
takemiyamakoto: *dest_account_id* is better or *dest_acct_id* is better.

schema/block.proto
MizukiSonoko: ok

PR #452, opened 2017-07-18
Feature/mizuki refactoring consensus

MizukiSonoko:
## What is this pull request?
A little fix & pass 
   
To pass CI
  
Do you provide tests, or use case examples?
Describe how to use features in your pull request.

## Details/Features
- Remove `set_test`
- `using Transaction = protocol::Transaction`


irohad/torii/command_service.cpp
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 #include "torii/command_service.hpp"
```
motxx: Fix `<torii/command_service.hpp>` ?

lebdron: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
We should include it with `"` symbol

motxx: If we'll use `"` symbol here, documents should be written at first, then all codes should follow the style (but currently, we use `<>` symbol for include path)

lebdron:
Closing this PR, since code in `feature/consensus` was replaced by force push

PR #453, opened 2017-07-18
Feature/stateful validation

grimadas:
## What is this pull request?
Stateful validation for all basic commands.
   
Stateful validation allows to filter out non-valid commands and  apply only valid ones. 
  
Each command has two methods: validate and execute. 
Validate method checks permissions and business. Execute is coupled with database commands allowing to perform checks for existence in database.

The usage of command validation is demonstrated in this file:
` irohad/validation/impl/stateful_validator_impl.cpp `


irohad/model/account_asset.hpp
```diff
@@ -37,7 +38,7 @@ namespace iroha {
       /**
        * Current balance
        */
-      std::string balance;
+      uint64_t balance;
```
satellitex: Why is balance 64bit?
It can not be expressed in 64 bit.

irohad/model/commands/transfer_asset.hpp
```diff
@@ -45,7 +46,7 @@ namespace iroha {
       /**
        * Amount of transferred asset
        */
-      std::string amount;
+      std::decimal::decimal64 amount;
```
motxx: `clang` doesn't support. maybe ok for now.
but in the future, how is importing `boost::multiprecision`?

irohad/model/account_asset.hpp
lebdron: It is a temporary solution, since we use `decimal64` now

irohad/model/commands/transfer_asset.hpp
grimadas: We are already fixing it. We will use our implementation of Amount, simply two ints.

irohad/model/account_asset.hpp
grimadas: for example 1$ 50 cents can be presented as 150.

satellitex: We must use 20-digit.
So it will replace after that.

PR #457, opened 2017-07-19
Feature/validation tests pure

grimadas:
## What is this pull request?
Fixed stateful validation with tests. 
   
Now for each stateful validation there is a test proving work of the validation.
For the test we use Mock object hence there is no dependency from other modules.  
  
For each command there are tests, that you can run.  
Each command has tests for valid and not valid cases. 
You are welcome to add new/missed cases for commands.

### DIFF
--- a/irohad/model/impl/stateful_command_validation.cpp
+++ b/irohad/model/impl/stateful_command_validation.cpp
@@ -82,15 +82,15 @@ namespace iroha {
     bool AssignMasterKey::validate(ametsuchi::WsvQuery &queries,
                                    const Account &creator) {
       auto signs = queries.getSignatories(account_id);
-
+      auto acc = queries.getAccount(account_id);
       return
           // Two cases - when creator assigns to itself, or system admin
           (creator.account_id == account_id ||
            creator.permissions.add_signatory) &&
-          // Check if account has at lest one signatory
-          signs &&
+          // Check if account exist and has at least one signatory
+          signs && acc.has_value() &&
           // Check if new master key is not the same
-          creator.master_key != pubkey &&
+          acc.value().master_key != pubkey &&
           // Check if new master key is in AccountSignatory relationship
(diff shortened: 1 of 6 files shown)


irohad/model/impl/stateful_command_validation.cpp
```diff
@@ -123,7 +123,7 @@ namespace iroha {
       // Creator must have permission to create assets
       return creator.permissions.create_assets &&
              // Name is within some range
-             asset_name.size() > 0 && asset_name.size() < 8 &&
+             asset_name.size() > 0 && asset_name.size() < 10 &&
```
l4l: wtf is 10?

grimadas: Just some number, can be changed any time

grimadas: Max length of asset name

irohad/model/impl/stateful_command_validation.cpp
```diff
@@ -211,13 +212,17 @@ namespace iroha {
...
           // Can account transfer assets
           creator.permissions.can_transfer &&
           // Creator can transfer only from their account
           creator.account_id == src_account_id &&
           // Balance in your wallet should be at least amount of transfer
           account_asset.value().balance >=
-              amount.get_joint_amount(asset.value().precision);
+              amount.get_joint_amount(asset.value().precision) &&
+          // Amount must be not zero
+          (amount.frac_part > 0 || amount.int_part > 0);
```
MizukiSonoko: Asset's amount will not be zero? why?

grimadas: If you allow transfer zero assets - you ddos the system

MizukiSonoko: ok i see

PR #464, opened 2017-07-23
Feature/missed stubs

muratovv:
## What is this pull request?
Add missed interfaces with respect to [current architecture](https://soramitsu.atlassian.net/wiki/spaces/IS/pages/1725585/Class+diagram+for+v0.95) 
   
Interfaces need to build up whole tx pipeline in current architecture


neewy:
I request to merge this PR since we need to move on to use current architecture and deliver components implementation

PR #471, opened 2017-07-24
Feature/torii query handlers

kamilsa:
## What is this pull request?
This PR includes:

1. Refactoring for stateless validation
2. Implementation of query processor
3. Update of tx response, so that it does not contain client anymore
4. Introduced query response
   
Query processor allows executing queries or return bad status if stateless validation is wrong
  
There is a test in test/module/irohad/torii/processor/transaction_processor_test

## Details/Features
Passes either query response or query stateless response to the observable


irohad/validation/impl/stateless_validator_impl.cpp
```diff
@@ -15,13 +15,11 @@
  * limitations under the License.
  */
 
+#include "stateless_validator_impl.hpp"
```
grimadas: Must be full name, i.e. `validation\impl\stateless_validator_impl.hpp`

kamilsa: Fixed

PR #474, opened 2017-07-24
Feature/update cryptoprovider

kamilsa:
## What is this pull request?
Adds functionality to cryptoprovider to sign blocks
   
To be able to sign blocks
  
There is a test in test/module/irohad/model/providers/model_crypto_provider_test.cpp

## Details/Features
Test on providers we moved from test/libs to test/module/irohad/model/providers


test/module/irohad/model/providers/model_crypto_provider_test.cpp
```diff
@@ -31,6 +38,18 @@ iroha::model::Transaction create_transaction() {
   return tx;
 }
 
+iroha::model::Block create_block() {
```
takemiyamakoto: It would be nice to have some kind of shared test_utils library that can be used to share sample blocks and transactions for use in various tests that we right.

PR #480, opened 2017-07-31
Fix/Refactor block serializer

grimadas:
## What is this pull request?
Checkers for ill-formed json added to json-serializer.  
   
To avoid runtime exceptions while serialising  json.
  
Additional tests for invalid cases: 
1) Invalid Transaction 
2) Invalid Block

### DIFF
--- a/irohad/ametsuchi/impl/block_serializer.cpp
+++ b/irohad/ametsuchi/impl/block_serializer.cpp
@@ -420,13 +420,14 @@ namespace iroha {
       }
 
       model::Block block{};
-      // Check if hash is presented
-      if (doc.HasMember("hash")) {
-        std::string hash_str(doc["hash"].GetString(),
-                             doc["hash"].GetStringLength());
-        auto hash_bytes = hex2bytes(hash_str);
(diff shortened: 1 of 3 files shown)


irohad/ametsuchi/impl/block_serializer.cpp
```diff
@@ -449,70 +458,99 @@ namespace iroha {
...
+          auto sig_sign_bytes = hex2bytes(sig_sign);
+          std::copy(sig_sign_bytes.begin(), sig_sign_bytes.end(),
+                    signature.signature.begin());
+
+          tx.signatures.push_back(signature);
+        }
+      }
+      if (json_tx.HasMember("created_ts")) {
+        // created_ts
+        tx.created_ts = json_tx["created_ts"].GetUint64();
```
kamilsa: Shall we return here false when created_ts is absent? I think this field is necessary for stateless validation, so we should not allow keep it empty

grimadas: Created ts will be created during signing process in iroha-cli

PR #488, opened 2017-08-02
Feature/raw block insertion

muratovv:
## What is this pull request?
This pr provide functionality to insert blocks in storage in unsafe mode, without validations. Those functionality required for put genesis block to ledger.   
  
- Do you provide tests, or use case examples?
Tests provides in module/irohad/common/block_insertion_tests.cpp

## Details/Features
- Add class with block insertion functionality
- Add useful blob converter.


irohad/common/blob_converter.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace iroha {
+  namespace common {
+
+    /**
+     * Convert string to blob vector
+     * @param source - string for conversion
+     * @return vector<blob>
+     */
+    std::vector <uint8_t> convert(std::string &source);
```
grimadas: You can use blob_t to hex_string and hex2bytes in common/types.hpp

irohad/main/impl/raw_block_insertion.cpp
```diff
@@ -0,0 +1,59 @@
...
+namespace iroha {
+  namespace main {
+
+    BlockInserter::BlockInserter(std::shared_ptr<ametsuchi::MutableFactory> factory)
+        : factory_(std::move(factory)) {
+    };
+
+    nonstd::optional<model::Block> BlockInserter::parseBlock(std::string &data) {
+      auto blob = stringToBytes(data);
+      ametsuchi::BlockSerializer serializer;
```
grimadas: Maybe it is better to write serilaizer from string ? There is already serliazer from string for transaction

libs/common/types.hpp
```diff
@@ -85,6 +82,44 @@ namespace iroha {
...
+    }
+    return bytes;
+  }
+
+  /**
+   * Convert string to blob vector
+   * @param source - string for conversion
+   * @return vector<blob>
+   */
+  inline std::vector<uint8_t> stringToBytes(std::string &source) {
```
grimadas: Maybe put const ref here ?

libs/common/types.hpp
```diff
@@ -85,6 +82,44 @@ namespace iroha {
...
+    }
+    return result;
+  }
+
+  /**
+   * blob vector to string
+   * @param source - vector for conversion
+   * @return result string
+   */
+  inline std::string bytesToString(std::vector<uint8_t> &source) {
```
grimadas: Maybe const ref ?

PR #505, opened 2017-08-10
added expected Timer behavior

x3medima17:
removed redundant std::move()
added expected Timer behavior


muratovv:
Merged into develop. [commit](https://github.com/hyperledger/iroha/commit/14e535c5b79d0dddf4af56fe1b888fdc5255f783)

PR #508, opened 2017-08-12
[WIP] IR279 - Add convenient initializers for blob_t

Warchant:
## What is this pull request?
- [x] add `blob_t<N>::from_string(std::string s);` which creates blob_t of size N from strings
- [x] add `blob_t<N>::from_hexstring(std::string s);` which creates blob_t of size N from string encoded in hex (like 0x1234567890abcdef)
- [x] add `blob_t<N>::from_string_var(std::string s);` which creates blob_t of size N from strings of variable size. Example `blob_t<32>::from_string_var("1"); // result is 1000...00 (32 bytes)`
- [x] add tests for each of method and constructor
- [x] fixed few bugs in tests related to "bad format" for pubkey and signature
- [x] refactored the code, to get rid of `std::copy`

Anything else?

### DIFF
--- a/irohad/ametsuchi/impl/postgres_wsv_query.cpp
+++ b/irohad/ametsuchi/impl/postgres_wsv_query.cpp
@@ -93,8 +93,7 @@ namespace iroha {
       std::vector<ed25519::pubkey_t> signatories;
       for (const auto &row : result) {
         pqxx::binarystring public_key_str(row.at("public_key"));
-        ed25519::pubkey_t pubkey =
-            std::string{public_key_str.begin(), public_key_str.end()};
+        ed25519::pubkey_t pubkey = public_key_str.str();
         signatories.push_back(pubkey);
       }
       return signatories;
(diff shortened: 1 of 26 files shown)


Warchant:
TODO: please, review tests once more. Some of them are broken after my update.
The reason is that public keys are created from strings of incorrect lengths, thus tests fail with exceltion `std::invalid_argument`

libs/crypto/ed25519_impl.cpp
```diff
@@ -16,12 +16,19 @@
  */
 
 #include <ed25519.h>
```
takemiyamakoto: I don't really like that this file is called ed25519_impl. We don't have hash_impl, etc. It feels somewhat inconsistent to me.

irohad/ametsuchi/impl/postgres_wsv_query.cpp
```diff
@@ -55,8 +55,7 @@ namespace iroha {
       row.at("account_id") >> account.account_id;
       row.at("domain_id") >> account.domain_name;
       pqxx::binarystring master_key(row.at("master_key"));
-      std::copy(master_key.begin(), master_key.end(),
-                account.master_key.begin());
+      account.master_key = std::string{master_key.begin(), master_key.end()};
```
lebdron: Use `pqxx::binarystring::str()`

neewy:
We have agreed to use optional type + monads for error handling instead of catching exceptions

neewy:
please, use monads the same way as in https://github.com/hyperledger/iroha/pull/552 @Warchant

neewy:
Assigned "candidate for closing" since https://github.com/hyperledger/iroha/pull/552 was accepted

neewy:
@Warchant is requested to review https://github.com/hyperledger/iroha/pull/552

neewy:
@Warchant please review this PR

PR #510, opened 2017-08-13
Feature/registration factory

muratovv:
## What is this pull request?
This pr aimed for solve problem with run-time parametric polymorphism. 
Way to solve it make set of interested objects, inflated on static time initialization.
  
This pr provide class for registration classes classed `ClassHandler`. Register interested classes in static initialization time and use for your purposes.

## Details/Features
Current pr provide registrations for:
* queries
* query responses
* transaction responses
* commands


irohad/model/registration/command_registration.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include "model/commands/transfer_asset.hpp"
+
+/**
+ * File contains registration for all command subclasses
+ */
+
+namespace iroha {
+  namespace model {
+
+    static ClassHandler command_handler;
```
lebdron: https://stackoverflow.com/a/3837513
Not sure if there is a profit of using static in this case. We can put `command_handler` inside `CommandRegistry`, and then create objects of this class where it is required. Same applies to other registries

PR #523, opened 2017-08-23
Add logging for yac and yac network;

muratovv:
## What is this pull request?
This pr provides logging for yac and for yac network.
  
All in case, just look for output when use consensus.

### DIFF
--- a/irohad/consensus/yac/impl/network_impl.cpp
+++ b/irohad/consensus/yac/impl/network_impl.cpp
@@ -19,11 +19,11 @@
 #include <grpc++/grpc++.h>
 #include "logger/logger.hpp"
 
-using namespace logger;
-
 namespace iroha {
   namespace consensus {
     namespace yac {
+
+      // ----------| Public API |----------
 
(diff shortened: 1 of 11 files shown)


irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -37,8 +37,8 @@ namespace iroha {
       };
 
       void YacGateImpl::vote(model::Block block) {
-        log_->info("vote for block");
         auto hash = hash_provider_->makeHash(block);
+        log_->info("vote for block", hash.block_hash);
```
lebdron: Variable output is missing.

irohad/consensus/yac/impl/network_impl.cpp
```diff
@@ -19,12 +19,12 @@
 #include <grpc++/grpc++.h>
 #include "logger/logger.hpp"
 
-using namespace logger;
```
lebdron: Why remove `using` here, but use it in other .cpp files?

PR #533, opened 2017-08-26
Fix genesis block generator yielding different blocks due to timestamp

lebdron:
`swarm-deploy.sh` used block generator for genesis block. Since nodes are deployed sequentially, timestamps of genesis blocks were different, which resulted in different genesis block hash.

This PR fixes the creation timestamp of genesis block to 0.

### DIFF
--- a/irohad/model/generators/impl/block_generator.cpp
+++ b/irohad/model/generators/impl/block_generator.cpp
@@ -23,9 +23,8 @@ namespace iroha {
     namespace generators {
       Block BlockGenerator::generateGenesisBlock(
           std::vector<std::string> peers_address) {
-        Block block;
-        block.created_ts =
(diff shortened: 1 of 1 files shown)


irohad/model/generators/impl/block_generator.cpp
```diff
@@ -23,9 +23,8 @@ namespace iroha {
     namespace generators {
       Block BlockGenerator::generateGenesisBlock(
           std::vector<std::string> peers_address) {
-        Block block;
-        block.created_ts =
-            (ts64_t)std::chrono::system_clock::now().time_since_epoch().count();
+        Block block{};
```
muratovv: What is the reason to set curly brackets?

lebdron: It default-initializes the fields of a struct.

PR #535, opened 2017-08-27
Add logging to command executor

muratovv:
## What is this pull request?
This pr provides logging in command executor.
   
Functionality gives useful information about fail case on stateful validation in system.

## Details/Features
* Logging in command executor
* Rework unused variable
* Replace logic symbols (&& ||) with modern
* Little style refactoring

### DIFF
--- a/irohad/model/execution/impl/command_executor.cpp
+++ b/irohad/model/execution/impl/command_executor.cpp
@@ -34,6 +34,10 @@
 using namespace iroha::model;
 using namespace iroha::ametsuchi;
 
+CommandExecutor::CommandExecutor() {
+  log_ = logger::log("CommandExecutor");
+}
+
 bool CommandExecutor::validate(const Command &command, WsvQuery &queries,
                                const Account &creator) {
   return hasPermissions(command, queries, creator) && isValid(command, queries);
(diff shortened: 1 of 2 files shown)


irohad/model/execution/impl/command_executor.cpp
```diff
@@ -48,40 +52,47 @@ bool AddAssetQuantityExecutor::execute(const Command &command,
 
   auto asset = queries.getAsset(add_asset_quantity.asset_id);
   if (not asset.has_value()) {
-    // No such asset
+    log_->info("[AddAssetQuantity] asset {} is absent",
```
lebdron: It would probably be less error-prone to initialize logger with new name in derived classes, than to copy command name every time.

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -194,18 +207,18 @@ bool AssignMasterKeyExecutor::isValid(const Command &command,
   auto acc = queries.getAccount(assign_master_key.account_id);
   // Check if account exist and new master key is not the same
```
lebdron: This also looks like a place for log.

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -147,11 +158,11 @@ bool AddSignatoryExecutor::hasPermissions(const Command &command,
...
   return
-      // Case 1. When command creator wants to add signatory to their
-      // account
-      add_signatory.account_id == creator.account_id ||
-      // Case 2. System admin wants to add signatory to account
-      creator.permissions.add_signatory;
+    // Case 1. When command creator wants to add signatory to their
+    // account
+      add_signatory.account_id == creator.account_id or
+          // Case 2. System admin wants to add signatory to account
```
lebdron: Is this indent required? Similar indents also appear below.

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -497,7 +519,7 @@ bool TransferAssetExecutor::isValid(const Command &command,
 
   // Amount must be not zero
```
lebdron: Also looks like logging can be added here.

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -367,6 +417,8 @@ bool SetAccountPermissionsExecutor::execute(const Command &command,
   auto account = queries.getAccount(set_account_permissions.account_id);
   if (not account.has_value()) {
     // There is no such account
```
lebdron: Can this comment be replaced with log? This also appears several times below

PR #540, opened 2017-08-28
Delete master key entity

neewy:
## What is this pull request?
Deletion of account master key
   
Now, account does not have to contain master key, as it is uniquely identified by account id.


neewy:
@muratovv should I also delete https://github.com/hyperledger/iroha/blob/86463eb8d5e5adffe55aac81d68d6385133cd181/irohad/model/account.hpp#L126?

neewy:
Consider using this code if needed, but I am stuck with master key wiping

neewy:
Made a mistake in changing status. Needs correction, I will change it a bit

PR #541, opened 2017-08-30
Update of maintainers list

neewy:
## What is this pull request?
Update of maintainers list

Previously, list of maintainers did not contain developer and had typo in GitHub name of our maintainer.


muratovv:
Are u really want to merge it in master branch?

PR #543, opened 2017-08-31
Add check to cli 

grimadas:
Add checks to cli to check if file defined in the flags is present.


iroha-cli/main.cpp
```diff
@@ -82,13 +82,19 @@ int main(int argc, char* argv[]) {
       logger->info("Send transaction to {}:{} ", FLAGS_address,
                    FLAGS_torii_port);
       std::ifstream file(FLAGS_json_transaction);
+      if (not file.is_open()){
```
lebdron: Execution should probably stop if such error occurs.

neewy:
It was proposed to introduce mechanism of handling critical situations both in irohad, iroha-cli. Consider changing this PR after we will decide on that

iroha-cli/main.cpp
```diff
@@ -91,6 +92,10 @@ int main(int argc, char *argv[]) {
```
x3medima17: curl

x3medima17: test

PR #549, opened 2017-09-04
Update README.md

neewy:
## What is this pull request?
Updated readme in master branch
   
As we changed script for Docker environment, we should inform developers on main page of repo. _However, this information is only valid for develop branch._ Please, tell me what do you think, and should we mention it to use develop branch for up-to-date code in README text?


neewy:
Since it is valid only for develop branch, we won't update main page (README.md in master branch) until next release

PR #550, opened 2017-09-04
Feature/test ir341

luckychess:
## What is this pull request?
Changes related to crypto:
- hash provider removed from everywhere
- crypto reworked (hashes and signatures)
- everything uses crypto were changed accordingly
- changed model for proto - transaction and query: both of them now have a payload field contained all the data and signature field for a payload's signature
- all code is changed according to new model (serializers, deserializers, converters and some other)

Other changes:
- cmake proto compilation improvements
- queries in serializers are const now
- std::abs compilation issue (again!)
   
Now to use crypto functions you have to serialize your model object to proto object and use its string representation (see code example below). This way allows us to get consistent hashes/signatures which do not depend from target platform internal binary data representation (endianess and so on). Proto model changes simplify crypto usage: now you always have a payload field and hashes/signs are calculated from it.
  
Typical workflow for hashing and signing now:
```
protocol::Transaction tx;
hash256_t hash = sha3_256(tx.payload().SerializeAsString());
iroha::sign(hash.to_string(), pubkey, privkey);
```

To get a signature from string there is from_string function available:
```
Signature sign{};
sign.pubkey = pubkey_t::from_string(pubkey_string);
sign.signature = sig_t::from_string(signature_string);
```


muratovv:
No reason to start review before we are not discussing, how model will looks like. I guess better to provide draft of model interface and after it make changes in code. Short review demonstrate that code now very dirty: changes not obvious and come pieces looks horrible. Another problem is no reason to review 4+k loc changes, separate pr to atomic parts - new model, crypto, etc.

libs/builder/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+message (STATUS "Hi there")
```
neewy: privet

irohad/ametsuchi/block_query.hpp
```diff
@@ -18,6 +18,8 @@
 #ifndef IROHA_BLOCK_QUERY_HPP
 #define IROHA_BLOCK_QUERY_HPP
 
+#include <cmath>
```
Warchant: why is this here?

irohad/model/converters/impl/pb_query_factory.cpp
```diff
@@ -30,91 +30,100 @@ namespace iroha {
       PbQueryFactory::PbQueryFactory() {
         log_ = logger::log("PbQueryFactory");
         serializers_[typeid(GetAccount)] = &PbQueryFactory::serializeGetAccount;
-        serializers_[typeid(GetAccountAssets)] = &PbQueryFactory::serializeGetAccountAssets;
-        serializers_[typeid(GetAccountTransactions)] = &PbQueryFactory::serializeGetAccountTransactions;
-        serializers_[typeid(GetAccountAssetTransactions)] = &PbQueryFactory::serializeGetAccountAssetTransactions;
-        serializers_[typeid(GetSignatories)] = &PbQueryFactory::serializeGetSignatories;
+        serializers_[typeid(GetAccountAssets)] =
+            &PbQueryFactory::serializeGetAccountAssets;
+        serializers_[typeid(GetAccountAssets)] =
```
Warchant: get account assets here is twice

libs/common/types.hpp
```diff
@@ -80,6 +82,17 @@ namespace iroha {
       }
       return res;
     }
+
+    static blob_t<size_> from_string(const std::string &data) {
+      if(data.size() != size_) {
+        throw BadFormatException("blob_t: input string has incorrect length");
```
Warchant: to other reviewers: this will be changed to optional as soon as Andrei will merge #552

irohad/model/CMakeLists.txt
```diff
@@ -3,15 +3,26 @@ add_subdirectory(converters)
 
 add_library(model
     model_crypto_provider_impl.cpp
-    model_hash_provider_impl.cpp
     impl/model_operators.cpp
+    converters/impl/pb_block_factory.cpp
+    converters/impl/pb_transaction_factory.cpp
+    converters/impl/pb_command_factory.cpp
```
lebdron: Looks like something went wrong, since these files are in `*_model_converters` targets in `irohad/model/converters/CMakeLists.txt`.

libs/crypto/crypto.hpp
```diff
@@ -63,8 +65,6 @@ namespace iroha {
    * @param seed
    * @return
    */
-  ed25519::keypair_t create_keypair(blob_t<32> seed);
-
-
+  keypair_t create_keypair(blob_t<32> seed);
```
l4l: Maybe `create_keypair(void)` will be a better solution (no issues with seed)?

test/integration/client_test.cpp
```diff
@@ -175,10 +175,13 @@ TEST_F(ClientServerTest, SendTxWhenStatelessInvalid) {
...
 
   auto doc = iroha::model::converters::stringToJson(json_tx).value();
-  iroha::model::converters::JsonTransactionFactory transactionFactory;
-  auto tx = transactionFactory.deserialize(doc).value();
-  iroha::model::HashProviderImpl hashProvider;
-  tx.tx_hash = hashProvider.get_hash(tx);
+  iroha::model::converters::JsonTransactionFactory jsonTransactionFactory;
+  auto tx = jsonTransactionFactory.deserialize(doc).value();
+  iroha::model::converters::PbTransactionFactory pbTransactionFactory;
+  auto hash = iroha::sha3_256(pbTransactionFactory.serialize(tx)
```
l4l: That will definitely not a proper way to use crypto stuff. Add todo for writing subfunctions like `hash(const Transaction&)` or `hash(const Block&)`

irohad/ametsuchi/block_query.hpp
luckychess: Well, that's really rx-cpp issue and this include should be somewhere inside their code. This place is bad but other places are not really better. And yes, without this line some systems (at least, my one) get a compilation error about ambiguous call of std::abs somewhere inside rc-cpp internals.

libs/crypto/crypto.hpp
luckychess: We could create an overload for ```create_keypair(void)``` but do not remove a version with seed because we have 2 different ```create_seed()``` implementations now (and both are used). I can add it if you think it will be useful.

test/integration/client_test.cpp
luckychess: Good point. I added TODO; actual work will require a lot of code changes so it should be done in separate PR I think.

irohad/model/CMakeLists.txt
luckychess: Thanks, fixed.

libs/crypto/crypto.hpp
l4l: Yep, don't need to be deleted and as long as pr related to reworking crypto imo this is fine to add

test/integration/client_test.cpp
l4l: Yes, thanks

test/module/irohad/peerservice/peerservice_test.cpp
```diff
@@ -33,7 +33,7 @@ class Service : public testing::Test {
   Service() {
     std::vector<ed25519::keypair_t> kps(NPEERS);
     std::generate_n(kps.begin(), NPEERS,
-                    []() { return create_keypair(create_seed()); });
+                    []() { return create_keypair(); });
```
l4l: can be just `std::generate_n(kps.begin(), NPEERS, create_keypair)`
and btw `std::generate(kps.begin(), kps.end(), create_keypair)` seems less error-prone for me

libs/crypto/ed25519_impl.cpp
```diff
@@ -81,4 +81,10 @@ namespace iroha {
 
     return keypair_t{.pubkey = pub, .privkey = priv};
   }
+
+  /**
+   * Create new keypair with a default seed (by create_seed())
```
l4l: Imo comments in header is enough

irohad/ametsuchi/block_query.hpp
l4l: Same problem, anyway this include doesn't help

l4l: Look at [this](https://gist.github.com/l4l/701b2d56dfc9b9714ab4955b1df16902) one

test/integration/client_test.cpp
neewy: @luckychess please tell me which TODO task to add to backlog 😄

PR #554, opened 2017-09-05
Added CI webhook to update upsource status

Warchant:
Adds a script to update build status on https://upsource.soramitsu.co.jp from circle CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - checkout
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -13,6 +14,10 @@ jobs:
     steps:
       - checkout
       - run:
+          name: notify upsource that about current build
```
neewy: "that" is redundant?

neewy:
Removed codacy warnings

PR #592, opened 2017-09-29
Feature/tx pipeline integration test

lebdron:
## What is this pull request?
Integration test which sends a transaction to the system and verifies that it achieves commit.

## Details/Features
- Add peer keypair to `CryptoProvider`, allowing `Simulator` to sign created blocks.
- Add block signature to `YacHash`, allowing `YacGate` to correctly merge block signatures from peers, which have voted for the block
- Fix stateless validator test for transaction timestamp


Warchant:
In [your test](http://upsource.soramitsu.co.jp/iroha/file/9026c34fcf4b8ffec21a926f8913e062ebb9c241/test/integration/pipeline/tx_pipeline_integration_test.cpp?nav=3177:3177:focused) -- specifically [here](http://upsource.soramitsu.co.jp/iroha/file/e540d390b31c38320a8953b231a69ff561a02ce4/irohad/main/impl/raw_block_insertion.cpp?nav=1445:1445:focused) a `SEGFAULT` happens, because `factory_` is `nullptr`.

PR #610, opened 2017-10-05
Add config module with gflags implementation

Warchant:
- [x] Add Config module, which allows users to access config through convenient interface.
- [x] Add IROHA_VERSION cmake variable, which helps to set up current build version during compile time. Then, it is possible to query it: `irohad --version`
    ```CMake
    cmake .. -DCMAKE_BUILD_TYPE=Release -DIROHA_VERSION="v0.95-build-${CI_BUILD_NUMBER}-${CI_COMMIT}"
    ```
- [x] Refactor code to use Config with gflags implementation.
    ```bash
    iroha@9d081052eb9a:/opt/iroha/build/bin$ ./irohad --helpshort
    irohad: 
    
    USING ENVIRONMENT VARIABLES: 
    	$ export FLAGS_flag1=val
    	$ export FLAGS_flag2=val
    	$ ./irohad -fromenv=flag1,flag2
    
    USING CLI: 
    	$ ./irohad -flag1=val -flag2=val
    
    USING FLAG FILE: 
    	$ cat /tmp/flags
    	-flag1=val
    	-flag2=val
    	$ ./irohad -flagfile=/tmp/flags
    ```

- [x] Fix `pb_query_factory` unit test
- [x] Add implementation for `initPeer()` (now `tx_intergration_pipeline_test` fails because of intentional `BOOST_ASSERT(false); ` in `initPeer()`
- [x] Fix tx_pipeline_integration_test

Main changes were made, so this PR is in ready-for-review state.


irohad/main/config/impl/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+add_library(irohad-flags
```
kamilsa: Please, move all this logic to irohad/main/config/CMakeLists.txt, so that there are no ../flags.hpp

irohad/main/config/impl/gflags_config.cpp
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "gflags_config.hpp"
+#include <sstream>
```
kamilsa: why it is here?

libs/util/impl/validator.hpp
```diff
@@ -0,0 +1,38 @@
+/**
```
kamilsa: rename class to validator impl

libs/util/string.hpp
```diff
@@ -0,0 +1,59 @@
+/**
```
kamilsa: don't like naming. When people write writing #include string.hpp looks like including system library

libs/util/string.hpp
```diff
@@ -0,0 +1,60 @@
...
+      /**
+       * Parse arbitrary typeT: string, int, long, double... from
+       * string-encoded format to T.
+       * @tparam T
+       * @throws std::exception if string can not be parsed.
+       * @param s any string
+       * @return
+       */
+      template <typename T>
+      T from_string(const std::string &s) {
```
kamilsa: Why not to use functions from std like stoi, stoul

test/module/irohad/ametsuchi/ametsuchi_fixture.hpp
```diff
@@ -69,28 +58,31 @@ DROP TABLE IF EXISTS peer;
...
-        iroha::remove_all(block_store_path);
+        iroha::remove_all(config_->blockStorage().path);
       }
 
-      std::string pgopt_ =
-          "host=localhost port=5432 user=postgres password=mysecretpassword";
-
-      std::string redishost_ = "localhost";
-      size_t redisport_ = 6379;
+      std::unique_ptr<iroha::config::Config> config_;
```
kamilsa: Can not find where `config_` is initialized. I think that the reason why `ametsuchi_test` fails. Please fix

libs/util/string.hpp
Warchant: Because `from_string` is generic method, which works for any type (double, float, uint8_t, int8_t, etc).
It is easy to generalize code which uses this method.

l4l:
`IROHA_VERSION` do nothing and it isn't a good idea to set it somehow (except maybe `git-checkout`)
So what's the point?

irohad/main/config/config.hpp
```diff
@@ -0,0 +1,191 @@
...
+      /**
+       * Reload config. Used to fill config with parsed data.
+       * Derived classes define a load strategy.
+       *
+       * When config is loaded, this->loaded_ should be setted to true.
+       */
+      virtual void load() = 0;
+
+      /// accessor for blockchain options
+      inline const BlockchainOptions &options() const noexcept;
```
l4l: Why you didn't combine def&decl?

irohad/main/config/flags.hpp
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * However, it is not recommended to use flags. Use iroha::config::Config.
+ */
+
+/// -new_ledger
+DECLARE_bool(create);
+DECLARE_string(genesis_block);
+
+/// Redis
+// redis host
```
l4l: remove this meaningless comments plz :( (section are fine)

irohad/main/config/impl/flags.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+  return true;
+}
+
+/// -new_ledger
+DEFINE_bool(create, false, "Create new ledger with given genesis block");
+
+DEFINE_string(genesis_block, "genesis.json", "Define path to genesis block");
+DEFINE_validator(genesis_block,
+                 [](const char *flagname, const std::string &path) {
```
l4l: why not exctracting to the function as well?

irohad/main/config/impl/flags.cpp
```diff
@@ -0,0 +1,132 @@
...
+                 });
+
+/// Redis
+// redis host
+DEFINE_string(redis_host,
+              "localhost",
+              "Define host on which redis is listening");
+DEFINE_validator(redis_host, &host_validator);
+
+// redis port
```
l4l: oh gosh, these comments D:

irohad/main/config/impl/gflags_config.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "main/config/flags.hpp"
+
+namespace iroha {
+  namespace config {
+
+    GFlagsConfig::GFlagsConfig() { load(); }
+
+    void GFlagsConfig::load() {
```
l4l: What's the point of moving code of constructor to `load`? `FLAGS_*` can only set once, I guess, and so the function should be called once

irohad/main/config/config.hpp
```diff
@@ -0,0 +1,191 @@
...
+       */
+      struct BlockchainOptions {
+        std::string genesis_block;  ///< path to the file with "genesis block"
+      };
+
+      /**
+       * Returns true, if config is loaded, false otherwise.
+       * @return
+       */
+      bool loaded() const noexcept;
```
l4l: Use case?

irohad/main/main.cpp
```diff
@@ -0,0 +1,103 @@
...
+using iroha::config::Config;
+using iroha::config::GFlagsConfig;
+using Iroha = Service;
+using iroha::filesystem::util::read_file;
+
+int main(int argc, char *argv[]) {
+  auto log = logger::log("MAIN");
+
+  try {
+#ifdef IROHA_VERSION
```
l4l: this better to be hardcoded into the binary

libs/util/impl/validator.cpp
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "validator.hpp"
+
+namespace iroha {
+
+  bool ValidatorImpl::isHostValid(const std::string &host) const noexcept {
+    return network::validation::is_host_valid(host);
```
l4l: Class for that? Sound as the story of the hashprovider

libs/util/string.hpp
l4l: they'll include `util/string.hpp` in iroha sources or smth like `iroha/util/string.hpp` outsourced
imo that's fine

libs/util/string.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <string>
+
+namespace iroha {
+  namespace string {
+    namespace validation {
+
+      /// @returns true if all chars in string are printable
+      inline bool is_printable(const std::string &s) {
+        return std::all_of(s.begin(), s.end(),
+                           [](const auto c) { return std::isprint(c); });
```
l4l: what's the point of lambda? `std::all_of(s.begin(), s.end(), std::isprint);` should work

clean.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+cd $(dirname $0)/
```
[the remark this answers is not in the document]
Warchant:
```
cd / 
$IROHA_HOME/clean.sh
```

will execute
```
rm -rf ./build
...
```

which may be potentially dangerous. Expected logic from this script is cleaning in project where it  is located.

irohad/main/config/config.hpp
Warchant: In current implementation `load()` is executed in constructor. It is very much possible that descendant of `Config` will not do this. Then, users of `Config` interface must check if config is loaded and react somehow if it is not.

irohad/main/config/config.hpp
Warchant: Cleaner interface.

irohad/main/config/impl/flags.cpp
Warchant: they are there only to find required flag quickly

irohad/main/config/impl/gflags_config.cpp
Warchant: if you change config from your program (change env var, for example), then you might want to reload your env var.

irohad/main/main.cpp
Warchant: It is hardcoded in binary. But is setted dynamically with cmake:
```
cmake .. -DIROHA_VERSION="v0.95-build-${CI_BUILD_NUMBER}-${CI_COMMIT}"
```

useful if we want to build redistributable binaries in circle ci

libs/util/string.hpp
Warchant: did not work for me when I wrote this

irohad/main/config/config.hpp
```diff
@@ -0,0 +1,191 @@
...
+       * @struct Config::Postgres
+       * @brief Postgres config defined here.
+       */
+      struct Postgres : public AuthService {
+        /**
+         * Returns preformatted "options":
+         * host=localhost port=$port user=$user password=$pwd
+         * @return
+         */
+        std::string options() const noexcept {
```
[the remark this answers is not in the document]
Warchant: its name `pg_opt`, so why should I name it `to_string`?

irohad/main/config/impl/gflags_config.cpp
Warchant: Also, you can not overload constructor. 
The purpose of  this class is to provide access to flags.
they way how it exactly executes `load()` should be  hidden in implementation class

libs/util/impl/validator.cpp
Warchant: Now it is not used and can be removed.

test/module/irohad/ametsuchi/ametsuchi_fixture.hpp
Warchant: https://github.com/hyperledger/iroha/blob/96523cb7dddd8c2c76ea388cf443560c16e5f3b5/test/module/irohad/ametsuchi/ametsuchi_fixture.hpp#L42

libs/util/string.hpp
Warchant: Yes, folder `util` works as namespace.

irohad/main/config/flags.hpp
Warchant: They are not meaningless. With these  comments it is easy to find the code related to specific flag.

irohad/main/config/impl/flags.cpp
Warchant: All extracted functions used more than once. This function is  used only once.

irohad/main/config/config.hpp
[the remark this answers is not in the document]
l4l: Because:
1. it's common for stl (represent smth human-readable)
2. it's more obvious/readable

irohad/main/config/flags.hpp
l4l:
```
//redis host
declare redis_host
```
orly? if you are able to find by the comment then you'll be able to find by the var

irohad/main/config/impl/flags.cpp
l4l: words of comments == words of variable

irohad/main/main.cpp
l4l: And someone may build the same source code with the different version. That shouldn't be done

libs/util/string.hpp
l4l: http://en.cppreference.com/w/cpp/string/byte/isprint#Notes
Yeah, seems that's I was wrong

docker/local-develop.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+function dependency(){
+	(command -v $1 > /dev/null || (echo "install $1" && exit 1))
+}
+
+dependency jq     || exit 1
```
l4l: this is the same as:
`(commanv -v jq >/dev/null || (echo "install jq" && exit 1)) || exit 1`
1. It has duplicates for exiting
2. It hard to read (look at the first mistake)

you basically can use `which` command and check those statements in the if statements, e.g (didn't test it):
```bash
if [ -x "$(which jq)" ]; then
    echo install jq
    exit 1
fi
```
or generify this block via your function

docker/local-develop.sh
```diff
@@ -0,0 +1,37 @@
...
+dependency docker || exit 2
+dependency grep   || exit 3
+dependency sed    || exit 4
+
+# usage
+echo "execute 'source $0' to get all env vars for testing"
```
l4l: shouldn't it be `\$0`?

docker/local-develop.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# cleanup
+docker rm -f pg rd 2>/dev/null
+
+# vars
+user=iroha
+password=iroha
```
l4l: it common practice to name bash vars in CAPITAL case, do you mind to fix?

docker/local-develop.sh
```diff
@@ -0,0 +1,37 @@
...
+
+echo "postgres: $pgip:5432 {login: $user, pwd: $password}"
+echo "redis:    $rdip:6379"
+
+export IROHA_POSTGRES_HOST=$pgip
+export IROHA_POSTGRES_PORT=5432
+export IROHA_POSTGRES_USER=$user
+export IROHA_POSTGRES_PASSWORD=$password
+export IROHA_REDIS_HOST=$rgip
+export IROHA_REDIS_PORT=6379
```
l4l: \n :(

test/integration/pipeline/tx_pipeline_integration_test.cpp
```diff
@@ -43,27 +43,41 @@ class TestIrohad : public Application {
...
 class TxPipelineIntegrationTest : public iroha::ametsuchi::AmetsuchiTest {
  public:
   TxPipelineIntegrationTest() { spdlog::set_level(spdlog::level::off); }
 
   void SetUp() override {
     iroha::ametsuchi::AmetsuchiTest::SetUp();
-    irohad = std::make_shared<TestIrohad>(std::move(this->config_));
+    genesis_block =
+        iroha::model::generators::BlockGenerator().generateGenesisBlock(
+            {"0.0.0.0:10001"});
```
l4l:
```cpp
#define HOST "0.0.0.0"
#define port "10001"

...
{HOST ":" PORT}
```
What do you think?

test/module/irohad/ametsuchi/ametsuchi_fixture.hpp
```diff
@@ -18,18 +18,43 @@
...
 
-#include <gtest/gtest.h>
-#include <cpp_redis/cpp_redis>
-#include <pqxx/pqxx>
-#include "config_fake.hpp"
+using namespace std::literals::string_literals;
+using iroha::string::util::from_string;
 
+template <typename T>
+T parseEnv(const char *name, T default_) {
```
l4l: add comments plz
looks like a useful function, maybe move it out from the tests?

test/module/irohad/ametsuchi/ametsuchi_fixture.hpp
```diff
@@ -57,32 +84,26 @@ DROP TABLE IF EXISTS signatory;
 DROP TABLE IF EXISTS peer;
 DROP TABLE IF EXISTS role;
 )";
+        auto config = std::make_unique<AmetsuchiConfig>();
```
l4l: why is it not possible to reuse _config? it will be good if comment is added

docker/local-develop.sh
Warchant: it should print current path to the script, so it should not be `\$0`

docker/local-develop.sh
Warchant: function can not exit with `exit 1` as I understand, so I have to write like this.

You propose to inline function instead of `dependency` name?

test/integration/pipeline/tx_pipeline_integration_test.cpp
Warchant: It is test.

test/module/irohad/ametsuchi/ametsuchi_fixture.hpp
Warchant: Not now. Now we don't need to parse env variables.

docker/local-develop.sh
l4l: Yep, that's it
also remove common tools, e.g sed, grep

docker/local-develop.sh
l4l: yeah, I see

irohad/main/config/config.hpp
```diff
@@ -197,10 +198,11 @@ namespace iroha {
     }
 
     inline keypair_t Config::Cryptography::keypair() const noexcept {
-      return keypair_t{iroha::pubkey_t::from_string(public_key),
-                       iroha::privkey_t::from_string(private_key)};
+      return keypair_t{
+          hexstringToArray<32>(public_key).value(),
+          hexstringToArray<64>(private_key).value()};
```
l4l: maybe it will be better to fix `from_string` method?

Warchant: from_string parses `bytes` and returns key. `hexstringToArray` is temporary solution here, it parses `hexstring`

docker/local-develop.sh
Warchant: No. :ok_man:

PR #613, opened 2017-10-05
Add unsigned postgres checks for precision and transaction count

neewy:
## What is this pull request?
PR adds checks for PostgreSQL tables
   
To guarantee that in database we don't store negative integers
  
## P.S. (optional)
Since we don't have checks for underflow or overflow — this is useless and guarantees only datatype validity for data storage layer. We need to fix corresponding logic in business layer (model classes).


neewy:
Regarded redundant as model classes aren't containing signed types

PR #619, opened 2017-10-08
Fix compilers warnings

l4l:
Checked with:
- gcc 7.2.0
- clang 4.0.1
- clang 5.0.0

Fixes:
- Remove redundant rapidjson include in json_commands_test.cpp
- Add default constructors to pb block and transaction factory
- Remove async redis disconnect from ametsuchi fixture
- Disable warnings for potentially evaluated expressions and missing override in tests

### DIFF
--- a/cmake/functions.cmake
+++ b/cmake/functions.cmake
@@ -10,7 +10,7 @@ function(strictmode target)
   if ((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") OR
   (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") OR
   (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang"))
-    target_compile_options(${target} PRIVATE -Wall -Wpedantic)
+    target_compile_options(${target} PRIVATE -Wall -Wpedantic -Werror -Wno-potentially-evaluated-expression)
   elseif ((CMAKE_CXX_COMPILER_ID STREQUAL "MSVC") OR
   (CMAKE_CXX_COMPILER_ID STREQUAL "Intel"))
     target_compile_options(${target} PRIVATE /W3 /WX)
--- a/cmake/functions.cmake
+++ b/cmake/functions.cmake
@@ -32,6 +32,13 @@ function(addtest test_name SOURCES)
       COMMAND $<TARGET_FILE:${test_name}> ${test_xml_output}
   )
   strictmode(${test_name})
+  if ((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") OR
+  (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") OR
+  (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang"))
+    target_compile_options(${test_name} PRIVATE -Wno-inconsistent-missing-override)
+  else ()
+    message(AUTHOR_WARNING "Unknown compiler: building target ${target} with default options")
+  endif ()
 endfunction()
 
 # Creates benchmark "bench_name", with "SOURCES" (use string as second argument)
--- a/irohad/ametsuchi/impl/redis_block_query.cpp
+++ b/irohad/ametsuchi/impl/redis_block_query.cpp
@@ -72,17 +72,16 @@ namespace iroha {
     std::vector<iroha::model::Block::BlockHeightType>
     RedisBlockQuery::getBlockIds(const std::string &account_id) {
       std::vector<uint64_t> block_ids;
-      client_.lrange(
-          account_id, 0, -1, [this, &block_ids](cpp_redis::reply &reply) {
-            for (const auto &block_reply : reply.as_array()) {
-              const auto &string_reply = block_reply.as_string();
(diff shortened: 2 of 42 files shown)


muratovv:
Add ignoring deprecated methods

neewy:
After review, this pull request could be merged into the codebase

irohad/main/CMakeLists.txt
```diff
@@ -17,20 +17,20 @@
 
 set(EXECUTABLE_OUTPUT_PATH ${PROJECT_BINARY_DIR}/bin)
 
-add_library(server_runner server_runner.cpp)
+addlibrary(server_runner server_runner.cpp)
```
muratovv: What is the difference between `add_library` and `addlibrary`?

irohad/main/CMakeLists.txt
```diff
@@ -55,7 +55,7 @@ target_link_libraries(application
     block_loader_service
     )
 
-add_executable(irohad irohad.cpp)
+addexecutable(irohad irohad.cpp)
```
muratovv: What is difference between `add_executable ` and `addexecutable `?

irohad/model/converters/impl/json_query_factory.cpp
```diff
@@ -78,7 +78,7 @@ namespace iroha {
             | des.String(&Query::creator_account_id, "creator_account_id")
             | des.Uint64(&Query::query_counter, "query_counter")
             | des.Object(&Query::signature, "signature") |
```
muratovv: Broken bind operator code style

libs/common/types.hpp
```diff
@@ -308,6 +310,11 @@ namespace iroha {
   using privkey_t = blob_t<64>;
 
   struct keypair_t {
+    keypair_t() = default;
```
muratovv: What is the reason to add default constructor?

irohad/model/converters/impl/json_query_factory.cpp
l4l: What do you mean?

muratovv: Last bind operator should be going to next line.

libs/common/types.hpp
l4l: There're issues in the monadic bind usage, e.g [here](https://github.com/hyperledger/iroha/blob/75bc558ef1c9533e46eedd357318e7cc74832be0/libs/crypto/keys_manager_impl.cpp#L68)

muratovv: Thank you, I see. But this stuff looks like a not so good design, isn't it?

irohad/model/converters/impl/json_query_factory.cpp
l4l: Ah, I see. Because of lambda, clang-format doesn't want to deal with it

libs/common/types.hpp
l4l: Yeah, definitely
Anyway I'd better rewrite keys loading directly via `operator>>`, but don't really want to mix it up

muratovv: Sure, I guess, we should talk with @Warchant about the status of crypto and maybe create the task for improving the design of it.

irohad/ametsuchi/impl/redis_block_query.cpp
```diff
@@ -72,17 +72,16 @@ namespace iroha {
...
-            for (const auto &block_reply : reply.as_array()) {
-              const auto &string_reply = block_reply.as_string();
-
-              // check if reply is an integer
-              if (isdigit(string_reply.c_str()[0])) {
-                block_ids.push_back(std::stoul(string_reply));
-              }
-            }
-          });
+      client_.lrange(account_id, 0, -1, [&block_ids](cpp_redis::reply &reply) {
```
luckychess: Looks like we can do the same in a fragment just below (line 92).

test/module/irohad/ametsuchi/ametsuchi_fixture.hpp
```diff
@@ -79,7 +79,6 @@ DROP TABLE IF EXISTS role;
         client.connect(redishost_, redisport_);
         client.flushall();
         client.sync_commit();
```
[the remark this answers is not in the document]
lebdron: https://github.com/Cylix/cpp_redis/blob/master/sources/core/client.cpp#L59
Destructor seems to call `disconnect()` anyway.

cmake/functions.cmake
```diff
@@ -32,6 +32,13 @@ function(addtest test_name SOURCES)
       COMMAND $<TARGET_FILE:${test_name}> ${test_xml_output}
   )
   strictmode(${test_name})
+  if ((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") OR
+  (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") OR
+  (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang"))
+    target_compile_options(${test_name} PRIVATE -Wno-inconsistent-missing-override)
```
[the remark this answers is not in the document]
lebdron: This one is only for tests, not for other targets, and is done since gmock does not add override to mock methods.

PR #624, opened 2017-10-12
Replace peer number with peer key in swarm deploy script

lebdron:
`peer_number` was replaced with `keypair_name` in crypto PR.


neewy:
https://github.com/hyperledger/iroha/blob/develop/irohad/main/application.hpp#L58 please check this TODO

PR #633, opened 2017-10-15
Feature/mst/processor

muratovv:
## What is this pull request?
This pr provides effective implementation of MST processor.
   
Mst processor is the key part of MST pipeline because it provides final API for use in the system. Also, this component contains all required dependencies, such as transport, storage, and propagation strategy.

## Details/Features
* Add MST processor
* Add tests for it
* little rework of transport interface

## P.S. (optional)
In this PR I apply the new style of documentation of tests with `given`, `when` and `then` descriptions.

### DIFF
--- a/irohad/multi_sig_transactions/impl/mst_processor.cpp
+++ b/irohad/multi_sig_transactions/impl/mst_processor.cpp
@@ -22,7 +22,6 @@ namespace iroha {
   MstProcessor::MstProcessor() { log_ = logger::log("MstProcessor"); }
 
   void MstProcessor::propagateTransaction(ConstRefTransaction transaction) {
-    std::lock_guard<std::mutex> lock{this->mutex_};
     this->propagateTransactionImpl(transaction);
   }
 
--- a/irohad/multi_sig_transactions/impl/mst_processor.cpp
+++ b/irohad/multi_sig_transactions/impl/mst_processor.cpp
@@ -16,3 +16,23 @@
  */
 
 #include "multi_sig_transactions/mst_processor.hpp"
+
+namespace iroha {
+
+  MstProcessor::MstProcessor() {
+    log_ = logger::log("MstProcessor");
+  }
+
+  void MstProcessor::propagateTransaction(std::shared_ptr<model::Transaction> transaction) {
+    std::lock_guard<std::mutex> lock{this->mutex_};
+    this->propagateTransactionImpl(std::move(transaction));
+  }
+
+  rxcpp::observable<std::shared_ptr<MstState>> MstProcessor::onStateUpdate() const {
+    return this->onStateUpdateImpl();
+  }
+
+  rxcpp::observable<std::shared_ptr<model::Transaction>> MstProcessor::onPreparedTransactions() const {
+    return this->onPreparedTransactionsImpl();
+  }
+} // namespace iroha
--- a/irohad/multi_sig_transactions/impl/mst_processor.cpp
+++ b/irohad/multi_sig_transactions/impl/mst_processor.cpp
@@ -19,20 +19,20 @@
 
 namespace iroha {
 
-  MstProcessor::MstProcessor() {
-    log_ = logger::log("MstProcessor");
-  }
+  MstProcessor::MstProcessor() { log_ = logger::log("MstProcessor"); }
 
-  void MstProcessor::propagateTransaction(std::shared_ptr<model::Transaction> transaction) {
+  void MstProcessor::propagateTransaction(ConstRefTransaction transaction) {
     std::lock_guard<std::mutex> lock{this->mutex_};
     this->propagateTransactionImpl(std::move(transaction));
   }
 
-  rxcpp::observable<std::shared_ptr<MstState>> MstProcessor::onStateUpdate() const {
+  rxcpp::observable<std::shared_ptr<MstState>> MstProcessor::onStateUpdate()
+      const {
     return this->onStateUpdateImpl();
   }
 
-  rxcpp::observable<std::shared_ptr<model::Transaction>> MstProcessor::onPreparedTransactions() const {
+  rxcpp::observable<std::shared_ptr<model::Transaction>>
+  MstProcessor::onPreparedTransactions() const {
     return this->onPreparedTransactionsImpl();
   }
-} // namespace iroha
+}  // namespace iroha
--- a/irohad/multi_sig_transactions/impl/mst_processor.cpp
+++ b/irohad/multi_sig_transactions/impl/mst_processor.cpp
@@ -23,7 +23,7 @@ namespace iroha {
 
   void MstProcessor::propagateTransaction(ConstRefTransaction transaction) {
     std::lock_guard<std::mutex> lock{this->mutex_};
-    this->propagateTransactionImpl(std::move(transaction));
+    this->propagateTransactionImpl(transaction);
   }
 
   rxcpp::observable<std::shared_ptr<MstState>> MstProcessor::onStateUpdate()
@@ -34,5 +34,10 @@ namespace iroha {
   rxcpp::observable<std::shared_ptr<model::Transaction>>
   MstProcessor::onPreparedTransactions() const {
     return this->onPreparedTransactionsImpl();
+  }
+
+  rxcpp::observable<std::shared_ptr<model::Transaction>>
+  MstProcessor::onExpiredTransactions() const {
+    return this->onExpiredTransactionsImpl();
   }
 }  // namespace iroha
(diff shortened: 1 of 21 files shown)


irohad/multi_sig_transactions/impl/mst_processor.cpp
```diff
@@ -16,3 +16,28 @@
  */
 
 #include "multi_sig_transactions/mst_processor.hpp"
+
+namespace iroha {
+
+  MstProcessor::MstProcessor() { log_ = logger::log("MstProcessor"); }
+
+  void MstProcessor::propagateTransaction(ConstRefTransaction transaction) {
+    std::lock_guard<std::mutex> lock{this->mutex_};
```
luckychess: In my opinion, we should add some explanation about why this lock is needed, what shared data is, which threads can have simultaneous access to them and so on. Such description will make reading and especially debugging this part of code a way more easier.

irohad/multi_sig_transactions/mst_processor.hpp
```diff
@@ -18,34 +18,87 @@
...
+     * participants
      * @param transaction - transaction for propagation
+     * General note: implementation of method covered by lock
      */
-    virtual void propagate_transaction(std::shared_ptr<model::Transaction> transaction) = 0;
+    void propagateTransaction(ConstRefTransaction transaction);
 
     /**
      * Prove updating of state for handling status of signing
+     * General note: implementation of method covered by lock
```
luckychess: Looks like a forgotten lock or a wrong commentary. Same with methods just below.

irohad/multi_sig_transactions/mst_time_provider.hpp
```diff
@@ -0,0 +1,39 @@
...
+   * Interface provides current time for iroha
+   */
+  class MstTimeProvider {
+   public:
+
+    /**
+     * Fetching current time in system
+     * @return current time
+     */
+    virtual TimeType getCurrentTime() const = 0;
```
luckychess: I could not find a definition for this. Also looks suitable to move such method into utils or something like that.

irohad/multi_sig_transactions/mst_types.hpp
```diff
@@ -26,5 +26,15 @@ namespace iroha {
...
+
+  class MstState;
+
+  template<typename T>
+  using ConstRefT = const T &;
+
+  using ConstRefTransaction = ConstRefT<TransactionType>;
+  using ConstRefPeer = ConstRefT<iroha::model::Peer>;
+  using ConstRefTime = ConstRefT<TimeType>;
+  using ConstRefState = ConstRefT<MstState>;
```
l4l: This file looks quite odd here. These are more looks like just "types" rather than mst-ones

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+  shared_ptr<MockTimeProvider> time_provider;
+
+ protected:
+  void SetUp() override {
+    transport = make_shared<MockMstTransport>();
+    storage = make_shared<MstStorageStateImpl>(own_peer,
+                                               make_shared<TestCompleter>());
+
+    propagation_strategy = make_shared<MockPropagationStrategy>();
+    EXPECT_CALL(*propagation_strategy, emitter())
```
l4l: is it a good idea to place EXPECT_CALL in the fixture's `SetUp`?

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+  // ---------------------------------| given |---------------------------------
+  auto onStateUpdateWrapper =
+      make_test_subscriber<CallExact>(mst_processor->onStateUpdate(), 0);
+  onStateUpdateWrapper.subscribe();
+
+  auto onPreparedTransactionsWrapper = make_test_subscriber<CallExact>(
+      mst_processor->onPreparedTransactions(), 0);
+  onPreparedTransactionsWrapper.subscribe();
+
+  auto onExpiredTransactionsWrapper = make_test_subscriber<CallExact>(
```
l4l: just an advice: this triplet of make-subscibe calls can be extracted to function that return `std::tuple<.., .., ..>`
then code will be much easier to read

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+ * @when insert not-completed transaction
+ *
+ * @then check that:
+ * state not updated
+ * AND absent prepared transactions
+ * AND absent expired transactions
+ */
+TEST_F(MstProcessorTest, notCompletedTransactionUsecase) {
+  // ---------------------------------| given |---------------------------------
+  auto onStateUpdateWrapper =
```
l4l: Seems that you mixed codestyles

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+ * transport invoked for all peers
+ */
+
+TEST_F(MstProcessorTest, onNewPropagationUsecase) {
+  // ---------------------------------| given |---------------------------------
+  auto quorum = 2;
+  auto timeAfter = 112;
+  mst_processor->propagateTransaction(
+      makeTx("hash", "sign_one", quorum, timeAfter));
+  EXPECT_CALL(*transport, sendState(_, _)).Times(2);
```
l4l: maybe fix to `.Times(quorum)`?

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
muratovv: Sure, if this behaviour suitable for all test scenarios.

irohad/multi_sig_transactions/mst_types.hpp
muratovv: What is your suggestion in this case?

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
muratovv: `make_test_subscriber` return only one thing. Could you explain more details about your issue?

irohad/multi_sig_transactions/mst_types.hpp
l4l: Idk, just comment out the situation.
imo it's better to leave this as is
and then add issue for moving this to _libs/common/types.hpp_ and refactor whole projects

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
muratovv: This constant should be deduced, but not from quorum( this will be very confused), but from the size of `peers` vector.

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
l4l: I mean does this calls performed in the SetUp?
E.g if i write a new test that calls this functions (could i, btw?) it would be automatically be broken?

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
l4l:
```
std::tuple<T1, T2, T3> f(int a, int b, int c) {
  std::tuple t{make_test_subscriber<CallExact>(mst_processor->onStateUpdate(), a),
    make_test_subscriber<CallExact>(mst_processor->onPreparedTransactions(), b),
    make_test_subscriber<CallExact>(mst_processor->onExpiredTransactions(), c)};
  std::get<0>(t).subscribe();
  std::get<1>(t).subscribe();
  std::get<2>(t).subscribe();
}
```
something like this

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
muratovv: You may add explicit `EXPECT_CALL` in your test. Or inherit from the fixture and override  `SetUp`.

irohad/multi_sig_transactions/mst_processor_impl.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef IROHA_MST_PROCESSOR_IMPL_HPP
+#define IROHA_MST_PROCESSOR_IMPL_HPP
+
+#include <memory>
+#include <vector>
```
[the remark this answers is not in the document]
muratovv: Fixed

irohad/multi_sig_transactions/mst_types.hpp
muratovv: Those using useful only for MST. If you move it into a shared file, like types.hpp. All components will depend on MST. Not so good.

l4l: All except `ConstRefState` barely connected with mst

PR #637, opened 2017-10-18
CircleCI improvements

Warchant:
- [x] This PR adds new step to the CircleCI 2.0: `macos-build`.
       Mean time for build with cache: 2.30 min
       Mean time for build without cache: 11 min
- [x] Bugfixes in circle.yml related to sonar-pr and sonar-release steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,9 +129,6 @@ jobs:
 
 
   macos-build:
-    #docker:
-    #  - image: postgres:9.5
-    #  - image: redis:3.2.8
     macos:
       xcode: 9.0.0
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,9 +27,6 @@ jobs:
       - run:
           name: ccache setup
           command: |
-            ccache --version
-            ccache --show-stats
-            ccache --zero-stats
             ccache -F 0
             ccache -M 0
 
@@ -51,6 +48,8 @@ jobs:
           key: build-cache-{{ .Branch }}-{{ epoch }}
           paths:
             - ~/.ccache
+
+      - run: ccache --show-stats
 
       # test and coverage info
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -112,21 +73,21 @@ jobs:
       - run:
           name: execute sonar-scanner to analyze PR
           command: >
-            [ ! -z "$SONAR_TOKEN" ] && \
+            if [ ! -z "$SONAR_TOKEN" ] && \
```
l4l: btw `! -z ...` is same as `-n`

.circleci/config.yml
```diff
@@ -151,33 +112,77 @@ jobs:
...
-  #     - attach_workspace:
-  #         at: /opt/iroha/build
-  #     - run: TODO
-  #         name: todo
-  #         command: echo "I am happy albatross"
-
+  macos-build:
+    #docker:
+    #  - image: postgres:9.5
+    #  - image: redis:3.2.8
```
l4l: why commented?

.circleci/config.yml
```diff
@@ -151,33 +112,77 @@ jobs:
...
+            - external-{{ arch }}-{{ .Branch }}
+            - external-{{ arch }}
+          paths:
+            - external
+      - run: ccache --show-stats
+      - run:
+          name: build
+          command: |
+            cmake -H. -Bbuild
+            cmake --build build -- -j$(sysctl -n hw.ncpu)
```
l4l: Afair we have some env variable for job number, don't we?

.circleci/config.yml
```diff
@@ -151,33 +112,77 @@ jobs:
...
+          keys:
+            - external-{{ arch }}-{{ .Branch }}
+            - external-{{ arch }}
+          paths:
+            - external
+      - run: ccache --show-stats
+      - run:
+          name: build
+          command: |
+            cmake -H. -Bbuild
```
l4l: why not `cmake -H$IROHA_HOME -B$IROHA_BUILD`?

.circleci/config.yml
Warchant: I've read somewhere that it is not.

.circleci/config.yml
Warchant: Should I remove comments?

.circleci/config.yml
Warchant: there are possibly no $IROHA_HOME or $IROHA_BUILD

.circleci/config.yml
Warchant: in mac? are you sure?

.circleci/config.yml
l4l: I mean why those containers isn't launched?

.circleci/config.yml
l4l: In circleci, I guess it should share env vars to every arch, don't it?

Warchant: it does not

.circleci/config.yml
Warchant: MacOS builds for circle ci 2.0 are bugged. When I uncomment it, circle ci thinks that I want to start linux, not mac... Everything fails.

PR #641, opened 2017-10-18
Replace SHA3-512 with SHA-512 for ed25519; update examples in docs

lebdron:
[Ed25519](https://en.wikipedia.org/wiki/EdDSA#Ed25519) is defined using SHA-512. This PR replaces SHA3-512 with standard hashing algorithm for signatures.
- Remove obsolete signature tests
- Update examples for new hash algorithm


libs/crypto/ed25519/keypair.c
```diff
@@ -1,12 +1,12 @@
...
-#include "sha3.h"
+#include "sha512.h"
 #include "ge.h"
 
 
 void ed25519_create_keypair(unsigned char *public_key, unsigned char *private_key, const unsigned char *seed) {
     ge_p3 A;
 
-    sha3_512(seed, 32, private_key);
+    sha512(seed, 32, private_key);
```
l4l: does size of private_key 64 bytes?

libs/crypto/ed25519/add_scalar.c
```diff
@@ -1,21 +1,21 @@
...
+#include "sha512.h"
 
 
 /* see http://crypto.stackexchange.com/a/6215/4697 */
 void ed25519_add_scalar(unsigned char *public_key, unsigned char *private_key, const unsigned char *scalar) {
     const unsigned char SC_1[32] = {1}; /* scalar with value 1 */
-
-    unsigned char n[32];
+    
+    unsigned char n[32];
```
muratovv: Confused end space.

test/module/libs/crypto/signature_test.cpp
```diff
@@ -44,78 +44,3 @@ TEST(Signature, PrintkeyPair) {
   ASSERT_NO_THROW({ std::cout << keypair.pubkey.to_base64() << std::endl; });
   ASSERT_NO_THROW({ std::cout << keypair.privkey.to_base64() << std::endl; });
 }
-
-TEST(Signature, generatedByAndroid) {
```
muratovv: Better to add a signature test for cryptography. We should add tests for crypto or task about it.

libs/crypto/ed25519/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_library(ed25519 STATIC
+add_library(ed25519
```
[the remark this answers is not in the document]
lebdron: https://cmake.org/cmake/help/v3.0/command/add_library.html
> If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is ON.

Other project libraries also use this implicit static definition.

libs/crypto/ed25519/keypair.c
lebdron: Yes. The number here refers to seed length, and seed is 32 bytes.

test/module/libs/crypto/signature_test.cpp
lebdron: Currently there is test for both sign and verify

https://github.com/hyperledger/iroha/blob/17af47bdda507c7999670699a9f1abc089887b59/test/module/libs/crypto/signature_test.cpp#L30

Additional tests can be added in another task.

PR #643, opened 2017-10-18
Add gossip strategy

l4l:
Pull requests should have these details:

This pull request contains basic gossip strategy for propagation (basically emitting next peers for gossiping)

So far there's no comments, I'm going to add them after.
Also there's no explicit dep for current time, that also should be fixed

PR #648, opened 2017-10-21
Add gossip strategy

l4l:
Squashed and fixed #643

### DIFF
--- a/irohad/multi_sig_transactions/gossip_propagation_strategy.hpp
+++ b/irohad/multi_sig_transactions/gossip_propagation_strategy.hpp
@@ -18,8 +18,8 @@
 #ifndef IROHA_GOSSIP_PROPAGATION_STRATEGY_HPP
 #define IROHA_GOSSIP_PROPAGATION_STRATEGY_HPP
 
-#include <chrono>
 #include <boost/optional.hpp>
+#include <chrono>
 #include <rxcpp/rx.hpp>
 #include "ametsuchi/peer_query.hpp"
 #include "multi_sig_transactions/mst_propagation_strategy.hpp"
--- a/irohad/multi_sig_transactions/gossip_propagation_strategy.hpp
+++ b/irohad/multi_sig_transactions/gossip_propagation_strategy.hpp
@@ -19,7 +19,7 @@
 #define IROHA_GOSSIP_PROPAGATION_STRATEGY_HPP
 
 #include <chrono>
-#include <nonstd/optional.hpp>
+#include <boost/optional.hpp>
 #include <rxcpp/rx.hpp>
 #include "ametsuchi/peer_query.hpp"
 #include "multi_sig_transactions/mst_propagation_strategy.hpp"
@@ -34,7 +34,7 @@ namespace iroha {
   class GossipPropagationStrategy : public PropagationStrategy {
    public:
     using PeerProvider = std::shared_ptr<ametsuchi::PeerQuery>;
-    using OptPeer = nonstd::optional<PropagationData::value_type>;
+    using OptPeer = boost::optional<PropagationData::value_type>;
     /**
      * Initialize strategy with
      * @param query is a provider of peer list
--- a/irohad/multi_sig_transactions/gossip_propagation_strategy.hpp
+++ b/irohad/multi_sig_transactions/gossip_propagation_strategy.hpp
@@ -28,8 +28,9 @@ namespace iroha {
 
   /**
    * This class provides strategy for propagation states in network
-   * Choose at exactly (or nothing iff provider empty) amount of peers
-   * each period of time
+   * Emits exactly (or zero iff provider is empty) amount of peers
+   * at some period
+   * note: it can be inconsistent with the peer provider
    */
   class GossipPropagationStrategy : public PropagationStrategy {
    public:
@@ -38,7 +39,7 @@ namespace iroha {
     /**
      * Initialize strategy with
      * @param query is a provider of peer list
-     * @param period of emitting to observable in ms
+     * @param period of emitting data in ms
      * @param amount of peers emitted per once
      */
     GossipPropagationStrategy(PeerProvider query,
@@ -51,9 +52,6 @@ namespace iroha {
 
     // --------------------------| end override |---------------------------
    private:
-    /**
-     * Peer provider
-     */
     PeerProvider query;
     /**
      * Cache of peer provider's data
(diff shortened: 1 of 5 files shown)


irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
```diff
@@ -0,0 +1,75 @@
...
+                      }
+                      return vec;
+                    })) {}
+
+  rxcpp::observable<PropagationData> GossipPropagationStrategy::emitter() {
+    return emitent;
+  }
+
+  bool GossipPropagationStrategy::initQueue() {
+    // Grub peers from the storage
```
lebdron: Retrieve peers

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
```diff
@@ -0,0 +1,75 @@
...
+    auto data_opt = query->getLedgerPeers();
+    if (!data_opt || data_opt->size() == 0) {
```
lebdron: It seems possible to use bind operator, which will return `bool`.

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    // Generate random ordered vector of unique numbers
+    non_visited.resize(last_data.size());
+    std::iota(non_visited.begin(), non_visited.end(), 0);
+    std::random_shuffle(non_visited.begin(), non_visited.end());
+    return true;
+  }
+
+  OptPeer GossipPropagationStrategy::visit() {
+    if (non_visited.empty() && !initQueue()) {
```
lebdron: Use alternative tokens for logical operators. http://en.cppreference.com/w/cpp/language/operator_alternative

test/module/irohad/multi_sig_transactions/gossip_propagation_strategy_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+ * limitations under the License.
+ */
+
+#include "multi_sig_transactions/gossip_propagation_strategy.hpp"
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <algorithm>
+#include <iterator>
+#include <memory>
+#include <model/peer.hpp>
```
lebdron: `""` for project includes. Apply https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes to all files, so it is easier to separate different includes.

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
```diff
@@ -0,0 +1,75 @@
...
+  using PeerProvider = GossipPropagationStrategy::PeerProvider;
+  using std::chrono::steady_clock;
+
+  GossipPropagationStrategy::GossipPropagationStrategy(
+      PeerProvider query, std::chrono::milliseconds period, uint32_t amount)
+      : query(query),
+        emitent(rxcpp::observable<>::interval(steady_clock::now(), period)
+                    .map([this, amount](int) {
+                      PropagationData vec;
+                      for (uint32_t i = 0; i < amount; i++) {
```
lebdron: It would be great to rewrite this in rx with `flat_map`.

irohad/multi_sig_transactions/gossip_propagation_strategy.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include <nonstd/optional.hpp>
+#include <rxcpp/rx.hpp>
+#include "ametsuchi/peer_query.hpp"
+#include "multi_sig_transactions/mst_propagation_strategy.hpp"
+
+namespace iroha {
+
+  /**
+   * This class provides strategy for propagation states in network
+   * Choose unique amount of peers each period of time
```
lebdron: It always chooses at most `amount` of peers, so the number is not always unique. Maybe `Choose at most n peers` sounds better.

l4l: yep, thanks

l4l: Good point, thanks

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
l4l: typo with grAb
but retrieve seems better

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
l4l: What's the differ?

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
l4l: Why? Let's enforce its usage upon discussion and standardization

test/module/irohad/multi_sig_transactions/gossip_propagation_strategy_test.cpp
l4l: Yep, thanks

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
l4l: Not that easy in current implementation. Anyway imo `map` is fine for this, not a big problem to destruct `int`
Can mark with todo if you'd like

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
```diff
@@ -46,26 +46,25 @@ namespace iroha {
...
-    std::random_shuffle(non_visited.begin(), non_visited.end());
-    return true;
+    return static_cast<bool>(
+        query->getLedgerPeers() |
+            [this](PropagationData data) -> nonstd::optional<int> {
+          last_data = data;
+          if (data.size() == 0) return {};
+          this->non_visited.resize(this->last_data.size());
+          std::iota(this->non_visited.begin(), this->non_visited.end(), 0);
+          std::random_shuffle(this->non_visited.begin(),
```
motxx: `std::random_shuffle` is deprecated in C++14 and removed in C++17. `std::shuffle` can be used instead.
http://en.cppreference.com/w/cpp/algorithm/random_shuffle

l4l: Nice point, thanks!

PR #656, opened 2017-10-26
Fix check consistency method

kamilsa:
## What is this pull request?
Describe simply what this pull request is about.

We had a bug of deleting all blocks but the first one, when iroha initializes from the "hot" state, when ledger has several blocks commited
   
Describe features, what this pull request resolves, and reasons for this pull request.

This PR fixes aforementioned bug and adds tests to check if block store does not delete blocks when it initializes
  
Take a look at new test case in flat file test

## Details/Features
The problem was with `check_consistency` method, which wrongly removed consistent blocks

### DIFF
--- a/irohad/ametsuchi/impl/flat_file/flat_file.cpp
+++ b/irohad/ametsuchi/impl/flat_file/flat_file.cpp
@@ -100,19 +100,13 @@ nonstd::optional<Identifier> check_consistency(const std::string &dump_dir) {
   }
 
   auto n = static_cast<uint32_t>(status);
-  tmp_id++;
+  tmp_id = 2;
 
-  // get last available identifier
-  auto last = tmp_id;
-  for (auto i = 2u; i < n; ++i) {
-    if (id_to_name(tmp_id) != namelist[i]->d_name) {
-      for (auto j = i; j < n; ++j) {
-        remove(dump_dir, namelist[j]->d_name);
-      }
-      break;
-    }
-    tmp_id++;
-    last = name_to_id(namelist[i]->d_name);
+  while (tmp_id < n and id_to_name(tmp_id - 1) == namelist[tmp_id]->d_name) {
+    ++tmp_id;
+  }
+  for (auto j = tmp_id; j < n; ++j) {
+    remove(dump_dir, namelist[j]->d_name);
   }
 
   for (auto j = 0u; j < n; ++j) {
@@ -120,7 +114,7 @@ nonstd::optional<Identifier> check_consistency(const std::string &dump_dir) {
   }
   free(namelist);
 
-  return last;
+  return tmp_id - 2;
 }
 
 /**
--- a/irohad/ametsuchi/impl/flat_file/flat_file.cpp
+++ b/irohad/ametsuchi/impl/flat_file/flat_file.cpp
@@ -102,7 +102,7 @@ nonstd::optional<Identifier> check_consistency(const std::string &dump_dir) {
   auto n = static_cast<uint32_t>(status);
   tmp_id++;
 
-  // get last available identificator
+  // get last available identifier
   auto last = tmp_id;
   for (auto i = 2u; i < n; ++i) {
     if (id_to_name(tmp_id) != namelist[i]->d_name) {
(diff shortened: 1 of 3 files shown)


irohad/ametsuchi/impl/flat_file/flat_file.cpp
```diff
@@ -102,22 +104,26 @@ nonstd::optional<Identifier> check_consistency(const std::string &dump_dir) {
 
   auto n = static_cast<uint32_t>(status);
   tmp_id++;
+
+  // get last available identificator
```
muratovv: Better to write "last available identifier" because there is describes invariant for the variable, not an action.

test/module/irohad/ametsuchi/flat_file_test.cpp
```diff
@@ -101,3 +101,24 @@ TEST_F(BlStore_Test, BlockStoreWhenAbsentFolder) {
   }
   rmdir(target_path.c_str());
 }
+
+TEST_F(BlStore_Test, BlockStoreInitializationWhenPreviousStorageExist){
```
muratovv: Add `given, when, then` description of test

test/module/irohad/ametsuchi/flat_file_test.cpp
```diff
@@ -101,3 +101,24 @@ TEST_F(BlStore_Test, BlockStoreWhenAbsentFolder) {
...
+  auto id2 = 2u;
+  bl_store1->add(id2, block2);
+
+  // create second block storage from the same folder
+  auto bl_store2 = FlatFile::create(block_store_path);
+  ASSERT_TRUE(bl_store2);
+
+  // check that last ids of both block storages are the same
+  ASSERT_EQ(bl_store1->last_id(), bl_store2->last_id());
+}
```
muratovv: Is there required for remove storages of `bl_store1` and `bl_store2`  here?

muratovv: If it is true, make fixture for the test.

kamilsa: The block stores' content is removed in tear down

test/module/irohad/ametsuchi/flat_file_test.cpp
```diff
@@ -101,3 +101,29 @@ TEST_F(BlStore_Test, BlockStoreWhenAbsentFolder) {
...
+ * @then new block storage has all blocks from the folder
+ */
+TEST_F(BlStore_Test, BlockStoreInitializationFromNonemptyFolder){
+  auto bl_store1 = FlatFile::create(block_store_path);
+  ASSERT_TRUE(bl_store1);
+
+  // Add two blocks to storage
+  std::vector<uint8_t > block1(1000, 5);
+  auto id1 = 1u;
+  bl_store1->add(id1, block1);
```
l4l: as you use previous vars once you can inline call:
`b1_store1->add(1u, std::vector<uint8_t>(1000, 5));`
same for the following

irohad/ametsuchi/impl/flat_file/flat_file.cpp
```diff
@@ -102,22 +101,26 @@ nonstd::optional<Identifier> check_consistency(const std::string &dump_dir) {
 
   auto n = static_cast<uint32_t>(status);
   tmp_id++;
+
+  // get last available identifier
+  auto last = tmp_id;
```
lebdron: This loop seems more clear if you do it in two loops: find non-consecutive file name, and remove files after it, including itself.
```C++
Identifier tmp_id = 2;
while (tmp_id < n and id_to_name(tmp_id - 1) == namelist[tmp_id]->d_name) {
  ++tmp_id;
}
for (auto j = tmp_id; j < n; ++j) {
  remove(dump_dir, namelist[j]->d_name);
}
...
return tmp_id - 2;
```

kamilsa: why start with identifier equals 2?

kamilsa: ah, understand

PR #658, opened 2017-10-27
Update SQL schema: Add json data to account

grimadas:
## What is this pull request?
Add json data to account and update sql schema 
   
To allow storage of account personal info we need add JSONb field in account table.
  
- Add new test in wsv_command_query

## Details/Features
 - By default account is created with empty JSON data

### DIFF
--- a/irohad/ametsuchi/impl/postgres_wsv_command.cpp
+++ b/irohad/ametsuchi/impl/postgres_wsv_command.cpp
@@ -113,13 +113,19 @@ namespace iroha {
         transaction_.exec(
             "INSERT INTO account(\n"
             "            account_id, domain_id, quorum, "
-            "transaction_count \n"
+            "transaction_count, data \n"
             "            )\n"
-            "    VALUES (" +
-            transaction_.quote(account.account_id) + ", " +
-            transaction_.quote(account.domain_id) + ", " +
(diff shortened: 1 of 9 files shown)


irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -113,13 +113,19 @@ namespace iroha {
...
-            transaction_.quote(account.quorum) + ", " +
-            /*account.transaction_count*/ transaction_.quote(0) + ");");
+            "    VALUES ("
+            + transaction_.quote(account.account_id)
+            + ", "
+            + transaction_.quote(account.domain_id)
+            + ", "
+            + transaction_.quote(account.quorum)
+            + ", "
+            + transaction_.quote(0)
```
kamilsa: Add comment that this is the transaction count

irohad/model/generators/impl/command_generator.cpp
```diff
@@ -57,7 +57,7 @@ namespace iroha {
           const std::string &domain_id,
           const pubkey_t &key) {
         return generateCommand<CreateAccount>(
-            account_name, domain_id, key);
+            account_name, domain_id, key, R"({})");
```
lebdron: Raw string is not required here.

PR #664, opened 2017-10-30
Add dockerize and release steps

Warchant:
What is done:
- [x] Add dockerize step for circle ci
- [x] Add release step (triggered  when pushed to master)
- [x] Now it is possible define versioning policy in circle ci env var. 

Currently it looks like
```
${CIRCLE_TAG:-MANUAL_VERSION}-build-$CIRCLE_BUILD_NUM-0x$CIRCLE_SHA1
```

p.s. MANUAL_VERSION is set if CIRCLE_TAG is empty.


.circleci/config.yml
```diff
@@ -29,11 +29,15 @@ jobs:
             ccache --max-size=1G
       - run:
           name: cmake
-          command: cmake -DCOVERAGE=ON -H$IROHA_HOME -B$IROHA_BUILD
+          command: cmake -DCOVERAGE=ON -H$IROHA_HOME -B$IROHA_BUILD -DPACKAGE_DEB=ON -DCMAKE_BUILD_TYPE=Release
```
Warchant: >-DCMAKE_BUILD_TYPE=Release

I just realized, coverage info will not be collected, because of Release build type. We have to do 2 builds, one for debug and one for release.

.circleci/config.yml
```diff
@@ -48,12 +52,54 @@ jobs:
...
             - compile_commands.json
             - reports
+            - iroha.deb
+            - release_dockerfile.tmp
+
+  dockerize:
+    <<: *defaults
+    docker:
+      - image: hyperledger/iroha-docker-develop
+      - image: docker:17.05.0-ce-git
```
Warchant: image

Nice solution

Warchant:
I added new step `build-linux-release`, which produces `.deb`, after which `dockerize` step is executed.

.circleci/config.yml
```diff
@@ -56,6 +55,101 @@ jobs:
...
+            cmake --build $IROHA_BUILD --target package -- -j4
+      - save_cache:
+          key: build-cache-release-{{ arch }}-{{ .Branch }}-{{ epoch }}
+          paths:
+            - ~/.ccache
+      - run:
+          name: copy deb file
+          command: |
+            cd $IROHA_BUILD
+            cp *.deb ./iroha.deb
```
l4l: what if `*.deb` matches on >2 files?

.circleci/config.yml
```diff
@@ -56,6 +55,101 @@ jobs:
...
+          name: Copy deb
+          command: |
+            cp $IROHA_BUILD/iroha.deb $IROHA_HOME/docker/iroha.deb
+      - run:
+          name: Copy dockerfile
+          command: cp $IROHA_HOME/build/release_dockerfile.tmp $IROHA_HOME/docker
+      - run:
+          name: Docker login
+          command: |
+              docker login -u $DOCKER_USER -p "$DOCKER_PASS"
```
l4l: i think it's better wrap `$DOCKER_USER` as well

docker/release_dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN apt-get install libc-ares-dev -y
+
+#Install iroha
+RUN dpkg -i /tmp/iroha.deb || true //continue executing even if command fails
+
+#Install iroha dependecies
+RUN apt-get install -y -f
+
+#Insall iroha again
+RUN dpkg -i /tmp/iroha.deb
```
l4l: 1. new line missing
2. afaik after fixing deps via apt-get there's no need to run installation again, isn't it?

docker/release_dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ubuntu:16.04
+
+ENV IROHA_HOME /opt/iroha
+ENV IROHA_BUILD /opt/iroha/build
+
+COPY iroha.deb /tmp/iroha.deb
+RUN apt-get update
+
+RUN apt-get install libc-ares-dev -y
```
l4l: merge with the previous

docker/release_dockerfile
Warchant: first you try to dpkg install, it says that it can't install iroha because of missing dependencies.
then you write `apt install -f`, it installs all deps.
then you install with dpkg again.

x3medima17:
|`.circleci/config.yml:6`|
|---|
test

docker/release_dockerfile
l4l: Look at [this](https://unix.stackexchange.com/a/159103)

.circleci/config.yml
Warchant: On the CI we can guarantee that there will be only one deb, and one tar.gz file.

.circleci/config.yml
```diff
@@ -171,16 +272,28 @@ jobs:
           key: external-{{ arch }}-{{ .Branch }}-{{ epoch }}
           paths:
             - external
-
       - run: ccache --show-stats
+      - run:
+          name: rename artifacts
+          command: |
+            cp build/iroha-*.tar.gz build/iroha-darwin.tar.gz
```
lebdron: Maybe use specific filename here instead of wildcard?

.circleci/config.yml
```diff
@@ -56,6 +55,101 @@ jobs:
...
+            cmake --build $IROHA_BUILD --target package -- -j4
+      - save_cache:
+          key: build-cache-release-{{ arch }}-{{ .Branch }}-{{ epoch }}
+          paths:
+            - ~/.ccache
+      - run:
+          name: copy deb file
+          command: |
+            cd $IROHA_BUILD
+            cp iroha-*.deb ./iroha.deb
```
lebdron: How about using specific filename instead of wildcard?

.circleci/config.yml
```diff
@@ -56,6 +55,101 @@ jobs:
...
+            cp iroha-*.tar.gz ./iroha.tar.gz
+            # if IROHA_VERSION is not set, then CIRCLE_BUILD_NUM
+            echo ${IROHA_VERSION:-$CIRCLE_BUILD_NUM} > version.txt
```
lebdron: Same, why build number instead of commit hash?

docker/release_dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN apt-get update;\
+    apt-get upgrade; \
+    apt-get install -y \
+        libc-ares-dev \
+        gdebi-core
+
+#Install iroha
+RUN gdebi /tmp/iroha.deb
+
+CMD ["bash"]
```
lebdron: Consider such container: https://github.com/kylemanna/docker-openvpn#quick-start
We can also run `irohad` here and attach required files with volumes before running this container, so there is no need to manually run `irohad`.

.circleci/config.yml
```diff
@@ -56,6 +55,101 @@ jobs:
...
+
+
+
+  dockerize:
+    <<: *defaults
+    docker:
+      - image: hyperledger/iroha-docker-develop
+      - image: docker:17.05.0-ce-git
+    steps:
+      - checkout
```
lebdron: Checkout clones the project, so there is no need to make a copy of `release_dockerfile` to `build/release_dockerfile.tmp`.

.circleci/config.yml
Warchant: it is not possible, because name looks like 
`iroha-0xd4cce615d9aaa8b96dd938b5ab67062f3ccfab24-Linux.deb`

.circleci/config.yml
```diff
@@ -159,10 +253,17 @@ jobs:
             - external
       - run: ccache --show-stats
       - run:
-          name: build
+          name: cmake
+          command: >
+            cmake -H. -Bbuild \
```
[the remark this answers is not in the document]
Warchant: No, because paths are different

.circleci/config.yml
Warchant: name contains `version`, and we use git commit hash as a version, so it is easier to just use asterisk.
There are no name collisions, since it is built on CI, which creates clean machine for every build

.circleci/config.yml
```diff
@@ -171,16 +272,28 @@ jobs:
...
+          command: |
+            cp build/iroha-*.tar.gz build/iroha-darwin.tar.gz
+            # if IROHA_VERSION is not set, then CIRCLE_BUILD_NUM
+            echo ${IROHA_VERSION:-$CIRCLE_BUILD_NUM} > build/version.txt
+      - store_artifacts:
+          path: build/iroha-darwin.tar.gz
+          destination: iroha-darwin.tar.gz
+      - store_artifacts:
+          path: build/version.txt
+          destination: version.txt
```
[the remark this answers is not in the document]
Warchant: Yes. Builds are different.

.circleci/config.yml
lebdron: Use first 7 symbols, just like github?

.circleci/config.yml
lebdron: This intent might be more clear for the reader with specifying variable explicitly.

.circleci/config.yml
```diff
@@ -171,16 +272,28 @@ jobs:
...
      
 
 workflows:
   version: 2
   full_pipeline:
     jobs:
-      # macos-build/build -> sonar -> dockerize
-      - macos-build
+      # build-macos-release + build -> sonar(pr/release)
+      - build-macos-release
```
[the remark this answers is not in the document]
Warchant: It is for time-efficiency.
You should build iroha for linux both in debug (for coverage) and release (for dockerization). 
But, for macos you have to build only once (either debug/release is fine). Since we now collect build artifacts after every build, it is logical to build in release.

Now for every build:
- linux (debug) + coverage + analysis steps
- macos (release) + store artifacts (if you do not want to build branch by yourself, but want to use iroha binaries) steps

For release builds (push to develop, master, tag):
- linux (debug) + coverage + analysis
- linux (release) + dockerization + store artifacts
- macos (release) + store artifacts

docker/release_dockerfile
Warchant: Now we do not have opportunity to `just run` irohad. After config will be implemented this will be changed.

.circleci/config.yml
Warchant: we don't use IROHA_VERSION yet, because we do not have versioning policy. Now dummy commit hash is used. 

If we implement versioning policy, now we do not need to change circle files. We just specify env var in circle ci.

.circleci/config.yml
Warchant: resolved

.circleci/config.yml
```diff
@@ -56,6 +55,101 @@ jobs:
...
+          command: cp $IROHA_HOME/build/release_dockerfile.tmp $IROHA_HOME/docker
+      - run:
+          name: Docker login
+          command: |
+              docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
+      - run:
+          name: Build and push iroha image
+          command: |
+              tag=test
+              if [ "$CIRCLE_BRANCH" == "develop" ]; then
```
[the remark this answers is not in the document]
Warchant: wontfix

docker/release/Dockerfile
```diff
@@ -2,15 +2,19 @@ FROM ubuntu:16.04
...
     apt-get install -y \
         libc-ares-dev \
         gdebi-core
 
 #Install iroha
-RUN gdebi /tmp/iroha.deb
+RUN gdebi --non-interactive /tmp/iroha.deb
 
-CMD ["bash"]
+WORKDIR /op/iroha
```
l4l: typo here

docker/release/Dockerfile
```diff
@@ -2,15 +2,19 @@ FROM ubuntu:16.04
 
 ENV IROHA_HOME /opt/iroha
 ENV IROHA_BUILD /opt/iroha/build
+ARG node
 
 COPY iroha.deb /tmp/iroha.deb
+COPY entrypoint.sh /entrypoint.sh
```
l4l: Move these `COPY`s below, where it needed

First - exactly before the running gdebi
second - exactly before `ENTRYPOINT`

x3medima17: fixed

docker/release/Dockerfile
x3medima17: fixed

neewy:
@lebdron please review

docker/release/Dockerfile
Warchant: @x3medima17 don't forget to push the code

.circleci/config.yml
```diff
@@ -56,6 +64,103 @@ jobs:
...
+          name: copy deb file
+          command: |
+            IROHA_VERSION=$(eval echo "$IROHA_VERSION")
+            cd $IROHA_BUILD
+            # iroha-$IROHA_VERSION-Linux.{tar.gz, deb}, always matches 1 file
+            cp iroha-*.deb ./iroha.deb
+            cp iroha-*.tar.gz ./iroha.tar.gz
+            # if IROHA_VERSION is not set, then CIRCLE_SHA1
+            echo ${IROHA_VERSION:-$CIRCLE_SHA1} > version.txt
+            cd $IROHA_HOME
```
lebdron: Does this line affect next commands (`store_artifacts`, `persist_to_workspace`)? If not, change directory is not necessary here.

.circleci/config.yml
```diff
@@ -16,26 +15,35 @@ jobs:
       - image: redis:3.2.8
     steps:
       - checkout
+      - run: echo $(eval echo "$IROHA_VERSION")
```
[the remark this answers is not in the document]
Warchant:
```
echo $IROHA_VERSION
${CIRCLE_TAG:-MANUAL_VERSION}-build-$CIRCLE_BUILD_NUM-0x$CIRCLE_SHA1

echo $(eval echo "$IROHA_VERSION")
v0.95-build-6000-0x{hash}
```

.circleci/config.yml
```diff
@@ -56,6 +64,103 @@ jobs:
...
+          command: |
+            cmake --build $IROHA_BUILD --target package -- -j4
+      - save_cache:
+          key: build-cache-release-{{ arch }}-{{ .Branch }}-{{ epoch }}
+          paths:
+            - ~/.ccache
+      - run:
+          name: copy deb file
+          command: |
+            IROHA_VERSION=$(eval echo "$IROHA_VERSION")
```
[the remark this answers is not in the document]
Warchant: `run` spawns new terminal, this IROHA_VERSION will be assigned only there.
Read above why `eval`

.circleci/config.yml
```diff
@@ -56,6 +64,103 @@ jobs:
...
+          command: |
+            cp $IROHA_BUILD/iroha.deb $IROHA_HOME/docker/release/iroha.deb
+      - run:
+          name: Docker login
+          command: |
+              docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
+      - run:
+          name: Build and push iroha image
+          command: |
+              tag=test
```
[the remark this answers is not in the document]
Warchant: it was used for testing I guess

.circleci/config.yml
```diff
@@ -157,12 +262,26 @@ jobs:
...
+            cmake -H. -Bbuild \
+              -DCOVERAGE=OFF \
+              -DTESTING=OFF \
+              -DPACKAGE_TGZ=ON \
+              -DCMAKE_BUILD_TYPE=Release \
+              -DIROHA_VERSION=$(eval echo "$IROHA_VERSION")
+      - run:
+          name: make
+          command: |
+            cmake --build build --target package -- -j$(sysctl -n hw.ncpu)
```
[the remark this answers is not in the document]
Warchant: Experiments show it is safe.

.circleci/config.yml
```diff
@@ -205,3 +338,23 @@ workflows:
...
+
+      # build-linux-release -> dockerize
+      - build-linux-release:
+          filters:
+            branches:
+              only:
+                - develop
+                - master
+            tags:
+              only: /v[\.0-9]+.*/
```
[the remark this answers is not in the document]
Warchant: yes, it is necessary. 
We [already have](https://github.com/hyperledger/iroha/tree/v0.95)  v0.95, and it is good to continue naming like this.

.circleci/config.yml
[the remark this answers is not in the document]
Warchant: image

It literally costs nothing.

.circleci/config.yml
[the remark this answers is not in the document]
Warchant: UPD: it is written like this, because now we have a way to change *how versions are formed* **without changes of circle.yml**

.circleci/config.yml
Warchant: It should not affect, because `run` spawns new terminal. It is safe to remove, I guess.

.circleci/config.yml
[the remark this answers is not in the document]
lebdron: In this case we need reply from @x3medima17.

.circleci/config.yml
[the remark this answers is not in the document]
lebdron: Can we measure time of build step with tar package and without it?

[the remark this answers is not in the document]
l4l: Don't really sure that zipping would cost any significant time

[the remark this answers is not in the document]
Warchant: Yes, it costs like 10 seconds or so, for all `package` steps

.circleci/config.yml
```diff
@@ -167,15 +167,14 @@ jobs:
           name: Build and push iroha image
           command: |
             tag=
-            if [ -z "$CIRCLE_TAG" ]; then
+            if [ ! -z "$CIRCLE_TAG" ]; then
```
l4l: there's `-n` for that

PR #667, opened 2017-11-01
Fix make old model for query

muratovv:
## What is this pull request?
Pr provides fix for `makeOldModel` method for the Query. Provided `toString` methods for `Signable` and `Query`. Also, provided compare method for `Signable`.

- fix implementation for make old model
- add toString methods
- add total compare method for `Signable`
- the absence of tests, because providing only interfaces

### DIFF
--- a/shared_model/interfaces/queries/query.hpp
+++ b/shared_model/interfaces/queries/query.hpp
@@ -86,7 +86,7 @@ namespace shared_model {
         return detail::PrettyStringBuilder()
             .init("Query")
             .append("creatorId", creatorAccountId())
-            .append("queryCounter", std::to_string(this->queryCounter()))
+            .append("queryCounter", std::to_string(queryCounter()))
             .append(Signable::toString())
             .append(boost::apply_visitor(detail::ToStringVisitor(), get()))
             .finalize();
--- a/shared_model/interfaces/queries/query.hpp
+++ b/shared_model/interfaces/queries/query.hpp
@@ -85,26 +85,30 @@ namespace shared_model {
       std::string toString() const override {
         return detail::PrettyStringBuilder()
             .init("Query")
-            .append("creator_id", creatorAccountId())
-            .append("query_counter", std::to_string(this->queryCounter()))
+            .append("creatorId", creatorAccountId())
+            .append("queryCounter", std::to_string(this->queryCounter()))
             .append(Signable::toString())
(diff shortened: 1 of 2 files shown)


shared_model/interfaces/queries/query.hpp
```diff
@@ -65,17 +66,34 @@ namespace shared_model {
...
-        return boost::apply_visitor(
+         auto old_model = boost::apply_visitor(
             detail::OldModelCreatorVisitor<OldModelType *>(), get());
+        old_model->creator_account_id = creatorAccountId();
+        old_model->query_counter = queryCounter();
+        // signature related
+        old_model->created_ts = createdTime();
+        std::for_each(signatures().begin(), signatures().end(), [&old_model](auto &signature_wrapper) {
+          auto old_sig = signature_wrapper->makeOldModel();
+          old_model->signature = *old_sig;
```
luckychess: Old model has only one signature so this behavior will assign a last signature from a new model. I think we should decide, should we change query to be like an old model query and have only a single signature or at least make this assignment explicit.

shared_model/interfaces/queries/query.hpp
```diff
@@ -65,17 +66,34 @@ namespace shared_model {
...
 
       // ------------------------| Primitive override |-------------------------
 
       std::string toString() const override {
-        return boost::apply_visitor(detail::ToStringVisitor(), get());
+        return detail::PrettyStringBuilder()
+            .init("Query")
+            .append("creator_id", creatorAccountId())
+            .append("query_counter", std::to_string(this->queryCounter()))
+            .append(Signable::toString())
```
luckychess: Do you want to call implementation from Primitive or what?

shared_model/interfaces/queries/query.hpp
```diff
@@ -65,17 +66,38 @@ namespace shared_model {
...
+      virtual const QueryCounterType &queryCounter() const = 0;
 
       // ------------------------| Primitive override |-------------------------
 
       std::string toString() const override {
-        return boost::apply_visitor(detail::ToStringVisitor(), get());
+        return detail::PrettyStringBuilder()
+            .init("Query")
+            .append("creatorId", creatorAccountId())
+            .append("queryCounter", std::to_string(this->queryCounter()))
```
lebdron: Use consistent style: `this->` is not required.

PR #677, opened 2017-11-04
Add shared_model tx responses

muratovv:
## What is this pull request?
This PR provides shared model transaction statuses.
   
Transaction statuses are used for notification user about the current state of the transaction in the pipeline. Thus, we need to reflect it in the shared model.
  
## Details/Features
What is provided:
* Tx status - variant container
* Abstract tx status
* Concrete data specializations
* Fix code style in old-model


shared_model/interfaces/transaction_responses/abstract_tx_response.hpp
```diff
@@ -0,0 +1,63 @@
...
+     private:
+      /**
+       * @return string representation of class name
+       */
+      virtual std::string className() const = 0;
+
+      /**
+       * @return old model status
+       */
+      virtual iroha::model::TransactionResponse::Status modelStatus() const = 0;
```
luckychess: I think it could be better named as ```oldModelStatus``` and also marked as deprecated.

shared_model/interfaces/transaction_responses/unknown_tx_response.hpp
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Transaction not found
+     */
+    class UnknownTxResponse : public AbstractTxResponse<UnknownTxResponse> {
+     private:
+      std::string className() const override { return "UnknownTxResponse"; }
+
+      iroha::model::TransactionResponse::Status modelStatus() const {
+        return iroha::model::TransactionResponse::Status::
+            STATELESS_VALIDATION_FAILED;
```
luckychess: May be we could use ```NOT_RECEIVED``` here?

shared_model/interfaces/transaction_responses/tx_response.hpp
```diff
@@ -0,0 +1,83 @@
...
+        : public Hashable<TransactionResponse,
+                          iroha::model::TransactionResponse> {
+     private:
+      /// PolymorphicWrapper shortcut type
+      template <typename Value>
+      using w = detail::PolymorphicWrapper<Value>;
+
+     public:
+      /// Type of variant, that handle all concrete tx responses in the system
+      using ResponseVariantType = boost::variant<w<CommittedTxResponse>,
```
luckychess: It looks like we don't have equivalent statuses for old ```ON_PROCESS``` and ```NOT_RECEIVED```. The last one is important I think.

shared_model/interfaces/transaction_responses/abstract_tx_response.hpp
```diff
@@ -0,0 +1,63 @@
...
+     * @tparam Model - concrete model transaction response
+     */
+    template <typename Model>
+    class AbstractTxResponse
+        : public Primitive<Model, iroha::model::TransactionResponse> {
+     private:
+      /**
+       * @return string representation of class name
+       */
+      virtual std::string className() const = 0;
```
luckychess: Is it only for ```toString``` or it has a deeper meaning?

muratovv: Only for `toString`. I want to do it as a template parameter, but it is not available in an easy way.

shared_model/interfaces/transaction_responses/tx_response.hpp
muratovv: I fixed `NOT_RECEIVED`. `ON_PROCESS` intentionally ignored, with the new shared model design we don't have separated phase for stateless validation.

shared_model/interfaces/transaction_responses/abstract_tx_response.hpp
l4l: Why not just use rtti? `typeid(*this).name()`

muratovv: Because of no reasons to use RTTI here. `Name` may return not what you expected and the result may change on different platforms.

shared_model/interfaces/transaction_responses/committed_tx_response.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace shared_model {
+  namespace interface {
+    /**
+     * Tx pipeline succeeded, tx is committed in ledger
+     */
+    class CommittedTxResponse : public AbstractTxResponse<CommittedTxResponse> {
+     private:
+      std::string className() const override { return "CommittedTxResponse"; }
+
+      iroha::model::TransactionResponse::Status oldModelStatus() const {
```
l4l: I guess `override` is missing
check also the following classes

shared_model/interfaces/transaction_responses/committed_tx_response.hpp
```diff
@@ -0,0 +1,39 @@
...
+#define IROHA_COMMITTED_TX_RESPONSE_HPP
+
+#include "interfaces/transaction_responses/abstract_tx_response.hpp"
+
+namespace shared_model {
+  namespace interface {
+    /**
+     * Tx pipeline succeeded, tx is committed in ledger
+     */
+    class CommittedTxResponse : public AbstractTxResponse<CommittedTxResponse> {
```
l4l: Is there any profits from responses being implemented via classes, not just `enum {response1, response2, ...}`?

shared_model/interfaces/transaction_responses/stateful_failed_tx_response.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include "model/transaction_response.hpp"
+#include "utils/string_builder.hpp"
+
+namespace shared_model {
+  namespace interface {
+
+    /**
+     * Tx response of broken stateful validation
+     */
+    class StatefulFailedTxResponse
```
l4l: Maybe use macros? This code is supa-pupa-overduplicated D:
e.g:
`#define TX_RESPONCE_CLASS(class_name, old_status) {..class code..}`
and use:
```cpp
TX_RESPONCE_CLASS(StatefulFailedTxResponse, iroha::model::TransactionResponse::Status::STATEFUL_VALIDATION_FAILED);
TX_RESPONCE_CLASS(...);
etc
```
It definitely will be lesser copy-paste errors

shared_model/interfaces/transaction_responses/tx_response.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * limitations under the License.
+ */
+
+#ifndef IROHA_TX_RESPONSE_HPP
+#define IROHA_TX_RESPONSE_HPP
+
+#include <boost/variant.hpp>
+#include "interfaces/hashable.hpp"
+#include "interfaces/polymorphic_wrapper.hpp"
+#include "interfaces/transaction_responses/committed_tx_response.hpp"
```
l4l: maybe rename folder to `tx_responses`?

shared_model/interfaces/transaction_responses/stateful_failed_tx_response.hpp
l4l: *note: template hackish solution looks ugly, macros are good to use here imo

shared_model/interfaces/transaction_responses/unknown_tx_response.hpp
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Transaction not found
+     */
+    class UnknownTxResponse : public AbstractTxResponse<UnknownTxResponse> {
+     private:
+      std::string className() const override { return "UnknownTxResponse"; }
+
+      iroha::model::TransactionResponse::Status oldModelStatus() const {
+        return iroha::model::TransactionResponse::Status::
+            NOT_RECEIVED;
```
l4l: Is it a good mapping `unknown <-> not_received`? May you explain that a bit

irohad/model/tx_responses/stateless_response.hpp
```diff
@@ -21,18 +21,17 @@
...
-   */
-  bool passed{};
-};
-} // namespace model
-} // namespace iroha
-#endif //IROHA_STATELESS_RESPONSE_HPP
+    /**
+     * Transaction response that contains
+     */
+    struct TransactionStatelessResponse : TransactionResponse {
```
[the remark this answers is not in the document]
lebdron: The default accessibility is `public` for `struct`.

shared_model/interfaces/transaction_responses/abstract_tx_response.hpp
```diff
@@ -0,0 +1,64 @@
...
+        return detail::PrettyStringBuilder().init(className()).finalize();
+      }
+
+      iroha::model::TransactionResponse *makeOldModel() const override {
+        auto tx_response = new iroha::model::TransactionResponse();
+        tx_response->current_status = oldModelStatus();
+        return tx_response;
+      }
+
+      bool operator==(const Model &rhs) const override { return true; }
```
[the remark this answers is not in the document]
muratovv: No reasons, because you are not available to compare different classes.

shared_model/interfaces/transaction_responses/committed_tx_response.hpp
muratovv: Variant and clean code for generic functions, such as `toString` and `makeOldModel`.

shared_model/interfaces/transaction_responses/committed_tx_response.hpp
```diff
@@ -0,0 +1,39 @@
...
+     private:
+      std::string className() const override { return "CommittedTxResponse"; }
```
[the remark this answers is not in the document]
muratovv: Don't care about virtual dispatching, because it used only in `toString`, that used only for developing purposes.  

And could you explain how to solve this task with static methods? They don't have polymorphic behavior at all.

shared_model/interfaces/transaction_responses/stateful_failed_tx_response.hpp
muratovv: I guess macro in the current solution not so good because it will be broken when some classes need to be extended. Also, this approach decreases readability at all.

shared_model/interfaces/transaction_responses/tx_response.hpp
muratovv: Tx not so obvious abbreviation. I guess better use the full name for the folder instead.

shared_model/interfaces/transaction_responses/unknown_tx_response.hpp
muratovv: The idea is not_received possible to achieve in two situations: when transaction with those hash not received and transaction received but was rejected long time ago. `NOT_RECEIVED` not reflected name for the second case. For my opinion unknow covered both, but you may provide a more suitable name.

shared_model/interfaces/transaction_responses/abstract_tx_response.hpp
muratovv: Clean and consistent output better to read.

irohad/model/tx_responses/stateless_response.hpp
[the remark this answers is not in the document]
muratovv: I remove this class at all.

PR #683, opened 2017-11-07
Add lazy field initialization

muratovv:
## What is this pull request?
This request provides lazy initialization for fields.
   
Lazy functionality will be very useful for proto-backend for lazy initialization of fields.

## Details/Features
* Class Lazy
* Tests for it


shared_model/utils/lazy.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace shared_model {
+  namespace detail {
+
+    /**
+     * Lazy class for lazy converting one type to another
+     * @tparam Source - input type
+     * @tparam Target - output type
+     */
+    template <typename Source, typename Target>
+    class Lazy {
```
l4l: It's a bad name for that class. It's main purpose is converting (and actually caching) stuff, not "being lazy"
Imo it's fine to name it just `converter`

shared_model/utils/lazy.hpp
```diff
@@ -0,0 +1,73 @@
...
+      using TransformType = std::function<Target(const Source &)>;
+
+     public:
+      Lazy(const Source &source, const TransformType &transform)
+          : source_(source), transform_(transform) {}
+
+      /**
+       * @return value after transformation
+       */
+      const Target &get() {
```
l4l: Unintended copy is here
What's the problem with returning `shared_ptr`?

test/module/shared_model/lazy_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * @when Call get once
+ * @then Assert that transform invoked
+ */
+TEST(LazyTest, GetterTest) {
+  SourceValue v;
+  v.val = 100500;
+  auto lazy = shared_model::detail::makeLazy(v, [](auto source) {
+    TargetValue target_value;
+    target_value.target = std::to_string(source.val);
+    return target_value;
```
l4l: Replace lambda's body with simplier construction: `return TargetValue{std::to_string(source.val};`

test/module/shared_model/lazy_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+ */
+TEST(LazyTest, CheckLaziness) {
+  SourceValue v;
+  v.val = 100500;
+  auto call_counter = 0;
+  auto lazy = shared_model::detail::makeLazy(v, [&call_counter](auto source) {
+    call_counter++;
+    TargetValue target_value;
+    target_value.target = std::to_string(source.val);
+    return target_value;
```
l4l: same here

shared_model/utils/lazy.hpp
muratovv: No reasons to provide `shared_ptr` here.

shared_model/utils/lazy.hpp
muratovv: This is not simple `converter` because converter it is `TransformType`. Maybe `LazyConverter`. @lebdron maybe you have ideas about the name?

lebdron: `LazyInitializer` sounds good to me. In case it will be reused several times in class member definition, a separate private `using` can be defined in the class: `using l = LazyInitializer<>`

shared_model/utils/lazy_initializer.hpp
```diff
@@ -50,7 +50,7 @@ namespace shared_model {
       }
 
      private:
-      Source source_;
+      const Source &source_;
       TransformType transform_;
       std::shared_ptr<Target> target_value_;
```
l4l: btw, why the `shared_ptr`, not `unique_ptr`?

PR #685, opened 2017-11-09
Fix ccache problem

Warchant:
This PR *possibly* (I don't have a way to test this) fixes the problem with `ccache`. 

- [x] As was suggested, it changes `{{ arch }}` to hardcoded word (linux/macos).
- [x] Adds new cross-branch cache. If cache from branch A is saved, then at branch B it is possible to recover it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,9 @@ jobs:
       - run:
           name: ccache setup
           command: |
-            ccache --max-files=0
+            ccache --version
+            ccache --show-stats
+            ccache --zero-stats
             ccache --max-size=1G
       - run:
           name: cmake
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -136,44 +142,54 @@ jobs:
...
+            - external-macos-{{ .Branch }}-
+            - external-macos-
           paths:
             - external
-      - run: ccache --show-stats
+      - run:
+          name: ccache setup
+          command: |
+            ccache --version
+            ccache --show-stats
```
l4l: What's the point of showing stats in setup?

Warchant: For debug purposes.

lebdron:
https://github.com/hyperledger/iroha/pull/664 seems to fix the issue. This pull request can be reopened if the issue will reappear.

PR #686, opened 2017-11-09
Remove redundant files

kamilsa:
## What is this pull request?
It clears not-needed or wrongly placed files in iroha repository
   
1. Iroha contains old hakodate script and many related files to it. Now they are obsolete, so should be removed.
2. docs folder is intended to keep documentation files, but now it contains some files needed to launch irohad, which is not right

## Details/Features
Now users of iroha should follow tutorial in order to launch irohad. So the only needed file to run it is config sample


docs/admin@test.priv
```diff
@@ -1 +0,0 @@
-881e179b8d6900f892b43c7010c77e47731711735dae462ba9a8d12a9d956e58ab5bce5597a9914b0a743fccd3c56e521c1b59d73391d8ae33cc3667c5068539
```
x3medima17: no newline

Warchant:
@kamilsa fixed DCO.

PR #687, opened 2017-11-09
Fix/createdomain - CreateDomain.json contains invalid json

syilmaz:
## What is this pull request?
The CreateDomain.json contains invalid JSON. The "domain_name" key should be "domain_id" and "user_default_role" is missing. 
   
Without this fix the `hakodate.sh` shell script returns `Json transaction has wrong format.`
  
Running `./hakodate.sh` will now successfully create a domain.

## Details/Features
- Renamed "domain_name" to "domain_id" inside the CreateDomain.json
- Added "user_default_role" to CreateDomain.json

### DIFF
--- a/example/CreateDomain.json
+++ b/example/CreateDomain.json
@@ -11,7 +11,8 @@
     "commands": [
         {
           "command_type": "CreateDomain",
-          "domain_name": "soramitsu"
+          "domain_id": "soramitsu",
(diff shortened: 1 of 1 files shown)


l4l:
Thanks for your contribution!
Unfortunately these are deprecated examples and will be removed [here](https://github.com/hyperledger/iroha/pull/686)

PR #688, opened 2017-11-10
Feature/complex asset cli

grimadas:
## What is this pull request?
Add missing command to interactive cli
   
Set account key value was missing in cli.
 

## Details/Features
- Fix added for empty json in CreateAccount
- Set Account KV added to interactive cli :+1:

### DIFF
--- a/iroha-cli/interactive/impl/interactive_transaction_cli.cpp
+++ b/iroha-cli/interactive/impl/interactive_transaction_cli.cpp
@@ -26,12 +26,11 @@
 #include "model/commands/create_role.hpp"
 #include "model/commands/grant_permission.hpp"
 #include "model/commands/revoke_permission.hpp"
+#include "model/commands/set_account_detail.hpp"
 #include "model/converters/json_common.hpp"
 #include "model/converters/json_transaction_factory.hpp"
 #include "model/converters/pb_common.hpp"
-#include "model/generators/transaction_generator.hpp"
 #include "model/permissions.hpp"
-#include "parser/parser.hpp"
 
 using namespace iroha::model;
 
@@ -53,7 +52,8 @@ namespace iroha_cli {
           {CREATE_ROLE, "Create new role"},
           {APPEND_ROLE, "Add new role to account"},
           {GRANT_PERM, "Grant permission over your account"},
-          {REVOKE_PERM, "Revoke permission from account"}
+          {REVOKE_PERM, "Revoke permission from account"},
+          {SET_ACC_KV, "Set account key/value detail"}
           // commands_description_map_
       };
 
@@ -106,7 +106,8 @@ namespace iroha_cli {
             can_create_account}},
           {APPEND_ROLE, {acc_id, role}},
           {GRANT_PERM, {acc_id, perm}},
-          {REVOKE_PERM, {acc_id, perm}}
+          {REVOKE_PERM, {acc_id, perm}},
+          {SET_ACC_KV, {acc_id, "key", "value"}}
           // command parameters descriptions
       };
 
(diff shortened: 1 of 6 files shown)


irohad/model/commands/create_account.hpp
```diff
@@ -49,13 +49,16 @@ namespace iroha {
 
       bool operator==(const Command &command) const override;
 
-      CreateAccount() {}
+      CreateAccount() { json_data = "{}"; }
```
l4l: use
```cpp
CreateAccount() : json_data("{}") {}
```
instead

iroha-cli/interactive/impl/interactive_transaction_cli.cpp
```diff
@@ -236,9 +238,12 @@ namespace iroha_cli {
       auto create_account = parser::parseValue<bool>(params[8]);
 
       if (not read_self.has_value() or not edit_self.has_value()
-          or not read_all.has_value() or not transfer_receive.has_value()
-          or not asset_create.has_value() or not create_domain.has_value()
-          or not roles.has_value() or not create_account.has_value()) {
+          or not read_all.has_value()
```
l4l: maybe add a vararg fn for that?
```cpp
template<typename T>
bool has_no_value() {
  return not t.has_value();
}

template<typename T, typename... Args>
bool has_no_value(T t, Args... a) {
  return not t.has_value() or not has_no_value(a...);
}
```

Not sure but some c++14 feature may allow to write only one method

irohad/model/converters/impl/json_command_factory.cpp
```diff
@@ -213,17 +213,25 @@ namespace iroha {
...
-        return make_optional_ptr<CreateAccount>()
+        auto des_val = make_optional_ptr<CreateAccount>()
             | des.String(&CreateAccount::account_name, "account_name")
             | des.String(&CreateAccount::domain_id, "domain_id")
-            | des.String(&CreateAccount::pubkey, "pubkey") | toCommand;
+            | des.String(&CreateAccount::pubkey, "pubkey");
+
+        if (document.HasMember("json_data")) {
+          des_val =
+              des_val | des.String(&CreateAccount::json_data, "json_data");
```
l4l: looks quite awkward D:

iroha-cli/interactive/impl/interactive_transaction_cli.cpp
Warchant: To be more specific, the correct code is:
```C++
template <typename T> bool has_value(T &&t) { return t.has_value(); }

template <typename T, typename... Args> bool has_value(T &&t, Args &&... a) {
  return t.has_value() and has_value(std::forward<Args>(a)...);
}

if(not has_value(a, b, c, d, e)) { /* do something */ }
```

Warchant: Another way is actually just write:
```
// optionals can be safely casted to bool
if(not (a && b && c && d && e)) { /* do something */  }
```

l4l: Seems that only [c++17 feature] :(

irohad/model/converters/impl/json_command_factory.cpp
grimadas: Any suggestions to improve ?

iroha-cli/interactive/impl/interactive_transaction_cli.cpp
l4l: Implicit cast to bool looks even better for me

irohad/model/converters/impl/json_command_factory.cpp
l4l: No idea, just want to notice so maybe someone other will have

iroha-cli/interactive/impl/interactive_transaction_cli.cpp
neewy: Nice talk you have. Any outcome?

PR #694, opened 2017-11-13
Feature/rfc 1035 domain name syntax

mitsutaka-takeda:
## What is this pull request?
Relax the current restriction on account names. This commit supports account names of the domain-name syntax defined in RFC1035.
   
We needed to create an account based on an e-mail address.  For example, when a user register into our system with his/her email address, our system creates an account in iroha ledger from email-addresses' local part.
  
From the interface's point of view, this commit changes what sequence of characters are accepted as a valid account name, and does not change the current interface nor add new interface.

## Details/Features
- Add a utility class DnsParser which validates if a string conforms to domain-name syntax(4b134338)


neewy:
Thank you for your first time contribution! We are really pleased to see improvements like these
It would be nice to see also contribution to https://github.com/hyperledger/iroha-api in **docs** branch, to keep the docs about API consistent

takemiyamakoto:
Does sorabot have anything to say?

neewy:
@Warchant you see — for some reason sorabot haven't triggered on such PR, can you please check that?

irohad/model/execution/dns_parser.hpp
```diff
@@ -0,0 +1,38 @@
+#ifndef IROHA_DNS_PARSER_HPP
```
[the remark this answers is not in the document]
mitsutaka-takeda: You're right. I will change the class name to DomainNameParser.

marksta80:
Guten Tag

Ich bin aktuell nicht im Büro, Ihre Mail wird daher nicht umgehend bearbeitet.

Wenden Sie sich bei dringenden Angelegenheiten bitte an service@4eyes.ch oder 061 261 43 48.

Freundliche Grüsse
Markus Stauffiger

Warchant:
@neewy sorabot does not work for forks. If you want to trigger sorabot, then setup `SORABOT_TOKEN` env variable inside circle CI. It should contain `personal access token` with `repo` permission. Account which analyzes PR should have push access to this PR.

irohad/model/execution/domain_name_parser.hpp
```diff
@@ -0,0 +1,38 @@
...
+#define IROHA_DOMAIN_NAME_PARSER_HPP
+
+#include <string>
+
+namespace iroha {
+  namespace model {
+    /**
+     * Parse a string as a domain name as defined in RFC 1035.
+     */
+    class DomainNameParser {
```
l4l: I think there's no need in class.
Also I don't really sure that *irohad/model/execution* is best folder for it, at least for me it's not obvious. Maybe *libs/parser/*?

irohad/model/execution/impl/domain_name_parser.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <boost/spirit/include/qi_plus.hpp>
+
+namespace iroha {
+  namespace model {
+    bool DomainNameParser::isValid(const std::string &str) {
+      namespace qi = boost::spirit::qi;
+
+      qi::rule<decltype(str.begin())> letter, let_dig, end_with_let_dig,
+          let_dig_hyp, label, subdomain, domain;
+      domain = subdomain | ' ';
```
l4l: Does this allow `" "` as domain name? From rfc is fine, but maybe not a good thing here.
Any thoughts on that?

irohad/model/execution/domain_name_parser.hpp
Warchant: `libs/util`

mitsutaka-takeda: I can create a package `libs/util` and move the `domain_name_parser` there. What should I name the package? `util` sounds like too generic and might conflict with other packages. Do you have any suggestion?

irohad/model/execution/impl/domain_name_parser.cpp
mitsutaka-takeda: Maybe we should follow the RFC 1123 and disallow a whitespace.
https://serverfault.com/questions/342284/does-active-directory-support-dns-names-with-spaces

takemiyamakoto: We should disallow whitespaces. Also, it might be nice to support katakana.

cimadai: Why do we need to support katakana?
Could you illustrate the use case of katakana?

takemiyamakoto: Using domain names written in katakana, like アルトプラス. We have a good Japanese community and it is not such a big alphabet, so it might be interesting to support it. The project is called Iroha after all ^^

irohad/model/execution/domain_name_parser.hpp
Warchant: `validator` is fine. All it does is answers the question "is domain name valid" and imo `validator` sounds okay.

libs/validator/domain_name_validator.hpp
```diff
@@ -0,0 +1,55 @@
...
+#ifndef VALIDATOR_DOMAIN_NAME_VALIDATOR_HPP
+#define VALIDATOR_DOMAIN_NAME_VALIDATOR_HPP
+
+#include <string>
+
+namespace validator {
+  /**
+   * Parse a string as a domain name as defined in RFC 1035.
+   */
+  class DomainNameValidator {
```
l4l: One [thing](https://github.com/hyperledger/iroha/pull/694#discussion_r150551815) left. What's the point of the class? Imo that's fine to use just `validator::isValidDomainName(..)`

Warchant:
Next time please don't forget to rebase branch onto develop :)
```
# once add hyperledger/iroha repo as upstream
git remote add upstream git@github.com:hyperledger/iroha.git
git fetch upstream

# then, to rebase:
git checkout $yourbranch
git rebase -i upstream/develop
```

I did this for you this time.

PR #696, opened 2017-11-15
Feature/get transactions block query

motxx:
## What is this pull request?
`RedisBlockQuery::getTransactions()` API.

As a server, I want to get transactions from tx hashes.
  
```cpp
block_query_->getTransactions({tx_hash_1, tx_hash_2}).subscribe([](auto tx) {
    // get transactions which have specified transactions' hashes.
  });
```

## Details/Features
- Using `RedisBlockQuery::getTxByHashSync()` in `RedisBlockQuery::getTransactions()`

### DIFF
--- a/irohad/ametsuchi/impl/redis_block_query.cpp
+++ b/irohad/ametsuchi/impl/redis_block_query.cpp
@@ -170,17 +170,18 @@ namespace iroha {
           });
     }
 
-    rxcpp::observable<model::Transaction> RedisBlockQuery::getTransactions(
+    rxcpp::observable<nonstd::optional<model::Transaction>>
+    RedisBlockQuery::getTransactions(
         const std::vector<iroha::hash256_t> &tx_hashes) {
-      return rxcpp::observable<>::create<model::Transaction>(
+      return rxcpp::observable<>::create<nonstd::optional<model::Transaction>>(
           [this, tx_hashes](auto subscriber) {
-            std::for_each(
-                tx_hashes.begin(),
-                tx_hashes.end(),
-                [ that = this, &subscriber ](auto tx_hash) {
-                  auto tx = that->getTxByHashSync(tx_hash.to_string());
-                  subscriber.on_next(tx.value_or(iroha::model::Transaction{}));
-                });
+            std::for_each(tx_hashes.begin(),
+                          tx_hashes.end(),
+                          [ that = this, &subscriber ](auto tx_hash) {
+                            auto tx =
+                                that->getTxByHashSync(tx_hash.to_string());
+                            subscriber.on_next(tx);
+                          });
             subscriber.on_completed();
           });
     }
--- a/irohad/ametsuchi/impl/redis_block_query.cpp
+++ b/irohad/ametsuchi/impl/redis_block_query.cpp
@@ -170,6 +170,21 @@ namespace iroha {
           });
     }
 
+    rxcpp::observable<model::Transaction> RedisBlockQuery::getTransactions(
+        const std::vector<iroha::hash256_t> &tx_hashes) {
+      return rxcpp::observable<>::create<model::Transaction>(
+          [this, tx_hashes](auto subscriber) {
+            std::for_each(
+                tx_hashes.begin(),
+                tx_hashes.end(),
+                [ that = this, &subscriber ](auto tx_hash) {
+                  auto tx = that->getTxByHashSync(tx_hash.to_string());
+                  subscriber.on_next(tx.value_or(iroha::model::Transaction{}));
+                });
+            subscriber.on_completed();
+          });
+    }
+
     boost::optional<model::Transaction> RedisBlockQuery::getTxByHashSync(
         const std::string &hash) {
       return getBlockId(hash) |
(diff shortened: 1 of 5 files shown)


motxx:
sonarqube failed but can be reveiwed.

irohad/ametsuchi/impl/redis_block_query.cpp
```diff
@@ -170,6 +170,21 @@ namespace iroha {
...
     }
 
+    rxcpp::observable<model::Transaction> RedisBlockQuery::getTransactions(
+        const std::vector<iroha::hash256_t> &tx_hashes) {
+      return rxcpp::observable<>::create<model::Transaction>(
+          [this, tx_hashes](auto subscriber) {
+            std::for_each(
+                tx_hashes.begin(),
+                tx_hashes.end(),
+                [ that = this, &subscriber ](auto tx_hash) {
```
l4l: extra space after/before square brackets :(

irohad/ametsuchi/impl/redis_block_query.cpp
```diff
@@ -170,6 +170,21 @@ namespace iroha {
...
+                  auto tx = that->getTxByHashSync(tx_hash.to_string());
+                  subscriber.on_next(tx.value_or(iroha::model::Transaction{}));
```
l4l: Not really sure it's a good way. Default tx can be indistinguishable from the real one
Imo emitting optional is better

test/module/irohad/ametsuchi/block_query_test.cpp
```diff
@@ -120,3 +125,70 @@ TEST_F(BlockQueryTest, GetAccountTransactionsNonExistingUser) {
...
+/**
+ * @given block store with 2 blocks totally containing 3 txs created by
+ * user1@test
+ * AND 1 tx created by user2@test
+ * @when query to get transactions with transactions' hashes
+ * @then queried transactions
+ */
+TEST_F(BlockQueryTest, GetTransactionsExistingTxHashes) {
+  //TODO 15/11/17 motxx - Use EqualList VerificationStrategy
+  {
```
l4l: Shouldn't this block can be divided in separate tests?

l4l: same below

test/module/irohad/ametsuchi/block_query_test.cpp
```diff
@@ -120,3 +125,70 @@ TEST_F(BlockQueryTest, GetAccountTransactionsNonExistingUser) {
...
+    auto wrapper = make_test_subscriber<CallExact>(
+      blocks->getTransactions({}), 0);
+    wrapper.subscribe();
+    ASSERT_TRUE(wrapper.validate());
+  }
+  {
+    //TODO 15/11/17 motxx - Use EqualList VerificationStrategy
+    auto wrapper = make_test_subscriber<CallExact>(
+      blocks->getTransactions({invalid_tx_hash_1, tx_hashes[0]}), 2);
+    auto subs_cnt = 0;
```
l4l: 1. Cant' find any variable modification (though [there's if check](https://github.com/hyperledger/iroha/pull/696/files#diff-c855aee60870118507c2cd5177c150eaR185))
2. This var can be static inside the subscribing lambda

irohad/ametsuchi/impl/redis_block_query.cpp
motxx: `clang-format` does this ;(

irohad/ametsuchi/impl/redis_block_query.cpp
```diff
@@ -170,6 +170,22 @@ namespace iroha {
...
+    RedisBlockQuery::getTransactions(
+        const std::vector<iroha::hash256_t> &tx_hashes) {
+      return rxcpp::observable<>::create<boost::optional<model::Transaction>>(
+          [this, tx_hashes](auto subscriber) {
+            std::for_each(tx_hashes.begin(),
+                          tx_hashes.end(),
+                          [ that = this, &subscriber ](auto tx_hash) {
+                            auto tx =
+                                that->getTxByHashSync(tx_hash.to_string());
+                            subscriber.on_next(tx);
```
Warchant: is it necessary to copy `tx_hash` inside this lambda or it can be passed by ref?
Also, is it necessary to copy tx twice:
 1. `auto tx = ... `
 2. `subscriber.on_next(tx)` 

Maybe it is possible to omit one copy?

Warchant:
LOL, sonar-scanner has a bug, it can not parse from json `...id: 4317685969, ...` since it is out of `uint32` range. Github, well done. Do not worry about failing of sonar-pr.

irohad/ametsuchi/impl/redis_block_query.cpp
motxx: > is it necessary to copy tx_hash inside this lambda or it can be passed by ref?

It should be copied because test fails. The reason is maybe lazy-evaluation.

And, I am afraid that capturing `this` in this and other methods. Can we guarantee that the object is alive until begin subscribed?

> is it necessary to copy tx twice:

Thanks, that's right.

PR #708, opened 2017-11-20
Feature/develop dockerfile update

lebdron:
## What is this pull request?
Manual installation of CMake and boost libraries.
Bump versions of boost and cmake: grpc build passes with cmake 3.5.1, boost 1.65 is required for pedantic build.
   
It allows manual version configuration of cmake and boost, which is required in certain branches.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -77,7 +77,7 @@ find_package(tbb)
 ##########################
 #         boost          #
 ##########################
-find_package(Boost 1.58.0 REQUIRED
+find_package(Boost 1.65.0 REQUIRED
     COMPONENTS
     filesystem
     system
--- a/docker/develop/Dockerfile
+++ b/docker/develop/Dockerfile
@@ -9,8 +9,12 @@ ENV IROHA_BUILD /opt/iroha/build
 
 RUN apt-get update; \
     apt-get -y upgrade; \
-    apt-get -y --no-install-recommends install apt-utils; \
+    apt-get -y --no-install-recommends install apt-utils software-properties-common; \
     apt-get -y clean
+
+# add git repository
+RUN add-apt-repository -y ppa:git-core/ppa; \
+    apt-get update
 
 RUN apt-get -y --no-install-recommends install build-essential python-software-properties \
         automake libtool \
--- a/docker/develop/Dockerfile
+++ b/docker/develop/Dockerfile
@@ -21,12 +21,20 @@ RUN apt-get -y --no-install-recommends install build-essential python-software-p
         # code coverage
         lcov \
         # other
-        wget curl cmake file unzip gdb iputils-ping vim ccache \
+        wget curl file unzip gdb iputils-ping vim ccache \
         gcovr vera++ cppcheck doxygen graphviz graphviz-dev; \
     apt-get -y clean
 
 RUN pip3 install --upgrade pip; \
     pip3 install flask requests
+
+# install cmake 3.5.1
+RUN git clone https://gitlab.kitware.com/cmake/cmake.git /tmp/cmake; \
+    (cd /tmp/cmake ; git checkout 64130a7e793483e24c1d68bdd234f81d5edb2d51); \
+    (cd /tmp/cmake ; /tmp/cmake/bootstrap --parallel=${PARALLELISM} --enable-ccache); \
+    make -j${PARALLELISM} -C /tmp/cmake; \
+    make -C /tmp/cmake install; \
+    rm -rf /tmp/cmake
 
 # install protobuf
 RUN git clone https://github.com/google/protobuf /tmp/protobuf; \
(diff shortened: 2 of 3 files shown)


cmake/dependencies.cmake
```diff
@@ -77,7 +77,7 @@ find_package(tbb)
 ##########################
 #         boost          #
 ##########################
-find_package(Boost 1.58.0 REQUIRED
+find_package(Boost 1.65.0 REQUIRED
```
l4l: May you example why 1.65 (e.g not 1.59)?

docker/develop/Dockerfile
```diff
@@ -9,25 +9,47 @@ ENV IROHA_BUILD /opt/iroha/build
...
+RUN git clone https://gitlab.kitware.com/cmake/cmake.git /tmp/cmake; \
+    (cd /tmp/cmake ; git checkout 64130a7e793483e24c1d68bdd234f81d5edb2d51); \
+    (cd /tmp/cmake ; /tmp/cmake/bootstrap --parallel=${PARALLELISM} --enable-ccache); \
+    make -j${PARALLELISM} -C /tmp/cmake; \
+    make -C /tmp/cmake install; \
+    rm -rf /tmp/cmake
+
+# install boost
+RUN git clone https://github.com/boostorg/boost /tmp/boost; \
+    (cd /tmp/boost ; git checkout 436ad1dfcfc7e0246141beddd11c8a4e9c10b146); \
```
Warchant: The same.

docker/develop/Dockerfile
```diff
@@ -9,25 +9,47 @@ ENV IROHA_BUILD /opt/iroha/build
...
+        wget curl file unzip gdb iputils-ping vim ccache \
+        gcovr cppcheck doxygen graphviz graphviz-dev; \
     apt-get -y clean
 
 RUN pip3 install --upgrade pip; \
     pip3 install flask requests
 
+# install cmake
+RUN git clone https://gitlab.kitware.com/cmake/cmake.git /tmp/cmake; \
+    (cd /tmp/cmake ; git checkout 64130a7e793483e24c1d68bdd234f81d5edb2d51); \
```
Warchant: please, if you use hash as version, at least specify latest closest release version. 
Commit id does not give any information about this

docker/develop/Dockerfile
```diff
@@ -9,25 +9,47 @@ ENV IROHA_BUILD /opt/iroha/build
...
+    rm -rf /tmp/cmake
+
+# install boost
+RUN git clone https://github.com/boostorg/boost /tmp/boost; \
+    (cd /tmp/boost ; git checkout 436ad1dfcfc7e0246141beddd11c8a4e9c10b146); \
+    (cd /tmp/boost ; git submodule init); \
+    (cd /tmp/boost ; git submodule update --recursive -j ${PARALLELISM}); \
+    (cd /tmp/boost ; /tmp/boost/bootstrap.sh --with-libraries=system,filesystem); \
+    (cd /tmp/boost ; /tmp/boost/b2 headers); \
+    (cd /tmp/boost ; /tmp/boost/b2 cxxflags="-std=c++14" -j 8 install); \
```
Warchant: why 8?
use ${PARALLELISM}

docker/develop/Dockerfile
l4l: It can introduce some features that can be added after some version and the next one can be incompatible
Though there's probably no such problems in cmake

Warchant: I mean, it would be good to have a comment at least which references version.

cmake/dependencies.cmake
lebdron: I can't find the exact line now, but it's related to pedantic warning in boost multiprecision: some macro was terminated with `;`when it is not required. This issue was fixed in version 1.65.

docker/develop/Dockerfile
```diff
@@ -9,25 +9,47 @@ ENV IROHA_BUILD /opt/iroha/build
...
+    make -j${PARALLELISM} -C /tmp/cmake; \
+    make -C /tmp/cmake install; \
+    rm -rf /tmp/cmake
+
+# install boost 1.65.1
+RUN git clone https://github.com/boostorg/boost /tmp/boost; \
+    (cd /tmp/boost ; git checkout 436ad1dfcfc7e0246141beddd11c8a4e9c10b146); \
+    (cd /tmp/boost ; git submodule init); \
+    (cd /tmp/boost ; git submodule update --recursive -j ${PARALLELISM}); \
+    (cd /tmp/boost ; /tmp/boost/bootstrap.sh --with-libraries=system,filesystem); \
```
Warchant: do you think we need only system and filesystem? if it happens that we need more, we have to re-build whole image.
I guess it is better to include everything from boost here.

What do you think? @lebdron

l4l: I don't really think that it's necessary. There's no much really important think that we can easily introduce and require some significant work. I think it's fine to leave it like this

PR #709, opened 2017-11-20
Use condition variable to wait until previous tx sent

motxx:
## What is this pull request?
Fix sending transactions in integration test to be able to guarantee the order of transactions.

To solve the issue of sending multiple transactions in the test.
Being lack this fix, the test of https://github.com/hyperledger/iroha/pull/692 sometimes fails because of sending multiple transactions. 

```cpp
sendTxsInOrderAndValidate({tx1, tx2, tx3});
```

## Details/Features
- Remove useless debug.
- Use condition variable to wait previous txs are sent.
- One transaction in one block.
- Remove the feature of sending multiple transactions which are independent each other inside one block because it is not used.

### DIFF
--- a/test/integration/pipeline/transfer_asset_inter_domain_test.cpp
+++ b/test/integration/pipeline/transfer_asset_inter_domain_test.cpp
@@ -74,10 +74,6 @@ class TransferAssetInterDomainTest : public TxPipelineIntegrationTestFixture {
     ASSERT_TRUE(irohad->storage);
 
     iroha::main::BlockInserter(irohad->storage).applyToLedger({genesis_block});
-    irohad->storage->getBlockQuery()->getTopBlocks(1).subscribe([](auto block) {
-      std::cout << "hoge\n";
-      EXPECT_EQ(block.transactions.size(), 2);
-    });
     irohad->init();
     irohad->run();
   }
@@ -171,6 +167,5 @@ TEST_F(TransferAssetInterDomainTest, TransferAssetInterDomainTest) {
           getVal(iroha::Amount().createFromString("5.50")))});
   iroha::model::ModelCryptoProviderImpl(ivanKeypair_).sign(tx3);
 
-  sendTransactions({tx1, tx2, tx3});
-  validate();
+  sendTxsInOrderAndValidate({tx1, tx2, tx3});
 }
--- a/test/integration/pipeline/tx_pipeline_integration_test_fixture.hpp
+++ b/test/integration/pipeline/tx_pipeline_integration_test_fixture.hpp
@@ -18,6 +18,7 @@
 #ifndef TX_PIPELINE_INTEGRATION_TEST_FIXTURE_HPP
 #define TX_PIPELINE_INTEGRATION_TEST_FIXTURE_HPP
 
+#include <atomic>
 #include "cryptography/ed25519_sha3_impl/internal/sha3_hash.hpp"
 #include "crypto/keys_manager_impl.hpp"
 #include "datetime/time.hpp"
@@ -98,8 +99,7 @@ class TxPipelineIntegrationTestFixture
   void sendTxsInOrderAndValidate(
       const std::vector<iroha::model::Transaction> &transactions) {
     // test subscribers can't solve duplicate func call.
-    ASSERT_FALSE(duplicate_sent);
-    duplicate_sent = true;
+    ASSERT_FALSE(duplicate_sent.exchange(true));
 
     const auto num_blocks =
         transactions.size();  // Use one block per one transaction
@@ -147,7 +147,7 @@ class TxPipelineIntegrationTestFixture
 
   std::shared_ptr<iroha::KeysManager> manager;
 
-  bool duplicate_sent = false;
+  std::atomic_bool duplicate_sent {false};
   size_t next_height_count = 2;
 
  private:
@@ -167,7 +167,7 @@ class TxPipelineIntegrationTestFixture
       commit.subscribe([this](auto block) { blocks.push_back(block); });
     });
     irohad->getPeerCommunicationService()->on_commit().subscribe(
-        [this](auto) { cv.notify_all(); });
+        [this](auto) { cv.notify_one(); });
   }
 
   void sendTransaction(const iroha::model::Transaction &transaction) {
--- a/test/integration/pipeline/tx_pipeline_integration_test_fixture.hpp
+++ b/test/integration/pipeline/tx_pipeline_integration_test_fixture.hpp
@@ -93,65 +93,32 @@ class TxPipelineIntegrationTestFixture
   }
 
   /**
-   * @param transactions
-   * @param expected_block_height - change expected height when using additional
-   * initial blocks
+   * @param transactions in order
    */
-  void sendTransactions(std::vector<iroha::model::Transaction> transactions) {
-    // generate expected proposal
-    expected_proposal = std::make_shared<iroha::model::Proposal>(transactions);
-    expected_proposal->height = 2;
+  void sendTxsInOrderAndValidate(
+      const std::vector<iroha::model::Transaction> &transactions) {
+    // test subscribers can't solve duplicate func call.
+    ASSERT_FALSE(duplicate_sent);
+    duplicate_sent = true;
 
-    // generate expected block
-    expected_block = iroha::model::Block{};
-    expected_block.height = expected_proposal->height;
-    expected_block.prev_hash = genesis_block.hash;
-    expected_block.transactions = transactions;
-    expected_block.txs_number = transactions.size();
-    expected_block.created_ts = 0;
-    expected_block.merkle_root.fill(0);
-    expected_block.hash = iroha::hash(expected_block);
-    irohad->getCryptoProvider()->sign(expected_block);
+    const auto num_blocks =
+        transactions.size();  // Use one block per one transaction
+    setTestSubscribers(num_blocks);
+    std::for_each(
+        transactions.begin(), transactions.end(), [this](auto const &tx) {
+          // this-> is needed by gcc
+          this->sendTransaction(tx);
+          // wait for commit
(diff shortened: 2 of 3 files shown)


l4l:
It's fine that txs goes without order here. Imo it's better sort sent proposal and compare with the base one

motxx:
I think that's one of the options, on the other hand, this test is the integration test. So maybe we have to send transactions the same way as clients. i.e. we need to send transactions via Torii endpoint.

And, if sending transactions like:
```cpp
sendTransactions({tx1, tx2});
```
`tx2` depends on `tx1`. If `tx1` is not processed, `tx2` may be failed at stateful validation.

Also, we cannot send transactions like:
```cpp
sendTransactions({tx1});
sendTransactions({tx2});
```
Because `validate()` functions need to be called only once in one test. So,
```cpp
sendTransactions({tx1, tx2});
validate();
```
is ok, but
```cpp
sendTransactions({tx1});
validate();
sendTransactions({tx2});
validate();
```
causes the test to fail.

l4l:
The problem with the sleep is it doesn't solve the problem. Only increases chances of green test. This probably should be done either via sorting as I mentioned earlier, or via conditional variable.

And about your comment: maybe you're right, but it seems quite strange for me to predict order of asynchronous service. And dependent commands should be sent in one txs, so there's no "dependent txs"

That are my though, I also would like to hear any others from our team

motxx:
> That are my though, I also would like to hear any others from our team

Can anybody help?
`sendTransactions()` is created by @lebdron so could you help if possible? (ping)

lebdron:
That method is synchronous here, since Torii itself (gGRPC service) is not created, and the gRPC endpoint is called directly. What is asynchronous here - `OrderingGateTransportGrpc::propagate_transaction` method:
https://github.com/hyperledger/iroha/blob/15ffc458b9704e19353ddba58b3582cf2202eaa9/irohad/ordering/impl/ordering_gate_transport_grpc.cpp#L53-L58
Therefore, I would suggest to inherit from ordering service/ordering service transport, add condition variable to inherited class, and notify when `onTransaction` method is called to ensure transaction order.

motxx:
Sorry, my todo comment: `Replace this with StatusAsync() after we can use tx status ON_PROCESS.` is wrong.

If clients want to guarantee the order of transactions, they need to wait for `COMMITTED` from `StatusAsync()`. And we can use this method.

Is this not related to your comment? @lebdron

motxx:
This PR stopped because we may implement https://github.com/hyperledger/iroha/pull/709#issuecomment-346259310 modifying ordering service init code. This design is being discussed now.

motxx:
The scenario of sending transactions is no problem.

So I think:
1. Bad: Use`sleep()` to wait transactions are committed.
2. Not so good: Use `StatusAsync()` because it can guarantee transactions are committed.
3. Better: Change OrderingService (maybe OrderingInit?) implementation because polling is not used in the test.

I think it's better to use the way of 2 temporarily because I can quickly fix the problem.
And add TODO comment to replace it with the way of 3.

motxx:
In order to wait transactions are committed,
the way of `sleep(1)` is bad. And `StatusAsync()` requires polling so there are same problem with `sleep(1)`

The scenario is no problem but removing the test is better?

motxx:
@lebdron 
Custom implementation for OrderingService in Txpipeline test requires duplicate code and maybe it is difficult to manage. The custom implementation will be such like:
```cpp
// tx_pipeline_integration_test_fixture.hpp
class CustomOrderingServiceTransportGrpc final : public iroha::ordering::OrderingServiceTransportGrpc {
 public:
  grpc::Status OrderingServiceTransportGrpc::onTransaction(
      ......) {
    if (subscriber_.expired()) {
      log_->error("No subscriber");
    } else {
      subscriber_.lock()->onTransaction(*factory_.deserialize(*request));
      {
        std::lock_guard<std::mutex> lk(send_tx_mtx);
        send_tx_ready = true;
      }
      send_tx_cv.notify_one();
    }
    ......
  }
};

class CustomOrderingInit final : public iroha::network::OrderingInit {
 public:
  std::shared_ptr<ordering::OrderingGateImpl> initOrderingGate(
      const std::shared_ptr<ametsuchi::PeerQuery> &wsv,
      size_t max_size,
      std::chrono::milliseconds delay_milliseconds) override {
    ...
    ordering_service_transport = std::make_shared<ordering::CustomOrderingServiceTransportGrpc>();
    ...
  }
};
```

How to solve this problem?

Also, `sleep()` implementations is here, maybe better than polling:
https://github.com/hyperledger/iroha/blob/19651930fa6851586bfb9f09c3f6f2f3d3a51eaf/test/module/irohad/ordering/ordering_gate_service_test.cpp#L95

motxx:
@lebdron @l4l 
To sort transactions, I tried to implement `CustomOrderingServiceTransportGrpc` for test, but finally, I thought using `std::this_thread::sleep_for` is the best way.

The reasons are:
1. Sorting transactions requires change `OrderingInit` to an abstract class, which needs duplicate implementations only for initialization of ordering service classes. Using template class also changing codes tricky.
2. If we implement `CustomOrderingServiceTransportGrpc` inherited from `OrderingServiceTransportGrpc`, hiding `onTransaction()`, changing `OrderingServiceTransportGrpc` to an abstract class and duplicate implementations. Also, I think it's not a good way that changing `OrderingServiceTransportGrpc` functions to virtual ones like `application.hpp`, because of https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#alternative-to-mocking-concrete-classes

I found there is a similar case of this PR. It uses `this_thread::sleep_for` to confirm order. So changing this code is also needed if we need to implement sorting transactions in `OrderingServiceTransport`, which takes some cost and it's not so much effective.
https://github.com/hyperledger/iroha/blob/develop/test/module/irohad/ordering/ordering_gate_service_test.cpp#L94-L95

motxx:
I changed the code to use condition variable for waiting transactions.
When a transaction is committed, `cv.notify_all()` works via `PeerCommunicationService`

test/integration/pipeline/tx_pipeline_integration_test_fixture.hpp
```diff
@@ -167,13 +133,70 @@ class TxPipelineIntegrationTestFixture
...
+
+    // verify commit and block
+    commit_wrapper = std::make_unique<TestSubscriber<Commit>>(
+        make_test_subscriber<CallExact>(
+            irohad->getPeerCommunicationService()->on_commit(), num_blocks));
+    commit_wrapper->subscribe([this](auto commit) {
+      commit.subscribe([this](auto block) { blocks.push_back(block); });
+    });
+    irohad->getPeerCommunicationService()->on_commit().subscribe(
+        [this](auto) { cv.notify_all(); });
```
l4l: Why all? I guess it should be `notify_one()`

test/integration/pipeline/transfer_asset_inter_domain_test.cpp
```diff
@@ -74,10 +74,6 @@ class TransferAssetInterDomainTest : public TxPipelineIntegrationTestFixture {
     ASSERT_TRUE(irohad->storage);
 
     iroha::main::BlockInserter(irohad->storage).applyToLedger({genesis_block});
-    irohad->storage->getBlockQuery()->getTopBlocks(1).subscribe([](auto block) {
```
l4l: Why is this removed?

test/integration/pipeline/tx_pipeline_integration_test_fixture.hpp
```diff
@@ -93,65 +93,31 @@ class TxPipelineIntegrationTestFixture
...
-    ASSERT_EQ(1, proposals.size());
-    ASSERT_EQ(*expected_proposal, proposals.front());
-
-    ASSERT_TRUE(commit_wrapper.validate());
-    ASSERT_EQ(1, blocks.size());
-    ASSERT_EQ(expected_block, blocks.front());
+  void sendTxsInOrderAndValidate(
+      const std::vector<iroha::model::Transaction> &transactions) {
+    ASSERT_FALSE(duplicate_sent);
+    duplicate_sent = true;
```
l4l: Imo `std::atomic_bool` is useful here, just in case

test/integration/pipeline/transfer_asset_inter_domain_test.cpp
motxx: This is for debug so not needed.

testest
```diff
@@ -0,0 +1,37 @@
+Running main() from gtest_main.cc
```
l4l: Forgot to remove?

test/integration/pipeline/tx_pipeline_integration_test_fixture.hpp
```diff
@@ -93,65 +93,32 @@ class TxPipelineIntegrationTestFixture
...
-    ASSERT_EQ(1, proposals.size());
-    ASSERT_EQ(*expected_proposal, proposals.front());
-
-    ASSERT_TRUE(commit_wrapper.validate());
-    ASSERT_EQ(1, blocks.size());
-    ASSERT_EQ(expected_block, blocks.front());
+  void sendTxsInOrderAndValidate(
+      const std::vector<iroha::model::Transaction> &transactions) {
+    // test subscribers can't solve duplicate func call.
+    ASSERT_FALSE(duplicate_sent);
```
l4l: `ASSERT_FALSE(duplicate_sent.fetch_or(true));` or smth like that

testest
motxx: Yes. I'll just remove 🙇

test/integration/pipeline/tx_pipeline_integration_test_fixture.hpp
motxx: Thanks. If lack of this, atomic_bool is useless.

motxx:
Can anyone else review this?

PR #711, opened 2017-11-22
Add base class check for poly wrapper ctors; variant deserializer check

lebdron:
- Polymorphic wrapper now can be constructed only from base types or
 same type

 - Add variant types iteration for deserializer, allowing easier
 debugging of missing types from one of deserializers

### DIFF
--- a/shared_model/utils/variant_deserializer.hpp
+++ b/shared_model/utils/variant_deserializer.hpp
@@ -36,9 +36,10 @@ namespace shared_model {
         /**
          * Dummy deserializer
          * @tparam V variant type for deserialization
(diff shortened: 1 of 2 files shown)


shared_model/utils/variant_deserializer.hpp
```diff
@@ -54,19 +55,22 @@ namespace shared_model {
          * If type selector is 0, head type is required type, and it is used
          * Otherwise call helper without front element in type list
          * @tparam V variant type for deserialization
+         * @tparam T list of types in V, iterated as in S
```
muratovv: Rework documentation, pls.

PR #713, opened 2017-11-23
Feature/shared model bindings

luckychess:
## What is this pull request?
Bindings of Iroha crypto for Python and Java
   
Iroha uses custom signature algorithm (ed25519 with sha-3 inside) so we need support it for different languages. Bindings are the easiest way.
  
Bindings libraries are disabled by default. To generate them add ```-DSWIG=ON``` option to your cmake command, e. g. (if you are in build/ directory):
```cmake -DSWIG=ON ..```
Now you can build Iroha as usual. After the build process is completed navigate to ```build/shared_model/cryptography/ed25519_sha3_impl/bindings/```. You can use Iroha crypto from Python in a next way:
```
import irohacrypto
blob1 = irohacrypto.Blob("1")
blob2 = irohacrypto.Blob("2")
prov = irohacrypto.CryptoProvider()
keypair = prov.generateKeypair()
sig = prov.sign(blob1, keypair)
print(prov.verify(sig, blob1, keypair.publicKey()))
print(prov.verify(sig, blob2, keypair.publicKey()))
```
And from Java:
```
public class my {
  static {
    try {
        System.loadLibrary("irohacrypto");
    } catch (UnsatisfiedLinkError e) {
      System.err.println("Native code library failed to load. \n" + e);
      System.exit(1);
    }
  }

  public static void main(String argv[]) {
    Blob blob1 = new Blob("1");
    Blob blob2 = new Blob("2");
    CryptoProvider prov = new CryptoProvider();
    Keypair keypair = prov.generateKeypair();
    Signed sig = prov.sign(blob1, keypair);
    System.out.println(prov.verify(sig, blob1, keypair.publicKey()));
    System.out.println(prov.verify(sig, blob2, keypair.publicKey()));
  }
}
```
and then run
```
javac *.java
```
If everything has gone well you will receive the next result:
```
true
false
```
Also please note that shared object library for Python (_irohacrypto.so) is different from Java library (libirohacrypto.so) even though they provide exactly the same functionality.

**Note about Java:** if you run your Java code you've just built you will probably give the ```java.lang.UnsatisfiedLinkError``` exception. The reason is because Java runtime can't find native .so library (libirohacrypto.so in our case). Just use the next command:
```
LD_LIBRARY_PATH=. java my
```
where **my** is a name of your manually created file.

## Details/Features
- Add swig option into cmake
- Generate swig bindings if swig option is on for Java and Python

### CODE REVIEW

File: shared_model/cryptography/blob.hpp [hunk 1]
```diff
@@ -81,8 +81,13 @@ namespace shared_model {
        * Design note: this method is deprecated and should be removed after
        * migration to shared model in whole project
        */
+
       template <typename BlobType>
-      [[deprecated]] BlobType makeOldModel() const {
+#ifndef SWIG
+      [[deprecated]]
+#endif
```
Warchant: better to define a separate file `deprecated.hpp` with 
```C++
#ifndef SWIG
#define DEPRECATED [[deprecated]]
#else
#define DEPRECATED
#endif
```

and then use only DEPRECATED keyword

shared_model/cryptography/ed25519_sha3_impl/bindings/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+
+if (SWIG)
+
+add_definitions(-DSWIG)
+
+find_package(SWIG REQUIRED)
+INCLUDE(${SWIG_USE_FILE})
+
+find_package(PythonLibs REQUIRED)
```
Warchant: specify version, if possible

shared_model/cryptography/ed25519_sha3_impl/bindings/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+include_directories(${PYTHON_INCLUDE_PATH})
+
+find_package(JNI REQUIRED)
```
Warchant: specify version, if possible

shared_model/cryptography/ed25519_sha3_impl/bindings/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+include_directories(${JAVA_INCLUDE_PATH})
+include_directories(${JAVA_INCLUDE_PATH2})
```
Warchant: why 2?

shared_model/interfaces/primitive.hpp
```diff
@@ -45,7 +45,11 @@ namespace shared_model {
        * model
        * @return pointer for old-style object
        */
-      [[deprecated]] virtual OldModelType *makeOldModel() const = 0;
+#ifndef SWIG
+      [[deprecated]]
+#endif
```
Warchant: the same as above (update also here with `DEPRECATED`)

Warchant:
Is it possible to create a separate cmake target, instead of `SWIG` option?

And then change `include_directories(...)` to `target_include_directories`?

shared_model/cryptography/ed25519_sha3_impl/bindings/CMakeLists.txt
luckychess: JAVA_INCLUDE_PATH is the include path to jni.h, JAVA_INCLUDE_PATH2 is the include path to jni_md.h. Both are needed here.

shared_model/cryptography/ed25519_sha3_impl/bindings/CMakeLists.txt
luckychess: Unfortunately cmake provides no JNI version info.

luckychess:
About separate cmake target: SWIG option was introduced to make possible Iroha build without bindings (not every system has everything needed to it like our CI).

luckychess:
I agree about tests but we need install Swig and other libs into CI and turn in always on. Do you think all these changes worth it?

luckychess:
I think bindings should remain in shared_model because they are related only to shared model and their main function is providing easy access to shared model classes.

PR #716, opened 2017-11-24
Add generification of crypto provider;

muratovv:
## What is this pull request?
Now possible to use cryptography without explicit configuration of algorithm;

This PR provides separation of `crypto_provider` with `Signer` and `Verifier`. And make default algorithm specialization of it.

Additional changes:
* Rework hashable documentation
* Fix warnings in seed and keypair files
* Reworked crypto providers with static context
* Fix CMake dependencies
* Add usage tests

### DIFF
--- a/shared_model/cryptography/crypto_provider/crypto_signer.hpp
+++ b/shared_model/cryptography/crypto_provider/crypto_signer.hpp
@@ -15,34 +15,23 @@
  * limitations under the License.
  */
 
-#ifndef IROHA_HASH_PROVIDER_HPP
-#define IROHA_HASH_PROVIDER_HPP
+#ifndef IROHA_CRYPTO_SIGNER_HPP
+#define IROHA_CRYPTO_SIGNER_HPP
 
 #include "cryptography/blob.hpp"
-#include "cryptography/hash.hpp"
+#include "cryptography/crypto_provider/crypto_defaults.hpp"
+#include "cryptography/keypair.hpp"
+#include "cryptography/signed.hpp"
 
 namespace shared_model {
   namespace crypto {
-    /**
-     * Wrapper class for hashing.
-     */
-    class HashProvider {
+    template<typename Algorithm = DefaultCryptoAlgorithmType>
+    class CryptoSigner {
      public:
-      /**
-       * Hash with sha3-256
-       * @param blob - blob to hash
-       * @return Hash of provided blob
-       */
-      Hash sha3_256(const Blob &blob) const;
-
-      /**
-       * Hash with sha3-512
-       * @param blob - blob to hash
-       * @return Hash of provided blob
-       */
-      Hash sha3_512(const Blob &blob) const;
+      static Signed sign(const Blob &blob, const Keypair &keypair) {
+        return Algorithm::sign(blob, keypair);
+      }
     };
   }  // namespace crypto
 }  // namespace shared_model
-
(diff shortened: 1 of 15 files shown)


shared_model/cryptography/ed25519_sha3_impl/crypto_provider.cpp
```diff
@@ -17,36 +17,37 @@
...
+#include "cryptography/ed25519_sha3_impl/signer.hpp"
+#include "cryptography/ed25519_sha3_impl/verifier.hpp"
 
 namespace shared_model {
   namespace crypto {
 
-    Signed CryptoProvider::sign(const Blob &blob,
-                                const Keypair &keypair) const {
-      return signer_.sign(blob, keypair);
+    Signed CryptoProviderEd25519Sha3::sign(const Blob &blob, const Keypair &keypair) {
```
luckychess: Please apply clang-format.

shared_model/cryptography/ed25519_sha3_impl/verifier.hpp
```diff
@@ -28,9 +28,9 @@ namespace shared_model {
      */
     class Verifier {
```
luckychess: Clang-format feels abandoned =(

shared_model/cryptography/crypto_provider/crypto_signer.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace shared_model {
+  namespace crypto {
+    /**
+     * CryptoSigner - wrapper for generalization signing for different
+     * cryptographic algorithms
+     * @tparam Algorithm - cryptographic algorithm for singing
+     */
+    template <typename Algorithm = DefaultCryptoAlgorithmType>
+    class CryptoSigner {
```
l4l: I don't see any reason why the class is need here, can be just a static templated function. And if you do want to separate from other stuff use namespace instead 
Exactly the same below

lebdron: Using class instead of namespace restricts the methods to be defined in single file. This is applicable in this case since it is an interface.

muratovv: @l4l  Also, we have other benefits of using the class instead of proto object:
* We can't use methods without a class specifier, with namespaces we allow to use `using`;
* Separated functions provided different template interface because their template parameters are not bounded;

muratovv: >May you also add a test in CryptoInitialization for hard-coded values for signing and/or verifing (in case of both functions are broken test are going green)

This is not a test of cryptography, in fact, this test provides proof of concept for client-code, that there is not appears explicit specification of the crypto-algorithm.

l4l: @lebdron
>Using class instead of namespace restricts the methods to be defined in single file. This is applicable in this case since it is an interface.

just class methods are implicitly `inline`-d. It can be made explicitly, don't see any problems

@muratovv
> We can't use methods without a class specifier, with namespaces we allow to use using;

Don't see any profits, just more code in tests :(

> Separated functions provided different template interface because their template parameters are not bounded;

Agree, that can be pretty useful if we use some kind of global `using Default = CryptoSigner<DefaultCryptoAlgorithmType>`

PR #719, opened 2017-11-25
Fix missing boost link for validator; support ninja build with external

lebdron:
- Validator library was not linked with boost, leading to missing
 header files error. This error was not visible since most setups have
 boost installed in system directories, which are included by default

 - Replace INTERFACE_SYSTEM_INCLUDE_DIRECTORIES with
 INTERFACE_INCLUDE_DIRECTORIES, since prior does not add actual folder
 to include paths

 - Replace $(MAKE) with make for ninja build due to $ symbol error

 - Add BUILD_BYPRODUCTS to fix missing target error in ninja build


l4l:
Build fine for me :+1:

PR #723, opened 2017-11-27
Feature/txresp fix proto

l4l:
This pr introduces a hash inside ToriiResponse that require for #712

Besides there's a test that checks hash in torii service output

### How to check that it works

`make -j10 && test_bin/torii_service_test`
`test_bin/shared_proto_tx_response_test` remain unbroken (just added additional check for hash)

### DIFF
--- a/irohad/torii/impl/command_service.cpp
+++ b/irohad/torii/impl/command_service.cpp
@@ -14,12 +14,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include <endpoint.pb.h>
-#include "cryptography/ed25519_sha3_impl/internal/sha3_hash.hpp"
 #include "torii/command_service.hpp"
 #include "common/types.hpp"
+#include "cryptography/ed25519_sha3_impl/internal/sha3_hash.hpp"
+#include "endpoint.pb.h"
 
 namespace torii {
+
+  using iroha::model::TransactionResponse;
+  using namespace iroha::protocol;
 
   CommandService::CommandService(
       std::shared_ptr<iroha::model::converters::PbTransactionFactory>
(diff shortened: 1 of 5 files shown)


irohad/torii/impl/command_service.cpp
```diff
@@ -14,72 +14,70 @@ See the License for the specific language governing permissions and
...
+            case TransactionResponse::COMMITTED:
+              res->second.set_tx_status(TxStatus::COMMITTED);
+              break;
+            case TransactionResponse::ON_PROCESS:
+              res->second.set_tx_status(TxStatus::ON_PROCESS);
+              break;
+            case TransactionResponse::NOT_RECEIVED:
+              res->second.set_tx_status(TxStatus::NOT_RECEIVED);
+              break;
+          }
```
Warchant: default case?

l4l: fxd in c7a287e

irohad/torii/impl/command_service.cpp
```diff
@@ -14,72 +14,71 @@ See the License for the specific language governing permissions and
...
+  void CommandService::ToriiAsync(Transaction const &request,
                                   google::protobuf::Empty &empty) {
     auto iroha_tx = pb_factory_->deserialize(request);
 
     auto tx_hash = iroha::hash(*iroha_tx).to_string();
 
-    iroha::protocol::ToriiResponse response;
-    response.set_tx_status(iroha::protocol::TxStatus::ON_PROCESS);
+    ToriiResponse response;
+    response.set_tx_hash(tx_hash);
```
muratovv: Insertion of hash and status looks like a separated function, maybe refactor it?

PR #761, opened 2017-12-12
Feature/shared model bindings extend interface

luckychess:
## What is this pull request?
Extending Swig interface with generating keypair based on provided seed string and using existing keypair.
   
These features are needed for Iroha lib clients.
  
```
import iroha
crypto = iroha.ModelCrypto()
 # seed string length should be equal 32
kp_gen_test = crypto.generateKeypair("somerandomstring0123456789abcdef"); 
kp_test = crypto.convertFromExisting("227a458045f8d7d520907fb400461ae3c9101a68af2851d56a85b96fd791911a", "1011098aebf697597a59151a0ba1bb34bcf081f5195fa8de551b9a2b52be864613aae9771f35df7318e0d7b8571f043cd54581a0b1c5cf36fb05ac8956d291fe");
```
Same from Java. If provided keypair will not pass validation, exception will be thrown.

## Details/Features
- Extend Swig API with seed keypair generation
- Create convertFromExisting() method in ModelCrypto class
- Add static method to create Blobs from hex strings
- Add Blob argument constructors to key classes
- Add generate random string function
- Move some functions from types.hpp to byteutils.hpp

### DIFF
--- a/iroha-cli/interactive/impl/interactive_status_cli.cpp
+++ b/iroha-cli/interactive/impl/interactive_status_cli.cpp
@@ -17,6 +17,7 @@
 
 #include "interactive/interactive_status_cli.hpp"
 #include "client.hpp"
+#include "common/byteutils.hpp"
 
 namespace iroha_cli {
   namespace interactive {
@@ -152,5 +153,5 @@ namespace iroha_cli {
     std::string InteractiveStatusCli::parseGetHash(ActionParams params) {
       return params[0];
     }
-  }
-}
+  }  // namespace interactive
+}  // namespace iroha_cli
--- a/libs/common/byteutils.hpp
+++ b/libs/common/byteutils.hpp
@@ -18,12 +18,9 @@
 #ifndef IROHA_BYTEUTILS_H
 #define IROHA_BYTEUTILS_H
 
-#include <string>
-
+#include <algorithm>
 #include <nonstd/optional.hpp>
-
-#include "crypto/base64.hpp"
-
+#include <string>
 #include "common/types.hpp"
 
 namespace iroha {
@@ -97,11 +94,11 @@ namespace iroha {
   }
 
   /**
-  * Convert hexstring to array of given size
-  * @tparam size - output array size
-  * @param string - input string for transform
-  * @return array of given size if size matches, nullopt otherwise
-  */
+   * Convert hexstring to array of given size
+   * @tparam size - output array size
+   * @param string - input string for transform
+   * @return array of given size if size matches, nullopt otherwise
+   */
   template <size_t size>
   nonstd::optional<blob_t<size>> hexstringToArray(const std::string &string) {
     return hexstringToBytestring(string) | stringToBlob<size>;
--- a/libs/common/byteutils.hpp
+++ b/libs/common/byteutils.hpp
@@ -34,7 +34,7 @@ namespace iroha {
    * @param s - string to convert
    * @return blob, if conversion was successful, otherwise nullopt
    */
-  template<size_t size>
+  template <size_t size>
   nonstd::optional<blob_t<size>> stringToBlob(const std::string &string) {
     if (size != string.size()) {
       return nonstd::nullopt;
@@ -45,15 +45,68 @@ namespace iroha {
   }
 
   /**
-   * Convert hexstring to array of given size
-   * @tparam size - output array size
-   * @param string - input string for transform
-   * @return array of given size if size matches, nullopt otherwise
+   * Convert string of raw bytes to printable hex string
+   * @param str
+   * @return
    */
+  inline std::string bytestringToHexstring(const std::string &str) {
+    std::stringstream ss;
+    ss << std::hex << std::setfill('0');
+    for (const auto &c : str) {
+      ss << std::setw(2) << (static_cast<int>(c) & 0xff);
+    }
+    return ss.str();
+  }
+
+  /**
+   * Convert printable hex string to string of raw bytes
+   * @param str
+   * @return
+   */
+  inline nonstd::optional<std::string> hexstringToBytestring(
+      const std::string &str) {
+    if (str.empty() or str.size() % 2 != 0) {
+      return nonstd::nullopt;
+    }
+    std::string result(str.size() / 2, 0);
+    for (size_t i = 0; i < result.length(); ++i) {
+      std::string byte = str.substr(i * 2, 2);
+      try {
+        result.at(i) = std::stoul(byte, nullptr, 16);
+      } catch (const std::invalid_argument &e) {
+        return nonstd::nullopt;
+      } catch (const std::out_of_range &e) {
+        return nonstd::nullopt;
+      }
+    }
+    return result;
+  }
+
+  /**
+   * Generates new random string from lower-case letters
+   * @param len - size of string to generate
+   * @return generated string
+   */
+  inline std::string randomString(int len) {
+    std::string str(len, 0);
+    std::generate_n(str.begin(), len, []() {
+      char c = 'a';
+      return c + std::rand() % ('z' - 'a' + 1);
+    });
+    return str;
+  }
+
+  /**
+  * Convert hexstring to array of given size
+  * @tparam size - output array size
+  * @param string - input string for transform
+  * @return array of given size if size matches, nullopt otherwise
+  */
   template <size_t size>
   nonstd::optional<blob_t<size>> hexstringToArray(const std::string &string) {
     return hexstringToBytestring(string) | stringToBlob<size>;
   }
-}
+
+}  // namespace iroha
 
 #endif  // IROHA_BYTEUTILS_H
--- a/libs/common/byteutils.hpp
+++ b/libs/common/byteutils.hpp
@@ -43,8 +43,8 @@ namespace iroha {
 
   /**
    * Convert string of raw bytes to printable hex string
-   * @param str
-   * @return
+   * @param str - raw bytes string to convert
+   * @return - converted hex string
    */
   inline std::string bytestringToHexstring(const std::string &str) {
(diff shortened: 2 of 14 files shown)


libs/common/byteutils.hpp
```diff
@@ -44,6 +41,58 @@ namespace iroha {
     return array;
   }
 
+  /**
+   * Convert string of raw bytes to printable hex string
+   * @param str
+   * @return
+   */
+  inline std::string bytestringToHexstring(const std::string &str) {
```
l4l: What's the point?

libs/common/byteutils.hpp
```diff
@@ -44,6 +41,58 @@ namespace iroha {
...
+    }
+    return result;
+  }
+
+  /**
+   * Generates new random string from lower-case letters
+   * @param len - size of string to generate
+   * @return generated string
+   */
+  inline std::string randomString(int len) {
```
l4l: Shouldn't be here

libs/common/byteutils.hpp
```diff
@@ -44,6 +41,58 @@ namespace iroha {
...
+    std::string str(len, 0);
+    std::generate_n(str.begin(), len, []() {
+      char c = 'a';
```
l4l: redundant var, inline it

shared_model/bindings/model_crypto.cpp
```diff
@@ -21,8 +21,31 @@
 namespace shared_model {
   namespace bindings {
     crypto::Keypair ModelCrypto::generateKeypair() {
-      // TODO: 06.12.2017 luckychess create keypair generation interface IR-684
       return crypto::CryptoProviderEd25519Sha3::generateKeypair();
     }
+
+    crypto::Keypair ModelCrypto::generateKeypair(const std::string &seed) {
+      return crypto::CryptoProviderEd25519Sha3::generateKeypair(
+          crypto::Seed(seed));
```
l4l: Maybe just remove explicit keyword in Seed's constructor? It's quite obvious from the code what's happening

shared_model/bindings/model_crypto.cpp
```diff
@@ -21,8 +21,31 @@
 namespace shared_model {
   namespace bindings {
     crypto::Keypair ModelCrypto::generateKeypair() {
-      // TODO: 06.12.2017 luckychess create keypair generation interface IR-684
       return crypto::CryptoProviderEd25519Sha3::generateKeypair();
     }
+
+    crypto::Keypair ModelCrypto::generateKeypair(const std::string &seed) {
+      return crypto::CryptoProviderEd25519Sha3::generateKeypair(
```
l4l: Add some using stuff here, it's cpp file

shared_model/bindings/model_crypto.cpp
```diff
@@ -21,8 +21,31 @@
...
+          crypto::Seed(seed));
+    }
+
+    crypto::Keypair ModelCrypto::convertFromExisting(
+        const std::string &publicKey, const std::string &privateKey) {
+      crypto::Keypair keypair((crypto::Keypair::PublicKeyType(
```
l4l: redundant bracket?

shared_model/bindings/model_crypto.cpp
```diff
@@ -21,8 +21,31 @@
...
       return crypto::CryptoProviderEd25519Sha3::generateKeypair();
     }
+
+    crypto::Keypair ModelCrypto::generateKeypair(const std::string &seed) {
+      return crypto::CryptoProviderEd25519Sha3::generateKeypair(
+          crypto::Seed(seed));
+    }
+
+    crypto::Keypair ModelCrypto::convertFromExisting(
+        const std::string &publicKey, const std::string &privateKey) {
```
l4l: snake_case plz

shared_model/cryptography/blob.hpp
```diff
@@ -39,12 +40,19 @@ namespace shared_model {
...
+
+      /**
+       * Creates new Blob object from provided hex string
+       * @param hex - string in hex format to create Blob from
+       * @return created Blob object
+       */
+      static Blob fromHexString(const std::string &hex) {
+        using iroha::operator|;
+        Blob b("");
+        iroha::hexstringToBytestring(hex) | [&](auto &s){b = Blob(s);};
```
l4l: `auto &&s`?

shared_model/cryptography/keypair.hpp
```diff
@@ -38,8 +38,8 @@ namespace shared_model {
       /// Type of private key
       using PrivateKeyType = PrivateKey;
 
-      explicit Keypair(PublicKeyType publickKey, PrivateKeyType privateKey)
-          : publicKey_(publickKey), privateKey_(privateKey) {}
+      explicit Keypair(PublicKeyType publicKey, PrivateKeyType privateKey)
```
l4l: snake_case again

libs/common/byteutils.hpp
luckychess: Well, I think there are no better place in a whole project and I do not want to create one more file for just one stupid function.

shared_model/bindings/model_crypto.cpp
luckychess: I think we can create a template for generateKeypair() method with default parameter "Ed25519sha3".

shared_model/bindings/model_crypto.cpp
luckychess: Surprisingly, but no :) Without double brackets it will not compile, say hello to https://en.wikipedia.org/wiki/Most_vexing_parse

libs/common/byteutils.hpp
luckychess: Main point is to allow users to pass their keys in hex format while binary strings are really uncomfortable to work with.
Btw, tests for converters exist but in develop branch: https://github.com/hyperledger/iroha/pull/714/files.

l4l: Sorry for being vague, I mean what's the point of moving it out? It seems a bit out of that pr

shared_model/bindings/model_crypto.cpp
l4l: holy mango of charango
this lang is awful D:

libs/common/byteutils.hpp
luckychess: The problem was we need converters in Blob. Initially is contained only one of them and I decided to copy it from types.hpp - definitely not the best decision ever. Now I need one more converter so I decided to fix my old mistake and call function instead of copy its content. I really do not like their current place in types.hpp - it's really overloaded by functions of different purposes which makes almost all project parts dependent of types.hpp (you can do a simple test - change something there and almost all project will be rebuilt). So I decided to do a small refactor just here.

shared_model/bindings/model_crypto.cpp
luckychess: Codacy blames for one argument constructors without explicit keyword (in rare cases when it works =) )

shared_model/bindings/model_crypto.cpp
neewy: cppcoreguidelines or why?

luckychess: Google coding style

shared_model/bindings/model_crypto.hpp
```diff
@@ -34,8 +34,23 @@ namespace shared_model {
...
+       */
+      crypto::Keypair generateKeypair(const std::string &seed);
+
+      /**
+       * Retrieves Keypair object (ed25519) from existing keypair.
+       * @param publicKey - ed25519 hex-encoded public key
+       * @param privateKey - ed25519 hex-encoded private key
+       * @return keypair from provided keys
+       */
+      crypto::Keypair convertFromExisting(const std::string &publicKey,
```
l4l: Missed fix for the case here

l4l:
It also would be nice if you fix sorabot's blockers

libs/common/byteutils.hpp
muratovv: Literally, all think like you and this idea bring a chaos to project. =(

shared_model/cryptography/blob.hpp
```diff
@@ -39,12 +40,19 @@ namespace shared_model {
...
+      }
+
+      /**
+       * Creates new Blob object from provided hex string
+       * @param hex - string in hex format to create Blob from
+       * @return created Blob object
+       */
+      static Blob fromHexString(const std::string &hex) {
+        using iroha::operator|;
+        Blob b("");
```
muratovv: Need to write in the documentation about default case, when passed hex string is broken.

shared_model/cryptography/keypair.hpp
```diff
@@ -38,18 +38,18 @@ namespace shared_model {
       /// Type of private key
       using PrivateKeyType = PrivateKey;
 
-      explicit Keypair(PublicKeyType publickKey, PrivateKeyType privateKey)
-          : publicKey_(publickKey), privateKey_(privateKey) {}
+      explicit Keypair(PublicKeyType public_key, PrivateKeyType private_key)
```
muratovv: I think better to pass `const PublicKeyType&` here. Same with the private key.

iroha-cli/interactive/impl/interactive_status_cli.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include "interactive/interactive_status_cli.hpp"
 #include "client.hpp"
+#include "common/byteutils.hpp"
```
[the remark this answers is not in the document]
luckychess: Because hexstringToBytestring() was moved. Unfortunately old types.hpp still remains included indirectly somewhere deep in client.hpp.

l4l:
Rebase plz

PR #762, opened 2017-12-13
Fix/iroha docker

### LINKED ISSUE #749: Iroha daemon fails to start in Ansible environment
**Given** macOS Sierra with ansible installed via brew, and commit of development branch https://github.com/hyperledger/iroha/commit/c81dd577b9779de6a679b5d6ea3bc38b61bc2114

**When** the Ansible network of peers is deployed according to https://soramitsu.atlassian.net/wiki/spaces/IS/pages/127008772/How+to+run+Iroha+network+with+ansible guide and I am attaching to any of the peers running docker containers with Iroha, redis and Postgres,

**Then** docker container with Iroha does not start:
```
iroha@instance-group-4-1c06:~$ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
85956175c272        redis:3.2.8         "docker-entrypoint.s…"   2 minutes ago       Up 2 minutes        6379/tcp            iroha_redis
13e2b81030dc        postgres:9.5        "docker-entrypoint.s…"   3 minutes ago       Up 3 minutes        5432/tcp            iroha_postgres
```
**and** docker logs contain:
```
iroha@instance-group-4-1c06:~$ docker logs iroha
key=node0
/opt/iroha_data
iroha-cli: error while loading shared libraries: libboost_filesystem.so.1.65.1: cannot open shared object file: No such file or directory
irohad: error while loading shared libraries: libpq.so.5: cannot open shared object file: No such file or directory
```

x3medima17:
## What is this pull request?
This PR fixes #749 issue
It adds boost library into deb package, also it adds pq library as deb dependency.


cmake/dependencies.cmake
```diff
@@ -88,6 +88,12 @@ set_target_properties(boost PROPERTIES
     INTERFACE_LINK_LIBRARIES "${Boost_LIBRARIES}"
     )
 
+if(ENABLE_LIBS_PACKAGING)
+  foreach (library ${Boost_LIBRARIES})
+    add_install_step_for_lib(${library})
```
l4l: Looks weird for me. What's the point of packing boost libs?
Explain plz

x3medima17: Because when we build deb package, it does not install boost libraries, and on a fresh system it will lack boost (see #749 )

l4l: Quite strange way of fixing. Maybe it's possible to set dependencies in the deb package itself?

x3medima17: We need a specific version of boost that is used during build

PR #768, opened 2017-12-14
Fix transaction validation for empty commands

luckychess:
## What is this pull request?
Fix for empty transaction validation
   
Transaction without commands are not allowed
  
Test included

## Details/Features
- Add validation step for zero commands in transaction case
- Add test for empty transaction validation

### DIFF
--- a/shared_model/validators/commands_validator.hpp
+++ b/shared_model/validators/commands_validator.hpp
@@ -90,7 +90,7 @@ namespace shared_model {
 
         ReasonsGroupType operator()(
             const detail::PolymorphicWrapper<interface::CreateAccount> &ca)
-            const {;
+            const {
           ReasonsGroupType reason;
           reason.first = "CreateAccount";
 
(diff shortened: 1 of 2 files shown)


shared_model/validators/commands_validator.hpp
```diff
@@ -90,7 +90,8 @@ namespace shared_model {
 
         ReasonsGroupType operator()(
             const detail::PolymorphicWrapper<interface::CreateAccount> &ca)
-            const {;
+            const {
+          ;
```
l4l: Remove it plz D:

luckychess: Ooops )

test/module/shared_model/validators/commands_validator_test.cpp
```diff
@@ -193,6 +193,20 @@ iroha::protocol::RevokePermission generateRevokePermission(
...
+ * @then answer has error about empty transaction
+ */
+TEST(commandsValidatorTest, EmptyTransactionTest) {
+  auto tx = generateEmptyTransaction();
+  tx.mutable_payload()->set_created_time(iroha::time::now());
+  shared_model::validation::CommandsValidator commands_validator;
+  auto answer = commands_validator.validate(
+          detail::make_polymorphic<proto::Transaction>(tx));
+  ASSERT_TRUE(answer.hasErrors());
+}
```
kamilsa: We can also check how many errors did we get from answer. Like here https://github.com/hyperledger/iroha/pull/768/files#diff-c76782c8a0c46a09b1c21cbbe5dccb3bR332

PR #775, opened 2017-12-15
Make transaction builder const

muratovv:
## What is this pull request?
This PR provides a constant interface for transaction builder.

## Details/Features
* make all generation methods of builder constant
* fix bug (silently) when on setter copy default stateless validator
* fix interface of stateless validator as a constant

### DIFF
--- a/shared_model/builders/protobuf/transaction.hpp
+++ b/shared_model/builders/protobuf/transaction.hpp
@@ -99,14 +99,14 @@ namespace shared_model {
 
       NextBuilder<TxCounter> txCounter(
           Transaction::TxCounterType tx_counter) const {
-        return transform<CreatorAccountId>([&](auto &tx) {
+        return transform<TxCounter>([&](auto &tx) {
           tx.mutable_payload()->set_tx_counter(tx_counter);
         });
       }
 
       NextBuilder<CreatedTime> createdTime(
           interface::types::TimestampType created_time) const {
-        return transform<CreatorAccountId>([&](auto &tx) {
+        return transform<CreatedTime>([&](auto &tx) {
           tx.mutable_payload()->set_created_time(created_time);
         });
       }
--- a/shared_model/builders/protobuf/transaction.hpp
+++ b/shared_model/builders/protobuf/transaction.hpp
@@ -115,11 +115,11 @@ namespace shared_model {
           const interface::types::AccountIdType &account_id,
           const interface::types::AssetIdType &asset_id,
           const std::string &amount) const {
-        return addCommand([&](auto command) {
-          auto cmd = command->mutable_add_asset_quantity();
-          cmd->set_account_id(account_id);
-          cmd->set_asset_id(asset_id);
-          addAmount(cmd->mutable_amount(), amount);
+        return addCommand([&](auto proto_command) {
+          auto command = proto_command->mutable_add_asset_quantity();
+          command->set_account_id(account_id);
+          command->set_asset_id(asset_id);
+          addAmount(command->mutable_amount(), amount);
         });
       }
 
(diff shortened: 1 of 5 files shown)


shared_model/builders/protobuf/transaction.hpp
```diff
@@ -46,131 +47,171 @@ namespace shared_model {
...
           const interface::types::AssetIdType &asset_id,
-          const std::string &amount) {
-        auto command = proto_command()->mutable_add_asset_quantity();
-        command->set_account_id(account_id);
-        command->set_asset_id(asset_id);
-        addAmount(command->mutable_amount(), amount);
-        return *this;
+          const std::string &amount) const {
+        return addCommand([&](auto command) {
+          auto cmd = command->mutable_add_asset_quantity();
```
kamilsa: to be consistent with other commands could be renamed: command->proto_command, cmd->command

shared_model/builders/protobuf/transaction.hpp
```diff
@@ -46,131 +47,173 @@ namespace shared_model {
...
+       */
+      template <typename Transformation>
+      NextBuilder<Command> addCommand(Transformation t) const {
+        auto copy = transaction_;
+        t(copy.mutable_payload()->add_commands());
+        return {copy, stateless_validator_};
+      }
 
      public:
       TemplateTransactionBuilder() = default;
```
lebdron: The following is required to allow parametrized stateless validator:
```C++
TemplateTransactionBuilder(const SV &validator = SV())
          : stateless_validator_(validator) {}
```

shared_model/builders/protobuf/transaction.hpp
```diff
@@ -46,131 +47,173 @@ namespace shared_model {
...
-      SV stateless_validator_;
+      /**
+       * Make add command transformation on copied object
+       * @tparam Transformation - callable type for changing command
+       * @param f - transform function for proto command
+       * @return new builder with added command
+       */
+      template <typename Transformation>
+      NextBuilder<Command> addCommand(Transformation t) const {
+        auto copy = transaction_;
```
lebdron: Maybe copying `this`, modifying it, and returning will be a prettier solution, so that `stateless_validator_` will not explicitly appear.

shared_model/builders/protobuf/transaction.hpp
```diff
@@ -46,131 +47,173 @@ namespace shared_model {
...
+
+      /**
+       * Make transformation on copied content
+       * @tparam Transformation - callable type for changing the copy
+       * @param f - transform function for proto object
+       * @return new builder with updated state
+       */
+      template <int Fields, typename Transformation>
+      NextBuilder<Fields> transform(Transformation t) const {
+        auto copy = transaction_;
```
lebdron:
```C++
auto copy = *this;
t(copy.transaction_);
return copy;
```
Private constructor with prototx and sv is not required in this case.

shared_model/builders/protobuf/transaction.hpp
```diff
@@ -46,131 +47,173 @@ namespace shared_model {
...
+          tx.mutable_payload()->set_creator_account_id(account_id);
+        });
       }
 
-      NextBuilder<TxCounter> txCounter(Transaction::TxCounterType tx_counter) {
-        transaction_.mutable_payload()->set_tx_counter(tx_counter);
-        return *this;
+      NextBuilder<TxCounter> txCounter(
+          Transaction::TxCounterType tx_counter) const {
+        return transform<CreatorAccountId>([&](auto &tx) {
```
lebdron: Should be `transform<TxCounter>`. Same with `CreatedTime` below. Maybe change return type to `auto` to write this only once?

PR #778, opened 2017-12-17
Add dynamic port selection to several tests

nickaleks:
## What is this pull request?
This pull request removes hardcoded ports from several test classes, and replaces them with dynamic port selection
   
Hardcoded ports can be already taken by something else, which produces test failure. Such a failure is hard to detect and reproduce, so it is best to avoid such issues.
  

## Details/Features
- Add dynamic ports to block_loader_test and ordering_gate_test

## P.S. (optional)
I was not able to change several tests, because address needs to be known beforehand for them to work. This means that adding dynamic port requires major changes in core classes.

### DIFF
--- a/test/integration/pipeline/tx_pipeline_integration_test.cpp
+++ b/test/integration/pipeline/tx_pipeline_integration_test.cpp
@@ -17,6 +17,7 @@
 
 #include "integration/pipeline/tx_pipeline_integration_test_fixture.hpp"
 
+// TODO: refactor services to allow dynamic port binding IR-741
 class TxPipelineIntegrationTest : public TxPipelineIntegrationTestFixture {
  public:
   void SetUp() override {
--- a/test/module/irohad/network/block_loader_test.cpp
+++ b/test/module/irohad/network/block_loader_test.cpp
@@ -39,9 +39,6 @@ using testing::A;
 class BlockLoaderTest : public testing::Test {
  public:
   void SetUp() override {
-    peer = Peer();
-    peer.address = "0.0.0.0:50051";
-    peers.push_back(peer);
     peer_query = std::make_shared<MockPeerQuery>();
     storage = std::make_shared<MockBlockQuery>();
     provider = std::make_shared<MockCryptoProvider>();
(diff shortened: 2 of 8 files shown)


nickaleks:
https://soramitsu.atlassian.net/browse/IR-741

test/module/irohad/network/block_loader_test.cpp
```diff
@@ -50,14 +47,20 @@ class BlockLoaderTest : public testing::Test {
...
     grpc::ServerBuilder builder;
     int port = 0;
-    builder.AddListeningPort(peer.address, grpc::InsecureServerCredentials(),
+    builder.AddListeningPort(address, grpc::InsecureServerCredentials(),
                              &port);
     builder.RegisterService(service.get());
     server = builder.BuildAndStart();
+
+    peer = Peer();
+    peer.address = "0.0.0.0:" + std::to_string(port);
```
l4l: Why it's `0.0.0.0` by the way? I think it should be set to localhost

nickaleks: It was in all tests, so I decided not to change it

nickaleks: Sonarcube issue fixed

l4l: Yeah, I know. Question is for @lebdron

test/integration/pipeline/tx_pipeline_integration_test.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include "integration/pipeline/tx_pipeline_integration_test_fixture.hpp"
 
+// TODO: refactor services to allow dynamic port binding IR-741
```
muratovv: There is need to refactor test or code?

nickaleks: Firstly code then tests. I can add TODO to code as well, but I think this todo in tests will allow tracing problem better.

test/module/irohad/network/block_loader_test.cpp
lebdron: https://en.wikipedia.org/wiki/0.0.0.0
>A way to specify "any IPv4 address at all". It is used in this way when configuring servers (i.e. when binding listening sockets). 

This seems more applicable than a special address like 127.0.0.1

l4l: "Address at all" basically means that it binds to **every** available address. That means host is going to be publicly available in the all related networks. Don't really sure it's secure (nevertheless tests are quite fast)

muratovv:
Pls, fix the build. After it, I will approve changes.

PR #781, opened 2017-12-18
Integrate warchant/ed25519

Warchant:
This PR adds fixes related to integration of https://github.com/warchant/ed25519 library.

Users may install their own version of ed25519 and hash. 
Since different implementations are ABI-compatible, user is able to install the fastest implementation `amd64-64-24k` and use it, while other user (with armv7 arch, for example) can install different (portable ref10) implementation.


Warchant:
Sad :( It seems I have to add automatic downloading of the ed lib.

libs/crypto/CMakeLists.txt
```diff
@@ -15,31 +15,31 @@
...
+add_library(sha3_impl
     hash.cpp
     )
+target_link_libraries(sha3_impl PUBLIC
+    ed25519 # contains hash
 
-target_link_libraries(hash
-    ed25519
-    common
+    # TODO: should not be here
```
neewy: Put TODO in the backlog

Warchant:
I will reopen when finish with cmake script.

PR #788, opened 2017-12-19
Feature/apppend role check

x3medima17:
Thanks for the pull request!
Pull requests should have these details:

## What is this pull request?
This pull requests adds an additional check on Append Role Executor. It checks if assigner has enough permissions to give a certain role to assignee.

If assigner X has a set of permissions **A** and want to append role with set **B** of permissions to the user Y. 
This command must succeed if and only if **B** is a subset of **A**. This PR performs this check.

### DIFF
--- a/irohad/model/execution/command_executor.hpp
+++ b/irohad/model/execution/command_executor.hpp
@@ -94,6 +94,8 @@ namespace iroha {
 
       bool isValid(const Command &command,
                    ametsuchi::WsvQuery &queries) override;
+    private:
+      Account creator_;
     };
 
     class DetachRoleExecutor : public CommandExecutor {
--- a/irohad/model/execution/command_executor.hpp
+++ b/irohad/model/execution/command_executor.hpp
@@ -113,8 +113,6 @@ namespace iroha {
       bool isValid(const Command &command,
                    ametsuchi::WsvQuery &queries) override;
 
-     private:
-      Account creator_;
     };
 
     class CreateRoleExecutor : public CommandExecutor {
--- a/irohad/model/execution/command_executor.hpp
+++ b/irohad/model/execution/command_executor.hpp
@@ -110,6 +110,9 @@ namespace iroha {
 
       bool isValid(const Command &command,
                    ametsuchi::WsvQuery &queries) override;
+
+     private:
+      Account creator_;
     };
 
     class CreateRoleExecutor : public CommandExecutor {
--- a/irohad/model/execution/impl/command_executor.cpp
+++ b/irohad/model/execution/impl/command_executor.cpp
@@ -18,7 +18,6 @@
 #include "model/execution/command_executor.hpp"
 #include <algorithm>
 #include "model/commands/add_asset_quantity.hpp"
-#include "model/commands/subtract_asset_quantity.hpp"
 #include "model/commands/add_peer.hpp"
 #include "model/commands/add_signatory.hpp"
 #include "model/commands/append_role.hpp"
@@ -32,6 +31,7 @@
 #include "model/commands/revoke_permission.hpp"
 #include "model/commands/set_account_detail.hpp"
 #include "model/commands/set_quorum.hpp"
+#include "model/commands/subtract_asset_quantity.hpp"
 #include "model/commands/transfer_asset.hpp"
 #include "model/execution/common_executor.hpp"
 #include "model/permissions.hpp"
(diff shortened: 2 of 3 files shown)


irohad/model/execution/impl/command_executor.cpp
```diff
@@ -73,13 +72,34 @@ namespace iroha {
     bool AppendRoleExecutor::hasPermissions(const Command &command,
                                             ametsuchi::WsvQuery &queries,
                                             const Account &creator) {
+      creator_ = creator;
```
l4l: What's the point of copying?

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -73,13 +72,34 @@ namespace iroha {
...
 
     bool AppendRoleExecutor::isValid(const Command &command,
                                      ametsuchi::WsvQuery &queries) {
-      // TODO 26/09/17 grimadas: check. No additional checks required
+      auto creator = creator_;
+      auto cmd_value = static_cast<const AppendRole &>(command);
+      auto role_permissions = queries.getRolePermissions(cmd_value.role_name);
+      auto account_roles = queries.getAccountRoles(creator.account_id);
+
+      if (not role_permissions.has_value() or not account_roles.has_value())
```
l4l: Add brackets plz

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -73,13 +72,34 @@ namespace iroha {
...
+        return false;
+
+      std::set<std::string> account_permissions;
+      for (const auto &role : *account_roles) {
```
l4l: Rewrite in functional way plz with boost::filtered

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -73,13 +72,34 @@ namespace iroha {
...
+        auto permissions = queries.getRolePermissions(role);
+        if (not permissions.has_value())
+          continue;
+        for (const auto &permission : *permissions)
+          account_permissions.insert(permission);
+      }
+
+      for (const auto &role_permission : *role_permissions)
```
l4l: There's a [method](http://en.cppreference.com/w/cpp/algorithm/all_any_none_of) for this

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -73,13 +72,34 @@ namespace iroha {
...
                                             const Account &creator) {
+      creator_ = creator;
       return checkAccountRolePermission(
           creator.account_id, queries, can_append_role);
     }
 
     bool AppendRoleExecutor::isValid(const Command &command,
                                      ametsuchi::WsvQuery &queries) {
-      // TODO 26/09/17 grimadas: check. No additional checks required
+      auto creator = creator_;
```
l4l: Same here

grimadas: This is not even needed, since you have a private field in the same class

irohad/model/execution/command_executor.hpp
```diff
@@ -110,6 +112,9 @@ namespace iroha {
 
       bool isValid(const Command &command,
                    ametsuchi::WsvQuery &queries) override;
+
+     private:
+      Account creator_;
```
grimadas: This is not used anywhere. Consider removing

irohad/model/execution/impl/command_executor.cpp
x3medima17: Hm, reasonable

irohad/model/execution/command_executor.hpp
x3medima17: It is used  in isValid method

irohad/model/execution/impl/command_executor.cpp
x3medima17: Object `creator` is located on the stack, I have to save it for future use

irohad/model/execution/impl/command_executor.cpp
```diff
@@ -309,62 +309,63 @@ namespace iroha {
...
-      bool SubtractAssetQuantityExecutor::hasPermissions(const Command &command,
-                                                    WsvQuery &queries,
-                                                    const Account &creator) {
-        auto cmd_value = static_cast<const SubtractAssetQuantity &>(command);
-        return creator.account_id == cmd_value.account_id
-               and checkAccountRolePermission(
-          creator.account_id, queries, can_subtract_asset_qty);
-      }
+      // accountAsset.value().balance -= amount;
+      return commands.upsertAccountAsset(account_asset.value());
```
l4l: Btw, didn't notice it at the first time. Afair optional has implicit cast to bool so it's replace this:
https://github.com/hyperledger/iroha/blob/a1cacb1ac9435fd2946423cc8d250ec0625281b5/irohad/model/execution/impl/command_executor.cpp#L343-L353
with this (though didn't test it):
```cpp
return new_account_asset_value->balance - subtract_asset_quantity.amount
    | [&account_asset](auto new_balance) {
  account_asset->balance = new_balance;
  return account_asset;
} | [](auto account_asset) {
  commands.upsertAccountAsset(account_asset);
}
```

But not really sure if it's better, decide on your own

l4l:
Also when you will merge it plz squash redundant commit(s) (e.g add "A", del "A" should be in one commit)

irohad/model/execution/impl/command_executor.cpp
grimadas: Although it looks nice, we still need to add logging for that. When valdiation fails the reason is not clear

neewy: have we added it? @grimadas

PR #789, opened 2017-12-19
Deleted ugly space in payloads

neewy:
## What is this pull request?
Deleted two ugly spaces
   
Because other lines don't have it
  
Look at the file named block.proto


neewy:
for some reason it also includes hidden commit from bogdan

PR #764, opened 2017-12-14
Add python bindings test

l4l:
## What is this pull request?
Test of bindings:
- [x] Python
- [x] Java
  
## How to use the features provided in the pull request?
For python:
`ctest -R python --output-on-failure`
For java:
`ctest -R java --output-on-failure`

PR #792, opened 2017-12-19
Builders bindings tests for python&java

l4l:
Rebased #764 

So far in this pr:
- Java builders tests
- Python builders tests
- gradle integration for java


test/module/shared_model/bindings/BuilderTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class BuilderTest {
+    static {
+        try {
+            System.loadLibrary("iroha");
+        } catch (UnsatisfiedLinkError e) {
+            System.err.println("Native code library failed to load. \n" + e);
```
muratovv: Is it a part of the test? I mean if the iroha lib not loaded test should be failed.

test/module/shared_model/bindings/BuilderTest.java
```diff
@@ -0,0 +1,145 @@
...
+            System.exit(1);
+        }
+    }
+
+    private Keypair keys;
+    private ModelBuilder builder;
+
+    ModelBuilder base() {
+        return new ModelBuilder().txCounter(BigInteger.valueOf(123))
+                .createdTime(BigInteger.valueOf(System.currentTimeMillis()))
```
muratovv: Maybe we need to think about the possibility of setting counter automatically from the native library.

test/module/shared_model/bindings/BuilderTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Test
+    void emptyTx() {
+        ModelBuilder builder = new ModelBuilder();
+        assertThrows(IllegalArgumentException.class, builder::build);
+    }
+
+    @Test
+    void txWithoutCommand() {
+//        broken for now
```
muratovv: What does it mean? What there is need to do for fix it?

test/module/shared_model/bindings/builder-test.py
```diff
@@ -0,0 +1,89 @@
...
+
+  def setUp(self):
+    self.keys = iroha.ModelCrypto().generateKeypair()
+    self.builder = self.generate_base()
+
+  def set_add_peer(self):
+    self.builder.addPeer("123.123.123.123", self.keys.publicKey())
+
+  def test_tx_without_command(self):
+    return #broken for now
```
muratovv: Same issue as with Java.

test/module/shared_model/bindings/builder-test.py
```diff
@@ -0,0 +1,89 @@
+import iroha
```
muratovv: Is there possible that Iroha lib failed on initialization? If yes, this cases handled by the test?

test/module/shared_model/bindings/BuilderTest.java
l4l: There's exit(1) below that will fail test anyway. I'll replace it with an explicit assertion

test/module/shared_model/bindings/BuilderTest.java
l4l: For some reason current shared_model doesn't allow to build tx w/o any command. That's a point for discussion primarily. If it's fine I can just delete this test

test/module/shared_model/bindings/builder-test.py
l4l: Then it'll be exception with smth like "sorry cannot find this stuff" and tests fails as long as it doesn't be handled

test/module/shared_model/bindings/BuilderTest.java
luckychess: Some time ago we decided that tx without commands is invalid (in chat discussion I believe) and I did a check in validator for this case.

l4l: Ok, I'll delete it then

PR #793, opened 2017-12-19
Return vector in Blob::blob()

l4l:
## What is this pull request?
This pr is changes the iterface of `Blob::blob()` from `std::string` to `std::vector<uint8_t>`

There're issues with the swig string conversions, binary data is basically converted incorrectly

Also note, that this way is more or less bad in terms of efficiency (there's can be redundant copies in debug builds). I think it should be optimized: all binary data should be stored in the `std::vector<uint8_t>`. But that requires significant amount of work, so maybe there's some better approaches

### DIFF
--- a/shared_model/cryptography/blob.hpp
+++ b/shared_model/cryptography/blob.hpp
@@ -29,6 +29,8 @@
 namespace shared_model {
   namespace crypto {
 
+    class Blob;
+    static inline std::string toBinaryString(const Blob &b);
     /**
      * Blob class present user-friendly blob for working with low-level
      * binary stuff. Its length is not fixed in compile time.
@@ -64,13 +66,6 @@ namespace shared_model {
       virtual const Bytes &blob() const { return blob_; }
 
       /**
-       * @return provides raw representation of blob as string
-       */
-      virtual const std::string str() const {
-        return std::string(blob_.begin(), blob_.end());
-      }
-
-      /**
        * @return provides human-readable representation of blob without leading
        * 0x
        */
@@ -104,7 +99,7 @@ namespace shared_model {
 
       template <typename BlobType>
       DEPRECATED BlobType makeOldModel() const {
-        return BlobType::from_string(str());
+        return BlobType::from_string(toBinaryString(*this));
       }
 
      private:
@@ -112,6 +107,10 @@ namespace shared_model {
       Bytes blob_;
       std::string hex_;
     };
+
+    static inline std::string toBinaryString(const Blob &b) {
+      return std::string(b.blob().begin(), b.blob().end());
+    }
   }  // namespace crypto
 }  // namespace shared_model
 #endif  // IROHA_SHARED_MODEL_BLOB_HPP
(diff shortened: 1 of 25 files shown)


Warchant:
Look at sorabot's issues.

shared_model/cryptography/blob.hpp
```diff
@@ -34,11 +35,21 @@ namespace shared_model {
      */
     class Blob : public interface::ModelPrimitive<Blob> {
      public:
+      using bytes = std::vector<uint8_t>;
```
luckychess: It would be better to name a type from a capital letter I think.

test/module/shared_model/cryptography/blob_test.cpp
```diff
@@ -15,19 +15,37 @@
  * limitations under the License.
  */
 
-#include <gtest/gtest.h>
 #include "cryptography/blob.hpp"
+#include <gtest/gtest.h>
 
 using namespace shared_model::crypto;
 
+auto data = "Hello World";
```
luckychess: I think it would be great to use some string with "bad" characters inside (like '\0') because it was a reason we introduced this pr.

test/module/shared_model/backend_proto/shared_proto_util_test.cpp
```diff
@@ -0,0 +1,38 @@
...
+using namespace iroha;
+using namespace shared_model::proto;
+
+/**
+ * @given some protobuf object
+ * @when making a blob from it
+ * @then make sure that the deserialized from string is the same
+ */
+TEST(UtilTest, StringFromMakeBlob) {
+  protocol::Header base, deserialized;
```
luckychess: May be use another proto object? I did a quick check and haven't found any other place where protocol::Header is used which makes it a candidate for deletion in future.

luckychess:
Btw should it be merged directly into develop and not to the trunk?

luckychess:
API of Blob should be changed anyway because Blob is a type known outside C++. We only have an option to leave internal string representation as is or also change it and I think the second option is better.

test/module/shared_model/backend_proto/shared_proto_util_test.cpp
l4l: It's basically the simpliest, that's the reason

l4l:
> Btw should it be merged directly into develop and not to the trunk?

I think so, it noticeably changes shared model primitive

> Another possible problem I see is that str() method remains public and available to anyone. Could it be hided somehow?

Don't really sure that it possible. But it still possible to change it e.g to `binstr()` or smth more obvious

> I don't think that existing API of Blob should be changed, since it works fine in C++. Instead this behaviour should be changed only in bindings.

And I don't sure that semantically string is a proper type for storing binary data. It's better be changed as earlier as possible. Anyway swig&java doesn't work in such way and I don't really sure that we will be able to easily change the swig. Besides there's can be problems with other bindings.

As @luckychess said the change of interntal datatype is optional (and I agree that it's better be fixed as well)

PR #734, opened 2017-12-02
Feature/shared model swig

luckychess:
## What is this pull request?
Swig bindings for Python and Java. Scope is shared model builders and signing built objects.
  
## How to use the features provided in the pull request?
Bindings libraries are disabled by default. To generate them add ```-DSWIG_PYTHON=ON``` and/or ```-DSWIG_JAVA=ON``` option to your cmake command.
Then you can build Iroha as usual. After the build process is completed navigate to ```build/shared_model//bindings/```. You can use Iroha crypto from Python in a next way:
```
import iroha

builder = iroha.ModelBuilder()
crypto = iroha.ModelCrypto()

me_kp = crypto.generateKeypair()
peer_kp = crypto.generateKeypair()
signatory_kp = crypto.generateKeypair()
account_kp = crypto.generateKeypair()
time = 1512549580
startCounter = 1
creator = "me"
signatory = "fyodorkek-san"

commands = []

commands.append(builder.txCounter(startCounter).creatorAccountId(creator).createdTime(time).addPeer("127.0.0.1:50051", peer_kp.publicKey()).build())

commands.append(builder.txCounter(startCounter+1).creatorAccountId(creator).createdTime(time+1).createDomain("iroha", "admin").build())

commands.append(builder.txCounter(startCounter+2).creatorAccountId(creator).createdTime(time+2).createAccount("luckychess", "iroha", account_kp.publicKey()).build())

commands.append(builder.txCounter(startCounter+3).creatorAccountId(creator).createdTime(time+3).addSignatory(signatory, signatory_kp.publicKey()).build())

commands.append(builder.txCounter(startCounter+4).creatorAccountId(creator).createdTime(time+4).removeSignatory(signatory, signatory_kp.publicKey()).build())

commands.append(builder.txCounter(startCounter+5).creatorAccountId(creator).createdTime(time+5).setAccountQuorum(signatory, 2).build())

commands.append(builder.txCounter(startCounter+6).creatorAccountId(creator).createdTime(time+6).assetQuantity("luckychess@iroha", "irohacoin#iroha", "123.45").build())

commands.append(builder.txCounter(startCounter+7).creatorAccountId(creator).createdTime(time+7).transferAsset("luckychess@iroha", "kamil@iroha", "irohacoin#iroha", "hello world", "12.45").build())


protoHelper = iroha.ModelTransactionProto()

for command in commands:
    print(protoHelper.signAndAddSignature(command, me_kp).toString())
```
And from Java:
```
import java.math.BigInteger;
[...]

PR #796, opened 2017-12-20
Feature/swig command bindings

lebdron:
## What is this pull request?
Add missing commands to ModelTransactionBuilder. 
   
The usage is similar to prior pull requests for SWIG bindings, such as https://github.com/hyperledger/iroha/pull/734

## Details/Features
- Add missing commands
- Remove redundant fields from protobuf/common_objects
- Fix several methods in protobuf/transaction.hpp

### DIFF
--- a/shared_model/backend/protobuf/common_objects/asset.hpp
+++ b/shared_model/backend/protobuf/common_objects/asset.hpp
@@ -33,9 +33,6 @@ namespace shared_model {
       template <typename AssetType>
       explicit Asset(AssetType &&account)
           : CopyableProto(std::forward<AssetType>(account)),
-            assetId_(proto_->asset_id()),
-            domainId_(proto_->domain_id()),
-            precision_(proto_->precision()),
             blob_([this] { return BlobType(proto_->SerializeAsString()); }) {}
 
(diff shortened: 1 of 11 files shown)


shared_model/backend/protobuf/common_objects/asset.hpp
```diff
@@ -33,25 +33,22 @@ namespace shared_model {
...
       }
 
       const interface::types::DomainIdType &domainId() const override {
-        return domainId_;
+        return proto_->domain_id();
       }
 
-      const interface::types::PrecisionType &precision() const override {
-        return precision_;
+      interface::types::PrecisionType precision() const override {
```
kamilsa: why remove const reference?

grimadas: It is just simple int

kamilsa: Now yes, but in future we can potentially might change it with more complex object, can't we?
I thought that was the idea of using `using` for creating new types: so that in future we may change type in single place and don't worry about remaining code. Here it seems that if precision (for some unknown for me reason) would be changed with more complex type, redundant copying might happen

PR #817, opened 2017-12-26
Feature/java example

luckychess:
Go to example folder.
```
./prepare.sh
gradle build
gradle run
```

### DIFF
--- a/example/java/TransactionExample.java
+++ b/example/java/TransactionExample.java
@@ -44,23 +44,29 @@ class TransactionExample {
     private static ModelProtoTransaction protoTxHelper = new ModelProtoTransaction();
     private static ModelProtoQuery protoQueryHelper = new ModelProtoQuery();
 
+    public static byte[] toByteArray(ByteVector blob) {
+        byte bs[] = new byte[(int)blob.size()];
+        for (int i = 0; i < blob.size(); ++i) {
+            bs[i] = (byte)blob.get(i);
+        }
+        return bs;
+    }
 
-
-    public static void main(String[] args) {
-        String adminPub = "", adminPriv = "";
-
+    public static String readKeyFromFile(String path) {
         try {
-            adminPub = new String(Files.readAllBytes(Paths.get("../admin@test.pub")));
-            adminPriv = new String(Files.readAllBytes(Paths.get("../admin@test.priv")));
+            return new String(Files.readAllBytes(Paths.get(path)));
         } catch (IOException e) {
             System.err.println("Unable to read key files.\n " + e);
             System.exit(1);
         }
+        return "";
+    }
 
-        Keypair keys = crypto.convertFromExisting(adminPub, adminPriv);
+    public static void main(String[] args) {
+        Keypair keys = crypto.convertFromExisting(readKeyFromFile("../admin@test.pub"),
+            readKeyFromFile("../admin@test.priv"));
 
         long currentTime = System.currentTimeMillis();
-        //long startCounter = 1;
         String creator = "admin@test";
 
         // build transaction (still unsigned)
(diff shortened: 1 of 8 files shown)


example/java/TransactionExample.java
```diff
@@ -0,0 +1,120 @@
...
+    public static void main(String[] args) {
+        String adminPub = "", adminPriv = "";
+
+        try {
+            adminPub = new String(Files.readAllBytes(Paths.get("../admin@test.pub")));
+            adminPriv = new String(Files.readAllBytes(Paths.get("../admin@test.priv")));
+        } catch (IOException e) {
+            System.err.println("Unable to read key files.\n " + e);
+            System.exit(1);
+        }
```
Warchant: can this code be added to `irohajava` package?
so the user writes something like `irohajava.load()`

example/java/prepare.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+cd $(dirname $0)
+
+# generate iroha lib
+if [[ $1 -eq 0 && ${IROHA_HOME} -eq 0 ]] ; then
```
Warchant: "$1", "${IROHA_HOME}"

in what case IROHA_HOME will be equal to 0? 

for emptiness better use -z

```
if [ -z "$1" && -z "${IROHA_HOME}" ]; then
```

example/java/prepare.sh
```diff
@@ -0,0 +1,12 @@
...
+    echo 'Neither IROHA_HOME nor path to CMakeLists.txt is set'
+    exit 0
+fi
+cmake_path=${1,-${IROHA_HOME}}
```
Warchant: what a weird construction?
At what cases it is required to set `cmake_path` to 0?

example/java/TransactionExample.java
luckychess: User loads library via ```System.loadLibrary("irohajava");``` but how can library be aware of user's keys and their location?

example/java/build.gradle
```diff
@@ -0,0 +1,61 @@
+apply plugin: 'java'
+apply plugin: 'com.google.protobuf'
+apply plugin: 'application'
+
+def BUILD_DIR = "./shared_model/bindings"
```
kamilsa: not build/shared_model/bindings ?

example/java/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.3'
+    }
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compile 'com.google.protobuf:protobuf-java:3.4.0'
```
l4l: should be 3.5.0

example/java/build.gradle
```diff
@@ -0,0 +1,61 @@
+apply plugin: 'java'
+apply plugin: 'com.google.protobuf'
```
l4l: should be after buildscript

example/java/TransactionExample.java
```diff
@@ -0,0 +1,163 @@
...
+import java.nio.file.Paths;
+import java.nio.file.Files;
+import java.io.IOException;
+import java.math.BigInteger;
+import java.lang.Thread;
+
+class TransactionExample {
+    static {
+        try {
+            System.loadLibrary("irohajava");
```
l4l: Imo we should make a package that doesn't require manual loading

example/java/TransactionExample.java
```diff
@@ -0,0 +1,163 @@
...
+
+    private static ModelCrypto crypto = new ModelCrypto();
+    private static ModelTransactionBuilder txBuilder = new ModelTransactionBuilder();
+    private static ModelQueryBuilder queryBuilder = new ModelQueryBuilder();
+    private static ModelProtoTransaction protoTxHelper = new ModelProtoTransaction();
+    private static ModelProtoQuery protoQueryHelper = new ModelProtoQuery();
+
+
+
+    public static void main(String[] args) {
```
l4l: Would be nice to extract functions from main. E.g `readKeyFrom`, `bytesFromVector`, etc

example/java/TransactionExample.java
luckychess: Hmm, I have no idea how to to this. In Python we use exactly the same approach.

example/java/build.gradle
luckychess: Why not put all all "apply" in one place?

l4l: Hm, I thought it doesn't work, sorry

PR #830, opened 2017-12-28
Feature/get account asset txs cli

motxx:
### Description of the Change
Implement CLI implementation of `GetAccountAssetTransactions`

Remove useless optional return type in pager cli
Remove wrong pager empty code
Change MAX_PAGER_LIMIT in old model to global var to pass the linker

- [x] Merge https://github.com/hyperledger/iroha/pull/841 first, then rebase onto trunk and can be reviewed.

### Benefits
Missing CLI API will be implemented.

### Possible Drawbacks 
None

None

None

### DIFF
--- a/iroha-cli/interactive/impl/interactive_query_cli.cpp
+++ b/iroha-cli/interactive/impl/interactive_query_cli.cpp
@@ -16,7 +16,7 @@
  */
 
 #include "interactive/interactive_query_cli.hpp"
-
+#include <boost/algorithm/string.hpp>
 #include <boost/lexical_cast.hpp>
 #include <fstream>
 #include "byteutils.hpp"
@@ -41,6 +41,7 @@ namespace iroha_cli {
           {GET_ACC, "Get Account Information"},
           {GET_ACC_AST, "Get Account's Assets"},
           {GET_ACC_TX, "Get Account's Transactions"},
+          {GET_ACC_AST_TX, "Get Account Asset's Transactions"},
           {GET_TX, "Get Transactions by transactions' hashes"},
           {GET_ACC_SIGN, "Get Account's Signatories"},
           {GET_ROLES, "Get all current roles in the system"},
@@ -51,6 +52,7 @@ namespace iroha_cli {
 
       const auto acc_id = "Requested account Id";
       const auto ast_id = "Requested asset Id";
+      const auto asts_id = "Requested assets Id";
       const auto pager_tx_hash = "Requested pager tx hash";
       const auto pager_limit = "Requested pager limit";
       const auto role_id = "Requested role name";
@@ -60,6 +62,7 @@ namespace iroha_cli {
           {GET_ACC, {acc_id}},
           {GET_ACC_AST, {acc_id, ast_id}},
           {GET_ACC_TX, {acc_id, pager_tx_hash, pager_limit}},
+          {GET_ACC_AST_TX, {acc_id, asts_id, pager_tx_hash, pager_limit}},
           {GET_TX, {tx_hashes}},
           {GET_ACC_SIGN, {acc_id}},
           {GET_ROLES, {}},
@@ -72,6 +75,7 @@ namespace iroha_cli {
           {GET_ACC, &InteractiveQueryCli::parseGetAccount},
           {GET_ACC_AST, &InteractiveQueryCli::parseGetAccountAssets},
           {GET_ACC_TX, &InteractiveQueryCli::parseGetAccountTransactions},
+          {GET_ACC_AST_TX, &InteractiveQueryCli::parseGetAccountAssetTransactions},
           {GET_TX, &InteractiveQueryCli::parseGetTransactions},
           {GET_ACC_SIGN, &InteractiveQueryCli::parseGetSignatories},
           {GET_ROLE_PERM, &InteractiveQueryCli::parseGetRolePermissions},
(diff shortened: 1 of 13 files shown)


motxx:
someone's reviews are wanted.

motxx:
For the reason of rebasing develop shared model but this PR didn't follow, this build cannot.
So reivew can be done after https://github.com/hyperledger/iroha/pull/841 merged.

Warchant:
@motxx please rebase onto develop and fix build.

motxx:
@Warchant 
Field validator cannot be built in this branch and trunk branch because it's implementation is in #841
This branch is waiting for #841

motxx:
This PR is rebased onto trunk and can be reviewed.

iroha-cli/interactive/impl/interactive_query_cli.cpp
```diff
@@ -168,59 +173,67 @@ namespace iroha_cli {
           local_time_, creator_, counter_, account_id, asset_id);
     }
 
-    nonstd::optional<iroha::model::Pager> InteractiveQueryCli::parsePager(
+    iroha::model::Pager InteractiveQueryCli::parsePager(
```
muratovv: What is the reason to return the raw parameter without an optional wrapper?

iroha-cli/interactive/impl/interactive_query_cli.cpp
```diff
@@ -168,59 +173,67 @@ namespace iroha_cli {
...
+        pager.tx_hash.fill(0);
+      } else {
+        const auto decoded_hash = iroha::hexstringToBytestring(encoded_tx_hash);
+        if (decoded_hash) {
+          pager.tx_hash = iroha::hash256_t::from_string(*decoded_hash);
+        } else {
+          log_->warn(
+              "Specified hash size is invalid. Set empty hash as "
+              "default");
+          pager.tx_hash.fill(0);
```
muratovv: Default hash looks bad, because of the following reasons:
* class shouldn't contain the logic of creation hash by itself
* default behavior in this case worse - not obvious usage
* if there was implied to make default strategy - this should be customizable - passes as a dependency.

iroha-cli/interactive/impl/interactive_query_cli.cpp
```diff
@@ -168,59 +173,67 @@ namespace iroha_cli {
...
-        return pager;
-      };
+      }
+      try {
+        pager.limit = boost::lexical_cast<decltype(pager.limit)>(limit_str);
+      } catch (...) {
+        log_->warn(
+            "Specified tx limit is invalid. Set {} (max size) as default",
+            MAX_PAGER_LIMIT);
+        pager.limit = MAX_PAGER_LIMIT;
```
muratovv: Same as before.

iroha-cli/interactive/impl/interactive_query_cli.cpp
```diff
@@ -168,59 +173,67 @@ namespace iroha_cli {
...
       GetTransactions::TxHashCollectionType tx_hashes;
-      std::for_each(params.begin(), params.end(), [&tx_hashes](auto const& hex_hash){
-        if (auto opt =
-          iroha::hexstringToArray<GetTransactions::TxHashType::size()>(hex_hash)) {
-          tx_hashes.push_back(*opt);
-        }
-      });
+      std::for_each(
+          params.begin(), params.end(), [&tx_hashes](auto const &hex_hash) {
+            if (auto opt = iroha::hexstringToArray<
```
muratovv: Is it clang-format formatting?

irohad/model/generators/impl/query_generator.cpp
```diff
@@ -83,7 +83,7 @@ namespace iroha {
         return query;
       }
 
-      optional_ptr<GetAccountTransactions>
+      std::shared_ptr<GetAccountTransactions>
```
muratovv: Same question about changing the interface for raw usage

iroha-cli/interactive/impl/interactive_query_cli.cpp
motxx: Yes, it is bad to change the value from client outside of the API implementation.
Maybe customizing hash256_t -> pager_hash_t and allows 0 and 32 byte size?

motxx: There is no stateless validation in iroha-cli but converting hex to bytes is required to generate query because `iroha::hash256_t::from_string` throw exception if the size is not correct.

One solution seems that it is okay to use `optional` in the return value of `parsePager`. But because all queries return `std::shared_ptr` and `query_handlers_` referring such functions, it cannot propagate the error result when the conversion fails. (please see also: https://github.com/hyperledger/iroha/pull/830/files#diff-0810ba332c21013b016cd0dd5d525cdeL211)

So in CLI, I removed `nonstd::optional` and apply the default value of the hash and show warning log.

iroha-cli/interactive/impl/interactive_query_cli.cpp
motxx: Answered in https://github.com/hyperledger/iroha/pull/830#discussion_r173097242

iroha-cli/interactive/impl/interactive_query_cli.cpp
motxx: Answered in https://github.com/hyperledger/iroha/pull/830#discussion_r173097242

irohad/model/generators/impl/query_generator.cpp
motxx: Answered in https://github.com/hyperledger/iroha/pull/830#discussion_r173097242

PR #836, opened 2018-01-04
Improve stateless validation

nickaleks:
### Description of the Change

This Pull Request introduces several changes to stateless validation:
1. FieldValidator now returns value of invalid field
2. Amount is checked to be greater than 0
3. CommandsValidator now appends counter to Command name, which allows understanding which command has failed.

The biggest change of this pull request is tests for FieldValidator, which are organized in a table-like manner with various test cases for different fields. Because of a lot of boilerplate, file is quite large, but it allows to add various test cases for different fields very easily.

### Benefits

1. Better error message for stateless validation
2. Test structure for FieldValidator allows us to test it easier

### Possible Drawbacks 

1. Size of the test file is quite large, and it may take some time to understand how it all works. 

If you want to add a new constraint to some field, you can test it by adding a test case to appropriate map in file field_validator_test.cpp.

Google Test has parametrized test which allows running a test with different values however it would complicate code even, and require even more boilerplate code.

### DIFF
--- a/shared_model/validators/field_validator.hpp
+++ b/shared_model/validators/field_validator.hpp
@@ -74,7 +74,8 @@ namespace shared_model {
                           const interface::types::PubkeyType &pubkey) const {
         if (pubkey.blob().size() != 32) {
           auto message =
-              generateErrorMessage("Public key has wrong size", pubkey.hex());
+              generateErrorMessage("Public key has wrong size",
+                                   std::to_string(pubkey.blob().size()));
           reason.second.push_back(std::move(message));
         }
       }
--- a/shared_model/validators/field_validator.hpp
+++ b/shared_model/validators/field_validator.hpp
@@ -73,8 +73,8 @@ namespace shared_model {
       void validatePubkey(ReasonsGroupType &reason,
                           const interface::types::PubkeyType &pubkey) const {
         if (pubkey.blob().size() != 32) {
-          auto message = generateErrorMessage("Public key has wrong size",
-                                              pubkey.hex());
+          auto message =
+              generateErrorMessage("Public key has wrong size", pubkey.hex());
           reason.second.push_back(std::move(message));
         }
       }
--- a/shared_model/validators/field_validator.hpp
+++ b/shared_model/validators/field_validator.hpp
@@ -48,7 +48,10 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::types::AccountIdType &account_id) const {
         if (not std::regex_match(account_id, account_id_)) {
-          reason.second.push_back("Wrongly formed account_id");
+          auto message =
+              generateErrorMessage("Wrongly formed account_id",
+                                   static_cast<std::string>(account_id));
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -56,7 +59,9 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::types::AssetIdType &asset_id) const {
         if (not std::regex_match(asset_id, asset_id_)) {
-          reason.second.push_back("Wrongly formed asset_id");
+          auto message = generateErrorMessage(
+              "Wrongly formed asset_id", static_cast<std::string>(asset_id));
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -68,7 +73,9 @@ namespace shared_model {
       void validatePubkey(ReasonsGroupType &reason,
                           const interface::types::PubkeyType &pubkey) const {
         if (pubkey.blob().size() != 32) {
-          reason.second.push_back("Public key has wrong size");
+          auto message = generateErrorMessage("Public key has wrong size",
+                                              pubkey.hex());
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -77,14 +84,18 @@ namespace shared_model {
           const interface::AddPeer::AddressType &address) const {
         if (not(iroha::validator::isValidIpV4(address)
                 or iroha::validator::isValidHostname(address))) {
-          reason.second.push_back("Wrongly formed PeerAddress: " + address);
+          auto message = generateErrorMessage(
+              "Wrongly formed PeerAddress", static_cast<std::string>(address));
+          reason.second.push_back(std::move(message));
         }
       }
 
       void validateRoleId(ReasonsGroupType &reason,
                           const interface::types::RoleIdType &role_id) const {
         if (not std::regex_match(role_id, name_)) {
-          reason.second.push_back("Wrongly formed role_id");
+          auto message = generateErrorMessage(
+              "Wrongly formed role_id", static_cast<std::string>(role_id));
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -92,7 +103,10 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::types::AccountNameType &account_name) const {
         if (not std::regex_match(account_name, name_)) {
-          reason.second.push_back("Wrongly formed account_name");
+          auto message =
+              generateErrorMessage("Wrongly formed account_name",
+                                   static_cast<std::string>(account_name));
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -100,7 +114,9 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::types::DomainIdType &domain_id) const {
         if (not std::regex_match(domain_id, name_)) {
-          reason.second.push_back("Wrongly formed domain_id");
+          auto message = generateErrorMessage(
+              "Wrongly formed domain_id", static_cast<std::string>(domain_id));
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -108,7 +124,10 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::types::AssetNameType &asset_name) const {
         if (not std::regex_match(asset_name, name_)) {
-          reason.second.push_back("Wrongly formed asset_name");
+          auto message =
+              generateErrorMessage("Wrongly formed asset_name",
+                                   static_cast<std::string>(asset_name));
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -116,7 +135,9 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::SetAccountDetail::AccountDetailKeyType &key) const {
         if (not std::regex_match(key, detail_key_)) {
-          reason.second.push_back("Wrongly formed key");
+          auto message = generateErrorMessage("Wrongly formed key",
+                                              static_cast<std::string>(key));
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -150,7 +171,10 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::types::AccountIdType &account_id) const {
         if (not std::regex_match(account_id, account_id_)) {
-          reason.second.push_back("Wrongly formed creator_account_id");
+          auto message =
+              generateErrorMessage("Wrongly formed creator_account_id",
+                                   static_cast<std::string>(account_id));
+          reason.second.push_back(std::move(message));
         }
       }
 
@@ -160,25 +184,26 @@ namespace shared_model {
         iroha::ts64_t now = iroha::time::now();
         // TODO 06/08/17 Muratov: make future gap for passing timestamp, like
         // with old timestamps IR-511 #goodfirstissue
+
+        auto time_message =
+            (boost::format("%llu, now: %llu") % timestamp % now).str();
         if (now < timestamp) {
-          reason.second.push_back(boost::str(
-              boost::format(
-                  "timestamp broken: send from future (%llu, now %llu)")
-              % timestamp % now));
+          auto message = generateErrorMessage(
+              "timestamp broken: sent from future", time_message);
+          reason.second.push_back(std::move(message));
         }
 
         if (now - timestamp > MAX_DELAY) {
-          reason.second.push_back(boost::str(
-              boost::format("timestamp broken: too old (%llu, now %llu)")
-              % timestamp % now));
+          auto message =
+              generateErrorMessage("timestamp broken: too old ", time_message);
+          reason.second.push_back(std::move(message));
         }
       }
 
       void validateCounter(ReasonsGroupType &reason,
                            const interface::types::CounterType &counter) const {
         if (counter == 0) {
-          reason.second.push_back(
-              boost::str(boost::format("Counter should be > 0")));
+          reason.second.push_back("Counter should be > 0");
         }
       }
 
@@ -187,6 +212,12 @@ namespace shared_model {
       // max-delay between tx creation and validation
       static constexpr auto MAX_DELAY =
           std::chrono::hours(24) / std::chrono::milliseconds(1);
+
+      std::string generateErrorMessage(const std::string &error,
+                                       const std::string &field_value) const {
+        return (boost::format("%s, passed value: %s") % error % field_value)
+            .str();
+      }
     };
   }  // namespace validation
 }  // namespace shared_model
--- a/shared_model/validators/field_validator.hpp
+++ b/shared_model/validators/field_validator.hpp
@@ -67,7 +67,9 @@ namespace shared_model {
 
       void validateAmount(ReasonsGroupType &reason,
                           const interface::Amount &amount) const {
-        /* put here any validations*/
+        if (amount.intValue() == 0) {
+          reason.second.push_back("Amount must be greater than 0");
+        }
       }
 
       void validatePubkey(ReasonsGroupType &reason,
--- a/shared_model/validators/field_validator.hpp
+++ b/shared_model/validators/field_validator.hpp
@@ -36,15 +36,15 @@ namespace shared_model {
     class FieldValidator {
      public:
       FieldValidator()
-          : account_id_regex(R"([a-z]{1,9}\@[a-z]{1,9})"),
-            asset_id_regex(R"([a-z]{1,9}\#[a-z]{1,9})"),
-            name_regex(R"([a-z]{1,9})"),
-            detail_key_regex(R"([A-Za-z0-9_]{1,})") {}
+          : account_id_pattern_(R"([a-z]{1,9}\@[a-z]{1,9})"),
+            asset_id_pattern_(R"([a-z]{1,9}\#[a-z]{1,9})"),
+            name_pattern_(R"([a-z]{1,9})"),
+            detail_key_pattern_(R"([A-Za-z0-9_]{1,})") {}
 
       void validateAccountId(
           ReasonsGroupType &reason,
           const interface::types::AccountIdType &account_id) const {
-        if (not std::regex_match(account_id, account_id_regex)) {
+        if (not std::regex_match(account_id, account_id_pattern_)) {
           auto message =
               (boost::format("Wrongly formed account_id, passed value: '%s'")
                % account_id)
@@ -56,7 +56,7 @@ namespace shared_model {
       void validateAssetId(
           ReasonsGroupType &reason,
           const interface::types::AssetIdType &asset_id) const {
-        if (not std::regex_match(asset_id, asset_id_regex)) {
+        if (not std::regex_match(asset_id, asset_id_pattern_)) {
           auto message =
               (boost::format("Wrongly formed asset_id, passed value: '%s'")
                % asset_id)
@@ -102,7 +102,7 @@ namespace shared_model {
 
       void validateRoleId(ReasonsGroupType &reason,
                           const interface::types::RoleIdType &role_id) const {
-        if (not std::regex_match(role_id, name_regex)) {
+        if (not std::regex_match(role_id, name_pattern_)) {
           auto message =
               (boost::format("Wrongly formed role_id, passed value: '%s'")
                % role_id)
@@ -114,7 +114,7 @@ namespace shared_model {
       void validateAccountName(
           ReasonsGroupType &reason,
           const interface::types::AccountNameType &account_name) const {
-        if (not std::regex_match(account_name, name_regex)) {
+        if (not std::regex_match(account_name, name_pattern_)) {
           auto message =
               (boost::format("Wrongly formed account_name, passed value: '%s'")
                % account_name)
@@ -126,7 +126,7 @@ namespace shared_model {
       void validateDomainId(
           ReasonsGroupType &reason,
           const interface::types::DomainIdType &domain_id) const {
-        if (not std::regex_match(domain_id, name_regex)) {
+        if (not std::regex_match(domain_id, name_pattern_)) {
           auto message =
               (boost::format("Wrongly formed domain_id, passed value: '%s'")
                % domain_id)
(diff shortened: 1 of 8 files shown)


shared_model/validators/field_validator.hpp
```diff
@@ -48,27 +48,37 @@ namespace shared_model {
...
+        }
       }
 
       void validatePubkey(ReasonsGroupType &reason,
                           const interface::types::PubkeyType &pubkey) const {
-        if (pubkey.blob().size() != 32) {
-          reason.second.push_back("Public key has wrong size");
+        if (pubkey.blob().size() != key_size) {
+          auto message =
+              generateErrorMessage("Public key has wrong size",
```
Warchant: *couldhave*
IMO `generateErrorMessage` can be just alias for `boost::format`. It is not very good to use `std::to_string`. What if I, as a developer of validator, want to make some comprehensive error? Like, `str+number+str+hex+number` (I hope you get the idea)

shared_model/validators/field_validator.hpp
```diff
@@ -48,27 +48,37 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::types::AccountIdType &account_id) const {
         if (not std::regex_match(account_id, account_id_)) {
-          reason.second.push_back("Wrongly formed account_id");
+          auto message =
+              generateErrorMessage("Wrongly formed account_id",
```
Warchant: *couldhave*
As for error message, it would be good to return matching regex to the user in error message, so the user will be able to fix his code without having to go to the documentation.

shared_model/validators/field_validator.hpp
```diff
@@ -77,46 +87,60 @@ namespace shared_model {
           const interface::AddPeer::AddressType &address) const {
         if (not(iroha::validator::isValidIpV4(address)
```
Warchant: Are you sure that validation should look like this? Because, as I remember, `address` here is `ip:port`, not just address.
If this checks only for IPv4 or hostname compliance, then these will be failed, but should not:
```
1.1.1.1:50
domain.com:50
```

shared_model/validators/field_validator.hpp
```diff
@@ -48,27 +48,37 @@ namespace shared_model {
           ReasonsGroupType &reason,
           const interface::types::AccountIdType &account_id) const {
         if (not std::regex_match(account_id, account_id_)) {
```
Warchant: *musthave*

`account_id_` is bad name for regex. 

I understand that this is not part of this PR, but please, move all regexp out of constructor 
 (https://github.com/hyperledger/iroha/blob/67a82efcb776283f85a6b912f89d86cbba0a9837/shared_model/validators/field_validator.hpp#L39), make them as static members and give them meaningful names:
```
account_id_regexp
...
```

shared_model/validators/field_validator.hpp
```diff
@@ -160,33 +187,42 @@ namespace shared_model {
...
+
+        auto time_message =
+            (boost::format("%llu, now: %llu") % timestamp % now).str();
         if (now < timestamp) {
-          reason.second.push_back(boost::str(
-              boost::format(
-                  "timestamp broken: send from future (%llu, now %llu)")
-              % timestamp % now));
+          auto message = generateErrorMessage(
+              "timestamp broken: sent from future", time_message);
```
Warchant: **bad timestamp**: sent from future

shared_model/validators/field_validator.hpp
```diff
@@ -160,33 +187,42 @@ namespace shared_model {
...
+          reason.second.push_back(std::move(message));
         }
 
-        if (now - timestamp > MAX_DELAY) {
-          reason.second.push_back(boost::str(
-              boost::format("timestamp broken: too old (%llu, now %llu)")
-              % timestamp % now));
+        if (now - timestamp > max_delay) {
+          auto message =
+              generateErrorMessage("timestamp broken: too old", time_message);
```
Warchant: **bad timestamp**: too old

shared_model/validators/field_validator.hpp
```diff
@@ -160,33 +187,42 @@ namespace shared_model {
...
      private:
       std::regex account_id_, asset_id_, ip_address_, name_, detail_key_;
       // max-delay between tx creation and validation
-      static constexpr auto MAX_DELAY =
+      static constexpr auto max_delay =
           std::chrono::hours(24) / std::chrono::milliseconds(1);
+      // size of key
+      static constexpr auto key_size = 32;
+
+      std::string generateErrorMessage(const std::string &error,
```
Warchant: Now I am thinking that this method is not needed at all, thus making a developer who writes a validator responsible for the message.

shared_model/validators/transaction_validator.hpp
```diff
@@ -189,7 +189,8 @@ namespace shared_model {
       ReasonsGroupType operator()(
           const detail::PolymorphicWrapper<interface::SetQuorum> &sq) const {
         ReasonsGroupType reason;
-        reason.first = "SetQuorum";
+        addInvalidCommand(reason, "SetAccountQuorum");
```
Warchant: SetQuorum or SetAccountQuorum?

Command's name is SetQuorum

shared_model/validators/transaction_validator.hpp
```diff
@@ -226,6 +227,15 @@ namespace shared_model {
 
      private:
       FieldValidator validator_;
+      mutable int command_counter{0};
```
Warchant: why `mutable`?

shared_model/validators/field_validator.hpp
nickaleks: Isn't it insecure to expose regexes?

shared_model/validators/field_validator.hpp
```diff
@@ -48,27 +48,37 @@ namespace shared_model {
...
+          auto message = generateErrorMessage(
+              "Wrongly formed asset_id", static_cast<std::string>(asset_id));
+          reason.second.push_back(std::move(message));
         }
       }
 
       void validateAmount(ReasonsGroupType &reason,
                           const interface::Amount &amount) const {
-        /* put here any validations*/
+        if (amount.intValue() == 0) {
```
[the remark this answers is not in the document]
nickaleks: It is unsigned, but I agree to change the condition to > 0 and do that for tx_counter as well because it is clearer and I had the same question when I saw this code for tx_counter.

shared_model/validators/field_validator.hpp
nickaleks: address, in this case, will contain ip:port, or hostname. You can check out test/module/libs/validator,  there are tests for this functions

shared_model/validators/transaction_validator.hpp
nickaleks: Typo

shared_model/validators/field_validator.hpp
Warchant: 1. Software is opensource, users are able to look at them by themselves
2. It is better for UX

shared_model/validators/transaction_validator.hpp
nickaleks: Because all validation methods are const, and we want to modify counter in them.

shared_model/validators/field_validator.hpp
nickaleks: I had similar thoughts, it just makes everything more complicated, so I'll remove it.

shared_model/validators/field_validator.hpp
nickaleks: Okay, will add.

shared_model/validators/field_validator.hpp
nickaleks: Isn't it a bit unnecessary to add 'regexp' to a variable name, since its type is already std::regex?

shared_model/validators/field_validator.hpp
nickaleks: Why is std::to_string bad?

shared_model/validators/field_validator.hpp
Warchant: @nickaleks it is obvious for you in IDE, but when others read the code, it is quite hard to understand just by looking at the variable.

nickaleks: Okay, also I think since this class does not have any state, we can as well use free functions in a proper namespace, as already done for example in address validation. But it is debatable because we tend to use classes everywhere.

nickaleks: Also, this change would require its own pull request because right now if we declare regexes static, we have to define them after class declaration. So I will fix it when we separate declaration and definition of FieldValidator.

shared_model/validators/field_validator.hpp
nickaleks: Sadly, there seems to be no way to extract a pattern from a string, so this change would require putting pattern in a separate variable. I would do that in separate pull request along with separation of .hpp and .cpp file.

Warchant:
However, could you also fix minor issues from sorabot?

shared_model/validators/field_validator.hpp
lebdron: @Warchant `std::regex` has a limited set of operations, so that if the variable is referenced in the code with `regex_match` for example, it is clear for the reader that referenced variable is indeed a regex.

shared_model/validators/field_validator.hpp
```diff
@@ -36,39 +36,50 @@ namespace shared_model {
...
+                  .str();
+          reason.second.push_back(std::move(message));
         }
       }
 
       void validateAmount(ReasonsGroupType &reason,
                           const interface::Amount &amount) const {
-        /* put here any validations*/
+        if (amount.intValue() <= 0) {
+          reason.second.push_back("Amount must be greater than 0");
```
lebdron: It might be useful to print out received value just as In other validators, for the same reason.

shared_model/validators/field_validator.hpp
```diff
@@ -160,33 +200,39 @@ namespace shared_model {
...
-          reason.second.push_back(
-              boost::str(boost::format("Counter should be > 0")));
+        if (counter <= 0) {
+          reason.second.push_back("Counter should be > 0");
         }
       }
 
      private:
-      std::regex account_id_, asset_id_, ip_address_, name_, detail_key_;
+      std::regex account_id_regex, asset_id_regex, name_regex, detail_key_regex;
```
lebdron: https://google.github.io/styleguide/cppguide.html#Variable_Names

>Data members of classes, both static and non-static, are named like ordinary nonmember variables, but with a trailing underscore.

shared_model/validators/field_validator.hpp
```diff
@@ -160,33 +200,39 @@ namespace shared_model {
...
         }
       }
 
       void validateCounter(ReasonsGroupType &reason,
                            const interface::types::CounterType &counter) const {
-        if (counter == 0) {
-          reason.second.push_back(
-              boost::str(boost::format("Counter should be > 0")));
+        if (counter <= 0) {
+          reason.second.push_back("Counter should be > 0");
```
lebdron: Value can also be printed out.

shared_model/validators/field_validator.hpp
```diff
@@ -36,39 +36,50 @@ namespace shared_model {
...
 
       void validateAccountId(
           ReasonsGroupType &reason,
           const interface::types::AccountIdType &account_id) const {
-        if (not std::regex_match(account_id, account_id_)) {
-          reason.second.push_back("Wrongly formed account_id");
+        if (not std::regex_match(account_id, account_id_regex)) {
+          auto message =
+              (boost::format("Wrongly formed account_id, passed value: '%s'")
+               % static_cast<std::string>(account_id))
```
lebdron: It's not clear whether cast is required, since it seems unlikely that `AccountIdType` will have only explicit conversion to `std::string`. Same applies to other methods.

shared_model/validators/transaction_validator.hpp
```diff
@@ -144,7 +144,7 @@ namespace shared_model {
           const detail::PolymorphicWrapper<interface::GrantPermission> &gp)
           const {
         ReasonsGroupType reason;
-        reason.first = "GrantPermission";
+       addInvalidCommand(reason, "GrantPermission");
```
lebdron: Space is missing before `addInvalidCommand`.

shared_model/validators/transaction_validator.hpp
```diff
@@ -189,7 +189,8 @@ namespace shared_model {
       ReasonsGroupType operator()(
           const detail::PolymorphicWrapper<interface::SetQuorum> &sq) const {
         ReasonsGroupType reason;
-        reason.first = "SetQuorum";
+        addInvalidCommand(reason, "SetQuorum");
+
```
lebdron: Newline seems redundant.

shared_model/validators/transaction_validator.hpp
```diff
@@ -201,7 +202,7 @@ namespace shared_model {
           const detail::PolymorphicWrapper<interface::SubtractAssetQuantity>
               &saq) const {
         ReasonsGroupType reason;
-        reason.first = "SubtractAssetQuantity";
+        addInvalidCommand(reason,"SubtractAssetQuantity");
```
lebdron: Space between `,` and `"SubtractAssetQuantity"` is missing. Maybe run clang_format?

test/module/shared_model/validators/field_validator_test.cpp
```diff
@@ -0,0 +1,747 @@
...
+  // possible commands, some fields are checked several times
+  // because they are present in several commands. To prevent that, this set
+  // contains all already tested fields, and checked each time we try to test
+  // a field
+  std::unordered_set<std::string> checked_fields;
+
+  /************************** TEST CASES ***************************/
+
+  std::vector<FieldTestCase> account_id_test_cases{
+      {
```
lebdron: Replacing this with some sort of function which takes case name, account id, bool (which will in turn create correct expected error message) will improve readability and reduce code duplication. Same applies for code below.

test/module/shared_model/validators/field_validator_test.cpp
```diff
@@ -0,0 +1,747 @@
...
+  std::vector<FieldTestCase> tx_hashes_test_cases;
+
+  /**************************************************************************/
+
+  // register validation function and test cases
+  std::unordered_map<std::string, FieldTest> field_validators{
+      // Command fields
+      {"account_id",
+       {[&] {
+          validation::ReasonsGroupType reason;
```
lebdron: Also looks like a shortcut is required, so that `reason` lines are not duplicated.

test/module/shared_model/validators/field_validator_test.cpp
```diff
@@ -0,0 +1,747 @@
...
+        return iroha::protocol::Query::Payload::descriptor()->FindOneofByName(
+            "query");
+      },
+      [&](auto field) {
+        // Set concrete type for new query
+        return payload->GetReflection()->MutableMessage(payload, field);
+      },
+      [this](auto field, auto query) { this->runTestCases(field); },
+      [&] {});
+}
```
lebdron: It looks like test for query fields is missing. The purpose is the same as `TransactionFieldsValidation`, since `QueryFieldsValidation` only does validation of concrete query (such as GetAccount) fields. It could iterate `iroha::protocol::Query::descriptor()`.

nickaleks: What do you mean by missing? if a field of a query is not covered by tests, an exception will be thrown inside runTestCases()

shared_model/validators/field_validator.hpp
Warchant: And probably it is better to write all strings in separate lines, like:
```
std::regex account_id_regex_;
std::regex asset_id_regex_;
...
```

shared_model/validators/transaction_validator.hpp
Warchant: Just use `clang-format`

test/module/shared_model/validators/field_validator_test.cpp
nickaleks: Do you suggest define a function for each field, which allows test case registration, like "registerAccountIDTestcase(name, account id, should_pass, error_message)"?

test/module/shared_model/validators/field_validator_test.cpp
nickaleks: I was thinking about passing a reference to the method to function, but non-static class members cannot be passed that way, I can wrap them in lambda, but the amount of code would not decrease in this case.

test/module/shared_model/validators/validators_fixture.hpp
```diff
@@ -29,8 +29,11 @@
 class ValidatorsTest : public ::testing::Test {
  public:
   ValidatorsTest() {
-    valid_amount.set_precision(2);
-    valid_amount.mutable_value()->set_fourth(1000);
+    amount.set_precision(2);
+    amount.mutable_value()->set_fourth(1000);
+    public_key_size = 32;
+    hash_size = 32;
+    counter = 0;
```
l4l: Can't find any place where it is used

test/module/shared_model/validators/validators_fixture.hpp
```diff
@@ -117,28 +120,51 @@ class ValidatorsTest : public ::testing::Test {
...
+    account_name = "admin";
+    domain_id = "ru";
+    detail_key = "key";
+    public_key = std::string(public_key_size, '0');
+    hash = std::string(public_key_size, '0');
+    role_permission = iroha::protocol::RolePermission::can_append_role;
+    grantable_permission =
+        iroha::protocol::GrantablePermission::can_add_my_signatory;
+    quorum = 2;
+    precision = 42;
```
l4l: Is it fine that precision doesn't match with the one at the `ValidatorsTest::ValidatorsTest()`?

test/module/shared_model/validators/field_validator_test.cpp
```diff
@@ -0,0 +1,747 @@
...
+ public:
+  // To test one field, validation function is required,
+  // which is always the same, and several test cases, which represent
+  // various inputs
+  using FieldTest = std::pair<ValidationFunction, std::vector<FieldTestCase>>;
+
+  // Run all test cases for given field
+  void runTestCases(const google::protobuf::FieldDescriptor *field) {
+    auto field_name = field->name();
+    // skip field, if already tested
```
l4l: What's the problem if it runs twice?

shared_model/validators/field_validator.hpp
```diff
@@ -160,33 +203,46 @@ namespace shared_model {
...
         if (now < timestamp) {
-          reason.second.push_back(boost::str(
-              boost::format(
-                  "timestamp broken: send from future (%llu, now %llu)")
-              % timestamp % now));
+          auto message = (boost::format("bad timestamp: sent from future, "
+                                        "timestamp: %llu, now: %llu")
+                          % timestamp % now)
+                             .str();
+          reason.second.push_back(std::move(message));
```
Warchant: this branch is not covered by tests

shared_model/validators/field_validator.hpp
```diff
@@ -36,39 +36,54 @@ namespace shared_model {
...
      public:
       FieldValidator()
-          : account_id_(R"([a-z]{1,9}\@[a-z]{1,9})"),
-            asset_id_(R"([a-z]{1,9}\#[a-z]{1,9})"),
-            ip_address_(
-                "((([0-1]?\\d\\d?)|((2[0-4]\\d)|(25[0-5]))).){3}(([0-1]?\\d\\d?"
-                ")|((2[0-4]\\d)|(25[0-5])))"),
-            name_(R"([a-z]{1,9})"),
-            detail_key_(R"([A-Za-z0-9_]{1,})") {}
+          : account_id_pattern_(R"([a-z]{1,9}\@[a-z]{1,9})"),
```
motxx: account_id validation has changed to RFC1035.
It is difficult to express it by regex. `boost::spirit::qi`(EBNF) is used and the parser is implemented in `validator::isValidDomainName()`
https://github.com/hyperledger/iroha/pull/694

shared_model/validators/field_validator.hpp
```diff
@@ -160,33 +203,46 @@ namespace shared_model {
...
 
      private:
-      std::regex account_id_, asset_id_, ip_address_, name_, detail_key_;
+      std::regex account_id_pattern_;
+      std::regex asset_id_pattern_;
+      std::regex name_pattern_;
+      std::regex detail_key_pattern_;
       // max-delay between tx creation and validation
-      static constexpr auto MAX_DELAY =
+      static constexpr auto max_delay =
```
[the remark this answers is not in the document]
nickaleks: AFAIK, ALL_CAPS names are reserved for preprocessor macros

test/module/shared_model/validators/field_validator_test.cpp
nickaleks: Twice - no issue, but it can run as many times as number of commands that have this field. Even if it is fast, I would consider 10x increase in test time to be unwanted.

shared_model/validators/field_validator.hpp
nickaleks: I suggest we make that a separate pull request.

shared_model/validators/field_validator.hpp
nickaleks: Right now it is hard to test timestamps, because we cannot verify that message is correct. I will think of a solution and maybe do a separate pull request

shared_model/validators/transaction_validator.hpp
nickaleks: It could be used in different functions, and I do not think static variable provides any major benefit over class member, and class member expresses intent more clearly.

test/module/irohad/consensus/yac/network_test.cpp
```diff
@@ -43,13 +42,15 @@ namespace iroha {
 
           grpc::ServerBuilder builder;
           int port = 0;
-          builder.AddListeningPort(peer.address,
+          builder.AddListeningPort("0.0.0.0:0",
```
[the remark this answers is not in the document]
nickaleks: This change is not part of this PR, I don't know how these commits ended up here, I guess I need to rebase :)

Also I think removing hardcoded literals like this one should be a separate issue, so that they all get organized in one place and used by all tests.

test/module/shared_model/validators/validators_fixture.hpp
nickaleks: Yeah, they are redundant, will remove.

shared_model/validators/field_validator.hpp
```diff
@@ -160,33 +203,46 @@ namespace shared_model {
...
           std::chrono::hours(24) / std::chrono::milliseconds(1);
```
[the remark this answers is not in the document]
nickaleks: The problem is that we need to add 'using namespace std::chrono_literals;', which is not allowed in header files by sorabot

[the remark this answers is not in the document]
l4l: Ah, I see and the exact using cause warning as well. Ok then

test/module/irohad/consensus/yac/network_test.cpp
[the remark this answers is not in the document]
l4l: The problem that you've introduced magic values here. It's better write clean code so there's no need in additional issues for fixing tech debt

test/module/shared_model/validators/field_validator_test.cpp
l4l: Looks quite hackish, but let it be. Hopefully that can save time on debugging in future

shared_model/validators/transaction_validator.hpp
l4l: It's possible to encapsulate this var in separate function for that and I don't see any issues here. Moreover it can be helpful in enhancing with the async support

test/module/shared_model/validators/field_validator_test.cpp
nickaleks: This is hackish, because we are limited by protobuf reflection, if we can get rid of it in the future, this code will hopefully go away.

PR #839, opened 2018-01-06
Delete legacy

neewy:
### Description of the Change

Deleted legacy libraries, fuzzing and tests
- fuzzing
 - base 64
 - generator
 - handler_map
 - ip_tools
 - map_queue
 - random
 - timer
 - corresponding tests

### Benefits

Decrease of legacy code

### Possible Drawbacks 

None


Warchant:
Do not remove base64

neewy:
@Warchant it is used nowhere now

PR #848, opened 2018-01-11
Set iroha user in Dockerfile

x3medima17:
### Description of the Change
This PR enables CI build from non-privileged user, instead of root.
It builds iroha-docker-develop with pre-created iroha user, and simplifies run-iroha-dev script.

### Benefits
More realistic CI builds.
Simpler run-iroha-dev script

### Possible Drawbacks 
None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,6 @@ jobs:
     steps:
       - checkout
       - run: echo $(eval echo "$IROHA_VERSION")
-      - run: echo ~/.ccache; whoami
       - restore_cache:
           keys:
             - v1-build-cache-debug-{{ arch }}-{{ .Branch }}-
@@ -48,9 +47,6 @@ jobs:
               -B$IROHA_BUILD \
               -DCMAKE_BUILD_TYPE=Debug \
               -DIROHA_VERSION=$(eval echo "$IROHA_VERSION")
-      - run:
-          name: whoami
-          command: whoami
       - run:
           name: make
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ defaults: &defaults
   environment:
     IROHA_HOME: /opt/iroha
     IROHA_BUILD: /opt/iroha/build
-    CCHACHE_DIR: /opt/.ccache
+    CCACHE_DIR: /opt/.ccache
     SONAR_DIR: /opt/.sonar
 
 
@@ -30,7 +30,7 @@ jobs:
             - v1-build-cache-debug-{{ arch }}-{{ .Branch }}-
             - v1-build-cache-debug-{{ arch }}-
           paths:
-            - $CCHACHE_DIR
+            - $CCACHE_DIR
 
       - run:
           name: ccache setup
@@ -63,7 +63,7 @@ jobs:
       - save_cache:
           key: v1-build-cache-debug-{{ arch }}-{{ .Branch }}-{{ epoch }}
           paths:
-            - $CCHACHE_DIR
+            - $CCACHE_DIR
       - run:
           name: unit tests, generate xunit-*.xml
           command: cmake --build $IROHA_BUILD --target test
@@ -96,7 +96,7 @@ jobs:
             - v1-build-cache-release-{{ arch }}-{{ .Branch }}-
             - v1-build-cache-release-{{ arch }}-
           paths:
-            - $CCHACHE_DIR
+            - $CCACHE_DIR
       - run:
           name: ccache setup
           command: |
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -21,12 +24,14 @@ jobs:
...
+      - run: echo ~/.ccache; whoami
       - restore_cache:
           keys:
-            - build-cache-debug-{{ arch }}-{{ .Branch }}-
-            - build-cache-debug-{{ arch }}-
+            - v1-build-cache-debug-{{ arch }}-{{ .Branch }}-
+            - v1-build-cache-debug-{{ arch }}-
           paths:
-            - ~/.ccache
+            - $CCHACHE_DIR
```
Warchant: cc**H**ache?

maybe **CCACHE** ?

scripts/run-iroha-dev.sh
```diff
@@ -12,14 +12,10 @@ export COMPOSE_PROJECT_NAME=${PROJECT}
...
-    adduser --disabled-password --gecos "" ${USER} --uid ${USERID}
-  docker-compose -f ${IROHA_HOME}/docker/docker-compose.yml exec node \
-    chown -R ${USER}:${USER} /tmp/ccache
   docker-compose -f ${IROHA_HOME}/docker/docker-compose.yml exec \
-    --user ${USER} node /bin/bash
+    node /bin/bash
   docker-compose -f ${IROHA_HOME}/docker/docker-compose.yml down
 else
   docker-compose -f ${IROHA_HOME}/docker/docker-compose.yml exec \
-    --user ${USER} node /bin/bash
```
Warchant: these lines are needed because if you build iroha locally, then switch to docker environment, you will have troubles with rights

Warchant: these lines are needed because if you build iroha locally, then switch to docker environment, you will have troubles with rights

x3medima17: It will be already logged as iroha user

.circleci/config.yml
x3medima17: Fixed

scripts/run-iroha-dev.sh
Warchant: If I have `bogdan` user with 1000 id, modify some file, set permissions on this file 600, then after login to docker I will be `iroha` user with potentially another ID, and this user will not have rights to modify files inside a container.

.circleci/config.yml
```diff
@@ -15,18 +18,20 @@ jobs:
...
     <<: *version
     docker:
-      - image: hyperledger/iroha-docker-develop
+      - image: hyperledger/iroha-docker-develop:v1
       - image: postgres:9.5
       - image: redis:3.2.8
     steps:
       - checkout
       - run: echo $(eval echo "$IROHA_VERSION")
+      - run: echo ~/.ccache; whoami
```
l4l: `echo ccache`? that is redundant i guess

docker/develop/Dockerfile
```diff
@@ -194,5 +194,12 @@ RUN git clone git://github.com/warchant/ed25519.git /tmp/ed25519; \
     ldconfig; \
     rm -rf /tmp/ed25519
 
+RUN adduser iroha-ci
 WORKDIR /opt/iroha
+RUN chmod -R 777 /opt/iroha; \
+    mkdir -p /tmp/ccache; \
+    chmod -R 777 /tmp/
```
l4l: just `mkdir -m 777 -p /tmp/ccache` should be enough, why the whole dir?

.circleci/config.yml
Warchant: Debug output. Remove

.circleci/config.yml
```diff
@@ -43,6 +48,9 @@ jobs:
               -B$IROHA_BUILD \
               -DCMAKE_BUILD_TYPE=Debug \
               -DIROHA_VERSION=$(eval echo "$IROHA_VERSION")
+      - run:
+          name: whoami
+          command: whoami
```
Warchant: Debug output. Remove.

docker/develop/Dockerfile
```diff
@@ -194,5 +194,12 @@ RUN git clone git://github.com/warchant/ed25519.git /tmp/ed25519; \
     ldconfig; \
     rm -rf /tmp/ed25519
 
+RUN adduser iroha-ci
 WORKDIR /opt/iroha
+RUN chmod -R 777 /opt/iroha; \
```
Warchant: there is no iroha inside. it is mounted from host in CI and in script.

scripts/run-iroha-dev.sh
```diff
@@ -1,8 +1,14 @@
 #!/bin/bash
 
-USER=iroha
 USERID=$(id -u $(whoami))
 
+
```
Warchant: add check that `IROHA_HOME` is set.

if not set -- print error message and exit with non-0 exit code

x3medima17: We assign IROHA_HOME in this script

PR #851, opened 2018-01-13
Fix/consensus order

kamilsa:
### Description of the Change

Currently yac consensus's function currentLeader does not have check on whether order vector is empty or not. This PR adds this check by means of optional.

### Benefits

Error handling on when order vector is empty.

### Possible Drawbacks 

Too many places in code, needed to be updated because of that change

### DIFF
--- a/irohad/consensus/yac/cluster_order.hpp
+++ b/irohad/consensus/yac/cluster_order.hpp
@@ -31,7 +31,13 @@ namespace iroha {
       class ClusterOrdering {
        public:
 
-        explicit ClusterOrdering(std::vector<model::Peer> order);
+        /**
+         * Creates cluster ordering from the vector of peers
+         * @param order vector of peers
+         * @return false if vector is empty, true otherwise
+         */
+        static nonstd::optional<ClusterOrdering> create(
+            std::vector<model::Peer> order);
 
         /**
          * Provide current leader peer
@@ -49,13 +55,21 @@ namespace iroha {
          */
         bool hasNext();
 
-        std::vector<model::Peer> getPeers() { return order_; };
+        std::vector<model::Peer> getPeers() {
+          return order_;
+        };
 
-        auto getNumberOfPeers() { return order_.size(); }
+        auto getNumberOfPeers() {
+          return order_.size();
+        }
 
         virtual ~ClusterOrdering() = default;
 
        private:
+        // prohibit creation of the object not from create method
+        explicit ClusterOrdering(std::vector<model::Peer> order);
+        ClusterOrdering() = delete;
+
         std::vector<model::Peer> order_;
         uint32_t index_ = 0;
       };
--- a/irohad/consensus/yac/cluster_order.hpp
+++ b/irohad/consensus/yac/cluster_order.hpp
@@ -30,14 +30,13 @@ namespace iroha {
        */
       class ClusterOrdering {
        public:
-
         /**
          * Creates cluster ordering from the vector of peers
          * @param order vector of peers
          * @return false if vector is empty, true otherwise
          */
         static nonstd::optional<ClusterOrdering> create(
-            std::vector<model::Peer> order);
+            const std::vector<model::Peer> &order);
 
         /**
          * Provide current leader peer
--- a/irohad/consensus/yac/impl/cluster_order.cpp
+++ b/irohad/consensus/yac/impl/cluster_order.cpp
@@ -15,19 +15,26 @@
  * limitations under the License.
  */
 
-#include <utility>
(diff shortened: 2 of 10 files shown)


irohad/consensus/yac/cluster_order.hpp
```diff
@@ -36,7 +36,7 @@ namespace iroha {
         /**
          * Provide current leader peer
          */
-        model::Peer currentLeader();
+        nonstd::optional<model::Peer> currentLeader();
```
l4l: 1. Why not boost?
2. Header seems missing

irohad/consensus/yac/impl/yac.cpp
```diff
@@ -128,12 +128,15 @@ namespace iroha {
...
-          timer_->invokeAfterDelay(delay_,
-                                   [this, vote] { this->votingStep(vote); });
-        }
+
+        cluster_order_.currentLeader() | [&](auto leader){
+          network_->send_vote(leader, vote);
+          cluster_order_.switchToNext();
+          if (cluster_order_.hasNext()) {
+            timer_->invokeAfterDelay(delay_,
+                                     [this, vote] { this->votingStep(vote); });
```
l4l: capture `vote` by-ref

irohad/consensus/yac/cluster_order.hpp
kamilsa: Which header?

l4l: for optional

irohad/consensus/yac/cluster_order.hpp
```diff
@@ -49,13 +55,20 @@ namespace iroha {
...
-        auto getNumberOfPeers() { return order_.size(); }
+        auto getNumberOfPeers() {
+          return order_.size();
+        }
 
         virtual ~ClusterOrdering() = default;
 
        private:
+        explicit ClusterOrdering(std::vector<model::Peer> order);
+        ClusterOrdering() {}
```
muratovv: what is the reason for presence of empty ctor here?

irohad/consensus/yac/impl/cluster_order.cpp
```diff
@@ -15,19 +15,26 @@
...
-#include <utility>
 #include "consensus/yac/cluster_order.hpp"
 
 namespace iroha {
   namespace consensus {
     namespace yac {
 
+      nonstd::optional<ClusterOrdering> ClusterOrdering::create(
+          std::vector<model::Peer> order) {
+        if (order.empty()) {
```
muratovv: minor issue: `if` statement may be simplified with the ternary operator.

test/module/irohad/consensus/yac/yac_mocks.hpp
```diff
@@ -139,9 +145,9 @@ namespace iroha {
 
         MockYacPeerOrderer() = default;
 
-        MockYacPeerOrderer(const MockYacPeerOrderer &rhs) {};
+        MockYacPeerOrderer(const MockYacPeerOrderer &rhs){};
```
muratovv: Is it changes happen with clang-format formatting? If no, check the whole pr and fix it.

irohad/consensus/yac/cluster_order.hpp
kamilsa: To make sure that creation of this object is only possible with create method

test/module/irohad/consensus/yac/yac_mocks.hpp
kamilsa: Yes, because of clang-formatting

irohad/consensus/yac/cluster_order.hpp
Warchant: public:
ClusterOrdering() = delete;

Warchant: https://stackoverflow.com/questions/13654927/why-explicitly-delete-the-constructor

l4l:
Some test also will be nice to add here

irohad/consensus/yac/cluster_order.hpp
```diff
@@ -31,7 +31,13 @@ namespace iroha {
...
        public:
 
-        explicit ClusterOrdering(std::vector<model::Peer> order);
+        /**
+         * Creates cluster ordering from the vector of peers
+         * @param order vector of peers
+         * @return false if vector is empty, true otherwise
+         */
+        static nonstd::optional<ClusterOrdering> create(
+            std::vector<model::Peer> order);
```
l4l: Can be const-ref

PR #852, opened 2018-01-13
Add missing get tx swig tests

l4l:
Seems that swig test for GetTransaction was missed, here's the simple fix

### DIFF
--- a/test/module/shared_model/bindings/query-test.py
+++ b/test/module/shared_model/bindings/query-test.py
@@ -77,5 +77,9 @@ class BuilderTest(unittest.TestCase):
     query = self.builder.getRolePermissions("user").build()
     self.assertTrue(self.check_proto_query(self.proto(query)))
 
+  def test_get_transactions(self):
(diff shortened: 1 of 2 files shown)


test/module/shared_model/bindings/query-test.py
```diff
@@ -77,5 +77,9 @@ def test_get_role_permissions(self):
     query = self.builder.getRolePermissions("user").build()
     self.assertTrue(self.check_proto_query(self.proto(query)))
 
+  def test_get_transactions(self):
+    query = self.builder.getTransactions([iroha.Hash("1" * 256), iroha.Hash("2" * 256)]).build()
```
motxx: I don't know much about this code, but `"1" * 256` is 256 bytes? sha3_256 is 256 bit, so maybe `"1" * 32`?

l4l: Yeah, you're write. Anywhere it just builder test and it doesn't really care about the hash size

muratovv:
Pls, fix DCO.

PR #859, opened 2018-01-14
Feature/yac order impl

grimadas:
### Description of the Change

YAC relies on uniform peer order distribution for leaders selection, provided by getOrdering function.     
In this pr we implement pseudorandom ordering function which returns peer order uniformly distributed depending on the hash of the block.  

### Benefits

It is harder for adversary  to attack rotating leader committee. 

### Possible Drawbacks 

Current implementation relies on default random generation engine based on http://en.cppreference.com/w/cpp/numeric/random/linear_congruential_engine. 

A toy test for a fairness of distribution is in this pr.  A simple statistic for distribution test can be considered in future, like https://en.wikipedia.org/wiki/Kolmogorov–Smirnov_test 

It is interesting to implement for the future alternative order function with exponential distribution depending on various statistics, like availability time, probability of maliciousness etc.

### DIFF
--- a/irohad/consensus/yac/impl/peer_orderer_impl.cpp
+++ b/irohad/consensus/yac/impl/peer_orderer_impl.cpp
@@ -16,6 +16,7 @@
  */
 
 #include "consensus/yac/impl/peer_orderer_impl.hpp"
+#include <random>
 
 namespace iroha {
   namespace consensus {
@@ -33,8 +34,12 @@ namespace iroha {
 
       nonstd::optional<ClusterOrdering> PeerOrdererImpl::getOrdering(
           YacHash hash) {
-        // TODO 01/08/17 Muratov: implement effective ordering based on hash value IR-504
-        return getInitialOrdering();
+        return query_->getLedgerPeers() | [&hash](auto peers){
+          std::seed_seq seed(hash.block_hash.begin(),hash.block_hash.end());
+          std::default_random_engine gen(seed);
+          std::shuffle(peers.begin(), peers.end(), gen);
+          return nonstd::make_optional<ClusterOrdering>(peers);
+        };
       }
     }  // namespace yac
   }    // namespace consensus
--- a/irohad/consensus/yac/impl/peer_orderer_impl.cpp
+++ b/irohad/consensus/yac/impl/peer_orderer_impl.cpp
@@ -27,18 +27,18 @@ namespace iroha {
           : query_(std::move(peer_query)) {}
 
       nonstd::optional<ClusterOrdering> PeerOrdererImpl::getInitialOrdering() {
-        return query_->getLedgerPeers() | [](const auto &peers) {
-          return ClusterOrdering::create(peers);
-        };
+        return query_->getLedgerPeers() |
+            [](const auto &peers) { return ClusterOrdering::create(peers); };
       }
 
       nonstd::optional<ClusterOrdering> PeerOrdererImpl::getOrdering(
-          const YacHash& hash) {
-        return query_->getLedgerPeers() | [&hash](auto peers){
-          std::seed_seq seed(hash.block_hash.begin(),hash.block_hash.end());
+          const YacHash &hash) {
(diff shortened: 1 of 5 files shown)


test/module/irohad/consensus/yac/peer_orderer_test.cpp
```diff
@@ -82,3 +83,36 @@ TEST_F(YacPeerOrdererTest, PeerOrdererOrderingWhenInvokeFaillCase) {
...
+                                      [](std::string res, const auto &peer) {
+                                        return res + " " + peer.address;
+                                      });
+    histogram[res]++;
+  }
+  ASSERT_EQ(comb, histogram.size());
+
+  // Output historgram, should output uniform distribution
+  std::cout << "The histogram output: " << std::endl;
+  std::for_each(histogram.begin(),
```
l4l: Maybe it's better: to check whether `accumulate(..)/times` in some suburb of 1? Otherwise this test is pretty pointless

test/module/irohad/consensus/yac/peer_orderer_test.cpp
```diff
@@ -82,3 +83,36 @@ TEST_F(YacPeerOrdererTest, PeerOrdererOrderingWhenInvokeFaillCase) {
...
+                histogram.end(),
+                [](const std::pair<std::string, int> &element) {
+                  std::cout << element.first << " :: " << element.second
+                            << std::endl;
+                });
+}
```
l4l: newline

test/module/irohad/consensus/yac/peer_orderer_test.cpp
```diff
@@ -82,3 +83,36 @@ TEST_F(YacPeerOrdererTest, PeerOrdererOrderingWhenInvokeFaillCase) {
...
+TEST_F(YacPeerOrdererTest, FairnessTest) {
+  // Calculate number of permutations of peers
+  size_t comb = std::tgamma(N_PEERS+1);
+  // Run experiments 30 times for each combination
+  size_t times = comb * 30;
+  std::unordered_map<std::string, int> histogram;
+  EXPECT_CALL(*wsv, getLedgerPeers())
+      .Times(times)
+      .WillRepeatedly(Return(peers));
+  for (size_t t = 1; t <= times; ++t) {
```
l4l: `boost::counting_range(1, times)`?
Would be nice to use transform here

test/module/irohad/consensus/yac/peer_orderer_test.cpp
```diff
@@ -16,12 +16,11 @@
  */
 
 #include <iostream>
-#include <memory>
+#include <cmath>
```
l4l: Is that the clang-format's order?

irohad/consensus/yac/impl/peer_orderer_impl.cpp
```diff
@@ -33,8 +34,12 @@ namespace iroha {
 
       nonstd::optional<ClusterOrdering> PeerOrdererImpl::getOrdering(
           YacHash hash) {
```
l4l: By ref maybe?

test/module/irohad/consensus/yac/peer_orderer_test.cpp
```diff
@@ -82,3 +83,36 @@ TEST_F(YacPeerOrdererTest, PeerOrdererOrderingWhenInvokeFaillCase) {
   auto order = orderer.getOrdering(YacHash());
   ASSERT_EQ(order, nonstd::nullopt);
 }
+
+TEST_F(YacPeerOrdererTest, FairnessTest) {
```
muratovv: Need to add `given, when, then` notation for the test.

test/module/irohad/consensus/yac/peer_orderer_test.cpp
muratovv: But in this case will be potential fails of the test because of "bad" random.

PR #867, opened 2018-01-16
Add tests for not covered branches in command executor

nickaleks:
### Description of the Change

This pull request adds several tests to cover missed branches in command_executor.cpp.
### Benefits

Increased code coverage for command executor

### Possible Drawbacks 
Tests are very implementation dependent because use ametsuchi mocks.

### DIFF
--- a/test/module/irohad/model/command_validate_execute_test.cpp
+++ b/test/module/irohad/model/command_validate_execute_test.cpp
@@ -1213,6 +1213,9 @@ TEST_F(TransferAssetTest, InvalidWhenWrongPrecisionDuringExecute) {
 }
 
 TEST_F(TransferAssetTest, InvalidWhenAmountOverflow) {
+  Amount max_amount(std::numeric_limits<boost::multiprecision::uint256_t>::max(), 2);
+  src_wallet.balance = max_amount;
+
   EXPECT_CALL(*wsv_query, getAsset(transfer_asset->asset_id)).WillOnce(Return(asset));
   EXPECT_CALL(
       *wsv_query,
@@ -1223,9 +1226,8 @@ TEST_F(TransferAssetTest, InvalidWhenAmountOverflow) {
       getAccountAsset(transfer_asset->dest_account_id, transfer_asset->asset_id))
       .WillOnce(Return(dst_wallet));
 
-  // More than account
-  Amount max_amount(std::numeric_limits<boost::multiprecision::uint256_t>::max(), 2);
-  transfer_asset->amount = max_amount;
+  // More than account balance
+  transfer_asset->amount = (max_amount - Amount(100, 2)).value();
 
   auto executor = factory->getCommandExecutor(command);
   ASSERT_FALSE(executor->execute(
--- a/test/module/irohad/model/command_validate_execute_test.cpp
+++ b/test/module/irohad/model/command_validate_execute_test.cpp
@@ -79,6 +79,16 @@ class CommandValidateExecuteTest : public ::testing::Test {
                 *command, *wsv_query, *wsv_command, creator.account_id);
   }
 
+
+  // skip validation and execute command
+  bool execute() {
+    auto executor = factory->getCommandExecutor(command);
+    return executor->execute(
+        *command, *wsv_query, *wsv_command, creator.account_id);
+  }
+
+
+  Amount max_amount(std::numeric_limits<boost::multiprecision::uint256_t>::max(), 2);
   std::string admin_id = "admin@test", account_id = "test@test",
               asset_id = "coin#test", domain_id = "test",
               description = "test transfer";
@@ -228,7 +238,6 @@ TEST_F(AddAssetQuantityTest, InvalidWhenNoAsset) {
 
 TEST_F(AddAssetQuantityTest, InvalidWhenAssetAdditionFails) {
   // amount overflows
-  Amount max_amount(std::numeric_limits<boost::multiprecision::uint256_t>::max(), 2);
   add_asset_quantity->amount = max_amount;
 
   EXPECT_CALL(*wsv_query,
@@ -1170,9 +1179,7 @@ TEST_F(TransferAssetTest, InvalidWhenInsufficientFundsDuringExecute) {
   // More than account
   transfer_asset->amount = Amount(155, 2);
 
-  auto executor = factory->getCommandExecutor(command);
-  ASSERT_FALSE(executor->execute(
-      *command, *wsv_query, *wsv_command, creator.account_id));
+  ASSERT_FALSE(execute());
 }
 
 TEST_F(TransferAssetTest, InvalidWhenWrongPrecision) {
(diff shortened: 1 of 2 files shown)


l4l:
screenshot_20180117_005423
screenshot_20180117_005435
screenshot_20180117_005435

l4l:
Imo that's pretty hard and probably out of that pr, but we can also add tests for broken database so we can cover cases like [command_executor.cpp:94](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-94)

Also [value_or](https://github.com/martinmoene/optional-lite#interface-of-optional-lite) function can slightly increase coverage in some cases, e.g [here](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-282) or [here](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-807)
Beside that multiple `optional::value()` calls can be refactored. It also can increase coverage

Plz teview also these lines: [177-182](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-177), [436-438](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-436), [533-560](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-553), [596-603](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-596), [620-622](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-620), [695](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-695), [754-755](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-754), [830-831](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-830), [882-886](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-882)

And especially these lines: [637](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-637), [678-681](https://codecov.io/gh/hyperledger/iroha/pull/867/src/irohad/model/execution/impl/command_executor.cpp?before=irohad/model/execution/impl/command_executor.cpp#D1-678)

motxx:
I suggest renaming `command_validate_execute_test` to another, something like `command_mock_call_test`.
As mentioned in possible drawbacks, commands are not validated by real database, which causes misunderstandings to others in the future.

l4l:
Maybe, sounds logical but I don't really like a new name. Mock is associated with fixtures and call doesn't say anything

motxx:
Right

nickaleks:
@l4l What do you mean by broken database? inconsistent data or when database is down?

About using value_or - command executor and storage will be greatly refactored in the future, to remove code duplication and increase informativeness, that is why making any changes to code is unnecessary at this point.

About 637 - strangely sonar does not show this as uncovered branch
678 - 681 - hard to test, because we have no way of checking what value was added to database

About other lines - sonar shows these branches as covered and codecov as partially. It seems to be caused by using different methods to calculate coverage.

I will look at 637 and in case something is missing, add test if necessary

test/module/irohad/model/command_validate_execute_test.cpp
```diff
@@ -77,6 +79,15 @@ class CommandValidateExecuteTest : public ::testing::Test {
                 *command, *wsv_query, *wsv_command, creator.account_id);
   }
 
+  // skip validation and execute command
```
muratovv: What does this comment mean? Looks redundant.

test/module/irohad/model/command_validate_execute_test.cpp
```diff
@@ -198,8 +209,15 @@ TEST_F(AddAssetQuantityTest, InvalidWhenWrongPrecision) {
 }
 
 TEST_F(AddAssetQuantityTest, InvalidWhenNoAccount) {
```
muratovv: Pls, cover all tests, that you add and change with `given, when, then`.

nickaleks:
@muratovv fixed test description and comment for execute function

l4l:
> What do you mean by broken database? inconsistent data or when database is down?

Inconsistency, yes. We probably have some number of bugs and probably some of them are related to db, so if the code can be protected from this kind of bug is a significant security plus.
And unavailable database is problem of deploying/installing/configuring, not ours imo.

> About using value_or - command executor and storage will be greatly refactored in the future, to remove code duplication and increase informativeness, that is why making any changes to code is unnecessary at this point.

Ok, let it be like it now

>About 637 - strangely sonar does not show this as uncovered branch
> 678 - 681 - hard to test, because we have no way of checking what value was added to database
>
> About other lines - sonar shows these branches as covered and codecov as partially. It seems to be caused by using different methods to calculate coverage.
> 
> I will look at 637 and in case something is missing, add test if necessary

Sonar has pretty awful (well, maybe it's biased) info about the coverage, it's better to look at info from codecov. Partial basically means that not-all-cases of the line covered or most probably the following two cases:
- There's can be non-covered exceptions and it's mostly fine to leave as-is
- Missed some logic-dependent branching is not really cool

Note also that, e.g 637 can be due to exception from `std::string`'s constructor (out_of_memory or smth like that)

test/module/irohad/model/command_validate_execute_test.cpp
```diff
@@ -978,6 +1240,29 @@ TEST_F(TransferAssetTest, InvalidWhenInsufficientFunds) {
   ASSERT_FALSE(validateAndExecute());
 }
 
+/**
+ * @given TransferAsset
+ * @when command tries to transfer amount which is lesson than source balance
```
lebdron: lesson->less

test/module/irohad/model/command_validate_execute_test.cpp
```diff
@@ -978,6 +1240,29 @@ TEST_F(TransferAssetTest, InvalidWhenInsufficientFunds) {
...
+  EXPECT_CALL(
+      *wsv_query,
+      getAccountAsset(transfer_asset->src_account_id, transfer_asset->asset_id))
+      .WillOnce(Return(src_wallet));
+  EXPECT_CALL(*wsv_query,
+              getAccountAsset(transfer_asset->dest_account_id,
+                              transfer_asset->asset_id))
+      .WillOnce(Return(dst_wallet));
+
+  // More than account
```
lebdron: More than account balance

PR #869, opened 2018-01-16
Feature/itf send query

lebdron:
### Description of the Change
Add sendQuery method to Integration Test Framework.

### Benefits
More features to test.

### Possible Drawbacks 
None.


irohad/model/converters/impl/pb_query_response_factory.cpp
```diff
@@ -317,6 +351,43 @@ namespace iroha {
...
+            result.reason = ErrorResponse::NOT_SUPPORTED;
+            break;
+          case protocol::ErrorResponse_Reason_NO_ASSET:
+            result.reason = ErrorResponse::NO_ASSET;
+            break;
+          case protocol::ErrorResponse_Reason_NO_ROLES:
+            result.reason = ErrorResponse::NO_ROLES;
+            break;
+          default:
+            break;
```
Warchant: why empty default?

https://github.com/hyperledger/iroha/pull/869/files#diff-54506143c6fcf71e8b5954f1e82bf10fR118
Here it is not empty

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -107,6 +107,12 @@ namespace integration_framework {
     return *this;
   }
 
+  IntegrationTestFramework &IntegrationTestFramework::sendQuery(
+      const iroha::model::Query &qry) {
+    sendQuery(qry, [](const auto &) {});
+    return *this;
+  }
```
Warchant: Cover with test.

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -107,6 +107,12 @@ namespace integration_framework {
...
+
   IntegrationTestFramework &IntegrationTestFramework::skipProposal() {
     checkProposal([](const iroha::model::Proposal &) {});
     return *this;
```
Warchant: Cover with test

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -126,6 +131,36 @@ namespace integration_framework {
...
+    };
+    if (not pb_qry) {
+      log_->info("query serialization failed");
+    }
+    return *this;
+  }
+
   template <typename Lambda>
   IntegrationTestFramework &IntegrationTestFramework::checkBlock(
       Lambda validation) {
```
Warchant: this fn is also not covered

test/framework/integration_framework/integration_test_framework.cpp
lebdron: I really hope these messages are not automated. In this case a simple "Cover `sendQuery(qry)` method" message would have been enough.

test/framework/integration_framework/integration_test_framework.hpp
lebdron: Which function? If it was not added in this pull request, then it should be covered in a separate task.

irohad/model/converters/impl/pb_query_response_factory.cpp
lebdron: This legacy code was extended only for test purposes. This will be replaced with shared model soon.

test/framework/integration_framework/itf_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+        .setInitialState()
+        .sendQuery(
+            query,
+            [](const auto &res) {
+              try {
+                auto err_res =
+                    dynamic_cast<const iroha::model::ErrorResponse &>(res);
+                ASSERT_EQ(iroha::model::ErrorResponse::STATELESS_INVALID,
+                          err_res.reason);
+              } catch (const std::bad_cast &e) {
```
l4l: gtest can print something meaningful about exceptions so don't sure it's a good idea to add try-catch here
Imo it's only increase the test complexity (and reduce test coverage lol)

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -126,6 +131,36 @@ namespace integration_framework {
...
+
+  template <typename Lambda>
+  IntegrationTestFramework &IntegrationTestFramework::sendQuery(
+      const iroha::model::Query &qry, Lambda validation) {
+    using iroha::operator|;
+    log_->info("send query");
+    // serialize without calling destructor on passed reference
+    auto pb_qry = iroha::model::converters::PbQueryFactory().serialize(
+        std::shared_ptr<const iroha::model::Query>(&qry, [](auto) {}));
+    pb_qry | [&](const auto &pb_qry) {
```
l4l: Maybe just `if (pb_qry) {..}`? Bind only make it complicated
Or even
```cpp
if (not pb_qry) {
  log_->info(..);
  return *this;
}
auto query = pb_qry.value();
//work with query
```

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -126,6 +131,36 @@ namespace integration_framework {
...
+      // send
+      iroha::protocol::QueryResponse pb_response;
+      iroha_instance_->getIrohaInstance()->getQueryService()->FindAsync(
+          pb_qry, pb_response);
+      // deserialize
+      auto response =
+          iroha::model::converters::PbQueryResponseFactory().deserialize(
+              pb_response);
+      // check validation function
+      response | [&](const auto &p) { validation(*p); };
```
l4l: Same here

irohad/model/converters/impl/pb_query_response_factory.cpp
```diff
@@ -85,6 +85,40 @@ namespace iroha {
         return response;
       }
 
+      optional_ptr<model::QueryResponse> PbQueryResponseFactory::deserialize(
+          const protocol::QueryResponse &response) const {
+        auto p = [](auto &&o) {
+          std::shared_ptr<QueryResponse> result =
+              std::make_shared<std::remove_reference_t<decltype(o)>>(
```
l4l: I think `decltype(o)` can be extracted as additional variable

irohad/model/converters/pb_query_response_factory.hpp
```diff
@@ -20,6 +20,7 @@
 
 #include <responses.pb.h>
 #include <model/account_asset.hpp>
```
l4l: Fix these two headers as well plz

test/framework/integration_framework/itf_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace integration_framework {
+
+  /**
+   * @given unsigned empty GetAccount query
+   * AND default-initialized IntegrationTestFramework
+   * @when query is sent to the framework
+   * @then query response is STATELESS_INVALID
+   */
+  TEST(ItfTest, SendQueryWithValidation) {
```
l4l: Also i'm not sure that it's a good name here. This test actually tests the tx pipeline, not itf

irohad/model/converters/impl/pb_query_response_factory.cpp
lebdron: Not necessary, since it will be replaced with shared model in near future.

irohad/model/converters/pb_query_response_factory.hpp
lebdron: Same issue, this code should be deleted soon.

irohad/model/converters/impl/pb_query_response_factory.cpp
l4l: nevertheless the old code shouldn't be dirty

test/framework/integration_framework/itf_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "framework/integration_framework/integration_test_framework.hpp"
+
+namespace integration_framework {
```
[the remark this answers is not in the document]
lebdron: https://google.github.io/styleguide/cppguide.html#Namespaces
>Do not use using-directives (e.g. `using namespace foo`).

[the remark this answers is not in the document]
l4l: It used in 99% of our tests

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -126,6 +131,30 @@ namespace integration_framework {
...
+          // send
+          iroha::protocol::QueryResponse pb_response;
+          iroha_instance_->getIrohaInstance()->getQueryService()->FindAsync(
+              pb_qry, pb_response);
+          // deserialize
+          return iroha::model::converters::PbQueryResponseFactory().deserialize(
+              pb_response);
+          // check validation function
+        }
+        | [&](const auto &p) { validation(*p); };
```
muratovv: Validation doesn't cover when appears nullopt.

irohad/model/converters/pb_query_response_factory.hpp
```diff
@@ -18,18 +18,19 @@
...
 #include "model/queries/responses/account_response.hpp"
-#include <nonstd/optional.hpp>
+#include "model/queries/responses/asset_response.hpp"
 #include "model/queries/responses/error_response.hpp"
+#include "model/queries/responses/roles_response.hpp"
 #include "model/queries/responses/signatories_response.hpp"
 #include "model/queries/responses/transactions_response.hpp"
-#include "model/queries/responses/roles_response.hpp"
-#include "model/queries/responses/asset_response.hpp"
+#include "responses.pb.h"
```
Warchant: IMO it is better to create single header `model/responses.hpp`, which includes all these responses. And then, include it here and in other places.

test/framework/integration_framework/iroha_instance.hpp
```diff
@@ -89,9 +92,9 @@ namespace integration_framework {
 
     // config area
     const std::string block_store_dir_ = "/tmp/block_store";
-    const std::string redis_host_ = getRedisHost();
-    const size_t redis_port_ = getRedisPort();
-    const std::string pg_conn_ = getPostgreCreds();
+    const std::string redis_host_ = getRedisHostOrDefault();
+    const size_t redis_port_ = getRedisPortOrDefault();
+    const std::string pg_conn_ = getPostgreCredsOrDefault();
```
Warchant: You don't need many methods. Look how configs are implemented in `trunk/cliutils`: https://github.com/hyperledger/iroha/blob/trunk/cliutils/test/integration/pipeline/tx_pipeline_integration_test.cpp#L104

You can at least reuse this method:
https://github.com/hyperledger/iroha/blob/trunk/cliutils/libs/util/string.hpp#L73

And make better configs:
https://github.com/hyperledger/iroha/blob/trunk/cliutils/test/integration/pipeline/tx_pipeline_integration_test.cpp#L42-L46

lebdron: Separate task was created for this issue.

irohad/model/converters/pb_query_response_factory.hpp
lebdron: This is done in shared model in some sense, which will replace this code in near future.

PR #873, opened 2018-01-17
Add comments to irohad.cpp and application.cpp

victordrobny:

### Description of the Change
Add comments to irohad.cpp and application.cpp

### DIFF
--- a/irohad/main/application.cpp
+++ b/irohad/main/application.cpp
@@ -61,7 +61,7 @@ Irohad::Irohad(const std::string &block_store_dir,
 }
 
 Irohad::~Irohad() {
-  // Shutting down listing services used by internal server
+  // Shutting down services used by internal server
   if (internal_server) {
     internal_server->Shutdown();
   }
--- a/irohad/main/application.cpp
+++ b/irohad/main/application.cpp
@@ -28,6 +28,9 @@ using namespace iroha::torii;
 using namespace iroha::model::converters;
 using namespace iroha::consensus::yac;
 
+/*
+ * Configuring iroha daemon
+ */
 Irohad::Irohad(const std::string &block_store_dir,
                const std::string &redis_host,
                size_t redis_port,
@@ -52,24 +55,33 @@ Irohad::Irohad(const std::string &block_store_dir,
       keypair(keypair) {
   log_ = logger::log("IROHAD");
   log_->info("created");
+  // Initializing storage at this point in order to read genesis block before
+  // initialization of iroha deamon
   initStorage();
 }
 
 Irohad::~Irohad() {
+  // Shutting down grpc server
   if (internal_server) {
     internal_server->Shutdown();
   }
+  // Shutting down torii server
   if (torii_server) {
     torii_server->shutdown();
   }
+  // Waiting until grpc server thread dies
   if (internal_thread.joinable()) {
     internal_thread.join();
   }
+  // Waiting until torii server thread dies
   if (server_thread.joinable()) {
     server_thread.join();
   }
 }
 
+/*
+ * Initializing iroha daemon
+ */
 void Irohad::init() {
   initProtoFactories();
   initPeerQuery();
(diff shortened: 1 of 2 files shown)


irohad/main/application.cpp
```diff
@@ -28,6 +28,9 @@ using namespace iroha::torii;
 using namespace iroha::model::converters;
 using namespace iroha::consensus::yac;
 
+/*
```
x3medima17: First line should start with /**

irohad/main/application.cpp
```diff
@@ -52,24 +55,33 @@ Irohad::Irohad(const std::string &block_store_dir,
...
   if (internal_thread.joinable()) {
     internal_thread.join();
   }
+  // Waiting until torii server thread dies
   if (server_thread.joinable()) {
     server_thread.join();
   }
 }
 
+/*
```
x3medima17: Same as above

irohad/main/application.cpp
```diff
@@ -87,15 +99,26 @@ void Irohad::init() {
   initQueryService();
 }
 
-void Irohad::dropStorage() { storage->dropStorage(); }
+/*
```
x3medima17: Same

libs/crypto/keys_manager_impl.cpp
```diff
@@ -52,66 +52,51 @@ namespace iroha {
...
-    std::string client_priv_key;
+  nonstd::optional<iroha::keypair_t> KeysManagerImpl::loadKeys() {
+    std::string pub_key;
+    std::string priv_key;
 
-    //TODO 15/11/17 motxx - better way to suppress the evaluation?
-    auto blocking = nonstd::make_optional<iroha::keypair_t>()
-        | loadFile(account_name_ + ".pub", client_pub_key)
-        | loadFile(account_name_ + ".priv", client_priv_key);
+    if (!loadFile(account_name_ + ".pub", pub_key)
```
x3medima17: Use `not` instead of `!`, and `or` instead of `||`

irohad/main/irohad.cpp
```diff
@@ -82,17 +112,23 @@ int main(int argc, char *argv[]) {
...
   }
 
+  // Check if genesis block path was specified
   if (not FLAGS_genesis_block.empty()) {
+    // If it is so, read genesis block and store it to iroha storage
     iroha::main::BlockInserter inserter(irohad.storage);
     auto file = inserter.loadFile(FLAGS_genesis_block);
     auto block = inserter.parseBlock(file.value());
 
+    // Check if genesis block was empty
```
lebdron: `parseBlock` will verify that JSON is well formed and that it has all the fields that are required by the block structure. So it is more clear to write something like "Check that provided genesis block file was correct".

irohad/main/irohad.cpp
```diff
@@ -82,17 +112,23 @@ int main(int argc, char *argv[]) {
                 std::chrono::milliseconds(config[mbr::LoadDelay].GetUint()),
                 keypair);
 
+  // Check if iroha daemon storage is properly set
```
lebdron: Maybe "was successfully initialized" is more specific than "is properly set".

irohad/main/application.cpp
```diff
@@ -52,24 +55,33 @@ Irohad::Irohad(const std::string &block_store_dir,
       keypair(keypair) {
   log_ = logger::log("IROHAD");
   log_->info("created");
+  // Initializing storage at this point in order to read genesis block before
+  // initialization of iroha deamon
```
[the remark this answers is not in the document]
victordrobny: It is because all of the init* functions are placed in Irohad::init() except for initStorage. It seems a bit non consistent, but there is a reason for it. I tried to specify the reason.

irohad/main/irohad.cpp
```diff
@@ -26,50 +26,80 @@
 
 #include "logger/logger.hpp"
 
+/**
+ * Validator for the configuration file path input argument.
+ * Path is considered to be valid if it is not empty.
+ */
 bool validate_config(const char *flag_name, std::string const &path) {
```
l4l: Doxygen-based info is missing :( like `@param` or `@return`

victordrobny: You are right. Here is a new pull request with a fix https://github.com/hyperledger/iroha/pull/882

PR #875, opened 2018-01-17
Rework raw block insertion

muratovv:
### Description of the Change

* rework Irohad initialization with storage block insertions
* rename raw block insertion to raw block loader
* remove insertion stuff from block loader
* fix integration tests
* remove redundant tests from block insertion tests

### Benefits

Easier to manage Irohad because of simplification of initialization.

### DIFF
--- a/irohad/main/impl/raw_block_loader.cpp
+++ b/irohad/main/impl/raw_block_loader.cpp
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-#include "main/raw_block_insertion.hpp"
+#include "main/raw_block_loader.hpp"
 #include "model/converters/json_common.hpp"
 #include <fstream>
 #include <utility>
@@ -24,12 +24,7 @@
 namespace iroha {
   namespace main {
 
-    BlockInserter::BlockInserter(std::shared_ptr<ametsuchi::MutableFactory> factory)
-        : factory_(std::move(factory)) {
-      log_ = logger::log("BlockInserter");
-    }
-
-    nonstd::optional<model::Block> BlockInserter::parseBlock(std::string data) {
+    nonstd::optional<model::Block> BlockLoader::parseBlock(std::string data) {
       auto document = model::converters::stringToJson(data);
       if (not document.has_value()) {
         log_->error("Blob parsing failed");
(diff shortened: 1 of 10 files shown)


irohad/main/impl/raw_block_loader.cpp
```diff
@@ -38,16 +33,7 @@ namespace iroha {
...
-      auto storage = factory_->createMutableStorage();
-      for (auto &&block : blocks) {
-        storage->apply(block, [](const auto &current_block, auto &query,
-                                 const auto &top_hash) { return true; });
-      }
-      factory_->commit(std::move(storage));
-    }
-
-    nonstd::optional<std::string> BlockInserter::loadFile(std::string path) {
+    nonstd::optional<std::string> BlockLoader::loadFile(std::string path) {
```
kamilsa: Is there a case when nullopt can be returned? There should be either check on that file exists or function should return just string. IMO first option is better

test/module/irohad/common/block_insertion_test.cpp
```diff
@@ -17,81 +17,43 @@
...
+/**
+ * @given generated block
+ *
+ * @when convert block to JSON
+ * AND parseBlock() with BlockLoader API
+ *
+ * @then check that blocks are equal
+ */
+TEST(BlockLoaderTest, BlockLoaderWhenParseBlock) {
   cout << "----------| block => string_repr(block)"
```
kamilsa: Do we really need this cout here?

muratovv: No, I 'll fix it.

test/module/irohad/common/block_insertion_test.cpp
```diff
@@ -17,81 +17,43 @@
 
 #include <gmock/gmock.h>
```
lebdron: gmock is not required in this test.

irohad/main/raw_block_loader.hpp
```diff
@@ -35,9 +35,9 @@ namespace iroha {
      * This class will be useful for creating test environment
      * and testing pipeline.
      */
-    class BlockInserter {
+    class BlockLoader {
      public:
-      explicit BlockInserter(std::shared_ptr<ametsuchi::MutableFactory> factory);
+      BlockLoader() = default;
```
lebdron: Logger has to be initialized in constructor, `std::shared_ptr<ametsuchi::MutableFactory> factory_;` is not required now.

test/module/irohad/common/raw_block_loader_test.cpp
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
```
l4l: too old :(

PR #876, opened 2018-01-17
[Hotfix] Fix CI bug: trunk sonar analysis

Warchant:
### Description of the Change

This PR fixes circle CI bug: to interpret string as regexp we need to add `/` (slash) from both left and right side of expression, otherwise it is interpreted as a string (branch name).


l4l:
Merged in 60e0a58

PR #877, opened 2018-01-17
Split declaration and definition of FieldValidator and make it as a s…

victordrobny:
### Description of the Change
Defenition and declaration of FieldValidarot is now separated.

### Benefits
It will provide benefits in case we include in several places.

### DIFF
--- a/shared_model/bindings/CMakeLists.txt
+++ b/shared_model/bindings/CMakeLists.txt
@@ -23,6 +23,7 @@ target_link_libraries(bindings
         shared_model_proto_backend
         shared_model_ed25519_sha3
         generator
+        shared_model_field_validator
         )
 
 
--- a/shared_model/bindings/CMakeLists.txt
+++ b/shared_model/bindings/CMakeLists.txt
@@ -23,7 +23,7 @@ target_link_libraries(bindings
         shared_model_proto_backend
         shared_model_ed25519_sha3
         generator
-        shared_model_field_validator
+        shared_model_stateless_validation
         )
 
 
--- a/shared_model/validators/CMakeLists.txt
+++ b/shared_model/validators/CMakeLists.txt
@@ -12,11 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-add_library(shared_model_stateless_validation
-        default_validator.cpp
+
+add_library(shared_model_field_validator
+        field_validator.cpp)
+
+target_link_libraries(shared_model_field_validator
+        schema
+        model_interfaces
         )
+
+add_library(shared_model_stateless_validation
+        default_validator.cpp)
 
 target_link_libraries(shared_model_stateless_validation
         schema
         model_interfaces
+        shared_model_field_validator
         )
+
--- a/shared_model/validators/CMakeLists.txt
+++ b/shared_model/validators/CMakeLists.txt
@@ -13,20 +13,13 @@
 # limitations under the License.
 
 
-add_library(shared_model_field_validator
-        field_validator.cpp)
-
-target_link_libraries(shared_model_field_validator
-        schema
-        model_interfaces
-        )
-
 add_library(shared_model_stateless_validation
-        default_validator.cpp)
+        default_validator.cpp
+        field_validator.cpp
+        )
 
 target_link_libraries(shared_model_stateless_validation
         schema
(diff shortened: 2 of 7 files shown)


shared_model/validators/CMakeLists.txt
```diff
@@ -12,11 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-add_library(shared_model_stateless_validation
-        default_validator.cpp
+
+add_library(shared_model_field_validator
```
nickaleks: Do we need a separate target for that? Is it possible to add it to shared_model_stateless_validation?

victordrobny: You wrote it in the task description(IR-754) :)
It can be added to shared_model_stateless_validation, but it is used in different targets.

nickaleks: Can we change those targets to use shared_model_stateless_validation? I think it would be better to have all validators as one target, as they are very close to each other and are used in similar contexts.

muratovv: I agree with @nickaleks. Better to make one target for all stateless validation for now.

shared_model/validators/CMakeLists.txt
```diff
@@ -12,11 +12,21 @@
...
+        field_validator.cpp)
+
+target_link_libraries(shared_model_field_validator
+        schema
+        model_interfaces
         )
 
+add_library(shared_model_stateless_validation
+        default_validator.cpp)
```
muratovv: Pls, add an end line after the source. Such as:
```
default_validator.cpp
)
```
This approach useful for future changes in file - line will not appear until change a `default_validator.cpp` directly.

shared_model/validators/field_validator.cpp
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "validators/field_validator.hpp"
+
```
l4l: Move related headers as well

shared_model/validators/CMakeLists.txt
```diff
@@ -12,11 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
```
l4l: Why newline?

shared_model/validators/CMakeLists.txt
```diff
@@ -12,11 +12,14 @@
...
 add_library(shared_model_stateless_validation
         default_validator.cpp
+        field_validator.cpp
         )
 
 target_link_libraries(shared_model_stateless_validation
         schema
         model_interfaces
         )
+
```
l4l: same

shared_model/validators/field_validator.cpp
```diff
@@ -0,0 +1,226 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
```
l4l: year :(

shared_model/bindings/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ target_link_libraries(bindings
         shared_model_proto_backend
         shared_model_ed25519_sha3
         generator
+        shared_model_stateless_validation
```
l4l: I think "stateless" shouldn't be here, otherwise it quite mixing up with torii

shared_model/validators/CMakeLists.txt
Warchant: It would be better to create OBJECT target and then combine all targets in `shared_model_stateless_validation`

shared_model/bindings/CMakeLists.txt
victordrobny: What do you mean? Without this dependency build fails

l4l: I meant only naming

victordrobny: This name was already given to the target.

PR #845, opened 2018-01-10
Cover Redis BlockQuery with tests

Warchant:
### Description of the Change

- [x] move `check_consistency`, `Identifier`, `id_to_name` to class FlatFile as static members, instead of free members. I did this at least because I needed `id_to_name` in `FlatFile` interface.
- [x] Fix bug in https://github.com/hyperledger/iroha/blob/develop/irohad/ametsuchi/impl/redis_block_query.cpp#L30-L35 . Test for bug + description is at https://github.com/hyperledger/iroha/compare/develop...feature/cover-redis-negative-cases?expand=1#diff-c855aee60870118507c2cd5177c150eaR241
- [x] Increase coverage of `BlockQuery`.

### Benefits

Bugfixes, increase coverage

### Possible Drawbacks 

None.

PR #881, opened 2018-01-18
Use std::less instead lambda at the ordering

l4l:
Related to [that](https://github.com/hyperledger/iroha/pull/845#discussion_r162268651)


Warchant:
DCO

PR #882, opened 2018-01-18
Add missing doxygen-based info

victordrobny:

### Description of the Change
Add missing doxygen based info in a comments of irohad.cpp

### DIFF
--- a/irohad/main/irohad.cpp
+++ b/irohad/main/irohad.cpp
@@ -27,16 +27,24 @@
 #include "logger/logger.hpp"
 
 /**
+ * Gflag valigator.
  * Validator for the configuration file path input argument.
  * Path is considered to be valid if it is not empty.
+ * @param flag_name - flag name
+ * @param path      - file name
+ * @return true if argument is valid
  */
 bool validate_config(const char *flag_name, std::string const &path) {
   return not path.empty();
 }
 
 /**
+ * Gflag valigator.
  * Validator for the keypair files path input argument.
  * Path is considered to be valid if it is not empty.
+ * @param flag_name - flag name
+ * @param path      - file name
+ * @return true if argument is valid
  */
 bool validate_keypair_name(const char *flag_name, std::string const &path) {
   return not path.empty();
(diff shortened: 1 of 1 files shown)


irohad/main/irohad.cpp
```diff
@@ -27,16 +27,24 @@
 #include "logger/logger.hpp"
 
 /**
+ * Gflag valigator.
  * Validator for the configuration file path input argument.
  * Path is considered to be valid if it is not empty.
+ * @param flag_name - flag name
```
l4l: flag_name - flag name? looks a bit useless

victordrobny: Yes. It is standard gflag validator signature(falg_name and flag_value). In the following function flag_name is always be 'config'

l4l: I mean the comment itself, it's not descriptive at all

irohad/main/irohad.cpp
```diff
@@ -42,8 +42,8 @@ bool validate_config(const char *flag_name, std::string const &path) {
  * Gflag valigator.
  * Validator for the keypair files path input argument.
  * Path is considered to be valid if it is not empty.
- * @param flag_name - flag name
- * @param path      - file name
+ * @param flag_name - flag name. Must be 'keypair' in this case
+ * @param path      - file name. Should be path to the keypair files
```
l4l: Why params has different modals btw (must/should)?

victordrobny: Because flag name can only have one value since it is hardcoded. But the value can has anything inside.

irohad/main/irohad.cpp
```diff
@@ -27,16 +27,24 @@
 #include "logger/logger.hpp"
 
 /**
+ * Gflag valigator.
```
Alexey-N-Chernyshov: valiGator?

irohad/main/irohad.cpp
```diff
@@ -27,16 +27,24 @@
...
+ * @param flag_name - flag name. Must be 'config' in this case
+ * @param path      - file name. Should be path to the config file
+ * @return true if argument is valid
  */
 bool validate_config(const char *flag_name, std::string const &path) {
   return not path.empty();
 }
 
 /**
+ * Gflag valigator.
```
Alexey-N-Chernyshov: +1

PR #896, opened 2018-01-20
Use rttr for cpp reflection

Warchant:
Signed-off-by: Bogdan Vaneev <warchantua@gmail.com>

### Description of the Change

- [x] Integrate [`warchant/rttr`](https://github.com/warchant/rttr) - I had to fork it to fix few bugs related to installation of the library on Linux/Mac machines
- [x] Update develop Dockerfile
- [x] Update `hyperledger/iroha-docker-develop` from `v1` to `v2`
- [x] Add simple test.

### Benefits

This PR introduces C++ Runtime Reflection library. It can be used to write tests for repetitive objects, like commands, queries - etc.

### Possible Drawbacks 

I tested `Findrttr.cmake` locally on MAC, in docker, with or without lib. However, it is possible that script isn't perfect. Also, library seems good but sometimes it may give you a surprise. Be careful. But it is still the best option. 

[test_reflection.cpp](test/framework/test_reflection.cpp)
Documentation: http://www.rttr.org/doc/master/five_minute_tutorial_page.html

- https://github.com/Manu343726/siplasplas - this project offers simple cmake-time clang-based reflection system. I was unable to build it, however I was able to use precompiled distribution from releases. It works only in docker. Branch: https://github.com/hyperledger/iroha/tree/feature/cppreflection
- https://github.com/AustinBrunkhorst/CPP-Reflection - it also offers simple cmake-time clang- based reflection system. Again, I was unable to build it completely. I did not manage to build/run/test this library due to dependency hell.
- https://github.com/Cylix/Reflex - library provides in-place macroses, which can be used for reflection. Syntax looks worse than rttr. 
- https://github.com/billyquith/ponder - pretty much the same as `rttr`. Older. Not as active, as `rttr`.


l4l:
A lot of compile error like this:
```
rttr/detail/method/method_wrapper.h:213:7: error: mangled name for ‘virtual rttr::detail::method_wrapper<bool (std::__cxx11::basic_string<char>::*)() const noexcept, (rttr::access_levels)0, rttr::detail::default_invoke, rttr::detail::default_args<>, rttr::detail::parameter_infos<>, 0>::~method_wrapper()’ will change in C++17 because the exception specification is part of a function type [-Werror=noexcept-type]
```

Warchant:
@l4l how about gcc-6 and clang-3.8 ? They should work.

l4l:
We will move to c++17 soon, so they should work as well

Warchant:
as @lebdron explained, rttr can not be used for reflection, because for our case we need automatic reflection generation. closing.

PR #897, opened 2018-01-20
Fix/postgres wsv coverage

l4l:
### Description of the Change

1. Extraction quoting, concatenation and execution methods as long as they may throw
2. Refactor the PostgresWsvComman with such a fns

Note that sql queries **must not** be changed, if you find some plz write about it

### Benefits

Increased code coverage of the class

### Possible Drawbacks 

None

These functions may be set as a memeber one, though I didn't do as following because then it's possible to move template function definition to the *.cpp file. Also it's pretty generic so I think it can be moved as ametsuchi-wide code

### DIFF
--- a/irohad/ametsuchi/impl/postgres_wsv_command.cpp
+++ b/irohad/ametsuchi/impl/postgres_wsv_command.cpp
@@ -27,20 +27,37 @@ namespace iroha {
     PostgresWsvCommand::PostgresWsvCommand(pqxx::nontransaction &transaction)
         : transaction_(transaction), log_(logger::log("PostgresWsvCommand")) {}
 
+    const char *kUnsuccesfulAlloc = "Unable to alloc: {}";
+
     template <typename T>
     std::string try_quote(PostgresWsvCommand *that, const T &s) noexcept {
       try {
         return that->transaction_.quote(s);
       } catch (std::bad_alloc &e) {
-        that->log_->error("Unable to alloc: {}", e.what());
+        that->log_->error(kUnsuccesfulAlloc, e.what());
+        std::terminate();
+      }
+    }
+
+    std::string try_concat(PostgresWsvCommand *that,
+                           std::initializer_list<std::string> l) noexcept {
+      try {
+        return std::accumulate(
+            l.begin(), l.end(), std::string(), [](std::string s, auto next) {
+              return s + next;
+            });
+      } catch (std::bad_alloc &e) {
+        that->log_->error(kUnsuccesfulAlloc, e.what());
         std::terminate();
       }
     }
 
     bool PostgresWsvCommand::insertRole(const std::string &role_name) {
       try {
-        transaction_.exec("INSERT INTO role(role_id) VALUES ("
-                          + try_quote(this, role_name) + ");");
+        transaction_.exec(try_concat(this,
+                                     {"INSERT INTO role(role_id) VALUES (",
+                                      try_quote(this, role_name),
+                                      ");"}));
       } catch (const std::exception &e) {
         log_->error(e.what());
         return false;
@@ -51,10 +68,13 @@ namespace iroha {
     bool PostgresWsvCommand::insertAccountRole(const std::string &account_id,
                                                const std::string &role_name) {
       try {
-        transaction_.exec(
-            "INSERT INTO account_has_roles(account_id, role_id) VALUES ("
-            + try_quote(this, account_id) + ", " + try_quote(this, role_name)
-            + ");");
+        transaction_.exec(try_concat(
+            this,
+            {"INSERT INTO account_has_roles(account_id, role_id) VALUES (",
+             try_quote(this, account_id),
+             ", ",
+             try_quote(this, role_name),
+             ");"}));
       } catch (const std::exception &e) {
         log_->error(e.what());
         return false;
--- a/irohad/ametsuchi/impl/postgres_wsv_command.cpp
+++ b/irohad/ametsuchi/impl/postgres_wsv_command.cpp
@@ -17,6 +17,7 @@
 
 #include "ametsuchi/impl/postgres_wsv_command.hpp"
 
+#include <exception>
 #include <iterator>
 #include <numeric>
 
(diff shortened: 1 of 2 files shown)


irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
-                          + ");");
-      } catch (const std::exception &e) {
-        log_->error(e.what());
-        return false;
+        return try_quote(
+            that,
+            std::accumulate(l.begin(),
+                            l.end(),
+                            std::string(),
+                            [](std::string s, auto next) { return s + next; }));
```
Warchant: pass by ref

irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
+      } catch (std::bad_alloc &e) {
```
Warchant: const

irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
-            "INSERT INTO account_has_roles(account_id, role_id) VALUES ("
-            + transaction_.quote(account_id)
-            + ", "
-            + transaction_.quote(role_name)
-            + ");");
-      } catch (const std::exception &e) {
-        log_->error(e.what());
-        return false;
+        return that->transaction_.quote(s);
+      } catch (std::bad_alloc &e) {
```
Warchant: const

irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
+    std::string try_concat(PostgresWsvCommand *that,
+                           std::initializer_list<const char *> l) noexcept {
       try {
-        transaction_.exec("DELETE FROM account_has_roles WHERE account_id="
-                          + transaction_.quote(account_id)
-                          + "AND role_id="
-                          + transaction_.quote(role_name)
-                          + ";");
+        return std::accumulate(
+            l.begin(), l.end(), std::string(), [](std::string s, auto next) {
```
Warchant: pass by ref

irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
+              return s + next;
+            });
+      } catch (std::bad_alloc &e) {
```
Warchant: const

irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
+        that->log_->error(kUnsuccesfulAlloc, e.what());
+        std::terminate();
       }
-      return true;
     }
 
-    bool PostgresWsvCommand::deleteAccountRole(const std::string &account_id,
-                                               const std::string &role_name) {
+    std::string try_concat(PostgresWsvCommand *that,
+                           std::initializer_list<const char *> l) noexcept {
```
Warchant: are you sure that T=const char* ? maybe std::string is better?

irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
+                      ";"}));
+    }
+
     bool PostgresWsvCommand::insertRolePermissions(
         const std::string &role_id, const std::set<std::string> &permissions) {
-      auto entry = [this, &role_id](auto permission) {
-        return "(" + transaction_.quote(role_id) + ", "
-            + transaction_.quote(permission) + ")";
+      auto _role_id = try_quote(this, role_id);
+      auto entry = [this, &_role_id](auto permission) {
```
Warchant: const auto& permission

irohad/ametsuchi/impl/postgres_wsv_command.hpp
```diff
@@ -68,6 +68,15 @@ namespace iroha {
           const std::string &permission_id) override;
 
      private:
+      template <typename T>
+      friend std::string try_quote(PostgresWsvCommand *,
+                                   std::initializer_list<T>) noexcept;
```
Warchant: also pass initializer list by ref or const ref

irohad/ametsuchi/impl/postgres_wsv_command.hpp
```diff
@@ -68,6 +68,15 @@ namespace iroha {
...
+      template <typename T>
+      friend std::string try_quote(PostgresWsvCommand *, const T &) noexcept;
+      friend std::string try_concat(
+          PostgresWsvCommand *, std::initializer_list<const char *>) noexcept;
```
Warchant: also pass initializer list by ref or const ref

irohad/ametsuchi/impl/postgres_wsv_command.cpp
l4l: Not really possible, the first one is rvalue and the second is from the initializer list

irohad/ametsuchi/impl/postgres_wsv_command.cpp
l4l: Nope, std::string may throw somewhere

irohad/ametsuchi/impl/postgres_wsv_command.hpp
l4l: Are you sure that you're understand what you say?

irohad/ametsuchi/impl/postgres_wsv_command.cpp
Warchant: auto&& s, auto &&next ?

irohad/ametsuchi/impl/postgres_wsv_command.hpp
Warchant: ah, I missed the point. Okay, sorry.

irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
 #include <numeric>
 
 namespace iroha {
   namespace ametsuchi {
-
     PostgresWsvCommand::PostgresWsvCommand(pqxx::nontransaction &transaction)
         : transaction_(transaction), log_(logger::log("PostgresWsvCommand")) {}
 
-    bool PostgresWsvCommand::insertRole(const std::string &role_name) {
+    const char *kUnsuccesfulAlloc = "Unable to alloc: {}";
```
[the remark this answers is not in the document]
l4l: > Is it a good decision to name variables depending on their type? I know it is a part of Google Style guide, but we do not do that in any other part of the project.

We should start doing this, it's better to have some fixed style rather than doing on thing on its own

> Also constexpr instead of const?

Kinda insignificant

irohad/ametsuchi/impl/postgres_wsv_command.cpp
```diff
@@ -17,335 +17,379 @@
...
+
+    template <typename T>
+    std::string try_quote(PostgresWsvCommand *that,
```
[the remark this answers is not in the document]
l4l: It's all about the coverage, that's it

Warchant:
Rebase please

Warchant:
Why closed?

PR #899, opened 2018-01-22
Result Type

nickaleks:
### Description of the Change
Based on https://github.com/hyperledger/iroha/pull/603, this pull request introduces Resutl<V, E> type which is either a result or an error. It can be used in error handling as a replacement for exceptions, where failure is expected, for example, validation, database access.

### Benefits

1. Allows providing information about error (one of the main drawbacks of optional)
2. More performant than exception handling and does not complicate control flow

### Possible Drawbacks 

1. Functional style match() call is used to retrieve a value, and any other option is no available. It disallows writing procedural style code.

Usage examples are provided in code comments

Instead of match(), various utility functions such as value(), has_value() could be provided, but they encourage writing a lot of boilerplate code, and make code such as Result<bool, ...> too easy to use incorrectly.


l4l:
What about this: ![MINOR][MINOR] [result.hpp#L38](https://github.com/hyperledger/iroha/blob/505dd110a3b3e74fe774c0a08dc195183e7d263a/libs/common/result.hpp#L38): Remove this commented out code. [rule](https://sonar.soramitsu.co.jp/coding_rules#rule_key=cxx%3ACommentedCode)

[MINOR]: https://sonarsource.github.io/sonar-github/severity-minor.png 'Severity: MINOR'

test/module/libs/common/result_test.cpp
```diff
@@ -0,0 +1,154 @@
...
+ * @when bind operator is used to chain these 2 functions and first one return
+ *       error
+ * @then result of bind contains error and second function is not invoked
+ */
+TEST(ResultTest, ResultBindOperatorErrorFirstFunction) {
+  auto result = error_get_int() | never_used_negate_int;
+
+  result.match(
+      [](Value<int> v) { FAIL(); },
+      [](Error<std::string> e) { ASSERT_EQ("first function", e.error); });
```
l4l: [const duplication](https://google.github.io/styleguide/cppguide.html#Constant_Names)

test/module/libs/common/result_test.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+  auto result = get_int() | negate_int;
+
+  result.match([](Value<int> v) { ASSERT_EQ(-10, v.value); },
+               [](Error<std::string> e) { FAIL() << "Unexpected error case"; });
+}
+
+// function which must not be called for test to pass
+auto never_used_negate_int = [](int a) -> Result<int, std::string> {
+  EXPECT_TRUE(false);
```
l4l: why not just `FAIL() << "Function " __FUNCTION__ " was called";`?

test/module/libs/common/result_test.cpp
```diff
@@ -0,0 +1,154 @@
...
+ * @given two functions which return different results, but value and error
+ *        types can be cast to each other
+ * @when bind operator is used to chain these 2 functions
+ * @then result type is properly deduced
+ */
+TEST(ResultTest, ResultBindOperatorCompatibleTypes) {
+  auto result = error_get_int() | never_used_negate_int;
+
+  static_assert(
+      std::is_same<decltype(result), decltype(never_used_negate_int(1))>::value,
```
l4l: `result_of` again

test/module/libs/common/result_test.cpp
```diff
@@ -0,0 +1,154 @@
...
+      [](Error<std::string> e) { ASSERT_EQ("first function", e.error); });
+}
+
+/**
+ * @given Result with void value
+ * @when match function is invoked
+ * @then void type is correctly handled by compiler
+ */
+TEST(ResultTest, ResultVoidValue) {
+  Result<void, std::string> result = makeError("error message");
```
l4l: And again duplication

l4l: I think it's better prohibit result of void
@nickaleks, @lebdron, @muratovv what do you think?

Warchant: yes, you should not be able to create V=void, but you can create void*

lebdron: Any reasons why is it better?

libs/common/result.hpp
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Bind operator allows chaining several functions which return result. If
+     * result contains error, it returns this error, if it contains value,
+     * function f is called.
+     * @param f function which return type must be compatible with original
+     * result
+     */
+    template <typename T, typename E, typename Transform>
+    constexpr auto operator|(Result<T, E> r, Transform &&f)
+        -> decltype(f(std::declval<T>())) {
```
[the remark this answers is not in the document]
nickaleks: As far as I know, result_of is deprecated since C++17

http://en.cppreference.com/w/cpp/types/result_of

test/module/libs/common/result_test.cpp
nickaleks: Result<void> can be used to when you need to know if function was successful, and if it wasn't - know the reason of failure

test/module/libs/common/result_test.cpp
nickaleks: FAIL() does not work, as well as ASSERT_TRUE(false) because it leads to unreachable code, which makes compiler deduce return type as void, instead of Result. This leads to compile error

libs/common/result.hpp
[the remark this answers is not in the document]
l4l: @lebdron, much easier to understand

@nickaleks , that's because `invoke_result_t` was introduced

test/module/libs/common/result_test.cpp
l4l: @Warchant, well, probably `void*` isn't really suitable for us, we have interfaces in such a case. Nevertheless it's the different type.

@lebdron, I don't really sure there's any use-cases. And that probably may lead some problems in future as long as `void` is a quite special type for templates

@nickaleks, Instead of `Result<void, error>` it's possible use just an error, I see no problem here. Also as [I may guess](https://ideone.com/dxexFo) it's not possible to create an empty value type in this case

libs/common/result.hpp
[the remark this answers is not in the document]
nickaleks: If we have result_of -> will have to change to invoke_result_t, but if we leave it as it is, no changes are required

test/module/libs/common/result_test.cpp
nickaleks: And what to put in case there is no error? Default value? This is hard to work with

test/module/libs/common/result_test.cpp
```diff
@@ -24,16 +24,28 @@ using iroha::expected::Value;
 using iroha::expected::makeError;
 using iroha::expected::makeValue;
 
+constexpr auto error_case_message = "Unexpected error case";
```
l4l: maybe `kErrorCaseMsg`?

test/module/libs/common/result_test.cpp
```diff
@@ -24,16 +24,28 @@ using iroha::expected::Value;
...
+constexpr auto value_case_message = "Unexpected value case";
+constexpr auto first_function_call_message = "first function";
+constexpr auto error_message = "error message";
+
+// Function which returns function which when invoked, fails the test
+template <typename T>
+auto makeFailCase(const std::string &error_message) {
+  return [&error_message](T t) { FAIL() << error_message; };
```
l4l: Is it formatted with clang? Afair it enforces new lines for non empty blocks inside braces

test/module/libs/common/result_test.cpp
nickaleks: I do not like the idea of putting k in constant names, it is type information, and names should be type-independent

test/module/libs/common/result_test.cpp
nickaleks: yes it is formatted with clang format

test/module/libs/common/result_test.cpp
l4l: It's not about the type, it semantic info

```cpp
#define "asd"
const char *kAsd = "asd";
constexpr auto kAsd = "asd"s;
```

There's 3 different types and exactly same semantic

nickaleks: const is type specifier since int and const int are 2 different types.

Also appending k to const name is not present anywhere in our code base and I think it is up to debate whether or not it should be.

PR #910, opened 2018-01-25
Move cache from torii to libs.

Alexey-N-Chernyshov:
Signed-off-by: Alexey Chernyshov <alexey.n.chernyshov@gmail.com>

### Description of the Change
Move cache from torii to libs, tests also are moved.

### Benefits
Separate abstact logic and torii service.

### Possible Drawbacks 
None

None

None


libs/cache/cache.hpp
```diff
@@ -19,7 +19,8 @@
 #define IROHA_CACHE_HPP
 
 #include <unordered_map>
-#include "torii/cache/abstract_cache.hpp"
+
+#include "abstract_cache.hpp"
```
luckychess: Please use full path here also (cache/abstract_cache)

PR #911, opened 2018-01-25
Update CONTRIBUTING.md

neewy:
### Description of the Change

We have changed the CVS schema.

### Benefits

Increased awareness of contributors.

### Possible Drawbacks 

None

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,13 +101,16 @@ Read our [C++ Style Guide](#c++-style-guide) and start with beginner-friendly is
 
 - Include issue numbers in Pull Request body only.
 
-- When finished work, **rebase onto base branch** with 
+- When finished work, **merge into base branch** by following simple steps:
+    1. Checkout the base branch
+    2. Merge feature branch and write meaninful merge commit (include PR number and title):
     ```bash
-    $ git fetch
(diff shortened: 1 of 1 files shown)


l4l:
Merge is for maintainers, not for contributors, why it should be here?

muratovv:
Who is the consumer of CONTRIBUTING.md file? I mean this information for maintainers or for contributors only. If second, there are no reasons to write about merging to base branch.

Warchant:
CONTRIBUTORS.md file is for all contributors, especially for the new ones.

IMO this scheme is bad:
1. reviewer will not know that developer failed with merge (wrongly resolved merge conflicts, for instance), therefore develop will be much more often in "failed" condition.
2. for developer there is no way to run CI and check all postconditions (builds on mac, builds on linux, tests are executed, etc). Locally you can execute only one step amount many. Working on mac, how can you be sure that your changes will work properly on ARM?

neewy:
I have closed this PR since the decision about VCS handling schema was not fixed. IMO we have to make it clear in this file, and since we tried merges only and I find (not only me though) comments of @Warchant valid, we need to use old approach with rebases + maybe squash & merge for trunks at least per a sprint. 

I will assign the discussion

PR #916, opened 2018-01-26
Chore: Update of CMake in Docker

laSinteZ:
### Description of the Change
Update of Cmake from 3.5.1 to 3.10.2 in Developer Docker
https://gitlab.kitware.com/cmake/cmake/tags/v3.10.2

This is required for new SWIG versions.

### Benefits
No need to update CMake manually, saves time

### Possible Drawbacks 
None


Warchant:
I will do this change in my branch feature/replace-flat-file-with-nudb, since there I also update CMake

PR #917, opened 2018-01-27
Docker-related quickfixes for develop

x3medima17:
Signed-off-by: dumitru <dimasavva17@gmail.com>

UPD: @warchant

### Description of the Change
- [x] fix issue with permissions in some tests: when run as user - pass, as root - fail. Now it is user-independent. **Note**: please, do not use chmod in tests. Use non-existent paths or remove files/dirs.
- [x] change adduser to useradd (non-interactive)
- [x] add advanced user mapping inside `run-iroha-dev.sh`. It allows to map current UID:GID inside docker-develop without necessity of user creation, which simplifies script.

### DIFF
--- a/docker/develop/Dockerfile
+++ b/docker/develop/Dockerfile
@@ -194,11 +194,18 @@ RUN git clone git://github.com/warchant/ed25519.git /tmp/ed25519; \
     ldconfig; \
     rm -rf /tmp/ed25519
 
-RUN adduser --disabled-password  iroha-ci --uid 6535
+# non-interactive adduser
+#   -m = create home dir
+#   -s = set default shell
+#   iroha-ci = username
+#   -u = userid, default for Ubuntu is 1000
+#   -U = create a group same as username
+#   no password
+RUN useradd -ms /bin/bash iroha-ci -u 1000 -U
+
 WORKDIR /opt/iroha
 RUN chmod -R 777 /opt/iroha; \
     mkdir -p /tmp/ccache -m 777;
 
 USER iroha-ci
-
 CMD ["/bin/bash"]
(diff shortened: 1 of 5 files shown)


test/module/libs/crypto/keys_manager_test.cpp
```diff
@@ -100,27 +101,23 @@ TEST_F(KeyManager, CreateAndLoad) {
...
-  permissions(pub_key_path, no_perms);
+  KeysManagerImpl manager = KeysManagerImpl(inexistent);
   ASSERT_FALSE(manager.createKeys(passphrase));
 }
 
 TEST_F(KeyManager, CreateInaccessiblePrikey) {
-  create_file(pub_key_path, "");
-  create_file(pri_key_path, "");
-  permissions(pri_key_path, no_perms);
+  KeysManagerImpl manager = KeysManagerImpl(inexistent);
```
l4l: Doesn't conform with the test name

docker/develop/Dockerfile
```diff
@@ -194,11 +194,18 @@ RUN git clone git://github.com/warchant/ed25519.git /tmp/ed25519; \
...
 
-RUN adduser --disabled-password  iroha-ci --uid 6535
+# non-interactive adduser
+#   -m = create home dir
+#   -s = set default shell
+#   iroha-ci = username
+#   -u = userid, default for Ubuntu is 1000
+#   -U = create a group same as username
+#   no password
+RUN useradd -ms /bin/bash iroha-ci -u 1000 -U
```
l4l: Prefer `useradd --no-create-home --shell /bin/bash iroha-ci --uid 1000 --user-group` instead and remove comments plz

PR #921, opened 2018-01-28
Conform old model perms with the shared_model

l4l:
### Description of the Change

As @motxx noticed here's the shared_model permission converting:

https://github.com/hyperledger/iroha/blob/4e58f7780b8c918e56328231ec8f5a9fd67bb429/shared_model/builders/protobuf/transaction.hpp#L211-L215

That's pr updates old model permission string names, so it may work with the shared_model as well.

### Possible Drawbacks 

?

### Alternate Designs

As an additional way of fixing that is converting permissions and leaving the old model as is. Imo that not really good because it requires update such strings in multiple places.

### DIFF
--- a/irohad/model/permissions.hpp
+++ b/irohad/model/permissions.hpp
@@ -24,55 +24,51 @@
 namespace iroha {
   namespace model {
 
-    const std::string can_append_role = "CanAppendRole";
-    const std::string can_create_role = "CanCreateRole";
-    const std::string can_detach_role = "CanDetachRole";
-    const std::string can_add_asset_qty = "CanAddAssetQuantity";
-    const std::string can_subtract_asset_qty = "CanSubtractAssetQuantity";
-    const std::string can_add_peer = "CanAddPeer";
-    const std::string can_add_signatory = "CanAddSignatory";
-    const std::string can_create_account = "CanCreateAccount";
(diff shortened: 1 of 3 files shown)


motxx:
In shared model, protobuf's standard json converter will be used. https://github.com/hyperledger/iroha/pull/906
So, I think we may need to specify `can_grant_add_signatory` (not `can_grant_can_...`) (from `RolePermission`) in new JSON converter of `CreateRole`. But now maybe we need to specify `can_grant_can_add_signatory` (previously `"CanGrantCanAddSignatory"`) in JSON converter.

Maybe does deleting `permissions.hpp` and old model JSON converter make the code clear?
And sorry, but I'm not sure that above my comment is really correct. It will help if someone gives me an opinion.

irohad/model/permissions.hpp
```diff
@@ -98,7 +94,7 @@ namespace iroha {
                                                      can_get_domain_signatories,
                                                      can_get_my_acc_detail};
 
-    const std::string can_grant = "CanGrant";
+    const std::string can_grant = "can_grant_";
```
muratovv: What is the purpose of end underscore?

l4l: we has can_grant<b>_</b>can_do_smth permissions, that's the purpose

PR #935, opened 2018-02-01
Fix issues found by cppclean in libs, iroha-cli, irohad/ametsuchi, irohad/consensus

Warchant:
### Description of the Change

cppclean has found many issues (https://hackmd.io/IYTmBNwRmAGBaATANgOwICzIMaPgDgDMBWBfbYDAI1gGZbgrFag=?view). 

I spent around 6-7 hours to fix some. Here is the result: https://hackmd.io/KYNgrAjGDMAmDGBaUxiICwCYAMtEA5tp9EBDdAMxwCMBOCbMeCoA?view

Here is the diff:
https://www.diffchecker.com/Nl5cxDew

You may notice that some new issues are added, but do not worry, many of them are false-positives and there are no other way to fix the issue.

It is the first PR in series of PRs, which fix all problems.

Here I fixed only:
- iroha-cli
- libs
- ametsuchi
- consensus

### Benefits

This PR does not change any logic, it introduces only small code refactorings.

Generally cppclean looks for unused includes, or suggests to place forward declarations. Both of these techniques reduce interdependency between objects, which is good -- it reduces recompilation of targets, if certain headers are changed.

Also, it found functions defined in cpp without keyword `static`. Without this keyword, functions have `external linkage` by default and are accessible by the rest of program just through forward declaration. I fixed also this.

### Possible Drawbacks 

There are many issues with logger. Because logger is an alias to `spdlog::logger`, I need to write many lines to write forward declaration, which is bad. 
If we really need to fix these issues, I suggest to remove `logger::Logger` and use `std::unique_ptr<stdlog::logger>` everywhere. For now logger introduces many false positives.


Warchant:
storage.hpp before:
image

storage.hpp after:
image

Warchant:
storage_result.hpp before:
image

storage_result.hpp after:
image

irohad/ametsuchi/impl/mutable_storage_impl.cpp
```diff
@@ -19,7 +19,7 @@
 #include "ametsuchi/impl/postgres_wsv_query.hpp"
 #include "ametsuchi/impl/redis_block_index.hpp"
 #include "model/commands/transfer_asset.hpp"
-
+#include "model/execution/command_executor_factory.hpp"  // for iroha::model::CommandExecutorFactory
```
Alexey-N-Chernyshov: Do we need this comment?

irohad/ametsuchi/impl/peer_query_wsv.cpp
```diff
@@ -15,9 +15,10 @@
  * limitations under the License.
  */
 
-#include <utility>
-
 #include "ametsuchi/impl/peer_query_wsv.hpp"
+#include <utility>
```
Alexey-N-Chernyshov: I'd prefer to order includes like Google (https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

irohad/ametsuchi/impl/postgres_wsv_command.hpp
```diff
@@ -18,11 +18,24 @@
 #ifndef IROHA_POSTGRES_WSV_COMMAND_HPP
 #define IROHA_POSTGRES_WSV_COMMAND_HPP
 
-#include "ametsuchi/wsv_command.hpp"
-#include "postgres_wsv_common.hpp"
+#include <set>
+#include <string>
+#include "ametsuchi/impl/postgres_wsv_common.hpp"
+#include "ametsuchi/wsv_command.hpp" // for WsvCommand
```
Alexey-N-Chernyshov: Do we need a comment?

irohad/main/impl/consensus_init.cpp
```diff
@@ -16,12 +16,12 @@
  */
 
 #include "main/impl/consensus_init.hpp"
-
```
Alexey-N-Chernyshov: I'd keep this line to separate related header like (https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

irohad/model/commands/all.hpp
```diff
@@ -0,0 +1,38 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
```
Alexey-N-Chernyshov: 2018

irohad/ametsuchi/impl/mutable_storage_impl.cpp
Warchant: Comments help me to not forget why exactly this header is included. It is just a coincidence that header name and symbol name are equal. In many cases they are not.

irohad/ametsuchi/impl/postgres_wsv_command.hpp
Warchant: I answered above.

irohad/consensus/yac/storage/yac_block_storage.hpp
```diff
@@ -62,12 +63,12 @@ namespace iroha {
         /**
          * @return votes attached to storage
          */
-        auto getVotes() -> decltype(votes_);
+        std::vector<VoteMessage> getVotes();
```
nickaleks: Why is next method declared const and this one is not?

irohad/consensus/yac/cluster_order.hpp
```diff
@@ -54,20 +55,17 @@ namespace iroha {
          */
         bool hasNext();
 
-        std::vector<model::Peer> getPeers() {
-          return order_;
-        };
+        std::vector<model::Peer> getPeers();
```
nickaleks: This and the next method should be const?

irohad/consensus/yac/storage/impl/yac_block_storage.cpp
```diff
@@ -56,11 +54,11 @@ namespace iroha {
         return getState();
       }
 
-      auto YacBlockStorage::getVotes() -> decltype(votes_) {
+      std::vector<VoteMessage> YacBlockStorage::getVotes() {
```
nickaleks: This should be const as well

irohad/model/commands/add_asset_quantity.hpp
```diff
@@ -33,18 +32,18 @@ namespace iroha {
       /**
        * Account where to add assets
        */
-      std::string account_id{};
+      std::string account_id;
```
nickaleks: What is the reason behind removal of initialization?

Warchant: I see only one advantage of using that initialization -- it saves you some code and time when you have to initialize many vars in many constructors.

However, one disadvantage is that it decreases overall coverage -- std::string may not be constructed and this line is marked as partial.

PR #947, opened 2018-02-06
Torii streaming

luckychess:
### Description of the Change

Streaming grpc call for transaction statuses.

### Benefits

We want to have a push-based notifications for tx statuses so here we are.

### Possible Drawbacks 

Each streaming call can theoretically block one of grpc thread up to a second. This timeout can be decreased in a least 10 times but it allows to start streaming call before transaction sending.

Two tests were added to torii_service_test.

Two main possibilities were checked: use combined Torii request which will return stream of responses or use separate calls for that (the second one was actually implemented). The first approach was faced with thread-related difficulties (kick me and I explain details).

### DIFF
--- a/irohad/main/application.cpp
+++ b/irohad/main/application.cpp
@@ -28,6 +28,8 @@ using namespace iroha::torii;
 using namespace iroha::model::converters;
 using namespace iroha::consensus::yac;
 
+using namespace std::chrono_literals;
+
 /**
  * Configuring iroha daemon
  */
@@ -234,7 +236,7 @@ void Irohad::initTransactionCommandService() {
       std::make_shared<TransactionProcessorImpl>(pcs, stateless_validator);
 
   command_service = std::make_unique<::torii::CommandService>(
-      pb_tx_factory, tx_processor, storage);
+      pb_tx_factory, tx_processor, storage, 1s, 10s);
 
   log_->info("[Init] => command service");
 }
--- a/irohad/main/application.cpp
+++ b/irohad/main/application.cpp
@@ -236,7 +236,7 @@ void Irohad::initTransactionCommandService() {
       std::make_shared<TransactionProcessorImpl>(pcs, stateless_validator);
 
   command_service = std::make_unique<::torii::CommandService>(
-      pb_tx_factory, tx_processor, storage, 1s, 10s);
+      pb_tx_factory, tx_processor, storage, proposal_delay_);
 
   log_->info("[Init] => command service");
 }
--- a/irohad/torii/command_client.hpp
+++ b/irohad/torii/command_client.hpp
@@ -29,13 +29,13 @@ namespace torii {
    */
   class CommandSyncClient {
    public:
-    CommandSyncClient(const std::string& ip, size_t port);
+    CommandSyncClient(const std::string &ip, size_t port);
 
-    CommandSyncClient(const CommandSyncClient&);
-    CommandSyncClient& operator=(CommandSyncClient);
+    CommandSyncClient(const CommandSyncClient &);
+    CommandSyncClient &operator=(CommandSyncClient);
 
-    CommandSyncClient(CommandSyncClient&&) noexcept ;
-    CommandSyncClient&operator=(CommandSyncClient&&) noexcept ;
+    CommandSyncClient(CommandSyncClient &&) noexcept;
+    CommandSyncClient &operator=(CommandSyncClient &&) noexcept;
 
     /**
      * requests tx to a torii server and returns response (blocking, sync)
@@ -52,13 +52,22 @@ namespace torii {
     grpc::Status Status(const iroha::protocol::TxStatusRequest &tx,
                         iroha::protocol::ToriiResponse &response) const;
 
+    /**
+     * Acquires stream of transaction statuses from the request
+     * moment until final.
+     * @param tx - transaction to send.
+     * @param response - vector of all statuses during tx pipeline.
+     */
+    void StatusStream(
+        const iroha::protocol::TxStatusRequest &tx,
+        std::vector<iroha::protocol::ToriiResponse> &response) const;
+
    private:
-    void swap(CommandSyncClient& lhs, CommandSyncClient& rhs);
+    void swap(CommandSyncClient &lhs, CommandSyncClient &rhs);
     std::string ip_;
     size_t port_;
     std::unique_ptr<iroha::protocol::CommandService::Stub> stub_;
   };
-
 
 }  // namespace torii
 
--- a/irohad/torii/command_service.hpp
+++ b/irohad/torii/command_service.hpp
@@ -17,14 +17,13 @@ limitations under the License.
 #ifndef TORII_COMMAND_SERVICE_HPP
 #define TORII_COMMAND_SERVICE_HPP
 
+#include <endpoint.grpc.pb.h>
+#include <endpoint.pb.h>
 #include <iostream>
 #include <string>
 #include <unordered_map>
-
 #include "ametsuchi/storage.hpp"
 #include "cache/cache.hpp"
-#include "endpoint.grpc.pb.h"
-#include "endpoint.pb.h"
 #include "model/converters/pb_transaction_factory.hpp"
 #include "model/transaction_response.hpp"
 #include "torii/processor/transaction_processor.hpp"
@@ -32,9 +31,7 @@ limitations under the License.
 namespace torii {
 
   /**
-   * Actual implementation of async CommandService.
-   * ToriiServiceHandler::(SomeMethod)Handler calls a corresponding method in
-   * this class.
+   * Actual implementation of sync CommandService.
    */
   class CommandService : public iroha::protocol::CommandService::Service {
    public:
@@ -46,25 +43,36 @@ namespace torii {
 
     CommandService(const CommandService &) = delete;
     CommandService &operator=(const CommandService &) = delete;
+
     /**
-     * actual implementation of async Torii in CommandService
+     * actual implementation of sync Torii in CommandService
      * @param request - Transaction
      * @param response - ToriiResponse
      */
     void Torii(iroha::protocol::Transaction const &request,
-                    google::protobuf::Empty &response);
-
-    void Status(iroha::protocol::TxStatusRequest const &request,
-                     iroha::protocol::ToriiResponse &response);
+               google::protobuf::Empty &response);
 
     virtual grpc::Status Torii(grpc::ServerContext *context,
                                const iroha::protocol::Transaction *request,
                                google::protobuf::Empty *response) override;
 
+    void Status(iroha::protocol::TxStatusRequest const &request,
+                iroha::protocol::ToriiResponse &response);
+
     virtual grpc::Status Status(
         grpc::ServerContext *context,
         const iroha::protocol::TxStatusRequest *request,
         iroha::protocol::ToriiResponse *response) override;
+
+    void StatusStream(
+        iroha::protocol::TxStatusRequest const &request,
+        grpc::ServerWriter<iroha::protocol::ToriiResponse> &response_writer);
+
+    virtual grpc::Status StatusStream(
+        grpc::ServerContext *context,
+        const iroha::protocol::TxStatusRequest *request,
+        grpc::ServerWriter<iroha::protocol::ToriiResponse> *response_writer)
+        override;
 
    private:
     std::shared_ptr<iroha::model::converters::PbTransactionFactory> pb_factory_;
@@ -73,6 +81,8 @@ namespace torii {
     std::shared_ptr<
         iroha::cache::Cache<std::string, iroha::protocol::ToriiResponse>>
         cache_;
+    std::mutex wait_subscription_;
+    std::condition_variable cv_;
   };
 
 }  // namespace torii
--- a/irohad/torii/command_service.hpp
+++ b/irohad/torii/command_service.hpp
@@ -37,14 +37,21 @@ namespace torii {
     /**
      * Creates a new instance of CommandService
      * @param pb_factory - model->protobuf and vice versa converter
-     * @param txProcessor - processor of received transactions
+     * @param tx_processor - processor of received transactions
      * @param storage - storage to request transactions outside the cache
+     * @param start_tx_processing_duration - maximum time between getting a
+     * StatusStream request and receiving at least one status about processing
+     * of corresponding transaction
+     * @param finish_tx_processing_duration - maximum transaction processing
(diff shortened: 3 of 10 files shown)


irohad/torii/command_service.hpp
```diff
@@ -17,24 +17,21 @@ limitations under the License.
 #ifndef TORII_COMMAND_SERVICE_HPP
 #define TORII_COMMAND_SERVICE_HPP
 
+#include <endpoint.grpc.pb.h>
```
muratovv: Use `"` instead.

irohad/torii/command_service.hpp
```diff
@@ -46,33 +43,46 @@ namespace torii {
...
+        override;
+
    private:
     std::shared_ptr<iroha::model::converters::PbTransactionFactory> pb_factory_;
     std::shared_ptr<iroha::torii::TransactionProcessor> tx_processor_;
     std::shared_ptr<iroha::ametsuchi::Storage> storage_;
     std::shared_ptr<
         iroha::cache::Cache<std::string, iroha::protocol::ToriiResponse>>
         cache_;
+    std::mutex wait_subscription_;
```
muratovv: Looks like a redundant variable. Maybe better to rework with a local variable instead?

irohad/torii/command_service.hpp
```diff
@@ -46,33 +43,46 @@ namespace torii {
...
+    std::condition_variable cv_;
```
muratovv: Add documentation with reasons to use the conditional variable and write that data that guards under the cv.

irohad/torii/command_service.hpp
```diff
@@ -46,33 +43,46 @@ namespace torii {
...
-
-    void Status(iroha::protocol::TxStatusRequest const &request,
-                     iroha::protocol::ToriiResponse &response);
+               google::protobuf::Empty &response);
 
     virtual grpc::Status Torii(grpc::ServerContext *context,
                                const iroha::protocol::Transaction *request,
                                google::protobuf::Empty *response) override;
 
+    void Status(iroha::protocol::TxStatusRequest const &request,
```
muratovv: Add documentation for the methods and methods below.

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
     Status(*request, *response);
     return grpc::Status::OK;
   }
+
+  void CommandService::StatusStream(
```
muratovv: General issue: this function looks very large. Better to rework it with subfunctions.

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
...
+      iroha::protocol::TxStatusRequest const &request,
+      grpc::ServerWriter<iroha::protocol::ToriiResponse> &response_writer) {
+    auto resp = cache_->findItem(request.tx_hash());
+    if (resp) {
+      if (resp->tx_status()
```
muratovv: This block possible to move into own function.

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
...
+              == iroha::protocol::TxStatus::STATELESS_VALIDATION_FAILED
+          or resp->tx_status()
```
muratovv: Also, I guess more clear to hold all values into collection and perform `OR` function (accumulate for example) for the collection

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
...
+    auto subscription = rxcpp::composite_subscription();
+    auto request_hash = request.tx_hash();
+
+    tx_processor_->transactionNotifier()
+        .filter([&request_hash](auto response) {
+          return response->tx_hash == request_hash;
+        })
+        .subscribe(
+            subscription,
+            [&](std::shared_ptr<iroha::model::TransactionResponse>
```
muratovv: better to use `auto` instead.

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
...
+                    iroha_response) {
+              iroha::protocol::ToriiResponse resp_sub;
+              resp_sub.set_tx_hash(request_hash);
+
+              switch (iroha_response->current_status) {
```
muratovv: Maybe move the switch statement into a separate function?

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
...
+                case iroha::model::TransactionResponse::
+                    STATELESS_VALIDATION_FAILED:
+                  resp_sub.set_tx_status(
+                      iroha::protocol::TxStatus::STATELESS_VALIDATION_FAILED);
+                  response_writer.WriteLast(resp_sub, grpc::WriteOptions());
+                  subscription.unsubscribe();
+                  finished = true;
```
muratovv: The code inside case statements looks duplicated. Maybe better to make a template functions for this stuff?

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
...
+                  cv_.notify_one();
+                  break;
+              }
+            });
+
+    std::unique_lock<std::mutex> lock(wait_subscription_);
+    // One second should be enough to at least start tx processing.
+    // Otherwise we think there is no such tx at all.
+    cv_.wait_for(lock, 1s);
```
muratovv: Better to set time parameter as a class dependency.

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
...
+    if (not finished) {
+      if (not resp) {
+        subscription.unsubscribe();
+        iroha::protocol::ToriiResponse resp_none;
+        resp_none.set_tx_hash(request_hash);
+        resp_none.set_tx_status(iroha::protocol::TxStatus::NOT_RECEIVED);
+        response_writer.WriteLast(resp_none, grpc::WriteOptions());
+      } else {
+        cv_.wait_for(lock, 10s);
```
muratovv: Same about class dependency. But in this case, we may say, that we require the time of one proposal propagation as a parameter.

test/module/irohad/torii/query_service_test.cpp
```diff
@@ -88,6 +88,7 @@ TEST_F(QueryServiceTest, InvalidWhenDuplicateHash) {
 
   init();
 
+  //TODO luckychess 03.02.2018 is this OK?
```
muratovv: This comment looks not Ok. Better to explain the concerns that you found and make a task that solves it.

test/module/irohad/torii/torii_service_test.cpp
```diff
@@ -344,3 +344,79 @@ TEST_F(ToriiServiceTest, CheckHash) {
...
+  auto new_tx = iroha::protocol::Transaction();
+  auto payload = new_tx.mutable_payload();
+  payload->set_tx_counter(1);
+  payload->set_creator_account_id("accountA");
+
+  auto iroha_tx = tx_factory.deserialize(new_tx);
+  std::string txhash = iroha::hash(*iroha_tx).to_string();
+
+  std::vector<iroha::protocol::ToriiResponse> torii_response;
+  std::thread t([&]() {
```
muratovv: Add reasons to use another thread.

irohad/torii/impl/command_service.cpp
luckychess: It gives a compilation error unfortumately.

irohad/torii/command_client.hpp
```diff
@@ -52,14 +52,23 @@ namespace torii {
     grpc::Status Status(const iroha::protocol::TxStatusRequest &tx,
                         iroha::protocol::ToriiResponse &response) const;
 
+    /**
+     * Acquires stream of transaction statuses from the request
+     * moment until final.
+     * @param tx - transaction to send.
+     * @param response - vector of all statuses during tx pipeline.
+     */
+    void StatusStream(
```
x3medima17: Why don't we make `StatusStream` functional way?

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,126 @@ namespace torii {
...
+        break;
+      case iroha::model::TransactionResponse::NOT_RECEIVED:
+      default:
+        proto_status = iroha::protocol::TxStatus::NOT_RECEIVED;
+        break;
+    }
+    return proto_status;
+  }
+
+  bool CommandService::isFinalStatus(iroha::protocol::TxStatus &status) {
```
x3medima17: Maybe there is more elegant way to do it?

irohad/torii/impl/command_service.cpp
```diff
@@ -130,4 +136,117 @@ namespace torii {
...
+                  response_writer.WriteLast(resp_sub, grpc::WriteOptions());
+                  subscription.unsubscribe();
+                  finished = true;
+                  cv_.notify_one();
+                  break;
+              }
+            });
+
+    std::unique_lock<std::mutex> lock(wait_subscription_);
+    // One second should be enough to at least start tx processing.
```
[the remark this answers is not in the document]
luckychess: Oooops!

irohad/torii/command_service.hpp
```diff
@@ -17,72 +17,138 @@ limitations under the License.
...
    * Actual implementation of sync CommandService.
    */
   class CommandService : public iroha::protocol::CommandService::Service {
    public:
+    /**
+     * Creates a new instance of CommandService
+     * @param pb_factory - model->protobuf and vice versa converter
+     * @param tx_processor - processor of received transactions
+     * @param storage - storage to request transactions outside the cache
+     * @param start_tx_processing_duration - maximum time between getting a
```
muratovv: I guess this time could be also deduced from the proposal emitting time.

irohad/torii/command_service.hpp
```diff
@@ -17,72 +17,138 @@ limitations under the License.
...
+     * StatusStream request and receiving at least one status about processing
+     * of corresponding transaction
+     * @param finish_tx_processing_duration - maximum transaction processing
+     * time. Generally is should be a time of one proposal propagation.
+     */
```
muratovv: I guess, better to create one dependency with proposal propagation time.

irohad/main/application.cpp
```diff
@@ -234,7 +236,7 @@ void Irohad::initTransactionCommandService() {
       std::make_shared<TransactionProcessorImpl>(pcs, stateless_validator);
 
   command_service = std::make_unique<::torii::CommandService>(
-      pb_tx_factory, tx_processor, storage);
+      pb_tx_factory, tx_processor, storage, 1s, 10s);
```
muratovv: Use proposal propagation time as a dependency in main.

test/module/irohad/torii/query_service_test.cpp
```diff
@@ -88,6 +88,7 @@ TEST_F(QueryServiceTest, InvalidWhenDuplicateHash) {
...
   query_service->Find(query, response);
```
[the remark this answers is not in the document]
luckychess: Well, not me, I've just found this place during all these infinity merges. These calls are OK, you can read comments a few lines above them.

irohad/torii/impl/command_service.cpp
```diff
@@ -141,93 +141,46 @@ namespace torii {
...
-      }
-      response_writer.Write(*resp);
-    }
+    checkCacheAndSend(resp, response_writer);
 
     bool finished = false;
     auto subscription = rxcpp::composite_subscription();
     auto request_hash = request.tx_hash();
 
+    // condition variable to ensure that current method will not return before
```
muratovv: For documentation purposes use `///` instead.

irohad/torii/impl/command_service.cpp
luckychess: Well, @muratovv advised something similar in one of his comments. I think introduction a new collection of final statuses will just make code larger. If we decide to introduce one more final status both approaches will require a change anyway.

irohad/torii/impl/command_service.cpp
```diff
@@ -249,4 +203,59 @@ namespace torii {
     return grpc::Status::OK;
   }
 
+  void CommandService::checkCacheAndSend(
+      boost::optional<iroha::protocol::ToriiResponse> &resp,
+      grpc::ServerWriter<iroha::protocol::ToriiResponse> &response_writer)
+      const {
+    if (resp) {
+      if (resp->tx_status()
```
muratovv: So, can we combine `or` function for statuses in a functional way?

irohad/torii/impl/command_service.cpp
```diff
@@ -249,4 +203,59 @@ namespace torii {
...
+        return;
+      }
+      response_writer.Write(*resp);
+    }
+  }
+
+  iroha::protocol::TxStatus CommandService::convertStatusToProto(
+      iroha::model::TransactionResponse::Status &status) {
+    iroha::protocol::TxStatus proto_status;
+    switch (status) {
```
muratovv: better to use map instead.

irohad/torii/impl/command_service.cpp
```diff
@@ -249,4 +203,59 @@ namespace torii {
...
+      case iroha::model::TransactionResponse::NOT_RECEIVED:
+      default:
+        proto_status = iroha::protocol::TxStatus::NOT_RECEIVED;
+        break;
+    }
+    return proto_status;
+  }
+
+  bool CommandService::isFinalStatus(iroha::protocol::TxStatus &status) {
+    return status == iroha::protocol::TxStatus::STATELESS_VALIDATION_FAILED
```
muratovv: Same as before with a functional style.

irohad/torii/command_service.hpp
luckychess: One of StatusStream use cases is to do StatusStream call before Torii call to collect as much statuses as possible. Proposal emitting time will be not enough in case of more or less significant network delay.

irohad/torii/impl/command_service.cpp
luckychess: It was really stupid from my side to introduce isFinalStatus() function and do not use it here.

irohad/torii/impl/command_service.cpp
luckychess: I do not think that std::transform will be good here: we'll just move this switch inside lambda and also add several lines of boilerplate code. Or may be you mean something else?

PR #948, opened 2018-02-07
Fix/nptr to result

Solonets:
### Description of the Change

A bad thing about smart pointers is they may store nullptr. In project we may guarantee that smart pointer has valid value inside. For purposes where there is no guarantee we should change type to Result<smart_pointer, error>. And this type explicitly say that there is possible null value.

This Pr fixes the problem above in /iroha/irohad/ametsuchi/impl/storage_impl.cpp

### Benefits

null pointers are explicitly processed

### Possible Drawbacks 

in several places, auto type deduction can't be used as
`auto p = ...` is substituted with `type p; match([]() { p = ... }) `


irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -69,11 +77,11 @@ namespace iroha {
           "SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY;");
     }
 
-    std::unique_ptr<TemporaryWsv> StorageImpl::createTemporaryWsv() {
+    expected::Result<std::unique_ptr<TemporaryWsv>, std::string> StorageImpl::createTemporaryWsv() {
```
Warchant: `using iroha::expected::Result` but you use `expected::Result`. IMO it can be either Result, or you can remove using.

UPD: also I've noticed the same across all the file.

If you want to use `expected::Result` and others, you can do it without `using`.

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -69,11 +77,11 @@ namespace iroha {
...
       auto command_executors = model::CommandExecutorFactory::create();
       if (not command_executors.has_value()) {
         log_->error(kCommandExecutorError);
-        return nullptr;
+        return makeError(kCommandExecutorError);
```
Warchant: can makeError log that error? why do we need to log it every time? @lebdron

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -37,7 +37,13 @@ namespace iroha {
 
     void SynchronizerImpl::process_commit(iroha::model::Block commit_message) {
       log_->info("processing commit");
-      auto storage = mutableFactory_->createMutableStorage();
+      auto storageResult = mutableFactory_->createMutableStorage();
+      std::unique_ptr<ametsuchi::MutableStorage> storage;
+      storageResult.match(
+          [&](expected::Value<std::unique_ptr<ametsuchi::MutableStorage>> &_storage) {
+            storage = std::move(_storage.value);
+          }, [](expected::Error<std::string> &error) {}
```
Warchant: I don't think it is good idea to leave error branch empty. At least throw something.

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -54,7 +60,13 @@ namespace iroha {
         // Block can't be applied to current storage
         // Download all missing blocks
         for (auto signature : commit_message.sigs) {
-          storage = mutableFactory_->createMutableStorage();
+          auto storageResult = mutableFactory_->createMutableStorage();
+          std::unique_ptr<ametsuchi::MutableStorage> storage;
+          storageResult.match(
+              [&](expected::Value<std::unique_ptr<ametsuchi::MutableStorage>> &_storage) {
+                storage = std::move(_storage.value);
+              }, [](expected::Error<std::string> &error) {}
```
Warchant: Again, empty "error" branch

test/module/irohad/ametsuchi/ametsuchi_mocks.hpp
```diff
@@ -144,7 +147,7 @@ namespace iroha {
 
     class MockTemporaryFactory : public TemporaryFactory {
      public:
-      MOCK_METHOD0(createTemporaryWsv, std::unique_ptr<TemporaryWsv>());
+      MOCK_METHOD0(createTemporaryWsv, iroha::expected::Result<std::unique_ptr<TemporaryWsv>, std::string>(void));
```
Warchant: mock is  already in `iroha` namespace, you can safely use `expected::Result`.

irohad/synchronizer/impl/synchronizer_impl.cpp
Solonets: error is processed later: 
 ```
storageResult.match(
          [&](expected::Value<std::unique_ptr<ametsuchi::MutableStorage>> &_storage) {
            storage = std::move(_storage.value);
          }, [](expected::Error<std::string> &error) {}
      );
if (not storage) {
        log_->error("Cannot create mutable storage");
        return;
      }
```

I am not sure, how it should be rewritten

File: test/module/irohad/ametsuchi/kv_storage_test.cpp [hunk 6]
```diff
@@ -42,7 +42,12 @@ class KVTest : public AmetsuchiTest {
  protected:
   void SetUp() override {
     AmetsuchiTest::SetUp();
-    storage = StorageImpl::create(block_store_path, pgopt_);
+    auto storageResult = StorageImpl::create(block_store_path, pgopt_);
+    storageResult.match(
+        [&](iroha::expected::Value<std::shared_ptr<StorageImpl>> &_storage) {
+          storage = _storage.value;
+        },
+        [](iroha::expected::Error<std::string> &error) { ASSERT_TRUE(0); });
```
kamilsa: Change to `FAIL() << error;`

test/module/irohad/ametsuchi/kv_storage_test.cpp
```diff
@@ -92,7 +97,12 @@ class KVTest : public AmetsuchiTest {
     block1.txs_number = block1.transactions.size();
 
     {
-      auto ms = storage->createMutableStorage();
+      std::unique_ptr<MutableStorage> ms;
+      auto storageResult = storage->createMutableStorage();
+      storageResult.match(
+          [&](iroha::expected::Value<std::unique_ptr<MutableStorage>>
+                  &_storage) { ms = std::move(_storage.value); },
+          [](iroha::expected::Error<std::string> &error) { ASSERT_TRUE(0); });
```
kamilsa: same as above

irohad/simulator/impl/simulator.cpp
```diff
@@ -59,9 +59,17 @@ namespace iroha {
...
+      auto temporaryStorageResult = ametsuchi_factory_->createTemporaryWsv();
+      temporaryStorageResult.match(
+          [&](expected::Value<std::unique_ptr<ametsuchi::TemporaryWsv>>
+                  &temporaryStorage) {
+            notifier_.get_subscriber().on_next(
+                validator_->validate(proposal, *(temporaryStorage.value)));
+          },
+          [&](expected::Error<std::string> &error) {
+            log_->error(error.error);
+            throw std::runtime_error(error.error);
```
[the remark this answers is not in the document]
Solonets: @kamilsa previously it was the same behavior, but caused by null pointer reference, I think coming out with a better solution is another task

PR #950, opened 2018-02-07
Rework OrderingService with new model

l4l:
### Description of the Change

Change OrderingServices' interfaces to the new shared model

### Issues

There's some issues that've been solved as simple as possible, though it's probably not really a good solution

#### iface -> proto

There's need to convert `proto::Proposal` to `protocol::Proposal`, it's pretty easy to set all the fields via builder but transactions. SM's proposal returns an interface, so there's no simple way to convert to protobuf, so I just use `makeOldModel` + `from_old`:

https://github.com/hyperledger/iroha/blob/5b711d1d7348f853ab8cda74e4b5b87207a24198/irohad/ordering/impl/ordering_service_transport_grpc.cpp#L61-L71

#### Pretty hard to create a simple tx

I guess the example is self-explainable

https://github.com/hyperledger/iroha/blob/5b711d1d7348f853ab8cda74e4b5b87207a24198/test/module/irohad/ordering/ordering_service_test.cpp#L82-L89

#### tbb queue wrapper

The tbb's concurrent queue can't work with interfaces (and even with `unique_ptr`), so I just used `shared_ptr` instead. Imo that can be a significant performance issue

https://github.com/hyperledger/iroha/blob/5b711d1d7348f853ab8cda74e4b5b87207a24198/irohad/ordering/impl/ordering_service_impl.hpp#L94

### DIFF
--- a/irohad/network/ordering_service.hpp
+++ b/irohad/network/ordering_service.hpp
@@ -34,7 +34,8 @@ namespace iroha {
        * Transform model proposal to transport object and send to peers
        * @param proposal - object for propagation
        */
-      virtual void publishProposal(model::Proposal &&proposal) = 0;
+      virtual void publishProposal(
+          shared_model::proto::Proposal &&proposal) = 0;
     };
   }  // namespace network
 }  // namespace iroha
--- a/irohad/network/ordering_service.hpp
+++ b/irohad/network/ordering_service.hpp
@@ -19,7 +19,6 @@
 #define IROHA_ORDERINGSERVICE_H
 
 #include "network/ordering_service_transport.hpp"
-#include "utils/polymorphic_wrapper.hpp"
 
 namespace iroha {
   namespace network {
--- a/irohad/network/ordering_service.hpp
+++ b/irohad/network/ordering_service.hpp
@@ -19,6 +19,7 @@
 #define IROHA_ORDERINGSERVICE_H
 
 #include "network/ordering_service_transport.hpp"
+#include "utils/polymorphic_wrapper.hpp"
 
 namespace iroha {
   namespace network {
@@ -35,7 +36,8 @@ namespace iroha {
        * @param proposal - object for propagation
        */
       virtual void publishProposal(
-          shared_model::proto::Proposal &&proposal) = 0;
+          shared_model::detail::PolymorphicWrapper<
+              shared_model::interface::Proposal> proposal) = 0;
     };
   }  // namespace network
 }  // namespace iroha
--- a/irohad/network/ordering_service.hpp
+++ b/irohad/network/ordering_service.hpp
@@ -36,8 +36,7 @@ namespace iroha {
        * @param proposal - object for propagation
        */
       virtual void publishProposal(
-          shared_model::detail::PolymorphicWrapper<
-              shared_model::interface::Proposal> proposal) = 0;
+          std::unique_ptr<shared_model::interface::Proposal> proposal) = 0;
     };
   }  // namespace network
 }  // namespace iroha
--- a/irohad/network/ordering_service_transport.hpp
+++ b/irohad/network/ordering_service_transport.hpp
@@ -17,9 +17,9 @@
 #ifndef IROHA_ORDERING_SERVICE_TRANSPORT_H
 #define IROHA_ORDERING_SERVICE_TRANSPORT_H
 
+#include <memory>
 #include "interfaces/iroha_internal/proposal.hpp"
 #include "interfaces/transaction.hpp"
-#include "utils/polymorphic_wrapper.hpp"
 
 namespace iroha {
   namespace network {
--- a/irohad/network/ordering_service_transport.hpp
+++ b/irohad/network/ordering_service_transport.hpp
@@ -61,8 +61,7 @@ namespace iroha {
        * @param proposal : proposal to be published
        */
       virtual void publishProposal(
-          shared_model::detail::PolymorphicWrapper<
-              shared_model::interface::Proposal> proposal,
+          std::unique_ptr<shared_model::interface::Proposal> proposal,
           const std::vector<std::string> &peers) = 0;
 
       virtual ~OrderingServiceTransport() = default;
--- a/irohad/network/ordering_service_transport.hpp
+++ b/irohad/network/ordering_service_transport.hpp
@@ -35,8 +35,8 @@ namespace iroha {
        * @param transaction - transaction object itself
        */
       virtual void onTransaction(
-          const shared_model::detail::PolymorphicWrapper<
-              shared_model::interface::Transaction> transaction) = 0;
+          std::shared_ptr<shared_model::interface::Transaction>
+              transaction) = 0;
 
       virtual ~OrderingServiceNotification() = default;
     };
--- a/irohad/network/ordering_service_transport.hpp
+++ b/irohad/network/ordering_service_transport.hpp
@@ -17,9 +17,9 @@
 #ifndef IROHA_ORDERING_SERVICE_TRANSPORT_H
 #define IROHA_ORDERING_SERVICE_TRANSPORT_H
 
-#include "backend/protobuf/proposal.hpp"
-#include "backend/protobuf/transaction.hpp"
+#include "interfaces/iroha_internal/proposal.hpp"
 #include "interfaces/transaction.hpp"
+#include "utils/polymorphic_wrapper.hpp"
 
 namespace iroha {
   namespace network {
@@ -35,7 +35,8 @@ namespace iroha {
        * @param transaction - transaction object itself
        */
       virtual void onTransaction(
-          const shared_model::proto::Transaction &transaction) = 0;
+          const shared_model::detail::PolymorphicWrapper<
+              shared_model::interface::Transaction> transaction) = 0;
 
       virtual ~OrderingServiceNotification() = default;
     };
@@ -59,8 +60,10 @@ namespace iroha {
        * Publishes proposal over network
        * @param proposal : proposal to be published
        */
-      virtual void publishProposal(shared_model::proto::Proposal &&proposal,
-                                   const std::vector<std::string> &peers) = 0;
+      virtual void publishProposal(
+          shared_model::detail::PolymorphicWrapper<
+              shared_model::interface::Proposal> proposal,
+          const std::vector<std::string> &peers) = 0;
 
       virtual ~OrderingServiceTransport() = default;
     };
--- a/irohad/network/ordering_service_transport.hpp
+++ b/irohad/network/ordering_service_transport.hpp
@@ -17,8 +17,9 @@
 #ifndef IROHA_ORDERING_SERVICE_TRANSPORT_H
 #define IROHA_ORDERING_SERVICE_TRANSPORT_H
 
-#include "model/proposal.hpp"
-#include "model/transaction.hpp"
+#include "backend/protobuf/proposal.hpp"
+#include "backend/protobuf/transaction.hpp"
+#include "interfaces/transaction.hpp"
 
 namespace iroha {
   namespace network {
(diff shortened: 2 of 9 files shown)


irohad/ordering/impl/ordering_service_transport_grpc.cpp
```diff
@@ -50,18 +58,23 @@ void OrderingServiceTransportGrpc::publishProposal(
...
-        protocol::Transaction(factory_.serialize(tx));
-  }
+  auto pb_proposal =
+      shared_model::proto::ProposalBuilder()
+          .height(proposal.height())
+          .createdTime(iroha::time::now())
+          .transactions(proposal.transactions()
+                        | boost::adaptors::transformed([](auto iface_tx) {
+                            std::unique_ptr<iroha::model::Transaction> old_tx(
+                                iface_tx->makeOldModel());
```
muratovv: I think there is forbidden to use `makeOldModel` because u create the new-style proposal.

l4l: I've described issues in pr message. There's an issue for converting iface->proto

irohad/network/ordering_service_transport.hpp
```diff
@@ -17,8 +17,9 @@
 #ifndef IROHA_ORDERING_SERVICE_TRANSPORT_H
 #define IROHA_ORDERING_SERVICE_TRANSPORT_H
 
-#include "model/proposal.hpp"
-#include "model/transaction.hpp"
+#include "backend/protobuf/proposal.hpp"
+#include "backend/protobuf/transaction.hpp"
+#include "interfaces/transaction.hpp"
```
x3medima17: No need to include the whole file, you can use forward declaration

irohad/ordering/impl/ordering_service_transport_grpc.cpp
```diff
@@ -33,14 +39,16 @@ grpc::Status OrderingServiceTransportGrpc::onTransaction(
...
+    subscriber_.lock()->onTransaction(
+        shared_model::proto::Transaction(*request));
   }
 
   return ::grpc::Status::OK;
 }
 
 void OrderingServiceTransportGrpc::publishProposal(
-    model::Proposal &&proposal, const std::vector<std::string> &peers) {
+    shared_model::proto::Proposal &&proposal,
```
x3medima17: You can use shared_model namespace at the top of the file

irohad/ordering/impl/ordering_service_transport_grpc.cpp
```diff
@@ -16,6 +16,12 @@
...
 
+#include <boost/range/adaptor/transformed.hpp>
+
+#include "backend/protobuf/from_old_model.hpp"
+#include "builders/protobuf/proposal.hpp"
+#include "datetime/time.hpp"
+
 using namespace iroha::ordering;
 using namespace iroha::protocol;
 using namespace iroha::model;
```
x3medima17: Do we need `iroha::model` ?

irohad/network/ordering_service_transport.hpp
l4l: There's no particular descision so far, [google prefer](https://google.github.io/styleguide/cppguide.html#Forward_Declarations) its avoiding

irohad/ordering/impl/ordering_service_transport_grpc.cpp
l4l: Doesn't change much

irohad/ordering/impl/ordering_service_transport_grpc.cpp
l4l: Nope, thanks

irohad/network/ordering_service.hpp
```diff
@@ -34,7 +34,8 @@ namespace iroha {
        * Transform model proposal to transport object and send to peers
        * @param proposal - object for propagation
        */
-      virtual void publishProposal(model::Proposal &&proposal) = 0;
+      virtual void publishProposal(
```
lebdron: It does not seem necessary to use concrete transport in interface. PolymorphicWrapper for interface should be fine here.

irohad/network/ordering_service_transport.hpp
```diff
@@ -33,7 +34,8 @@ namespace iroha {
        * Callback on receiving transaction
        * @param transaction - transaction object itself
        */
-      virtual void onTransaction(const model::Transaction &transaction) = 0;
+      virtual void onTransaction(
```
lebdron: Same as above, PolymorphicWrapper should be used in abstract class.

irohad/network/ordering_service_transport.hpp
```diff
@@ -57,7 +59,7 @@ namespace iroha {
        * Publishes proposal over network
        * @param proposal : proposal to be published
        */
-      virtual void publishProposal(model::Proposal &&proposal,
+      virtual void publishProposal(shared_model::proto::Proposal &&proposal,
```
lebdron: Same as above with PolymorphicWrapper for interface.

irohad/ordering/impl/ordering_service_impl.cpp
```diff
@@ -44,19 +46,21 @@ namespace iroha {
     }
 
     void OrderingServiceImpl::generateProposal() {
-      auto txs = decltype(std::declval<model::Proposal>().transactions)();
-      for (model::Transaction tx;
-           txs.size() < max_size_ and queue_.try_pop(tx);) {
-        txs.push_back(std::move(tx));
+      std::vector<shared_model::proto::Transaction> poped_tx;
```
lebdron: Not sure if popped is a great choice here, maybe dequeued?

irohad/ordering/impl/ordering_service_transport_grpc.cpp
```diff
@@ -16,31 +16,36 @@
  */
 #include "ordering/impl/ordering_service_transport_grpc.hpp"
 
+#include <boost/range/adaptor/transformed.hpp>
```
lebdron: These headers does not seem to be used here.

test/module/irohad/ordering/ordering_service_test.cpp
```diff
@@ -74,6 +79,15 @@ class OrderingServiceTest : public ::testing::Test {
     fake_transport = std::make_shared<MockOrderingServiceTransport>();
   }
 
+  auto empty_tx() {
+    constexpr auto kTotal = 15;
+    return shared_model::proto::TemplateTransactionBuilder<
```
lebdron: There was a TestTransactionBuilder for that.

irohad/network/ordering_service_transport.hpp
```diff
@@ -17,8 +17,9 @@
 #ifndef IROHA_ORDERING_SERVICE_TRANSPORT_H
 #define IROHA_ORDERING_SERVICE_TRANSPORT_H
 
-#include "model/proposal.hpp"
-#include "model/transaction.hpp"
+#include "backend/protobuf/proposal.hpp"
```
lebdron: Backend should not be required in interfaces.

test/module/irohad/ordering/ordering_service_test.cpp
```diff
@@ -95,7 +109,8 @@ TEST_F(OrderingServiceTest, SimpleTest) {
 
   EXPECT_CALL(*fake_transport, publishProposal(_, _)).Times(1);
 
-  fake_transport->publishProposal(model::Proposal({}), {});
+  fake_transport->publishProposal(
+      shared_model::proto::TemplateProposalBuilder<7>().build(), {});
```
lebdron: Can this be replaced with TestTransactionBuilder?

l4l: Validation can be applied, don't think so

irohad/ordering/impl/ordering_service_impl.cpp
l4l: There's no deque in tbb, containers from standard lib don't support concurrency

irohad/network/ordering_service.hpp
```diff
@@ -34,7 +35,9 @@ namespace iroha {
...
+          shared_model::detail::PolymorphicWrapper<
```
lebdron: It was a mistake to suggest using PolymorphicWrapper, since it does deep copy of object when a copy constructor is called. It is enough to use `shared_ptr` in component interfaces.

irohad/network/ordering_service_transport.hpp
```diff
@@ -33,7 +34,9 @@ namespace iroha {
        * Callback on receiving transaction
        * @param transaction - transaction object itself
        */
-      virtual void onTransaction(const model::Transaction &transaction) = 0;
+      virtual void onTransaction(
+          const shared_model::detail::PolymorphicWrapper<
```
lebdron: PW -> shared_ptr

irohad/network/ordering_service_transport.hpp
```diff
@@ -57,8 +60,10 @@ namespace iroha {
        * Publishes proposal over network
        * @param proposal : proposal to be published
        */
-      virtual void publishProposal(model::Proposal &&proposal,
-                                   const std::vector<std::string> &peers) = 0;
+      virtual void publishProposal(
+          shared_model::detail::PolymorphicWrapper<
```
lebdron: PW -> shared_ptr

irohad/ordering/impl/ordering_service_transport_grpc.cpp
```diff
@@ -16,31 +16,36 @@
  */
 #include "ordering/impl/ordering_service_transport_grpc.hpp"
 
+#include "backend/protobuf/from_old_model.hpp"
+#include "builders/protobuf/proposal.hpp"
+#include "datetime/time.hpp"
```
lebdron: `now()` is not used in this file.

irohad/ordering/impl/ordering_service_transport_grpc.cpp
```diff
@@ -16,31 +16,36 @@
...
     ::grpc::ServerContext *context,
-    const protocol::Transaction *request,
+    const iroha::protocol::Transaction *request,
     ::google::protobuf::Empty *response) {
   if (subscriber_.expired()) {
     log_->error("No subscriber");
   } else {
-    subscriber_.lock()->onTransaction(*factory_.deserialize(*request));
+    subscriber_.lock()->onTransaction(
+        shared_model::detail::makePolymorphic<shared_model::proto::Transaction>(
```
lebdron: It is necessary to pass rvalue to shared_model proto constructor here, because the object is owned by grpc. So a copy of request should be made.

irohad/ordering/impl/ordering_service_impl.cpp
```diff
@@ -34,8 +35,11 @@ namespace iroha {
     }
 
     void OrderingServiceImpl::onTransaction(
-        const model::Transaction &transaction) {
-      queue_.push(transaction);
+        const shared_model::detail::PolymorphicWrapper<
+            shared_model::interface::Transaction> transaction) {
+      auto proto = static_cast<const shared_model::proto::Transaction *>(
```
lebdron: Is it necessary to cast it to proto here? A collection of pointers to interface should be possible to use.

test/module/irohad/ordering/ordering_service_test.cpp
```diff
@@ -28,36 +28,39 @@
 #include "ordering/impl/ordering_service_impl.hpp"
 #include "ordering/impl/ordering_service_transport_grpc.hpp"
 
+#include "builders/protobuf/proposal.hpp"
+#include "builders/protobuf/transaction.hpp"
+#include "module/shared_model/builders/protobuf/test_proposal_builder.hpp"
+#include "module/shared_model/builders/protobuf/test_transaction_builder.hpp"
+#include "module/shared_model/validators//validators.hpp"
+#include "validators/default_validator.hpp"
```
lebdron: Is it used in this file?

irohad/ordering/impl/ordering_service_transport_grpc.cpp
l4l: It's const param

irohad/network/ordering_service.hpp
l4l: Quite inconsistent interface then

irohad/ordering/impl/ordering_service_impl.cpp
l4l: There's no other way to cast it from PW

PR #955, opened 2018-02-08
Fix flags in cmake

Warchant:
This fix adds caching for default flags, so now it is possible to set them in cmake time.

Before:
```
cmake -DCMAKE_CXX_FLAGS="-O3" ..
```
will be overwritten with `set(CMAKE_CXX_FLAGS "...")

Now -- CMAKE_CXX_FLAGS will be cached from cmake arguments.


Warchant:
It has been decided that @bakhtin will pull these changes to his branch first.

PR #958, opened 2018-02-08
Add created_time in model::Proposal

l4l:
### Description of the Change

Add missing field `created_time` at the `model::Proposal`, also initializing that field where it possible, fixes ir-501 todo

### Possible Drawbacks and Alternate Designs

Can be a problem with block initialization here:
https://github.com/hyperledger/iroha/blob/92d329b3ec88ed7ebe6310fd9c34313c3d09bd18/irohad/simulator/impl/simulator.cpp#L74

Imo it's better use time::now() instead and remove the timestamp field from proposal completely

### DIFF
--- a/irohad/validation/impl/stateful_validator_impl.cpp
+++ b/irohad/validation/impl/stateful_validator_impl.cpp
@@ -68,7 +68,7 @@ namespace iroha {
       model::Proposal validated_proposal(
           std::accumulate(txs.begin(), txs.end(), valid, filter));
       validated_proposal.height = proposal.height;
-      validated_proposal.height = iroha::time::now();
+      validated_proposal.created_time = iroha::time::now();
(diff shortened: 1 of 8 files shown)


irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -66,6 +68,7 @@ namespace iroha {
       model::Proposal validated_proposal(
           std::accumulate(txs.begin(), txs.end(), valid, filter));
       validated_proposal.height = proposal.height;
+      validated_proposal.height = iroha::time::now();
```
kamilsa: what?

l4l: oops, typical copy-paste mistake

PR #959, opened 2018-02-09
Refactor logger format

grimadas:
### Description of the Change

Add date to logger and refactor the format

### Benefits

Now we have full date in logger

### Possible Drawbacks 
Check the new format: 
Non debug mode
image
Debug mode:
image

### DIFF
--- a/libs/logger/logger.cpp
+++ b/libs/logger/logger.cpp
@@ -38,11 +38,20 @@ namespace logger {
   }
 
   static void setGlobalPattern() {
-    spdlog::set_pattern("%Y-%m-%d %H:%M:%S th:%t %l %n %v");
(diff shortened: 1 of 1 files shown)


l4l:
Would be nice to:
1. Separate date&thread info, e.g in `[` `]` brackets as it was before
2. Make it configurable, e.g thread info can be useless sometime (moreover it's useful for debuging i tihnk)

luckychess:
Could you also please print thread ids in hex format? It would be a way more comfortable to read.

bakhtin:
LGTM from log processing point of view. Please, fix as suggested to make it more readable.

neewy:
can we also please add milliseconds to track time more precisely?

PR #961, opened 2018-02-09
[WIP] Hunterize iroha

Warchant:
# Motivation
Current approach with dependencies is bad, because we are forced to hardcode paths for libs and include headers. This is error prone is some toolchains, such as XCode, which we need to support.
The solution we've selected is [Hunter](https://github.com/ruslo/hunter) -- CMake-only script, which manages all dependencies in configuration time.

# Comments

- **This trunk and PR is used to test CI and keep track of progress**. 
- **do not push directly to this branch**, only via PRs. 

After we finish hunterization of all dependencies, we will forget about headache with them forever.
This is a good way to manage dependencies and good chance to upgrade their versions to the latest.

# Current state

This branch will be used as trunk. This trunk will be merged into develop as soon as all dependencies will be hunterized. 

Please do not update the following list by yourself. It helps to keep track of tasks required to finish migration to Hunter.

List of dependencies. (x - hunterized, empty - not hunterized)
- [ ] protobuf
- [ ] postgresql + pqxx
- [ ] grpc
- [x] rapidjson 1.1.0
- [x] boost 1.65.1
    - [x] filesystem
    - [x] system
    - [ ] spirit
- [x] gtest and gmock 1.8.0
- [ ] gflags 2.2 - needs to be tested, may be buggy, so probably we will need to hunterize different version.
- [ ] c-ares
- [x] spdlog 0.13 -- but we need 0.16, so we need to hunterize new version.
- [ ] rxcpp
- [ ] ed25519 -- repository hyperledger/iroha-ed25519 is ready to be hunterized
- [ ] tbb
- [ ] optional (martinmoene)
- [ ] swig
- [ ] google benchmark

Also, all CI systems and scripts should be cleaned and fixed:
- [ ] CircleCI
- [ ] Jenkins

# Other changes

- organize cmake files and scripts in better structure
- add new configuration type `-DCMAKE_BUILD_TYPE=Coverage`, which adds necessary flags to collect more quality coverage metrics
- add ccache support for XCode
- add hunter binary cache: 
    Now you can specify 2 environment variables: `GITHUB_USER` with your username and `GITHUB_USER_PASSWORD` with your personal access token with permission `public_repo`, then specify `cmake .. -DHUNTER_RUN_UPLOAD=ON` and it will download, build and upload binary files to https://github.com/soramitsu/hunter-binary-cache . Cache hit will be if and only if hunter version, toolchain and architecture are same.
- fix bug in `clang-format` target: it fixed by mistake all files inside `external` directory.
- update minimal required cmake to 3.7


Warchant:
This PR is going to be splitted onto multiple PRs.

First is https://github.com/hyperledger/iroha/pull/983

PR #963, opened 2018-02-11
Update stateful validation to shm

kamilsa:
### Description of the Change

Changes API of stateful validation to use new model. 

### Benefits

Stateful validation now uses shared model objects

### Possible Drawbacks 

Getting validated proposal from original proposal uses conversion to shared_model::proto::Proposal, which should not be there, because iroha internally should not know about transport.

Tests were updated to use new model.


irohad/simulator/impl/simulator.cpp
```diff
@@ -63,8 +65,12 @@ namespace iroha {
...
           [&](expected::Value<std::unique_ptr<ametsuchi::TemporaryWsv>>
                   &temporaryStorage) {
-            notifier_.get_subscriber().on_next(
-                validator_->validate(proposal, *(temporaryStorage.value)));
+            auto validated_proposal =
+                validator_->validate(shared_model::detail::makePolymorphic<
+                                         shared_model::proto::Proposal>(
+                    shared_model::proto::from_old(proposal)),
+                                     *(temporaryStorage.value));
+            notifier_.get_subscriber().on_next(*validated_proposal->makeOldModel());
```
l4l: leak here, use `unique_ptr`

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -18,7 +18,14 @@
 #include "validation/impl/stateful_validator_impl.hpp"
 #include <numeric>
 #include <set>
+
+#include "datetime/time.hpp"
```
l4l: [Should be below](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -18,7 +18,14 @@
...
+#include <boost/range/adaptor/transformed.hpp>
+
+#include "backend/protobuf/from_old_model.hpp"
+#include "builders/protobuf/proposal.hpp"
 #include "model/account.hpp"
+#include "validation/impl/stateful_validator_impl.hpp"
```
l4l: It was at 18th line

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -27,11 +34,15 @@ namespace iroha {
...
+    shared_model::detail::PolymorphicWrapper<interface::Proposal>
+    StatefulValidatorImpl::validate(
+        const shared_model::detail::PolymorphicWrapper<interface::Proposal>
+            &proposal,
         ametsuchi::TemporaryWsv &temporaryWsv) {
-      log_->info("transactions in proposal: {}", proposal.transactions.size());
-      auto checking_transaction = [this](auto &tx, auto &queries) {
+      log_->info("transactions in proposal: {}",
+                 proposal->transactions().size());
+      auto checking_transaction = [this](const model::Transaction &tx,
```
l4l: Why moved from `auto` type?

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -42,9 +53,18 @@ namespace iroha {
                 }
                 |
                 [&](const auto &signatories) {
+                  auto shm_signatories =
```
l4l: Avoid using shared model in variable names, otherwise can have some inconsistency if [we decided renaming](https://github.com/hyperledger/iroha/issues/967#issuecomment-365215540)

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -42,9 +53,18 @@ namespace iroha {
...
+                      signatories
+                      | boost::adaptors::transformed([](const auto &signatory) {
+                          return crypto::PublicKey(signatory.to_string());
+                        });
                   // Check if signatures in transaction are account signatory
-                  return this->signaturesSubset(tx.signatures, signatories)
-                      ? nonstd::make_optional(signatories)
+                  return this->signaturesSubset(
+                             shared_model::proto::from_old(tx).signatures(),
+                             std::vector<crypto::PublicKey>(
```
l4l: Why is an additional allocation here?

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -53,41 +73,54 @@ namespace iroha {
       // Filter only valid transactions
       auto filter = [&temporaryWsv, checking_transaction](auto &acc,
                                                           const auto &tx) {
-        auto answer = temporaryWsv.apply(tx, checking_transaction);
+        auto answer =
+            temporaryWsv.apply(*tx->makeOldModel(), checking_transaction);
```
l4l: Leak again

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -53,41 +73,54 @@ namespace iroha {
...
-      model::Proposal validated_proposal(
-          std::accumulate(txs.begin(), txs.end(), valid, filter));
-      validated_proposal.height = proposal.height;
+      auto valid_txs = std::accumulate(txs.begin(), txs.end(), valid, filter);
+
+      // rework validation logic, so that this conversion is not needed
+      auto valid_proto_txs =
+          valid_txs
+          | boost::adaptors::transformed([](const auto &polymorphic_tx) {
+              return *static_cast<const shared_model::proto::Transaction *>(
```
l4l: maybe just `return *polymorphic_tx.operator->();`?

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -53,41 +73,54 @@ namespace iroha {
...
+                                    .createdTime(proposal->created_time())
+                                    .height(proposal->height())
+                                    .transactions(valid_proto_txs)
+                                    .build();
+
       log_->info("transactions in verified proposal: {}",
-                 validated_proposal.transactions.size());
-      return validated_proposal;
+                 validated_proposal.transactions().size());
+      return shared_model::detail::makePolymorphic<decltype(
```
l4l: Use `interface::Proposal` instead, it isn't deduced from args

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -53,41 +73,54 @@ namespace iroha {
...
     bool StatefulValidatorImpl::signaturesSubset(
-        const model::Transaction::SignaturesType &signatures,
-        const std::vector<pubkey_t> &public_keys) {
+        const shared_model::interface::Transaction::SignatureSetType
+            &signatures,
+        const std::vector<crypto::PublicKey> &public_keys) {
       // TODO 09/10/17 Lebedev: simplify the subset verification IR-510
       // #goodfirstissue
-      std::set<pubkey_t> txPubkeys;
+      std::unordered_set<std::string> txPubkeys;
```
l4l: What was the issue with the ordered one?

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -53,41 +73,54 @@ namespace iroha {
...
-        accPubkeys.insert(pubkey);
+        txPubkeys.insert(sign->publicKey().toString());
       }
-      return std::includes(accPubkeys.begin(),
-                           accPubkeys.end(),
-                           txPubkeys.begin(),
-                           txPubkeys.end());
+      return std::all_of(public_keys.begin(),
+                         public_keys.end(),
+                         [&txPubkeys](const auto &public_key) {
```
l4l: There's comparator for `std::includes` as well, use it. It more intuitive (and faster) to use it here

irohad/validation/impl/stateful_validator_impl.hpp
```diff
@@ -17,7 +17,12 @@
 #ifndef IROHA_STATEFUL_VALIDATIOR_IMPL_HPP
 #define IROHA_STATEFUL_VALIDATIOR_IMPL_HPP
 
+#include "backend/protobuf/transaction.hpp"
+#include "interfaces/base/signable.hpp"
+#include "builders/protobuf/transport_builder.hpp"
```
l4l: Unordered includes

irohad/validation/impl/stateful_validator_impl.cpp
lebdron: There are no `time_t` or `now()` calls in the file. Is this required?

irohad/validation/impl/stateful_validator_impl.cpp
lebdron: `makePolymorphic` also does `make_shared`, which will not work on abstract type.

irohad/validation/impl/stateful_validator_impl.hpp
```diff
@@ -17,7 +17,12 @@
 #ifndef IROHA_STATEFUL_VALIDATIOR_IMPL_HPP
 #define IROHA_STATEFUL_VALIDATIOR_IMPL_HPP
 
+#include "backend/protobuf/transaction.hpp"
+#include "interfaces/base/signable.hpp"
```
lebdron: Is signable used here? protobuf transaction also does not seem to be used.

irohad/validation/impl/stateful_validator_impl.cpp
kamilsa: No, then we will allocate interface::Transaction which is abstract. And yet, ProposalBuilder below requires list of proto::Transactions

irohad/validation/impl/stateful_validator_impl.cpp
kamilsa: None. I updated this function, so that now it goes through public key and checks if all of them are in this set. Checking presence of element works faster with unordered sets

irohad/validation/impl/stateful_validator_impl.cpp
kamilsa: Not suitable here. std::includes work with ordered sets. In order to get ordered sets from passed containers we will need to iterate through both containers and then perform `std::includes`. Using my approach we iterate through 1st container, and while we are iterating through second we return bool result

l4l: It's only the matter of performance?

irohad/validation/impl/stateful_validator_impl.cpp
l4l: Yeah, sorry. I mean proto::Proposal, ofc

irohad/validation/impl/stateful_validator_impl.cpp
l4l: It's only the matter of performance?

irohad/simulator/impl/simulator.cpp
```diff
@@ -63,8 +65,13 @@ namespace iroha {
...
           [&](expected::Value<std::unique_ptr<ametsuchi::TemporaryWsv>>
                   &temporaryStorage) {
+            auto shm_proposal = shared_model::detail::makePolymorphic<
+                shared_model::proto::Proposal>(
+                shared_model::proto::from_old(proposal));
+            auto validated_proposal =
+                validator_->validate(shm_proposal, *temporaryStorage.value);
             notifier_.get_subscriber().on_next(
-                validator_->validate(proposal, *(temporaryStorage.value)));
+                *validated_proposal->makeOldModel());
```
l4l: Leakage here

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -53,41 +71,55 @@ namespace iroha {
       // Filter only valid transactions
       auto filter = [&temporaryWsv, checking_transaction](auto &acc,
                                                           const auto &tx) {
-        auto answer = temporaryWsv.apply(tx, checking_transaction);
+        auto *old_tx = tx->makeOldModel();
```
l4l: better to use unique_ptr, always deallocated (even at the exception) and almost no overhead

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -29,11 +30,14 @@ namespace iroha {
       log_ = logger::log("SFV");
     }
 
-    model::Proposal StatefulValidatorImpl::validate(
-        const model::Proposal &proposal,
+    shared_model::detail::PolymorphicWrapper<shared_model::interface::Proposal>
+    StatefulValidatorImpl::validate(
+        const shared_model::detail::PolymorphicWrapper<
```
lebdron: Polymorphic wrapper -> shared_ptr

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -55,42 +69,55 @@ namespace iroha {
...
+      auto &txs = proposal->transactions();
       decltype(txs) valid = {};
 
-      model::Proposal validated_proposal(
-          std::accumulate(txs.begin(), txs.end(), valid, filter));
-      validated_proposal.height = proposal.height;
-      validated_proposal.created_time = proposal.created_time;
+      auto valid_txs = std::accumulate(txs.begin(), txs.end(), valid, filter);
+
+      // rework validation logic, so that this conversion is not needed
```
lebdron: Looks like a TODO, so a Jira id is required.

irohad/validation/impl/stateful_validator_impl.hpp
```diff
@@ -40,8 +43,10 @@ namespace iroha {
        * all changes after removing wsv will be ignored
        * @return proposal with valid transactions
        */
-      model::Proposal validate(const model::Proposal &proposal,
-                               ametsuchi::TemporaryWsv &temporaryWsv) override;
+      std::shared_ptr<shared_model::interface::Proposal> validate(
+          const shared_model::detail::PolymorphicWrapper<
```
lebdron: const shared_model::detail::PolymorphicWrapper -> std::shared_ptr

irohad/validation/stateful_validator.hpp
```diff
@@ -38,8 +40,9 @@ namespace iroha {
        * all changes after removing wsv will be ignored
        * @return proposal with valid transactions
        */
-      virtual model::Proposal validate(
-          const model::Proposal &proposal,
+      virtual std::shared_ptr<shared_model::interface::Proposal> validate(
+          const shared_model::detail::PolymorphicWrapper<
```
lebdron: const shared_model::detail::PolymorphicWrapper -> std::shared_ptr

irohad/validation/impl/stateful_validator_impl.hpp
```diff
@@ -17,7 +17,10 @@
 #ifndef IROHA_STATEFUL_VALIDATIOR_IMPL_HPP
 #define IROHA_STATEFUL_VALIDATIOR_IMPL_HPP
 
+#include "backend/protobuf/transaction.hpp"
+#include "model/converters/pb_transaction_factory.hpp"
```
lebdron: It does not seem that these newly added (backend, model, validators) headers are required here.

irohad/simulator/impl/simulator.cpp
```diff
@@ -63,8 +65,14 @@ namespace iroha {
       temporaryStorageResult.match(
           [&](expected::Value<std::unique_ptr<ametsuchi::TemporaryWsv>>
                   &temporaryStorage) {
-            notifier_.get_subscriber().on_next(
-                validator_->validate(proposal, *(temporaryStorage.value)));
+            auto shm_proposal = shared_model::detail::makePolymorphic<
```
lebdron: makePolymorphic should not be required.

irohad/simulator/impl/simulator.cpp
```diff
@@ -16,7 +16,9 @@
  */
 
 #include "simulator/impl/simulator.hpp"
+#include "backend/protobuf/from_old_model.hpp"
 #include "model/sha3_hash.hpp"
+#include "utils/polymorphic_wrapper.hpp"
```
lebdron: polymorphic_wrapper.hpp is also not required since PW will not be used.

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -55,42 +68,55 @@ namespace iroha {
...
+      auto valid_proto_txs =
+          valid_txs
+          | boost::adaptors::transformed([](const auto &polymorphic_tx) {
+              return static_cast<const shared_model::proto::Transaction&>(
+                  *polymorphic_tx.operator->());
```
lebdron: *polymorphic_tx.operator->() -> *tx

l4l: We have no `operator*` actually

irohad/validation/impl/stateful_validator_impl.hpp
```diff
@@ -33,15 +34,16 @@ namespace iroha {
 
       /**
        * Function perform stateful validation on proposal
-       * and return proposal with valid transactions
+       * and return proposal with valid
```
lebdron: "transactions" are missing.

test/module/irohad/validation/validation_mocks.hpp
```diff
@@ -19,6 +19,8 @@
 #define IROHA_VALIDATION_MOCKS_HPP
 
 #include <gmock/gmock.h>
+#include "interfaces/iroha_internal/proposal.hpp"
+#include "utils/polymorphic_wrapper.hpp"
```
lebdron: PolymorphicWrapper is unused.

irohad/validation/stateful_validator.hpp
```diff
@@ -17,7 +17,9 @@ limitations under the License.
 #define IROHA_VALIDATION_STATEFUL_VALIDATOR_HPP
 
 #include "ametsuchi/temporary_wsv.hpp"
+#include "interfaces/iroha_internal/proposal.hpp"
 #include "model/proposal.hpp"
+#include "utils/polymorphic_wrapper.hpp"
```
lebdron: `model/proposal.hpp` and `utils/polymorphic_wrapper.hpp` are unused.

irohad/validation/impl/stateful_validator_impl.hpp
```diff
@@ -18,6 +18,7 @@
 #define IROHA_STATEFUL_VALIDATIOR_IMPL_HPP
 
 #include "validation/stateful_validator.hpp"
+#include "validators/default_validator.hpp"
```
lebdron: Is `default_validator` used?

PR #964, opened 2018-02-11
Cache with custom type

kamilsa:
### Description of the Change

Allows using cache with custom type as a key when corresponding hasher is defined.
### Benefits

Makes cache class more flexible

### Possible Drawbacks 

None

Test is provided

### DIFF
--- a/libs/cache/cache.hpp
+++ b/libs/cache/cache.hpp
@@ -24,9 +24,9 @@
 
 namespace iroha {
   namespace cache {
-    template <typename KeyType, typename ValueType>
+    template <typename KeyType, typename ValueType, typename KeyHash = std::hash<KeyType>>
     class Cache
-        : public AbstractCache<KeyType, ValueType, Cache<KeyType, ValueType>> {
+        : public AbstractCache<KeyType, ValueType, Cache<KeyType, ValueType, KeyHash>> {
      public:
       uint32_t getIndexSizeHighImpl() const {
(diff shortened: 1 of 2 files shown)


test/module/libs/cache/cache_test.cpp
```diff
@@ -132,3 +132,44 @@ TEST(CacheTest, FindVeryOldTransaction) {
...
+// Test with custom type
+struct Key {
+  std::string info;
+
+  bool operator==(const Key &a) const {
+    return info == a.info;
+  }
+};
+
+// Hash stratefy for key type
```
luckychess: Typo f->g

libs/cache/cache.hpp
```diff
@@ -24,9 +24,9 @@
 
 namespace iroha {
   namespace cache {
-    template <typename KeyType, typename ValueType>
+    template <typename KeyType, typename ValueType, typename KeyHash = std::hash<KeyType>>
```
muratovv: Add the documentation for the class.

test/module/libs/cache/cache_test.cpp
```diff
@@ -132,3 +132,41 @@ TEST(CacheTest, FindVeryOldTransaction) {
   }
   ASSERT_EQ(cache.findItem("0"), boost::none);
 }
+
+// Test with custom type
```
muratovv: use `///` for documentation instead

PR #965, opened 2018-02-11
Fix/pip

x3medima17:
### Description of the Change
This PR enables bindings compatibility with Python 3, but Python 2 is still supported.
It also improves code style in python script, and removes unneeded dependencies.

### Benefits
Python 2 and Python 3 support

### Possible Drawbacks 
None


l4l:
screenshot_20180212_182756

something wrong here

l4l:
Test failing as well

example/python/prepare.sh
```diff
@@ -5,9 +5,9 @@ cd $(dirname $0)
 # generate iroha lib
 CURDIR="$(cd "$(dirname "$0")"; pwd)"
 IROHA_HOME="$(dirname $(dirname "${CURDIR}"))"
-cmake -H$IROHA_HOME -Bbuild -DSWIG_PYTHON=ON -DSHARED_MODEL_DISABLE_COMPATIBILITY=ON;
+cmake -H$IROHA_HOME -Bbuild -DSWIG_PYTHON=ON -DSHARED_MODEL_DISABLE_COMPATIBILITY=ON -DSUPPORT_PYTHON2=OFF;
```
l4l: Better use: `python -c "print('OFF' if __import__('sys').version_info[:][0] == 3 else 'ON')"`

example/python/prepare.sh
```diff
@@ -5,9 +5,9 @@ cd $(dirname $0)
...
 cmake --build build/ --target irohapy -- -j"$(getconf _NPROCESSORS_ONLN)"
 
 # generate proto files in current dir
 protoc --proto_path=../../schema --python_out=. block.proto primitive.proto commands.proto queries.proto responses.proto endpoint.proto
-python2 -m grpc_tools.protoc --proto_path=../../schema --python_out=. --grpc_python_out=. endpoint.proto yac.proto ordering.proto loader.proto
+python3 -m grpc_tools.protoc --proto_path=../../schema --python_out=. --grpc_python_out=. endpoint.proto yac.proto ordering.proto loader.proto
```
l4l: Use default available python, there's may be no `python3` binary

example/python/prepare.sh
x3medima17: `python` is by default 2, to run the 3rd one you have to do it explicitly: `python3`

PR #980, opened 2018-02-15
Fix/itf

Solonets:
### Description of the Change

FIX: https://soramitsu.atlassian.net/browse/IR-974

### Benefits

now checkProposal and checkBlock in Integration Test Framework work

### Possible Drawbacks 

None

### DIFF
--- a/test/framework/integration_framework/integration_test_framework.cpp
+++ b/test/framework/integration_framework/integration_test_framework.cpp
@@ -45,6 +45,7 @@ namespace integration_framework {
                 // permssion-dependent proto building
                 std::vector<std::string>{iroha::model::can_create_domain,
                                          iroha::model::can_create_account,
+                                         iroha::model::can_add_asset_qty,
                                          iroha::model::can_add_peer,
                                          iroha::model::can_receive,
                                          iroha::model::can_transfer})
@@ -71,7 +72,7 @@ namespace integration_framework {
     log_->info("init state");
     // peer initialization
     std::shared_ptr<iroha::keypair_t> old_key(keypair.makeOldModel());
-    iroha_instance_->initPipeline(*old_key);
+    iroha_instance_->initPipeline(*old_key, maximum_block_size_);
     log_->info("created pipeline");
     // iroha_instance_->clearLedger();
     // log_->info("cleared ledger");
--- a/test/framework/integration_framework/integration_test_framework.cpp
+++ b/test/framework/integration_framework/integration_test_framework.cpp
@@ -38,7 +38,7 @@ namespace integration_framework {
             .creatorAccountId("admin@test")
             .txCounter(1)
             .createdTime(iroha::time::now())
-            .addPeer("0.0.0.0:123", key.publicKey())
+            .addPeer("0.0.0.0:10001", key.publicKey())
             .createRole(
                 default_role,
                 // TODO (@l4l) IR-874 create more confort way for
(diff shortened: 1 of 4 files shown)


test/integration/pipeline/tx_pipeline_integration_test.cpp
```diff
@@ -187,6 +187,7 @@ TEST(PipelineIntegrationTest, SendQuery) {
  * @given some user
  * @when sending sample AddAssetQuantity transaction to the ledger
  * @then receive STATELESS_VALIDATION_SUCCESS status on that tx
+ * @then wait for proposal and block
```
l4l: Multiple then here, use just `and` instead

test/integration/pipeline/tx_pipeline_integration_test.cpp
```diff
@@ -199,13 +200,21 @@ TEST(PipelineIntegrationTest, SendTx) {
...
+  auto checkStatelessValid = [](auto &status) {
     ASSERT_NO_THROW(
         boost::get<shared_model::detail::PolymorphicWrapper<
             shared_model::interface::StatelessValidTxResponse>>(status.get()));
   };
+  auto checkProposal = [](auto &proposal) {
+    ASSERT_EQ(proposal->transactions.size(), 1);
+  };
+  auto checkBlock = [](auto &block) {
+    ASSERT_EQ(block->transactions.size(), 0);
```
l4l: What's the problem with the commit?

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -176,9 +180,8 @@ namespace integration_framework {
       ObjectType &ref_for_insertion,
       const WaitTime &wait,
       const std::string &error_reason) {
-    if (!queue.try_pop(ref_for_insertion)) {
-      std::this_thread::sleep_for(wait);
-    }
+    std::unique_lock<std::mutex> lk(queue_mu);
+    queue_cond.wait_for(lk, wait, [&]() { return not queue.empty(); });
```
l4l: What's the point of the unused predicate? Maybe assert is missing
also `()` are redundant

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -96,7 +98,11 @@ namespace integration_framework {
           commit_observable.subscribe([this](auto committed_block) {
             block_queue_.push(
                 std::make_shared<iroha::model::Block>(committed_block));
+            log_->info("block");
+            queue_cond.notify_all();
```
l4l: Why is that the same `cv` that is used for a proposal? I think they should be different

l4l: E.g in theory there can be multiple proposals without the commits but `checkProposal(..).checkBlock(..)` will succeed

Solonets: @l4l as `checkProposal` and `checkBlock` are called sequentially in one thread there are no reason to use different `cv` as they going to block each other any way

test/framework/integration_framework/integration_test_framework.hpp
Solonets: @l4l wat?

test/integration/pipeline/tx_pipeline_integration_test.cpp
Solonets: @l4l no such user

Solonets: @l4l ask the one who wrote this test

Solonets: UPD: no permisions

test/integration/pipeline/tx_pipeline_integration_test.cpp
kamilsa: Please use and as @l4l suggested

PR #987, opened 2018-02-16
Feature/graceful shutdown

luckychess:
### Description of the Change
Adds graceful shutdown for iroha daemon.

### Benefits

Better shutdown management: closing grpc connections, database connections and so on.

### Possible Drawbacks 

None.

Run irohad and then hit ctrl+c or use kill <pid> command: you will see ```shutting down...``` message.


irohad/simulator/impl/simulator.cpp
```diff
@@ -28,17 +28,28 @@ namespace iroha {
         std::shared_ptr<ametsuchi::TemporaryFactory> factory,
         std::shared_ptr<ametsuchi::BlockQuery> blockQuery,
         std::shared_ptr<model::ModelCryptoProvider> crypto_provider)
-        : validator_(std::move(statefulValidator)),
+        : proposal_subscription_(rxcpp::composite_subscription()),
```
lebdron: No need to call empty constructor, it will be called before constructor body. Same for `verified_proposal_subscription_`

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -29,10 +29,16 @@ namespace iroha {
         std::shared_ptr<network::BlockLoader> blockLoader)
         : validator_(std::move(validator)),
           mutableFactory_(std::move(mutableFactory)),
-          blockLoader_(std::move(blockLoader)) {
+          blockLoader_(std::move(blockLoader)),
+          subscription_(rxcpp::composite_subscription()) {
```
lebdron: Same for this constructor as above.

test/integration/client_test.cpp
```diff
@@ -65,56 +65,51 @@ class ClientServerTest : public testing::Test {
...
+        .run();
 
     runner->waitForServersReady();
   }
 
-  virtual void TearDown() {
-    runner->shutdown();
-    th.join();
-  }
+  virtual void TearDown() {}
```
lebdron: No need for empty override.

PR #989, opened 2018-02-17
Fix -Wreturn-type

Warchant:
Because function returns `template V` (some object) and does not return anything, it caused warning, which in turn, with -Werror resulted in compilation error.

UPD: when you use BOOST_ASSERT in such function, and program is compiled with -DNDEBUG (any release mode), it actually disables BOOST_ASSERT and warning `-Wreturn-type` appears.

### DIFF
--- a/shared_model/utils/variant_deserializer.hpp
+++ b/shared_model/utils/variant_deserializer.hpp
@@ -20,6 +20,8 @@
 
 #include <boost/serialization/variant.hpp>
 
+#define NORETURN [[noreturn]]
+
 namespace shared_model {
   namespace detail {
     /**
@@ -40,8 +42,9 @@ namespace shared_model {
          * @tparam Archive container type
          */
         template <class V, class T = typename V::types, class Archive>
-        static V invoke(Archive &&, int) {
+        NORETURN static V invoke(Archive &&, int) {
           BOOST_ASSERT_MSG(false, "Required type not found");
+          std::abort();
         }
       };
 


shared_model/utils/variant_deserializer.hpp
```diff
@@ -18,8 +18,16 @@
...
 #define IROHA_VARIANT_DESERIALIZER_HPP
 
+#include <stdnoreturn.h>
 #include <boost/serialization/variant.hpp>
 
+// http://en.cppreference.com/w/c/language/_Noreturn
+#ifdef noreturn
+#define NORETURN noreturn
+#else
+#define NORETURN [[noreturn]]
```
l4l: [It safe](http://en.cppreference.com/w/cpp/language/attributes) to use only that

shared_model/utils/variant_deserializer.hpp
```diff
@@ -18,8 +18,16 @@
 #ifndef IROHA_VARIANT_DESERIALIZER_HPP
 #define IROHA_VARIANT_DESERIALIZER_HPP
 
+#include <stdnoreturn.h>
 #include <boost/serialization/variant.hpp>
 
+// http://en.cppreference.com/w/c/language/_Noreturn
```
l4l: why not cpp docs D:

shared_model/utils/variant_deserializer.hpp
```diff
@@ -40,8 +48,9 @@ namespace shared_model {
          * @tparam Archive container type
          */
         template <class V, class T = typename V::types, class Archive>
-        static V invoke(Archive &&, int) {
+        NORETURN static V invoke(Archive &&, int) {
           BOOST_ASSERT_MSG(false, "Required type not found");
+          std::exit(1);
```
l4l: Imo it's better to use `std::abort`, so 1) either all of the would be closed 2) it's possible to catch SIGABRT
Also there's a constant for that `EXIT_FAILURE`

PR #996, opened 2018-02-20
Feature/irohad test

luckychess:

### Description of the Change

Simple test which tries to run irohad daemon.

### Benefits

It's possible that after some changes irohad could become unable to run even though all of tests are passed. This test gives a protection for such cases.

### Possible Drawbacks 

None.

Same pr as https://github.com/hyperledger/iroha/pull/969 (without RunIrohadWithoutArgs test due to its inconsistency).


Solonets:
don't integration tests solve it?

luckychess:
Actually no, this is a test for a whole system and it should guarantee that we can run Iroha daemon itself without any problems.

test/system/irohad_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "common/types.hpp"
+#include "main/iroha_conf_loader.hpp"
+
+using namespace boost::process;
+using namespace std::chrono_literals;
+using iroha::operator|;
+
+class IrohadTest : public testing::Test {
+ public:
+  virtual void SetUp() {
```
l4l: It's already virtual, use override instead

test/system/irohad_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+    timeout = 1s;
+    setPaths();
+    auto config = parse_iroha_config(path_config_.string());
+    blockstore_path_ = config[config_members::BlockStorePath].GetString();
+    pgopts_ = config[config_members::PgOpt].GetString();
+  }
+  virtual void TearDown() {
```
l4l: same

test/system/irohad_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+#include <rapidjson/document.h>
```
l4l: Why is that here?

luckychess: We read json config in SetUp().

test/system/irohad_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+    return params(
+        path_config_.string(), path_genesis_.string(), path_keypair_.string());
+  }
+
+ private:
+  void setPaths() {
+    path_irohad_ = boost::filesystem::path(PATHIROHAD);
+    irohad_executable = path_irohad_ / "irohad";
+    path_example_ = path_irohad_.parent_path().parent_path() / "example";
+    path_config_ = path_example_ / "config.sample";
```
Warchant: test will not work locally with such config.

it contains "pg host=localhost", however when you run ./scripts/run-iroha-dev.sh postgres host is "postgres"

luckychess:
@Warchant I fixed an issue you mentioned, could you check it please?

test/system/irohad_test.cpp
```diff
@@ -26,7 +26,14 @@
 #include "common/types.hpp"
 #include "main/iroha_conf_loader.hpp"
 
+// workaround for redefining -WERROR problem
+#undef RAPIDJSON_HAS_STDSTRING
```
Warchant: what kind of error does it show?

luckychess: Warning about redefinition of RAPIDJSON_HAS_STDSTRING. Since we build tests with -WError build fails.

luckychess:
Commit message was renamed after rebase so github hasn't released about merge happened.

PR #998, opened 2018-02-20
Replace flat_file.cpp with NuDB implementation

Warchant:
### Description of the Change

- [x] replaces flat_file.cpp with NuDB implementation.
- [x] removes white-box tests for flatfile
- [x] updates boost (adds additional dependencies) and adds NuDB to docker develop


Warchant:
Closing, because I faced some issues with tests on Linux. Will be opened after I fix them.

PR #1000, opened 2018-02-20
Move cache step to the end

Warchant:
### Description of the Change

Move "save_cache" step in the end of "build" step.

### Benefits

When tests could potentially fail you typically push commit and wait until tests will succeed in CI. 
When save_cache is in the end, test step and analysis steps are executed earlier, so you, as a developer, can save 2-3 min.


bakhtin:
LGTM

PR #1003, opened 2018-02-21
Fix/jenkins ctest output

lebdron:
### Description of the Change
Add CTest error output to Jenkins
Fix destructor order according to C++ standard: 12.4 8, because transaction depends on connection
>Bases and members are destroyed in the reverse order of the completion of their constructor

Fix libpqxx support

### Benefits
Jenkins support

### Possible Drawbacks 
None

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,6 +33,8 @@ pipeline {
         IROHA_REDIS_HOST = "redis-${GIT_COMMIT}-${BUILD_NUMBER}"
         IROHA_POSTGRES_PORT = 5432
         IROHA_REDIS_PORT = 6379
+
+        CTEST_OUTPUT_ON_FAILURE = 1
     }
     agent {
         label 'docker_1'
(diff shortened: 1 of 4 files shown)


Warchant:
To clarify:
Because of destruction order, when connection is destroyed first, transaction becomes dangling pointer. When shared_ptr executes destroy(), test will fail with SIGSEGV.

In https://github.com/hyperledger/iroha/pull/999 I **explicitly** close connections in correct order in TearDown().
In this PR @lebdron changes member order to destroy connection and transaction in correct order, thus **implicitly** solving this problem.

IMO close order should be explicit, because other programmers, when write new tests may reuse code from these tests or write new code based on these tests and *may make the same mistake again*. In case of explicit connection destruction, it is obvious for the reader that connection should be destroyed and he/she will not make this mistake again.

IMO we need to merge both PRs and only then merge into develop.

muratovv:
@Warchant 
We discuss a lot about solutions with @lebdron about this issue.
In my opinion better to fix this problem in Andrei's way because both of solutions may be broken but this him approach more clean and consistent with code environment.
Also, by the way, if similar problems appear this notes that there are architectural problems with encapsulation. And class required to be reworked.

PR #1009, opened 2018-02-23
Pass compiler generator and launchers to dependencies

Warchant:
Signed-off-by: Bogdan Vaneev <warchantua@gmail.com>

### Description of the Change

Pass to dependencies:
- compiler generator
- launcher C, CXX (for ccache)
- add gpr lib to packaging, it is used by `grpc`
- remove grpc++_reflection from packaging -- we do not use it.

Change library output path to ${CMAKE_BUILD_DIR} -- means that all external libs will be compiled in `build` directory so we do not need to write full path to these libs.

Fix identations.

**Note**: `iroha_get_lib_name` only does preformatting of library name.
```
Example:
iroha_get_lib_name(OUT iroha SHARED) -> OUT=libiroha.so
iroha_get_lib_name(OUT iroha STATIC) -> OUT=libiroha.a
```

This PR is split part of https://github.com/hyperledger/iroha/pull/983

### DIFF
--- a/cmake/Modules/Findgrpc.cmake
+++ b/cmake/Modules/Findgrpc.cmake
@@ -34,34 +34,56 @@ set(URL https://github.com/grpc/grpc)
 set(VERSION bfcbad3b86c7912968dc8e64f2121c920dad4dfb)
 set_target_description(grpc "Remote Procedure Call library" ${URL} ${VERSION})
 
+iroha_get_lib_name(GPRLIB       gpr               SHARED)
+iroha_get_lib_name(GRPCLIB      grpc              SHARED)
+iroha_get_lib_name(GRPCPPLIB    grpc++            SHARED)
+iroha_get_lib_name(GRPCPPREFLIB grpc++_reflection SHARED)
+
 if (NOT grpc_FOUND)
   find_package(Git REQUIRED)
   externalproject_add(grpc_grpc
       GIT_REPOSITORY ${URL}
       GIT_TAG        ${VERSION}
-      CMAKE_ARGS -DgRPC_PROTOBUF_PROVIDER=package -DgRPC_PROTOBUF_PACKAGE_TYPE=CONFIG -DProtobuf_DIR=${EP_PREFIX}/src/google_protobuf-build/lib/cmake/protobuf -DgRPC_ZLIB_PROVIDER=package -DBUILD_SHARED_LIBS=ON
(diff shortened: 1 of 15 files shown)


cmake/Modules/Findgrpc.cmake
```diff
@@ -34,34 +34,56 @@ set(URL https://github.com/grpc/grpc)
 set(VERSION bfcbad3b86c7912968dc8e64f2121c920dad4dfb)
 set_target_description(grpc "Remote Procedure Call library" ${URL} ${VERSION})
 
+iroha_get_lib_name(GPRLIB       gpr               SHARED)
+iroha_get_lib_name(GRPCLIB      grpc              SHARED)
+iroha_get_lib_name(GRPCPPLIB    grpc++            SHARED)
+iroha_get_lib_name(GRPCPPREFLIB grpc++_reflection SHARED)
```
l4l: quite useless

cmake/Modules/Findgrpc.cmake
```diff
@@ -34,34 +34,56 @@ set(URL https://github.com/grpc/grpc)
...
-  set(grpc_INCLUDE_DIR ${source_dir}/include)
-  set(grpc_LIBRARY ${binary_dir}/${CMAKE_SHARED_LIBRARY_PREFIX}grpc${CMAKE_SHARED_LIBRARY_SUFFIX})
-  set(grpc_grpc++_LIBRARY ${binary_dir}/${CMAKE_SHARED_LIBRARY_PREFIX}grpc++${CMAKE_SHARED_LIBRARY_SUFFIX})
-  set(grpc_grpc++_reflection_LIBRARY ${binary_dir}/${CMAKE_SHARED_LIBRARY_PREFIX}grpc++_reflection${CMAKE_SHARED_LIBRARY_SUFFIX})
-  set(grpc_CPP_PLUGIN ${binary_dir}/grpc_cpp_plugin)
+  set(grpc_INCLUDE_DIR               ${source_dir}/include)
+  set(gpr_LIBRARY                    ${CMAKE_BINARY_DIR}${XCODE_EXT}/${GPRLIB})
+  set(grpc_LIBRARY                   ${CMAKE_BINARY_DIR}${XCODE_EXT}/${GRPCLIB})
+  set(grpc_grpc++_LIBRARY            ${CMAKE_BINARY_DIR}${XCODE_EXT}/${GRPCPPLIB})
+#  set(grpc_grpc++_reflection_LIBRARY ${CMAKE_BINARY_DIR}${XCODE_EXT}/${GRPCPPREFLIB})
```
l4l: I think that and related comments can be removed

PR #1018, opened 2018-02-28
Feature/shared model cmake

victordrobny:
### Description of the Change
CMake for shared model was reworked. Now shared model is a separate cmake project.

### Benefits
Now shared model can be compiled almost independently from irohad.

### Possible Drawbacks 
There are still dependencies from iroha project(schema, libs). 

#> cd shared_model
#> cmake -H. -Bbuild;
#> cmake --build build -- -j4


CMakeLists.txt
```diff
@@ -98,6 +98,8 @@ if(SHARED_MODEL_DISABLE_COMPATIBILITY)
     add_definitions(-DDISABLE_BACKWARD)
 endif()
 
+SET(IROHA_ROOT_PROJECT ON)
```
Warchant: `set(IROHA_ROOT_PROJECT ON CACHE BOOL "Description")`

or 

`option(IROHA_ROOT_PROJECT "Description" ON)`

currently you have no way to change this var from terminal without changing CMakeLists.txt

shared_model/clean.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+rm -rf external
+rm -rf build
```
Warchant: newline

schema/CMakeLists.txt
```diff
@@ -21,60 +21,72 @@ compile_proto_to_cpp(primitive.proto)
...
-    grpc++
+    add_library(loader_grpc
+        loader.pb.cc
+        loader.grpc.pb.cc
+        )
+    target_link_libraries(loader_grpc
+        schema
+        grpc++
     )
+endif()
```
Warchant: newline

schema/CMakeLists.txt
```diff
@@ -21,60 +21,72 @@ compile_proto_to_cpp(primitive.proto)
...
     proposal.pb.cc
     )
+
+
+
 target_link_libraries(schema
     protobuf
     )
 target_include_directories(schema PUBLIC
     .
```
Warchant: never use `.`, use `${CMAKE_CURRENT_SOURCE_DIR}`

shared_model/cmake/Modules/Finded25519.cmake
```diff
@@ -0,0 +1,56 @@
...
+  set(ed25519_LIBRARY     ${binary_dir}${XCODE_EXT}/${EDLIB})
+  file(MAKE_DIRECTORY ${ed25519_INCLUDE_DIR})
+
+  add_dependencies(ed25519 hyperledger_ed25519)
+endif ()
+
+set_target_properties(ed25519 PROPERTIES
+    INTERFACE_INCLUDE_DIRECTORIES ${ed25519_INCLUDE_DIR}
+    IMPORTED_LOCATION ${ed25519_LIBRARY}
+    )
```
Warchant: add after this line:
```
if(ENABLE_LIBS_PACKAGING)
  add_install_step_for_lib(${ed25519_LIBRARY})
endif()
```

CMakeLists.txt
victordrobny: But I don't want to change it from terminal.

Warchant: Otherwise you do not have any control is n this variable, it will be always ON

l4l:
Maybe use symlinks for the same files?

shared_model/CMakeLists.txt
```diff
@@ -12,6 +12,34 @@
...
+
+    SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules)
+
+    add_definitions(-DDISABLE_BACKWARD)
+
+    include(FeatureSummary)
+    include(cmake/functions.cmake)
+    include(cmake/dependencies.cmake)
+        
+    add_subdirectory(../schema build)
```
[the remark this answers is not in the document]
lebdron: Also `build` can be named `schema` for consistency.

lebdron: Because `../libs` is not added, `generator` library is not built, which will probably result in unresolved symbols. It seems that testing this behaviour will require moving shared library tests to shared_library directory, adding gtest, and TESTING flag.

lebdron: Also `add_subdirectory(../libs libs)` seems necessary.

CMakeLists.txt
victordrobny: Yes, it is correct.

shared_model/clean.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
+rm ../schema/*.{cc,h}
```
l4l: `*.pb.{cc,h}`
Also use `rm $(dirname $0)/../schema/...`

schema/CMakeLists.txt
```diff
@@ -21,60 +21,70 @@ compile_proto_to_cpp(primitive.proto)
...
+
+if (IROHA_ROOT_PROJECT)
+    compile_proto_to_grpc_cpp(endpoint.proto)
+    compile_proto_to_grpc_cpp(yac.proto)
+    compile_proto_to_grpc_cpp(ordering.proto)
+    compile_proto_to_grpc_cpp(loader.proto)
+endif()
+
+if (IROHA_ROOT_PROJECT)
+    SET(endpoint_pb_cc endpoint.pb.cc)
```
l4l: prefer lower case function names (e.g `set` here)

PR #1019, opened 2018-02-28
Refactor permissions and add set account detail in genesis block

neewy:
### Description of the Change

This pull requests brings prettified file permissions.hpp, refactored order in primitive.proto and addition of domain-related permissions to the primitive.proto and pb_command_factory. The order is persistent across modified files, so it is convenient now to check what is the status of permissions.

### Benefits

Now genesis block has almost all possible permissions in it. The code looks a bit better.

### Possible Drawbacks 

1. Genesis block was modified manually. CLI code has to be modified as well.
2. Grant permission for TransferAsset is not checked.

pb_commands_tests.cpp
CommandTest, create_role

I also executed a manual test against modified genesis.block, and it works.

### DIFF
--- a/irohad/model/permissions.hpp
+++ b/irohad/model/permissions.hpp
@@ -24,122 +24,186 @@
 namespace iroha {
   namespace model {
 
+    /* ~~~~~~~~       Command-related permissions        ~~~~~~~~ */
+
+    //  The set of permissions below refer to the specific commands.
+    //  During stateful validations, these permissions are checked
+    //  to be assigned to transaction creator.
+
+    /*                            Role                            */
     const std::string can_append_role = "can_append_role";
     const std::string can_create_role = "can_create_role";
     const std::string can_detach_role = "can_detach_role";
+
+    /*                       Asset quantity                       */
     const std::string can_add_asset_qty = "can_add_asset_qty";
     const std::string can_subtract_asset_qty = "can_subtract_asset_qty";
+
+    /*                            Peer                            */
     const std::string can_add_peer = "can_add_peer";
+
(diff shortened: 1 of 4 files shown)


neewy:
For now, if you add such permissions directly to the genesis block the daemon will not launch

irohad/model/permissions.hpp
```diff
@@ -24,122 +24,186 @@
...
+    /*                           Domain                           */
+    const std::string can_create_domain = "can_create_domain";
+
+    /* ~~~~~~~~       Query-related permissions        ~~~~~~~~   */
+
+    //  The set of permissions below refer to the specific queries.
+    //  During stateful validations, these permissions are checked
+    //  to be assigned to query creator.
+    //  These permissions are divided into three groups:
+    // * my — query creator can only query its data
```
grimadas: It think 'only' is redundant here

schema/primitive.proto
```diff
@@ -1,72 +1,103 @@
...
+ */
+
 syntax = "proto3";
+
+
 package iroha.protocol;
 
+
+/**
+ * Represents any possible value for permission field,
```
grimadas: ❤️

schema/primitive.proto
```diff
@@ -1,72 +1,103 @@
...
+ * then the account is capable of performing such action.
+ *
+ * Command and Query related permissions are included in the role,
+ * which an Account can have.
+ *
+ * Grantable permission is being set as a relation between two accounts:
```
grimadas: 'is being set' - seem to be overcomplicated phrase

PR #1027, opened 2018-03-02
Remove poly wrapper from shared_model::validator

Solonets:

### Description of the Change

Currently, validator interface requires `poly_wrapper` for command and query validation. This is a redundant and not convenient way for the interface.

The pr reworks interface of stateless validators for queries and commands with const ref.

### Benefits

no need to wrap object before passing

### Possible Drawbacks 

none


l4l:
Is it merged?

l4l:
Merged in 110377c

PR #1032, opened 2018-03-02
Tests for GetTransactions query

l4l:
### Description of the Change

Tests for GetTransactions query:

- query self tx without **can_get_{my,all}_txs** perms
- query self tx with **can_get_my_txs**
- query self tx with **can_get_all_txs**
- query inexistent hash
- query other user tx

Moreover there's additional fixes, so most of the tests are passing (or looks quite better):
- making ITF consts public
- create admin account with `all_perm_group` (otherwise no permission for user creation)
- builder for new user with specific permissions (can be useful in other queries tests)
- type checker for `ErrorQueryResponse` (can be useful in other queries tests)
- `SpecifiedVisitor`, which casts variant to optional of specified type (can be useful for most of the shared_model types usage)

### Benefits

Code coverage increased, useful code


Warchant:
Please write meaningful names for PR name

test/framework/base_tx.hpp
```diff
@@ -0,0 +1,51 @@
...
+            key)
+        .txCounter(1)
+        .creatorAccountId(
+            integration_framework::IntegrationTestFramework::kAdminId)
+        .createdTime(iroha::time::now());
+  }
+
+  static inline auto createUserWithPerms(
+      const std::string &user,
+      const std::string &user_id,
```
grimadas: It seems you can get rid of user_id, as you are using kDefaultDomain anyway. Possible solution is to pass domain name to `createUserWithPerms` and  `createUser`

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -31,27 +32,58 @@ using namespace shared_model::crypto;
 using namespace std::literals::string_literals;
 
 namespace integration_framework {
+
+  const std::string IntegrationTestFramework::kDefaultDomain = "test";
```
grimadas: Why it starts with `k`?

l4l: [Due to code style](https://google.github.io/styleguide/cppguide.html#Constant_Names)

test/framework/base_tx.hpp
```diff
@@ -0,0 +1,51 @@
...
+      const std::string role_id,
+      std::vector<std::string> perms) {
+    return createUser(user, key)
+        .detachRole(
+            user_id,
+            integration_framework::IntegrationTestFramework::kDefaultRole)
+        .createRole(role_id, perms)
+        .appendRole(user_id, role_id);
+  }
+}  // namespace framework
```
luckychess: |`line 60`|
|---|
Is detachRole() call is necessary here?

test/framework/base_tx.hpp
```diff
@@ -0,0 +1,51 @@
...
+    return shared_model::proto::TransactionBuilder()
+        .createAccount(
+            user,
+            integration_framework::IntegrationTestFramework::kDefaultDomain,
+            key)
+        .txCounter(1)
+        .creatorAccountId(
+            integration_framework::IntegrationTestFramework::kAdminId)
+        .createdTime(iroha::time::now());
+  }
```
luckychess: I think it could be better to pass domain as an argument.

test/framework/base_tx.hpp
luckychess: Same about creator id.

test/framework/base_tx.hpp
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "backend/protobuf/transaction.hpp"
+#include "builders/protobuf/transaction.hpp"
+#include "cryptography/public_key.hpp"
+#include "datetime/time.hpp"
+#include "framework/integration_framework/integration_test_framework.hpp"
+
```
luckychess: Return, not param.

test/framework/base_tx.hpp
```diff
@@ -0,0 +1,66 @@
...
+namespace framework {
+  /**
+   * Creates a set of transactions for user creation
+   * @param user is username of new user
+   * @param key is a public key of new user
+   * @return pre-built transaction
+   */
+  static inline auto createUser(const std::string &user,
```
luckychess: I still think it would be better just to pass it as a parameter.

shared_model/interfaces/utils/query_error_response_visitor.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef IROHA_QUERY_ERROR_RESPONSE_VISITOR_HPP
+#define IROHA_QUERY_ERROR_RESPONSE_VISITOR_HPP
+
```
luckychess: I do not think interfaces directory is a good place for this. Same for the second class. Maybe just utils?

luckychess:
Missed param.

luckychess:
Looks painful.

test/framework/base_tx.hpp
l4l: Yes, the point is give the user only specified permission, not to append

test/framework/base_tx.hpp
l4l: All of the work inside the only domain, I think for now it's not necessary. Also have a look at https://github.com/hyperledger/iroha/pull/1032#discussion_r172098353

shared_model/interfaces/utils/query_error_response_visitor.hpp
l4l: > I do not think interfaces directory is a good place for this

Why? It works over the ifaces

> Also do we need such utils in other places besides GetTransactions test?

I think it can be useful as a general API (well, maybe just more generic)

test/framework/base_tx.hpp
l4l: Methods becoming more complicated then.
Moreover it's better define for tests with inter-domain logic custom functions, it's not really common I think

shared_model/interfaces/utils/query_error_response_visitor.hpp
luckychess: It works over interfaces but it's not an interface by itself. Also at the moment interface directory contains only interfaces and I think it's a good design so let's keep it.

l4l: Ok, agree

PR #1036, opened 2018-03-04
Update query service

kamilsa:

### Description of the Change

QueryService performs redundant operations. Particularly, the following part of the code contained problems:

https://github.com/hyperledger/iroha/blob/def7bc4f7efdf0c2d1c3caecccd2cf3efe0c889e/irohad/torii/impl/query_service.cpp#L53-L77

1. Here on line 60 response's status is set to `STATELESS_INVALID` but this status is not put into cache, whereas response variable is overwritten on line 74 by previous response for the same query hash.
2. On line 64 empty response was put into cache.
3. On line 73 value of optional is taken without checking its existence
4. No tests reflected aforementioned issues
 
### Benefits

All problems above were fixed.

### Possible Drawbacks 

Now if after query is passed into `QueryProcessor::queryHandle` method, but afterwards corresponding query response is not in the cache(empty optional is returned from cache_), query service will return `NOT_SUPPORTED` status.

### Usage Examples or Tests

Tests, reflecting solution of above problems were introduced.

### DIFF
--- a/irohad/torii/impl/query_service.cpp
+++ b/irohad/torii/impl/query_service.cpp
@@ -16,10 +16,8 @@
  */
 
 #include "torii/query_service.hpp"
-#include "backend/protobuf/from_old_model.hpp"
-#include "backend/protobuf/transaction_responses/proto_tx_response.hpp"
-#include "common/types.hpp"
-#include "model/sha3_hash.hpp"
+#include "backend/protobuf/query_responses/proto_query_response.hpp"
+#include "validators/default_validator.hpp"
 
 namespace torii {
 
@@ -30,21 +28,28 @@ namespace torii {
     query_processor_->queryNotifier().subscribe(
         [this](const std::shared_ptr<shared_model::interface::QueryResponse>
                    &iroha_response) {
-          // Find client to respond
+          auto proto_response =
+              static_cast<shared_model::proto::QueryResponse &>(*iroha_response)
+                  .getTransport();
+
           auto hash = iroha_response->queryHash();
-          auto res = cache_.findItem(hash);
-
-          if (res) {
-            cache_.addItem(hash, iroha_response);
-          }
-
+          cache_.addItem(hash, proto_response);
         });
   }
 
   void QueryService::Find(iroha::protocol::Query const &request,
                           iroha::protocol::QueryResponse &response) {
-    //    shared_model::proto::QueryResponse model_response(response);
     shared_model::crypto::Hash hash;
+    auto blobPayload = shared_model::proto::makeBlob(request.payload());
+    hash = shared_model::proto::Query::HashProviderType::makeHash(blobPayload);
+
+    if (cache_.findItem(hash)) {
+      // Query was already processed
+      response.mutable_error_response()->set_reason(
+          iroha::protocol::ErrorResponse::STATELESS_INVALID);
+      return;
+    }
+
     shared_model::proto::TransportBuilder<
         shared_model::proto::Query,
         shared_model::validation::DefaultQueryValidator>()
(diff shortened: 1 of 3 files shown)


test/module/irohad/torii/query_service_test.cpp
```diff
@@ -29,54 +29,112 @@ using namespace shared_model::detail;
...
+      std::shared_ptr<shared_model::interface::QueryResponse>>
+      notifier;
   EXPECT_CALL(*query_processor, queryNotifier())
-      .WillOnce(
-          Return(rxcpp::observable<>::empty<std::shared_ptr<QueryResponse>>()));
-
+      .WillOnce(Return(notifier.get_observable()));
+  EXPECT_CALL(*query_processor,
+              queryHandle(Truly(
+                  [this](std::shared_ptr<shared_model::interface::Query> rhs) {
```
kamilsa: Note, without Truly and this expression, GMOCK would match query by comparing two `shared_ptr`s using `operator==` which for `shared_ptr` compares `lhs.get() == rhs.get()` i.e. pointer values

irohad/torii/query_service.hpp
```diff
@@ -60,7 +58,7 @@ namespace torii {
     std::shared_ptr<iroha::torii::QueryProcessor> query_processor_;
 
     iroha::cache::Cache<shared_model::crypto::Hash,
-                        std::shared_ptr<shared_model::interface::QueryResponse>,
+                        iroha::protocol::QueryResponse,
```
kamilsa: Note, query service is aware about the transport, so keeping shared model object in cache only makes life harder by adding redundant conversions from model to transport objects

test/module/irohad/torii/query_service_test.cpp
Alexey-N-Chernyshov: mb leave comment in the code?

test/module/irohad/torii/query_service_test.cpp
```diff
@@ -29,54 +29,112 @@ using namespace shared_model::detail;
...
+                    return rhs->operator==(*query);
```
Alexey-N-Chernyshov: why not `*rhs == *query`?

test/module/irohad/torii/query_service_test.cpp
```diff
@@ -29,54 +29,112 @@ using namespace shared_model::detail;
...
+      notifier;
   EXPECT_CALL(*query_processor, queryNotifier())
-      .WillOnce(
-          Return(rxcpp::observable<>::empty<std::shared_ptr<QueryResponse>>()));
-  EXPECT_CALL(*query_processor, queryHandle(_)).WillOnce(Return());
+      .WillOnce(Return(notifier.get_observable()));
+  EXPECT_CALL(*query_processor,
+              queryHandle(Truly(
+                  [this](std::shared_ptr<shared_model::interface::Query> rhs) {
+                    return rhs->operator==(*query);
```
Alexey-N-Chernyshov: the same as above

irohad/torii/impl/query_service.cpp
```diff
@@ -30,21 +28,28 @@ namespace torii {
...
+          auto hash = iroha_response->queryHash();
+          cache_.addItem(hash, proto_response);
         });
   }
 
   void QueryService::Find(iroha::protocol::Query const &request,
                           iroha::protocol::QueryResponse &response) {
-    //    shared_model::proto::QueryResponse model_response(response);
     shared_model::crypto::Hash hash;
+    auto blobPayload = shared_model::proto::makeBlob(request.payload());
```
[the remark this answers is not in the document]
kamilsa: No, request is iroha::protocol::Query object and returns message type

PR #1038, opened 2018-03-05
Fix/bindings

luckychess:
### Description of the Change
Currently bindings do not work because of incompatible changes were done in files related to Swig. This PR fixes this problem (tested for Java and Python). Also it adds getAccountDetail() call into bindings interface and fixes type problem with setAccountDetail().

### Benefits
Bindings targets (like irohajava or irohapy) cannot be compiled at the moment. Also [gs]etAccountDetail methods are pretty unusable at the moment.

### Possible Drawbacks 
None.

Old examples are valid.

### DIFF
--- a/shared_model/interfaces/common_objects/types.hpp
+++ b/shared_model/interfaces/common_objects/types.hpp
@@ -71,10 +71,14 @@ namespace shared_model {
       using AccountNameType = std::string;
       /// Type of asset name
       using AssetNameType = std::string;
-      /// Type of asset name
+      /// Type of detail
       using DetailType = std::string;
       /// Type of JSON data
       using JsonType = std::string;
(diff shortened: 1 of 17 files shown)


shared_model/interfaces/common_objects/types.hpp
```diff
@@ -71,10 +71,14 @@ namespace shared_model {
       using AccountNameType = std::string;
       /// Type of asset name
       using AssetNameType = std::string;
-      /// Type of asset name
+      /// Type of detail
       using DetailType = std::string;
       /// Type of JSON data
       using JsonType = std::string;
+      /// Type of account detail key
+      using AccountDetailKeyType = std::string;
```
nickaleks: Nice!

PR #1043, opened 2018-03-06
Generate key pair from hex string seed in bindings

Solonets:

### Description of the Change

shared_model::bindings::generateKeypair(const std::string &) takes a 32-byte binary string, which is not convenient for end users. Now it instead take a hex string which can be converted to 32-byte binary string.

### Benefits
more convenient for users?

### Possible Drawbacks 

None

### DIFF
--- a/test/module/shared_model/bindings/CMakeLists.txt
+++ b/test/module/shared_model/bindings/CMakeLists.txt
@@ -15,10 +15,17 @@
 addtest(model_query_builder_test
     model_query_builder_test.cpp
     )
+
 target_link_libraries(model_query_builder_test
     bindings
     )
 
+addtest(model_crypto_test
+    model_crypto_test.cpp
+    )
+target_link_libraries(model_crypto_test
+    bindings
+    )
 if (SWIG_PYTHON OR SWIG_JAVA)
   get_property(SWIG_BUILD_DIR GLOBAL PROPERTY SWIG_BUILD_DIR)
 endif()
--- a/test/module/shared_model/bindings/model_crypto_test.cpp
+++ b/test/module/shared_model/bindings/model_crypto_test.cpp
@@ -1,0 +1,53 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2018 All Rights Reserved.
+ * http://soramitsu.co.jp
+ *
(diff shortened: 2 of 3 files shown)


test/module/shared_model/bindings/CMakeLists.txt
```diff
@@ -15,10 +15,17 @@
...
 target_link_libraries(model_query_builder_test
     bindings
     )
 
+addtest(model_crypto_test
+    model_crypto_test.cpp
+    )
+target_link_libraries(model_crypto_test
+    bindings
+    )
```
l4l: add space below plz

test/module/shared_model/bindings/model_crypto_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+#include "bindings/model_crypto.hpp"
+
+#include <gtest/gtest.h>
+
+/**
+ * @given ModelCrypto module
+ * @when Receive 32 byte hex string
+ * @then assertion is not thrown
```
l4l: `... not thrown on keypair generation`

test/module/shared_model/bindings/model_crypto_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+ */
+TEST(ModelCryptoTest, GenerateKeypair) {
+  ASSERT_NO_THROW(
+  shared_model::bindings::ModelCrypto().generateKeypair(std::string(64, 'a'));
```
l4l: Broken formatting here, and below

test/module/shared_model/bindings/model_crypto_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+      shared_model::bindings::ModelCrypto().generateKeypair(std::string(64, 'A'));
+  );
+}
+/**
+ * @given ModelCrypto module
+ * @when Receive invalid hex byte string
+ * @then assertion is thrown
+ */
+TEST(ModelCryptoTest, GenerateKeypairInvalidSeed) {
+  ASSERT_ANY_THROW(
```
l4l: Use `ASSERT_THROW(..., const std::runtime_error&)` instead

PR #1054, opened 2018-03-08
Feature/rework simulator with shared model

Alexey-N-Chernyshov:
### Description of the Change
Rework simulator with shared_model.

### Benefits
Less old model dependencies.

### Possible Drawbacks 
No

no

no


irohad/simulator/impl/simulator.cpp
```diff
@@ -92,21 +94,36 @@ namespace iroha {
           });
     }
 
-    void Simulator::process_verified_proposal(model::Proposal proposal) {
+    void Simulator::process_verified_proposal(
+        const shared_model::interface::Proposal &proposal) {
       log_->info("process verified proposal");
+
       model::Block new_block;
```
Alexey-N-Chernyshov: Block will be replaced with crypto_provider relocation. Now it depends on signature and IR-968 will resolve this dependency.

irohad/simulator/impl/simulator.cpp
```diff
@@ -33,15 +37,18 @@ namespace iroha {
           block_queries_(std::move(blockQuery)),
           crypto_provider_(std::move(crypto_provider)) {
       log_ = logger::log("Simulator");
-      ordering_gate->on_proposal().subscribe(proposal_subscription_,
-                                             [this](model::Proposal proposal) {
-                                               this->process_proposal(proposal);
-                                             });
+      ordering_gate->on_proposal().subscribe(
+          proposal_subscription_, [this](model::Proposal old_proposal) {
+            auto proposal = shared_model::proto::Proposal(
```
kamilsa: why constructor here?

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -42,8 +42,9 @@ namespace iroha {
             block_loader_(std::move(block_loader)),
             delay_(delay) {
         log_ = logger::log("YacGate");
-        block_creator_->on_block().subscribe(
-            [this](auto block) { this->vote(block); });
+        block_creator_->on_block().subscribe([this](auto block) {
```
Warchant: one block copy

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -42,8 +42,9 @@ namespace iroha {
...
+          this->vote(*std::unique_ptr<iroha::model::Block>(block->makeOldModel()));
```
Warchant: this->vote is another block copy

irohad/consensus/yac/impl/yac_gate_impl.cpp
kamilsa: block is `shared_ptr` passing by value is acceptable and more thread-safe

PR #1056, opened 2018-03-08
Feature/improve itf

igor-egorov:
### Description of the Change

This PR represents the first bunch of Integration Test Framework improvements.

### Benefits

- Docstrings are added to public methods of ITF.
- Destruction of ITF now will cause call of done (dropStorage) if no custom deleter was provided.
- Added regression test that uses the rest previously unused ITF methods.
- Added test for sequential initialization of two Iroha instances (test simulates the crash of the first instance - ITF destructed without a call of done).
- Added test for double call of done.

### DIFF
--- a/test/framework/integration_framework/integration_test_framework.cpp
+++ b/test/framework/integration_framework/integration_test_framework.cpp
@@ -18,13 +18,22 @@
 #include "framework/integration_framework/integration_test_framework.hpp"
 
 #include <memory>
+
+#include "backend/protobuf/block.hpp"
+#include "backend/protobuf/queries/proto_query.hpp"
+#include "backend/protobuf/query_responses/proto_query_response.hpp"
+#include "backend/protobuf/transaction.hpp"
+#include "backend/protobuf/transaction_responses/proto_tx_response.hpp"
 #include "builders/protobuf/block.hpp"
 #include "builders/protobuf/proposal.hpp"
 #include "builders/protobuf/transaction.hpp"
 #include "common/files.hpp"
-#include "cryptography/hash_providers/sha3_256.hpp"
+#include "cryptography/crypto_provider/crypto_defaults.hpp"
+#include "cryptography/default_hash_provider.hpp"
 #include "datetime/time.hpp"
-// TODO (@l4l) IR-874 create more comfort way for permssion-dependent proto
+#include "framework/integration_framework/iroha_instance.hpp"
+#include "framework/integration_framework/test_irohad.hpp"
+// TODO (@l4l) IR-874 create more comfort way for permission-dependent proto
 // building
 #include "validators/permissions.hpp"
 
@@ -39,35 +48,23 @@ namespace integration_framework {
   const std::string IntegrationTestFramework::kAdminId = "admin@test";
   const std::string IntegrationTestFramework::kAssetName = "coin";
 
+  IntegrationTestFramework::IntegrationTestFramework(
+      size_t maximum_proposal_size,
+      std::function<void(integration_framework::IntegrationTestFramework &)>
+          deleter)
+      : maximum_proposal_size_(maximum_proposal_size), deleter_(deleter) {}
+
   IntegrationTestFramework::~IntegrationTestFramework() {
-    pqxx::lazyconnection connection(iroha_instance_->pg_conn_);
-    const auto drop = R"(
-DROP TABLE IF EXISTS account_has_signatory;
-DROP TABLE IF EXISTS account_has_asset;
-DROP TABLE IF EXISTS role_has_permissions;
-DROP TABLE IF EXISTS account_has_roles;
-DROP TABLE IF EXISTS account_has_grantable_permissions;
-DROP TABLE IF EXISTS account;
-DROP TABLE IF EXISTS asset;
-DROP TABLE IF EXISTS domain;
-DROP TABLE IF EXISTS signatory;
-DROP TABLE IF EXISTS peer;
-DROP TABLE IF EXISTS role;
-DROP TABLE IF EXISTS height_by_hash;
-DROP TABLE IF EXISTS height_by_account_set;
-DROP TABLE IF EXISTS index_by_creator_height;
-DROP TABLE IF EXISTS index_by_id_height_asset;
-)";
-
-    pqxx::work txn(connection);
-    txn.exec(drop);
-    txn.commit();
-    connection.disconnect();
-
-    iroha::remove_dir_contents(iroha_instance_->block_store_dir_);
+    if (deleter_) {
+      deleter_(*this);
+    }
+    // the code below should be executed anyway in order to prevent app hang
+    if (iroha_instance_ and iroha_instance_->instance_) {
+      iroha_instance_->instance_->terminate();
+    }
   }
 
-  IntegrationTestFramework &IntegrationTestFramework::setInitialState(
+  shared_model::proto::Block IntegrationTestFramework::defaultBlock(
       const shared_model::crypto::Keypair &key) {
     auto genesis_tx =
         shared_model::proto::TransactionBuilder()
@@ -75,8 +72,8 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
             .createdTime(iroha::time::now())
             .addPeer("0.0.0.0:50541", key.publicKey())
             .createRole(kDefaultRole,
-                        // TODO (@l4l) IR-874 create more confort way for
-                        // permssion-dependent proto building
+                        // TODO (@l4l) IR-874 create more comfort way for
+                        // permission-dependent proto building
                         std::vector<std::string>{
                             shared_model::permissions::role_perm_group.begin(),
                             shared_model::permissions::role_perm_group.end()})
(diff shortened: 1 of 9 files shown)


test/regression/regression_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+                .txCounter(1)
+                .addAssetQuantity(kUser, kAsset, "1.0")
+                .build()
+                .signAndAddSignature(
+                    shared_model::crypto::DefaultCryptoAlgorithmType::
+                        generateKeypair());
+
+  auto checkStatelessValid = [](auto &status) {
+    ASSERT_NO_THROW(
+        boost::get<shared_model::detail::
```
l4l: You may use [SpecifiedVisitor](https://github.com/hyperledger/iroha/blob/develop/shared_model/interfaces/utils/specified_visitor.hpp#L28), like here:
https://github.com/hyperledger/iroha/blob/afd5bd1abed98524dd63662a6deb579ac8e2f87e/test/integration/acceptance/invalid_fields_test.cpp#L59-L61

test/regression/regression_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+        .sendTx(tx, checkStatelessValid)
+        .checkProposal(checkProposal)
+        .checkBlock(checkBlock)
+        .done();
+  }
+}
+
+/**
+ * @given ITF instance with Iroha
+ * @when done method is called twice
```
l4l: Actually it is called 3 times, the last one at the dtor

igor-egorov: Correct. The point is to call done() twice explicitly.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -52,40 +50,131 @@ namespace integration_framework {
...
+     * If specified, then will be called instead of default destructor's code
+     */
+    explicit IntegrationTestFramework(
+        size_t maximum_proposal_size = 10,
+        void (*destructor_lambda)(IntegrationTestFramework *) = nullptr)
+        : maximum_proposal_size_(maximum_proposal_size),
+          destructor_lambda_(destructor_lambda) {}
+
+    /**
+     * Initialize Iroha instance with default genesis block and provided signing
```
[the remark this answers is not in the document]
igor-egorov: Default block generation will be moved to a separate function, but two versions of setInitialState are about to exist (because defaultBlock also uses keypair as an argument).

test/regression/regression_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+ * @then no errors are caused as the result
+ */
+TEST(RegressionTest, DoubleCallOfDone) {
+  integration_framework::IntegrationTestFramework itf;
```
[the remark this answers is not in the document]
igor-egorov: The used approach was chosen intentionally. done() by design has void return type.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -60,15 +61,23 @@ namespace integration_framework {
...
+        const std::function<void(IntegrationTestFramework *)> &deleter =
+            nullptr)
+        : maximum_proposal_size_(maximum_proposal_size), deleter_(deleter) {}
+
+    /**
+    * Construct default genesis block
+    * @param key - signing key
+    * @return signed genesis block
+    */
+    virtual shared_model::proto::Block defaultBlock(
```
l4l: 1. Why virtual?
2. Why isn't static?

test/regression/regression_test.cpp
```diff
@@ -0,0 +1,79 @@
+//
+// Created by igor on 07.03.18.
+//
+
+#include <gtest/gtest.h>
+
+#include "builders/protobuf/transaction.hpp"
+#include "cryptography/ed25519_sha3_impl/internal/ed25519_impl.hpp"
```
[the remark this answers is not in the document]
l4l: what about this?

l4l:
Also update your branch?

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -60,15 +61,23 @@ namespace integration_framework {
      */
     explicit IntegrationTestFramework(
         size_t maximum_proposal_size = 10,
-        void (*destructor_lambda)(IntegrationTestFramework *) = nullptr)
-        : maximum_proposal_size_(maximum_proposal_size),
-          destructor_lambda_(destructor_lambda) {}
+        const std::function<void(IntegrationTestFramework *)> &deleter =
```
muratovv: What is the reason to pass the pointer? Maybe rework it with a reference?

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -113,66 +123,59 @@ namespace integration_framework {
 
     /**
      * Send query to Iroha and validate the response
-     * @tparam Lambda - type of function that handles query result
-     * @param qry - query
+     * @param qry - query to be requested
      * @param validation - callback for query result check that receives object
-     * of type shared_model::proto::QueryResponse
-     * @return ITF instance
+     * of type \relates shared_model::proto::QueryResponse by reference
```
muratovv: Does \relates work for types?

\relates <name>
>This command can be used in the documentation of a non-member function <name>. It puts the function inside the 'related function' section of the class documentation. This command is useful for documenting non-friend functions that are nevertheless strongly coupled to a certain class. It prevents the need of having to document a file, but only works for functions.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -60,15 +61,23 @@ namespace integration_framework {
      */
     explicit IntegrationTestFramework(
         size_t maximum_proposal_size = 10,
-        void (*destructor_lambda)(IntegrationTestFramework *) = nullptr)
-        : maximum_proposal_size_(maximum_proposal_size),
-          destructor_lambda_(destructor_lambda) {}
+        const std::function<void(IntegrationTestFramework *)> &deleter =
+            nullptr)
+        : maximum_proposal_size_(maximum_proposal_size), deleter_(deleter) {}
```
muratovv: Move the implementation to cpp.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -113,66 +123,59 @@ namespace integration_framework {
...
-     * @return ITF instance
+     * Request next proposal from queue and serve it with custom handler
+     * @param validation - callback that receives object of type \relates
+     * std::shared_ptr<shared_model::interface::Proposal> by reference
+     * @return this
      */
-    template <typename Lambda>
-    IntegrationTestFramework &checkProposal(Lambda validation);
+    IntegrationTestFramework &checkProposal(
+        const std::function<void(ProposalType &)> &validation);
```
muratovv: Just pass function object.

muratovv: Same in different methods from ITF.

test/regression/regression_test.cpp
```diff
@@ -46,12 +58,7 @@ TEST(RegressionTest, SequentialInitialization) {
   };
   {
     integration_framework::IntegrationTestFramework(
-        10,
-        [](integration_framework::IntegrationTestFramework *this_) {
-          /* TODO Igor Egorov, 08.03.2018, IR-1085
-           * find another way to shutdown ITF without hang */
-          std::this_thread::sleep_for(std::chrono::milliseconds(200));
-        })
+        10, [](integration_framework::IntegrationTestFramework *) {})
```
muratovv: there is possible to use auto instead of itf type

test/framework/integration_framework/integration_test_framework.hpp
l4l: That may cause an ub, you pass reference to `delete`

test/regression/regression_test.cpp
l4l: up

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -39,35 +47,23 @@ namespace integration_framework {
...
+      deleter_(*this);
+    }
+    // the code below should be executed anyway in order to prevent app hang
+    if (iroha_instance_ and iroha_instance_->instance_) {
+      iroha_instance_->instance_->terminate();
+    }
   }
 
-  IntegrationTestFramework &IntegrationTestFramework::setInitialState(
+  shared_model::proto::Block IntegrationTestFramework::defaultBlock(
```
muratovv: Better to return const ref for interface here. There are no reasons to pass concrete implementation here.

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -95,14 +91,20 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
             .createdTime(iroha::time::now())
             .build()
             .signAndAddSignature(key);
-    return setInitialState(key, genesis_block);
+    return genesis_block;
```
muratovv: The `prevHash` field looks outdated, rework it with shared model.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -51,41 +55,139 @@ namespace integration_framework {
...
+     */
+    explicit IntegrationTestFramework(
+        size_t maximum_proposal_size = 10,
+        std::function<void(IntegrationTestFramework &)> deleter =
+            [](IntegrationTestFramework &itf) { itf.done(); });
+
     ~IntegrationTestFramework();
+
+    /**
+    * Construct default genesis block
```
muratovv: We have to write here what it means. E.g. domains, accounts and other entities.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -51,41 +55,139 @@ namespace integration_framework {
...
+    * @param key - signing key
+    * @return signed genesis block
+    */
+    static shared_model::proto::Block defaultBlock(
+        const shared_model::crypto::Keypair &key);
+
+    /**
+     * Initialize Iroha instance with default genesis block and provided signing
```
muratovv: Maybe rework method with default parameter with `defaultBlock` call?

test/framework/integration_framework/integration_test_framework.cpp
l4l: It is a static function, thus for your case there's a need in static object, that may be not a good idea

igor-egorov: Decided with @muratovv and @l4l left this place "as is".

PR #1082, opened 2018-03-18
Add detailed error at config parsing

### LINKED ISSUE #1081: JSON parse error
I got an error when I run **irohad**
`irohadh --config example/config.sample --genesis_block example/genesis.block --keypair_name example/node0` 
error message:
`what():    JSON parse error: eample.sample`

I built iroha at ubuntu docker environment：
`git clone -b develop --depth=1 https://github.com/hyperledger/iroha`
`sh run-iroha-dev.sh`
`cmake -H. -Bbuild; cmake --build build -- -j$(nproc)`
`make`

How can I solve this problem?

Thanks

l4l:
### Description of the Change

- Missed empty line in config, that may resolve #1081 
- More detailed error at config parsing

### DIFF
--- a/example/config.sample
+++ b/example/config.sample
@@ -8,3 +8,4 @@
   "vote_delay" : 5000,
   "load_delay" : 5000
 }
+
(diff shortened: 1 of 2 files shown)


Warchant:
Add test for this case

l4l:
> Add test for this case

Actually that should be handled by the only one current system test, no idea why it is passing. It may be system-dependent behavior

PR #1093, opened 2018-03-19
Change port from 10001 to 50541 in tests

kamilsa:

### Description of the Change

Some ides use port `10001` while the same port is used in tests in Iroha. That leads to failing tests. That PR changes port `10001` to `50541` in such tests. 
### Possible Drawbacks 

None

### DIFF
--- a/test/integration/pipeline/transfer_asset_inter_domain_test.cpp
+++ b/test/integration/pipeline/transfer_asset_inter_domain_test.cpp
@@ -15,8 +15,8 @@
  * limitations under the License.
  */
 
-#include "integration/pipeline/tx_pipeline_integration_test_fixture.hpp"
 #include "backend/protobuf/from_old_model.hpp"
+#include "integration/pipeline/tx_pipeline_integration_test_fixture.hpp"
 
 using namespace std::chrono_literals;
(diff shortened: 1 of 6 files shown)


test/integration/pipeline/transfer_asset_inter_domain_test.cpp
```diff
@@ -28,7 +28,7 @@ class TransferAssetInterDomainTest : public TxPipelineIntegrationTestFixture {
 
     // creates node and admin keys and generate default genesis transaction
     auto genesis_tx1 =
-        TransactionGenerator().generateGenesisTransaction(0, {"0.0.0.0:10001"});
+        TransactionGenerator().generateGenesisTransaction(0, {"0.0.0.0:50541"});
```
x3medima17: You are using this number twice in the same file, maybe it will be better to declare a constant?

x3medima17:
Why 50541? :))

lebdron:
@x3medima17 probably because 50541 falls into 49152–65535 range of ports, which are not registered. https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports

PR #1103, opened 2018-03-20
Jenkinsfile fixes

bakhtin:
#### Jenkinsfile
- Fix ccache not being used during builds on MacOS
- Fix full path to binaries being used instead of relative ones on MacOS. Jenkins now complements PATH with additional lookup paths
- Fix post step may stuck forever on random builds. It now has a timeout. Post step will be forcibly terminated after a certain timeout (60s)
- Fix `cppcheck` target enabled even when coverage is disabled for that build
- Fix local `parallelism` param in Groovy build scripts was always null on the FIRST build. It is a known  [Jenkins bug](https://issues.jenkins-ci.org/browse/JENKINS-41929)

**[UPD]**
- FIx dependencies in release Dockefile not installed properly

### Benefits
Increase stability of CI pipeline. Improve Jenkinsfile readability.

### Possible Drawbacks 
None

### DIFF
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -1,7 +1,14 @@
 #!/usr/bin/env groovy
 
 def doDebugBuild(coverageEnabled=false) {
-  def parallelism = env.PARALLELISM
+  def parallelism = params.PARALLELISM
+  // params are always null unless job is started
+  // this is the case for the FIRST build only.
+  // So just set this to same value as default. 
+  // This is a known bug. See https://issues.jenkins-ci.org/browse/JENKINS-41929
+  if (parallelism == null) {
+    parallelism = 4
+  }
   if ("arm7" in env.NODE_NAME) {
     parallelism = 1
   }
@@ -16,7 +23,7 @@ def doDebugBuild(coverageEnabled=false) {
   def platform = sh(script: 'uname -m', returnStdout: true).trim()
   sh "curl -L -o /tmp/${env.GIT_COMMIT}/Dockerfile --create-dirs https://raw.githubusercontent.com/hyperledger/iroha/${env.GIT_COMMIT}/docker/develop/${platform}/Dockerfile"
   // pull docker image in case we don't have one
-  // speeds up consequent image builds as we simply tag them 
+  // speeds up consequent image builds as we simply tag them
   sh "docker pull ${DOCKER_BASE_IMAGE_DEVELOP}"
   if (env.BRANCH_NAME == 'develop') {
     iC = docker.build("hyperledger/iroha:${GIT_COMMIT}-${BUILD_NUMBER}", "--build-arg PARALLELISM=${parallelism} -f /tmp/${env.GIT_COMMIT}/Dockerfile /tmp/${env.GIT_COMMIT}")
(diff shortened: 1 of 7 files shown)


.jenkinsci/debug-build.groovy
```diff
@@ -36,6 +43,10 @@ def doDebugBuild(coverageEnabled=false) {
     + " -v /var/jenkins/ccache:${CCACHE_DIR}") {
 
     def scmVars = checkout scm
+    def cmakeOptions = ""
```
l4l: Maybe build type?

.jenkinsci/debug-build.groovy
```diff
@@ -48,19 +59,24 @@ def doDebugBuild(coverageEnabled=false) {
...
+        -DIROHA_VERSION=${env.IROHA_VERSION} \
+        ${cmakeOptions}
     """
     sh "cmake --build build -- -j${parallelism}"
     sh "ccache --show-stats"
-    sh "cmake --build build --target test"
-    sh "cmake --build build --target cppcheck"
-
     if ( coverageEnabled ) {
+      sh "lcov -i --capture --directory build --config-file .lcovrc --output-file build/reports/coverage.init.info"
```
l4l: May you explain the purpose of that line?

bakhtin: According to `lcov` manual, we have to generate so-called 'baseline' report prior running tests. That is to ensure proper line coverage percentage. [See lcov man page](http://ltp.sourceforge.net/coverage/lcov/lcov.1.php)

Jenkinsfile
```diff
@@ -150,47 +156,54 @@ pipeline {
...
                   -H. \
                   -Bbuild \
                   -DCMAKE_BUILD_TYPE=${params.BUILD_TYPE} \
-                  -DIROHA_VERSION=${env.IROHA_VERSION}
+                  -DIROHA_VERSION=${env.IROHA_VERSION} \
+                  ${cmakeOptions}
               """
-              sh "/usr/local/bin/cmake --build build -- -j${params.PARALLELISM}"
-              sh "/usr/local/bin/ccache --show-stats"
+              sh "cmake --build build -- -j${params.PARALLELISM}"
```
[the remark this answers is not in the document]
bakhtin: Accessing `params.PARALLELISM` directly works just fine even in case of the initial build. However, if you assign this value to a local variable inside Groovy script and then use it for a build, it's value always null

Jenkinsfile
```diff
@@ -300,23 +321,22 @@ pipeline {
               env.IROHA_VERSION = "0x${scmVars.GIT_COMMIT}"
               env.IROHA_HOME = "/opt/iroha"
               env.IROHA_BUILD = "${env.IROHA_HOME}/build"
-              env.CCACHE_DIR = "${env.IROHA_HOME}/.ccache"
```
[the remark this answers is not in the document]
bakhtin: Exactly. We store build cache in a separate directory

.jenkinsci/debug-build.groovy
bakhtin: Entering this function we already know which build type is selected for the job. So, no, no need to define build type here

PR #1104, opened 2018-03-21
Feature/unique ptr in shm copy ir 1115

Alexey-N-Chernyshov:
### Description of the Change
This PR introduces Cloneable interface and clone() function that returns unique_ptr described here (https://github.com/CppCodeReviewers/Covariant-Return-Types-and-Smart-Pointers) to replace shared_model *copy() method that returns raw pointer.

### Benefits
Safety with unique_ptr.

### Possible Drawbacks 
no

```
 // Inherit from clonable with CRTP
 struct Base : object::cloneable<Base>
 {
   // ... other methods
 };

 // override clone() where it is needed
 struct Derived : Base
 {
   // ... other methods
 protected:
   Derived* clone() const override
   {
     return new Derived(*this);
   }
 };

 // and clone with function clone(obj) or clone(obj *)
 Derived derived;
 auto c1 = clone(derived);
 auto c2 = clone(&derived);
```

1. Just override copy() - this approach will hide base.
2. copy() returns unique_ptr<Base> - static_cast needed every time.
3. The same as implemented, but use method instead of function - less general and need to hide base method in every class in hierarchy of shared_model.

### DIFF
--- a/example/config.sample
+++ b/example/config.sample
@@ -8,3 +8,4 @@
   "vote_delay" : 5000,
   "load_delay" : 5000
 }
+
--- a/example/config.sample
+++ b/example/config.sample
@@ -8,4 +8,3 @@
   "vote_delay" : 5000,
   "load_delay" : 5000
 }
-
--- a/libs/common/cloneable.hpp
+++ b/libs/common/cloneable.hpp
@@ -21,21 +21,18 @@
 #include <memory>
 
 /**
- * Functions and interface to create cloneable classes with
+ * Functions and interface for creation cloneable classes with
  * smart pointers.
  *
  * Usage:
- * struct Base : object::cloneable<Base>
- * {
+ * struct Base : object::cloneable<Base> {
  *   // ... other methods
  * };
  *
- * struct Derived : Base
- * {
+ * struct Derived : Base {
  *   // ... other methods
  * protected:
- *   Derived* clone() const override
- *   {
+ *   Derived* clone() const override {
  *     return new Derived(*this);
  *   }
  * };
@@ -45,7 +42,7 @@
  */
 
 /**
- * Function to clone Cloneable.
+ * Function to clone from Cloneable.
  * @tparam T - derived from Cloneable
  * @param object - object to clone
  * @return clone of object
@@ -85,7 +82,7 @@ class Cloneable {
   /**
    * Polymorphic clone constructor.
    * Method guarantees deep-copy.
-   * @return pointer to clone object
+   * @return pointer to cloned object
    */
   virtual T *clone() const = 0;
 
--- a/libs/common/cloneable.hpp
+++ b/libs/common/cloneable.hpp
@@ -1,5 +1,5 @@
 /**
- * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
+ * Copyright Soramitsu Co., Ltd. 2018 All Rights Reserved.
  * http://soramitsu.co.jp
  *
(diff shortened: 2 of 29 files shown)


example/config.sample
```diff
@@ -8,4 +8,3 @@
   "vote_delay" : 5000,
   "load_delay" : 5000
 }
-
```
l4l: Revert that plz

libs/common/cloneable.hpp
```diff
@@ -0,0 +1,96 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
```
l4l: 2018

libs/common/cloneable.hpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Helper function to copy from pointer to Cloneable.
+ * @tparam T - derived from Cloneable
+ * @param object - object to clone
+ * @return clone of object
+ */
+template <typename T>
+auto clone(T *object) -> decltype(clone(*object)) {
```
l4l: It doesn't work w/o decltype? Afair c++14 allow omit it

shared_model/interfaces/base/model_primitive.hpp
```diff
@@ -29,7 +30,7 @@ namespace shared_model {
      * @tparam Model - your new style model
      */
     template <typename Model>
-    class ModelPrimitive {
+    class ModelPrimitive : public Cloneable<ModelPrimitive<Model>> {
```
l4l: What is the point in moving away cloning function from ModelPrimitive?

test/integration/client_test.cpp
```diff
@@ -218,9 +218,8 @@ TEST_F(ClientServerTest, SendQueryWhenValid) {
   auto account_admin = iroha::model::Account();
   account_admin.account_id = "admin@test";
 
-  auto account_test = std::shared_ptr<shared_model::interface::Account>(
-      shared_model::proto::AccountBuilder().accountId("test@test").build().copy()
-  );
+  std::shared_ptr<shared_model::interface::Account> account_test = clone(
```
l4l: move?

test/module/irohad/model/command_validate_execute_test.cpp
```diff
@@ -1514,7 +1506,7 @@ TEST_F(TransferAssetTest, InvalidWhenAmountOverflow) {
                               transfer_asset->asset_id))
       .WillOnce(Return(dst_wallet));
 
-  // More than account balance
+  // More than account balance)
```
l4l: .

shared_model/interfaces/base/model_primitive.hpp
Alexey-N-Chernyshov: To use it somewhere else besides shared_model

l4l: Not sure that it is needed

test/integration/client_test.cpp
Alexey-N-Chernyshov: moving a temporary object prevents copy elision

libs/common/cloneable.hpp
```diff
@@ -0,0 +1,96 @@
...
+ public:
+  using base_type = T;
+
+  virtual ~Cloneable() = default;
+
+ protected:
+  /**
+   * Polymorphic clone constructor.
+   * Method guarantees deep-copy.
+   * @return pointer to clone object
```
l4l: clone**d**

Issue #628, opened 2017-10-13
Improve check of signatures subset in stateful validator


neewy:
Currently set inclusion is verified in non-effective way. New solution should not increase time complexity of existing solution. This is a good start for a newcomer to contribute to the code:

https://github.com/hyperledger/iroha/blob/5fd84ec8191ceaf69bd49991ffdcbb4f1fb0440d/irohad/validation/impl/stateful_validator_impl.cpp#L74

darsh7807:
```
std::map<pubkey_t,bool>map_keys;
for (auto sign : signatures) {
        map_keys[sign.pubkey] = True;
  }
for (auto pubkey : public_keys) {
       if(map_keys[pubkey] == False) return False;
      }
return True;
```
Is this approach acceptable?

Warchant:
@darsh7807 please look at the latest code, approach may be different:

https://github.com/hyperledger/iroha/blob/develop/irohad/validation/impl/stateful_validator_impl.cpp#L97-L112

Please, always look at the latest code (develop branch) when you're studying iroha :-)

darsh7807:
@Warchant  Ok, I looked at it, I think it is pretty much same thing with the previous one.
I have discussed on the telegram channel and found this approach suitable for this condition:
`for( auto pubkey : pulic_keys){
   for(auto sign : signatures){
         if( sign.pubkey == pubkey)  {
             count++; break;
          }     
    }
}
if ( count == public_key.size()) return True;
return False;
`

darsh7807:
I made a PR please review it 
[PR](https://github.com/hyperledger/iroha/pull/1108)

steephengeorge:
I am new here. Can I make this change as follows. So I can learn the process and take part in this collaborative effort:

 bool StatefulValidatorImpl::signaturesSubset(
        const shared_model::interface::SignatureSetType &signatures,
        const std::vector<shared_model::crypto::PublicKey> &public_keys) {
      // TODO 09/10/17 Lebedev: simplify the subset verification IR-510
      // #goodfirstissue
      return std::all_of(public_keys.begin(),
                         public_keys.end(),
                         [&signatures](const auto &public_key) {
                           return signatures.find(public_key)
                               != signatures.end();
                         });
    }

l4l:
@steephengeorge looks good, couple minors that need to be fixed are:
- the todo comment removal
- adding [dco](https://github.com/probot/dco#how-it-works) for your commit

l4l:
Fixed in #1339

PR #1108, opened 2018-03-22
Improved check of signatures subset in stateful validator

darsh7807:
### Description of the Change
I made two loops to check the whether the public key of signatures exists in the public_keys.

### Benefits
This is making code readable 

### Possible Drawbacks 
It might slow down the runtime of the function


neewy:
@darsh7807 
> It might slow down the runtime of the function

Can you please tell more?

darsh7807:
@neewy Previous implementation was using set to store keys which is O(nlogn) operation. all_of()  function uses O(n) 
So,thats O(nlogn) complexity.
My implementation is O(n*m) so it can slow down for larger data.

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -98,16 +98,16 @@ namespace iroha {
         const std::vector<shared_model::crypto::PublicKey> &public_keys) {
       // TODO 09/10/17 Lebedev: simplify the subset verification IR-510
```
l4l: remove the comment as well

darsh7807: Okay I have removed the comments

lebdron:
@darsh7807 According to https://github.com/hyperledger/iroha/issues/628 , the idea was to improve the efficiency in terms of algorithmic runtime complexity. Since current approach has greater runtime complexity, it does not satisfy the requirements described in the issue.

darsh7807:
@lebdron  We discussed this on Iroha channel on the telegram and what I was informed was the code should be more easy to read and the complexity was not a big issue. So I increased the complexity. 
You can see my attempt to improve complexity in the issue comments but on a channel, they gave feedback that it was not more readable and they wanted a good readable solution.

lebdron:
@darsh7807 Unfortunately there was a miscommunication in developer chat, and no comments were provided on the proposed solution. After revision of issue solutions requirements, we came to conclusion that `O(n^2)` might be unacceptable if we increase the maximum number of signatures allowed for transactions and accounts.

darsh7807:
@lebdron , okay it's fine then.  I will propose another solution using unordered  map which I was discussing on channel. I am out for 9 days. I will submit the solution once I am back.

lebdron:
Please reopen the pull request when it will be completed.

PR #1109, opened 2018-03-22
Fix problem with overloaded method unavailability

luckychess:
### Description of the Change

This pull request eliminates overloading of generateKeypair() function (renames one of overloads to fromPrivateKey).

### Benefits

For languages which don't support function overload (like Javascript) behaviour is unspecified in general. In case of Javascript it resulted in impossibility to access generateKeypair(seed). Swig does not contain any documented ways to deal with this issue.

### Possible Drawbacks 

Client libraries should be rebuilt.

  let keypairFromSeed = crypto.fromPrivateKey(...)


luckychess:
Merged.

PR #1114, opened 2018-03-22
Lcov coverage as a CMake target

bakhtin:

### Description of the Change

Make coverage step in CI a separate Cmake target instead of plain scripting

### Benefits
Less spaghetti code in Jenkinsfile makes it more readable. Further changes to the coverage strategy will not require editing Jenkinsfile, but only the CMake file instead

### Possible Drawbacks 
None

### DIFF
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -1,7 +1,14 @@
 #!/usr/bin/env groovy
 
 def doDebugBuild(coverageEnabled=false) {
-  def parallelism = env.PARALLELISM
+  def parallelism = params.PARALLELISM
+  // params are always null unless job is started
+  // this is the case for the FIRST build only.
+  // So just set this to same value as default. 
+  // This is a known bug. See https://issues.jenkins-ci.org/browse/JENKINS-41929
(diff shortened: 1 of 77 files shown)


.jenkinsci/debug-build.groovy
```diff
@@ -69,7 +69,7 @@ def doDebugBuild(coverageEnabled=false) {
     sh "cmake --build build -- -j${parallelism}"
     sh "ccache --show-stats"
     if ( coverageEnabled ) {
-      sh "lcov -i --capture --directory build --config-file .lcovrc --output-file build/reports/coverage.init.info"
+      sh "cmake --build build --target coverage.init.info"
```
l4l: Looks as quite a strange target imo. Maybe it's better rename to `coverage-init` & `coverage-update`?

bakhtin: This is the way the target was previously defined in [this commit](https://github.com/hyperledger/iroha/commit/87b60b9428778a0e07a6ba7b6060460954d0f50c). Moreover, the target defined in such a way helps to deduct report name that is used in subsequent steps

PR #1118, opened 2018-03-23
Order cli

kamilsa:
### Description of the Change

This PR fixes issue in CLI when order of options to add commands to transaction or create query may differ because unordered map was used
### Benefits

Now order in cli is fixed and determined

### Possible Drawbacks 

Required to change `findInHandlerMap`  so it can take both `map` and `unordered_map`

### DIFF
--- a/irohad/network/impl/block_loader_impl.cpp
+++ b/irohad/network/impl/block_loader_impl.cpp
@@ -86,7 +86,7 @@ rxcpp::observable<std::shared_ptr<Block>> BlockLoaderImpl::retrieveBlocks(
               std::make_shared<shared_model::proto::Block>(std::move(block));
 
           // stateless validation of block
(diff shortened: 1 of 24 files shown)


iroha-cli/interactive/interactive_common_cli.hpp
```diff
@@ -157,22 +158,22 @@ namespace iroha_cli {
...
      * @tparam K key type
      * @tparam V value type
      * @param command_name - key to find in the map
      * @param params_map - map
      * @return nullopt if key not found, value if found
      */
-    template <typename K, typename V>
-    boost::optional<V> findInHandlerMap(K command_name,
-                                         std::unordered_map<K, V> params_map) {
+    template <typename MapType>
```
kamilsa: To accept arbitrary maps

l4l:
Why is this closed?

PR #1121, opened 2018-03-24
Fix dependencies installation in release docker image

bakhtin:

### Description of the Change

Jenkinsfile:
- Fix dependencies not installed when building release docker image

### Benefits
Docker images will come with preinstalled dependencies required to run Iroha.

### Possible Drawbacks 
None

### DIFF
--- a/docker/release/x86_64/Dockerfile
+++ b/docker/release/x86_64/Dockerfile
@@ -7,7 +7,9 @@ RUN apt-get update;\
 
 #Install iroha
 COPY iroha.deb /tmp/iroha.deb
-RUN apt-get install -y /tmp/iroha.deb; rm -f /tmp/iroha.deb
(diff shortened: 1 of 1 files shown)


docker/release/x86_64/Dockerfile
```diff
@@ -7,7 +7,9 @@ RUN apt-get update;\
 
 #Install iroha
 COPY iroha.deb /tmp/iroha.deb
-RUN apt-get install -y /tmp/iroha.deb; rm -f /tmp/iroha.deb
+RUN dpkg -i /tmp/iroha.deb
+
+RUN apt-get -fy install;rm -f /tmp/iroha.deb
```
tyvision: That approach is more lightweight because `gdebi` adds 50+ Mb to the image size.

PR #1138, opened 2018-03-27
iOS build script

lebdron:
### Description of the Change
Iroha bindings build script for iOS

### Benefits
Builds bindings libraries and dependencies with iOS specific parameters - bitcode, ARC, ARM architecture.

### Possible Drawbacks 
None.

### Usage Examples or Tests 
`./example/ios/ios-build.sh OS Debug`


lebdron:
Added +x and `CORES` parameter.

PR #1139, opened 2018-03-27
Feature/java doc

laSinteZ:
### Description of the Change
Added "Java library" section. Kudos to @muratovv  for script and explanations :)
[Here is the compiled page](https://html-xrwpptoamt.now.sh/guides/libraries.html#java-library) to simplify the review process

### Benefits
Now we have an explanation of how to add the Java Client library to the Java project.

### Possible Drawbacks
None

<!-- If no drawbacks, explicitly mention this (write Non

### DIFF
--- a/docs/source/guides/libraries.rst
+++ b/docs/source/guides/libraries.rst
@@ -19,14 +19,126 @@ Troubleshooting
 Java Library
 ------------
 
-Where to Get
-^^^^^^^^^^^^
+Building Native Library
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Prerequisites
+"""""""""""""
+- Java 6
+- Gradle
+
+Build Process
+"""""""""""""
+
+Clone Iroha repository
+
+.. code-block:: shell
+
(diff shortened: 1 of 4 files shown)


l4l:
Rebase plz

docs/source/guides/libraries.rst
```diff
@@ -19,17 +19,150 @@ Troubleshooting
...
+
+Now we need to build our Java native library. Run 
+``example/java/build_library.sh``, which will take care of it.
+
+.. code-block:: shell
+
+  cd example/java
+  ./build_library.sh
+
+.. note:: ``build_library.sh`` script creates a ``dist`` folder with
```
neewy: why dist by the way?

neewy:
There is one thing missing:  in `example/java/TransactionExample.java` we don't have code which explains how to get status of a transaction using gRPC streaming. @muratovv should this example be added in a separate PR?

docs/source/guides/libraries.rst
muratovv: Short form of `distribution`. What is more suitable for it? @neewy

PR #1141, opened 2018-03-28
Use http and check hash after download for swig.

luckychess:

### Description of the Change

Https can be not supported by cmake (it depends of does libcurl support ssl or not). This pr replaces https for http for swig package download and checks its hash after to prevent from downloading something malicious. 

### Benefits

More universal way to download swig with comparable level of security.

### Possible Drawbacks 

None.


luckychess:
Doesn't solve a problem since we http url redirects to another http which redirects to https.

PR #1059, opened 2018-03-09
 Replace old model to shared model in synchronizer

luckychess:
Signed-off-by: luckychess <toobwn@gmail.com>

### Description of the Change

Synchronizer: replace old model to shared model.

### Benefits

Old model is a subject to remove so we are one step closer to success now.

### Possible Drawbacks 
None.

### Usage Examples or Tests *[optional]*
Tests were changed accordingly.

PR #1145, opened 2018-03-28
Remove storage implementation dependency from chain validator

lebdron:
### Description of the Change
Removes postgres dependency from chain validator, because chain validator is not linked with postgres.

### Benefits
Project builds successfully.

### Possible Drawbacks 
None.

### Testing
A test that builds every CMake target independently will be written soon.


lebdron:
Now present in https://github.com/hyperledger/iroha/pull/1059

PR #1147, opened 2018-03-29
Move Android build script from examples

luckychess:

### Description of the Change

Android (and IOS) build scripts are actually not examples so let's move them to better place.

### Benefits

Cleaner and less confusing directory structure.

### Possible Drawbacks 

None.


l4l:
Move the rest packages as well plz, no need to add more similar prs

luckychess:
Rest are actually not packages but examples.

neewy:
спасибо

PR #1148, opened 2018-03-29
Feature/stateless validation rules

kamilsa:
### Description of the Change

Introduces stateless validation rules and updates corresponding tests tests.

| domain_id    | RFC1035, RFC1123 (disallows the root white space)                             |
|--------------|-------------------------------------------------------------------------------|
| address      | IP address of added peer                                                      |
| peer_key     | Public-key of added peer                                                      |
| asset_name   | [a-z_0-9]{1,32}                                                               |
| asset_id     | <asset_name>#<domain_id>                                                      |
| account_name | [a-z_0-9]{1,32}                                                               |
| account_id   | <account_name>@<domain_id>                                                    |
| role_id      | [a-z_0-9]{1,32}                                                               |
| permissions  | See 5.2 Permission                                                            |
| key          | Key of set account detail  [A-Za-z0-9_]{1,64}                                 |
| detail       | ≤ 4096                                                                        |
| quorum       | Number, 0 < quorum ≤ public-key set up to account ≤ 128                       |
| public_key   | Public-key, 32 digits                                                         |
| hash         | Hash value of sha3, 32 digits                                                 |
| created_time | UNIX epoch time, 13 digits in milliseconds, ≤ 24 hours including current time |
| amount       | amount,≥ 0                                                                    |

### Benefits

We have clarified stateless validation rules and updated iroha correspondingly

### Possible Drawbacks 

None


test/module/shared_model/validators/field_validator_test.cpp
```diff
@@ -153,9 +167,9 @@ class FieldValidatorTest : public ValidatorsTest {
       if (!testcase.value_is_valid) {
         ASSERT_TRUE(!reason.second.empty())
             << testFailMessage(field_name, testcase.name);
-        ASSERT_THAT(reason.second.at(0),
-                    testing::MatchesRegex(testcase.expected_message))
-            << testFailMessage(field_name, testcase.name);
+        //        ASSERT_THAT(reason.second.at(0),
```
Alexey-N-Chernyshov: Why dead code here? Please, leave a comment with TODO or fix it.

neewy:
there is one thing we should also limit -- this is a message in transfer asset. there should be a decent limit for it, like 4096 is enough.

l4l:
disclaimer: I haven't look at the patch, only the pr-post, so I'm considering its up-to-date

address: domain name might be useful
detail: why is that small? e.g consider storing additional keys, like rsa; it might be more than 4K
created_time: number that can be represented in 32 bits, there is quite a difference
amount: why can be equal to 0

test/integration/acceptance/transfer_asset_test.cpp
```diff
@@ -285,29 +285,21 @@ TEST_F(TransferAsset, EmptyDesc) {
...
-        auto txes = block->transactions();
-        ASSERT_EQ(txes.size(), 1);
-        auto transfer = *boost::apply_visitor(
-            interface::SpecifiedVisitor<interface::TransferAsset>(),
-            txes[0]->commands()[0]->get());
-        ASSERT_EQ(transfer->message(), long_desc);
-      })
-      .done();
+          kUser1Id, kUser2Id, kAsset, long_desc, kAmount)));
+  ASSERT_ANY_THROW(itf.skipProposal());
```
muratovv: Call done() and check txStatus

PR #1149, opened 2018-03-29
Fix/query execution wsv rework

victordrobny:
### Description of the Change

Parallel queries in torii use the same postgres connection in wsv
 

### Benefits

Query processing factory race condition is fixed.

### DIFF
--- a/irohad/ametsuchi/impl/postgres_wsv_query.cpp
+++ b/irohad/ametsuchi/impl/postgres_wsv_query.cpp
@@ -40,6 +40,15 @@ namespace iroha {
           log_(logger::log("PostgresWsvQuery")),
           execute_{makeExecuteOptional(transaction_, log_)} {}
 
+    PostgresWsvQuery::PostgresWsvQuery(
+        std::unique_ptr<pqxx::lazyconnection> connection,
+        std::unique_ptr<pqxx::nontransaction> transaction)
+        : connection_ptr_(std::move(connection)),
+          transaction_ptr_(std::move(transaction)),
+          transaction_(*transaction_ptr_),
+          log_(logger::log("PostgresWsvQuery")),
+          execute_{makeExecuteOptional(transaction_, log_)} {}
+
     bool PostgresWsvQuery::hasAccountGrantablePermission(
         const AccountIdType &permitee_account_id,
         const AccountIdType &account_id,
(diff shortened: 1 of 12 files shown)


irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -271,7 +271,19 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
     }
 
     std::shared_ptr<WsvQuery> StorageImpl::getWsvQuery() const {
-      return wsv_;
+      auto postgres_connection =
+          std::make_unique<pqxx::lazyconnection>(postgres_options_);
+      try {
+        postgres_connection->activate();
+      } catch (const pqxx::broken_connection &e) {
+        exit(1);
```
lebdron: It is better to replace this with `throw` and TODO with task to change method signature.

test/integration/client_test.cpp
```diff
@@ -221,6 +220,8 @@ TEST_F(ClientServerTest, SendQueryWhenValid) {
   std::shared_ptr<shared_model::interface::Account> account_test = clone(
       shared_model::proto::AccountBuilder().accountId("test@test").build());
 
+  EXPECT_CALL(*storage, getWsvQuery()).WillRepeatedly(Return(wsv_query));
```
lebdron: Since there are no failure cases, these `EXPECT`s can be moved to `SetUp`.

test/module/irohad/torii/torii_queries_test.cpp
```diff
@@ -146,10 +143,13 @@ TEST_F(ToriiQueriesTest, FindAccountWhenNoGrantPermissions) {
   account.account_id = "b@domain";
   auto creator = "a@domain";
 
+  EXPECT_CALL(*storage, getWsvQuery()).WillRepeatedly(Return(wsv_query));
```
lebdron: Same as before, since there are no tests that return `nullptr` for `WsvQuery` or `BlockQuery`, this can be moved to `SetUp`.

irohad/ametsuchi/impl/postgres_wsv_query.cpp
```diff
@@ -209,5 +209,13 @@ namespace iroha {
         return peers;
       };
     }
+    PostgresWsvQuery::PostgresWsvQuery(
```
lebdron: Please place this definition just below definition of another constructor, so that order of definition matches order of declarations in `hpp`.

PR #1150, opened 2018-03-29
fix dependencies for command_execution

victordrobny:

### Description of the Change
Fix dependencies for command_execution target


victordrobny:
The purpose of this pr is to fix command_execution target, not chain_validator

PR #1166, opened 2018-03-30
Smart contracts

rishavpiku:
There  must be a design for the smart contracts and it has been proposed that we need to be aware of using a proper compatible language instead of using either of the Solidity or Go. We might go for scala or else we can define a development environment for the same so that programmers feel free to write, irrespective of anything.


l4l:
Consider opening a new issue

PR #1170, opened 2018-03-31
Uploading of the build artifacts

bakhtin:

### Description of the Change
- A feature to upload artifacts produced after a successful build to the artifact server
- Replace Release build stage in Jenkinsfile with a function

### Benefits
- Artifacts will now be available on external server (artifact.soramitsu.co.jp) for downloading

### Possible Drawbacks 
None

### DIFF
--- a/.jenkinsci/artifacts.groovy
+++ b/.jenkinsci/artifacts.groovy
@@ -1,0 +1,44 @@
+#!/usr/bin/env groovy
+
+def uploadArtifacts(filePaths, uploadPath, artifactServers=['artifact.soramitsu.co.jp']) {
+  def baseUploadPath = 'files'
+  def filePathsConverted = []
+  agentType = sh(script: 'uname', returnStdout: true).trim()
+  uploadPath = baseUploadPath + uploadPath
+  filePaths.each {
+    fp = sh(script: "for i in \$(ls -d ${it}); do FILES=\${i},\${FILES};done;echo \$FILES", returnStdout: true).trim()
+    filePathsConverted.addAll(fp.split(','))
+  }
+  def shaSumBinary = 'sha256sum'
+  def md5SumBinary = 'md5sum'
+  if (agentType == 'Darwin') {
+    shaSumBinary = 'shasum -a 256'
+    md5SumBinary = 'md5 -r'
+  }
+  sh "> \$(pwd)/batch.txt"
+  filePathsConverted.each {
+    sh "echo put ${it} $uploadPath >> \$(pwd)/batch.txt;"
+    sh "$shaSumBinary ${it} | cut -d' ' -f1 > \$(pwd)/\$(basename ${it}).sha256"
+    sh "$md5SumBinary ${it} | cut -d' ' -f1 > \$(pwd)/\$(basename ${it}).md5"
+    sh "echo put \$(pwd)/\$(basename ${it}).sha256 $uploadPath >> \$(pwd)/batch.txt;"
+    sh "echo put \$(pwd)/\$(basename ${it}).md5 $uploadPath >> \$(pwd)/batch.txt;"
+  }
+  // mkdirs recursively
+  uploadPath = uploadPath.split('/')
+  def p = ''
+  sh "> \$(pwd)/mkdirs.txt"
+  uploadPath.each {
+    p += "/${it}"
+    sh("echo -mkdir $p >> \$(pwd)/mkdirs.txt")
+  }
+
+  sshagent(['jenkins-artifact']) {
+    sh "ssh-agent"
+    artifactServers.each {
+      sh "sftp -b \$(pwd)/mkdirs.txt jenkins@${it} || true"
+      sh "sftp -b \$(pwd)/batch.txt jenkins@${it}"
+    }
+  }
+}
+
+return this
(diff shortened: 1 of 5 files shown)


.jenkinsci/artifacts.groovy
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env groovy
+
+def uploadArtifacts(filePaths, uploadPath, artifactServers=['artifact.soramitsu.co.jp']) {
+  def baseUploadPath = 'files'
+  def filePathsConverted = []
+  agentType = sh(script: 'uname', returnStdout: true).trim()
+  uploadPath = baseUploadPath + uploadPath
+  filePaths.each {
+    fp = sh(script: "for i in \$(ls -d ${it}); do FILES=\${i},\${FILES};done;echo \$FILES", returnStdout: true).trim()
```
l4l: Is there any problem with `tr`?

l4l: Prefer `'` instead of `"`, so there is no escapes

l4l: same below

.jenkinsci/artifacts.groovy
```diff
@@ -0,0 +1,44 @@
...
+  sh "> \$(pwd)/mkdirs.txt"
+  uploadPath.each {
+    p += "/${it}"
+    sh("echo -mkdir $p >> \$(pwd)/mkdirs.txt")
+  }
+
+  sshagent(['jenkins-artifact']) {
+    sh "ssh-agent"
+    artifactServers.each {
+      sh "sftp -b \$(pwd)/mkdirs.txt jenkins@${it} || true"
```
l4l: why `|| true?`

.jenkinsci/debug-build.groovy
```diff
@@ -40,7 +40,8 @@ def doDebugBuild(coverageEnabled=false) {
     + " -e IROHA_POSTGRES_USER=${env.IROHA_POSTGRES_USER}"
     + " -e IROHA_POSTGRES_PASSWORD=${env.IROHA_POSTGRES_PASSWORD}"
     + " --network=${env.IROHA_NETWORK}"
-    + " -v /var/jenkins/ccache:${CCACHE_DIR}") {
+    + " -v /var/jenkins/ccache:${CCACHE_DIR}"
+    + " -v /tmp/${GIT_COMMIT}-${BUILD_NUMBER}:/tmp/${GIT_COMMIT}") {
```
l4l: Isn't `GIT_COMMIT` enough?

.jenkinsci/linux-post-step.groovy
```diff
@@ -0,0 +1,24 @@
+def linuxPostStep() {
+  timeout(time: 600, unit: "SECONDS") {
+    try {
+      if (currentBuild.result != "UNSTABLE") {
+        if (BRANCH_NAME ==~ /(master|develop)/) {
```
l4l: can't it be like `if (currentBuild.result != "UNSTABLE" && BRANCH_NAME ==~ /(master|develop)/)`?

Jenkinsfile
```diff
@@ -93,7 +93,13 @@ pipeline {
           steps {
             script {
               debugBuild = load ".jenkinsci/debug-build.groovy"
-              debugBuild.doDebugBuild(true)
+              coverage = load ".jenkinsci/selected-branches-coverage.groovy"
+              if (coverage.selectedBranchesCoverage(['develop', 'master'])) {
```
l4l: just `debugBuild.doDebugBuild(coverage.selectedBranchesCoverage(['develop', 'master']))`?

Jenkinsfile
```diff
@@ -325,44 +328,36 @@ pipeline {
...
-                //archive(includes: 'build/bin/,compile_commands.json')
+              def releaseBuild = load ".jenkinsci/mac-release-build.groovy"
+              releaseBuild.doReleaseBuild()
+            }
+          }
+          post {
+            always {
+              script {
+                timeout(time: 600, unit: "SECONDS") {
+                  if (BRANCH_NAME ==~ /(master|develop)/) {
```
l4l: Maybe generalize linux-post-step?

.jenkinsci/artifacts.groovy
bakhtin: What do you mean? Replace `trim()` with `tr -d`?

bakhtin: `${it}` here refers to an element of `for ... each` loop and should not be escaped. However, escaped variables refer to the shell variables and not defined previously in Groovy script. Usage of single quotes would work for intra-shell vars (like `FILES` here) but `${it}` will be interpreted like a literal string and not a Groovy variable.

.jenkinsci/artifacts.groovy
bakhtin: SFTP does not support recursive directories creation. The only way to create directory tree is creating each non-existing path level-by-level. SFTP exits with non-zero code if path already exists. To prevent this, we exit with zero code if such path already exists.

.jenkinsci/debug-build.groovy
bakhtin: We're using such construction to more easily correlate file system path with actual Jenkins job on a build machine

.jenkinsci/artifacts.groovy
l4l: no, I mean shell script replacement (for + concatenation)

luckychess:
I think libed25519 should be linked statically (as far as I remember) so first case could be OK. About second - you have to use same protobuf version as we use in Findprotobuf.cmake - 3.4.0.

bakhtin:
Please, continue the discussion about bindings in respective PR #1179. I think this PR should be merged regardless the bindings build status

PR #1176, opened 2018-04-02
Validation

rishavpiku:
The check of the signatures could be improved using the following solution, by changing the stateful validator


neewy:
> +1,361 −855

please include only relevant changes

neewy:
I am sorry, this code cannot compile. So far we cannot merge this change, please change your code accordingly, and explain your intention to improve this code and why it is efficient now.

PR #1180, opened 2018-04-04
Fix consensus getting locked due to synchronous subscribers

lebdron:
### Description of the Change
Consensus was getting locked on two nodes because commit was sent to network after performing synchronous actions of subscribers to commit notification. Sending commit to network before notification fixes this issue.

### Benefits
Less locks due to consensus subscribers.

### Possible Drawbacks 
It is hard to test at the moment, because it requires integration test with 2+ nodes.

### DIFF
--- a/irohad/consensus/yac/impl/yac.cpp
+++ b/irohad/consensus/yac/impl/yac.cpp
@@ -210,8 +210,8 @@ namespace iroha {
                              // IR-497
                            },
                            [&](const CommitMessage &commit) {
-                             notifier_.get_subscriber().on_next(commit);
                              this->propagateCommit(commit);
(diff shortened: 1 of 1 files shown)


irohad/consensus/yac/impl/yac.cpp
```diff
@@ -210,8 +210,8 @@ namespace iroha {
                              // IR-497
                            },
                            [&](const CommitMessage &commit) {
-                             notifier_.get_subscriber().on_next(commit);
                              this->propagateCommit(commit);
+                             notifier_.get_subscriber().on_next(commit);
```
muratovv: Nice move.
Also, I wonder to add a task with the test.

PR #1181, opened 2018-04-04
Feature/ordering gate order

lebdron:
### Description of the Change
Replace regular queue with priority queue in ordering gate, so that proposals are propagated in order to the system if they are received in random order from network.

### Benefits
Less issues with Simulator warnings about last block height being inconsistent with proposal height.

### Possible Drawbacks 
None.

### DIFF
--- a/test/module/irohad/ordering/ordering_gate_test.cpp
+++ b/test/module/irohad/ordering/ordering_gate_test.cpp
@@ -224,3 +224,54 @@ TEST(OrderingGateQueueBehaviour, SendManyProposals) {
 
   ASSERT_TRUE(wrapper_after.validate());
 }
+
+/**
+ * @given Initialized OrderingGate
+ * AND MockPeerCommunicationService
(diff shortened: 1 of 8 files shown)


test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -224,3 +224,54 @@ TEST(OrderingGateQueueBehaviour, SendManyProposals) {
...
+    ordering_gate.onProposal(std::make_shared<shared_model::proto::Proposal>(
+        TestProposalBuilder().height(height).build()));
+  };
+
+  std::vector<decltype(ordering_gate.on_proposal())::value_type> messages;
+  ordering_gate.on_proposal().subscribe([&](auto val) {
+    messages.push_back(val);
+  });
+
+  // set flag to false
```
Alexey-N-Chernyshov: what does it mean?

PR #1182, opened 2018-04-04
Fix simulator getting locked when triggered by OrderingGate queue

lebdron:
### Description of the Change
Simulator is getting locked in `getTopBlocks` call, possibly due to incorrect RxCpp usage.

### Benefits
No lock in Simulator.

### Possible Drawbacks 
Solution looks like a crutch, so it requires additional study of RxCpp.

### DIFF
--- a/irohad/main/application.cpp
+++ b/irohad/main/application.cpp
@@ -67,7 +67,6 @@ void Irohad::init() {
   initWsvRestorer();
   restoreWsv();
 
-  initPeerQuery();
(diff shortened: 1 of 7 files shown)


lebdron:
@muratovv @luckychess please check the latest commit

lebdron:
Simulator lock fix is applied in https://github.com/hyperledger/iroha/pull/1243 , other changes should be moved to a separate pull request.

Issue #1183, opened 2018-04-04
Resolve licensing issues discovered by Fossology scan


tkuhrt:
## Context

Hyperledger Iroha source files were scanned by The Linux Foundation utilizing Fossology to determine license for files contained in the following repositories:
  * https://github.com/hyperledger/iroha.git (develop branch)
  * https://github.com/hyperledger/iroha-api.git (docs branch)
  * https://github.com/hyperledger/iroha-scala.git (master branch)
  * https://github.com/hyperledger/iroha-ed25519.git (master branch)
  * https://github.com/hyperledger/iroha-dotnet.git (master branch)

From the Hyperledger Charter (https://www.hyperledger.org/about/charter) Section 13, all files that are included in Hyperledger projects should either be licensed with the Apache License, Version 2.0 for source code or Creative Commons Attribution 4.0 International License for documentation.

## Expected Behavior

All files should have license text included either directly or via a LICENSE file in the directory for files that cannot be modified to include a comment with the license text (think binary files).

## Actual Behavior

See results of scan: 
[iroha-2018-04.xlsx](https://github.com/hyperledger/iroha/files/1876860/iroha-2018-04.xlsx)

  * The most notable finding is at iroha/docs/source/conf.py. This file inaccurately describes Iroha's license as "CC-BY-NC-3.0". In addition to being the wrong license, this is a noncommercial-use-only license. If this file hasn't been modified or removed, it should be promptly changed by its author to list Iroha's license as Apache-2.0.
  * There are 2 files (sha3.c, in two directories) with a notice saying that they are partially public domain, but partially based on code from http://keccak.noekeon.org/. I haven't been able to find any details on that site about licenses for this "keccak" software. The developers will want to either confirm what license governs the "keccak" portions of this file, or else find an alternative with a specified open source license.
  * I did not see any references to CC-BY-4.0, Creative Commons Attribution 4.0. This is the Hyperledger project license for documentation files. The team will likely want to add CC-BY-4.0 notices to documentation files, in particular, the top-level README file for each repo. An example of how Fabric does this can be seen at the bottom of https://github.com/hyperledger/fabric/blob/release-1.1/README.md.
  * There are a very small number of files with non-Apache third party license (MIT, Zlib and public domain). These will likely not be particularly concerning, but will still be subject to GB approval as license exceptions.

## Possible Fix

[...]

PR #1184, opened 2018-04-04
Fix of licenses

laSinteZ:
### Description of the Change
Fixed license info in our readme and conf.py files according to https://github.com/hyperledger/iroha/issues/1183

### Benefits
Partially eliminating problems with licensing.

### Possible Drawbacks 
We still need to license all other source code files :)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -52,7 +52,7 @@ master_doc = 'index'
 # General information about the project.
 project = u'Iroha'
 documentation = u'Iroha Documentation'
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -53,8 +53,8 @@
 project = u'Iroha'
 documentation = u'Iroha Documentation'
 description = u'Distributed ledger technology platworm, written in C++'
```
tkuhrt: I know this is not part of this change, but `platform` is spelled incorrectly in this line.

PR #1192, opened 2018-04-08
Rework Docker images pull mechanism

bakhtin:

### Description of the Change
Before this PR, Jenkins has been building Docker images prior running every build. The procedure is the following:
- forcibly build Docker image referenced by current commit. In majority of cases Docker reuses image layers local cache and simply tags already existing image
- in case any changes were introduced in Docker file, the image is rebuilt and used while building Iroha
- in case the current build is recognized as Master or Develop branch commit, resulting Docker image is uploaded to Dockerhub
The problem is that in the very beginning of Dockerfile we run `apt-get update` command followed by `apt-get upgrade`

```
RUN apt-get update; \
    apt-get -y upgrade; \
    apt-get -y --no-install-recommends install apt-utils software-properties-common; \
    apt-get -y clean
[snip]
```

Docker decides whether the `RUN` section is in cache or not totally depends on the similarity of commands, i.e. if no changes were introduced into a `RUN` section, Docker will reuse local cache. This works fine as long as we do not need to update `apt` cache. And sometimes we must do it because repositories are referenced by DNS names and not by IP addresses. However, those resolved domain names are stored in APT cache as IP addresses, which may become unavailable over time.
This leads to a situation when modified Dockerfile, that was added with additional packages installed from APT repositories cannot install them, as IP addresses not available any more. This requires APT cache to be updated with `apt-get update` command. And as Docker reuses local cache this operation will never run.
We can force Docker to always update cache before every build, but this leads to rebuilding the whole image, because `apt-get update` is the very first command in Dockerfile

This PR introduces a new mechanism for building Docker images during pipeline run. We compare Dockerfile of two consecutive commits: current and the previous one. If they differ, then simply rebuild an an image. Otherwise, reuse the image already built in the previous commit.

### Benefits
Docker images that are built during pipeline run will now avoid stale APT repositories, while still benefit from already built images during previous pipeline runs.

### Possible Drawbacks 
This may increase pipeline run time in cases a Dockerfile has been changed as we now do not use local layer cache, but rebuilding the entire image. But doing so allows to avoid stale APT repositories, which would have failed a build anyways.


.jenkinsci/docker-pull-or-build.groovy
```diff
@@ -0,0 +1,49 @@
...
+  if (diffExitCode == 0) {
+    return false
+  }
+  return true
+}
+
+def dockerPullOrUpdate() {
+  def platform = sh(script: 'uname -m', returnStdout: true).trim()
+  def commit = sh(script: "echo ${BRANCH_NAME} | md5sum | cut -c 1-8", returnStdout: true).trim()
+  if (remoteFilesDiffer("https://raw.githubusercontent.com/hyperledger/iroha/${env.GIT_COMMIT}/docker/develop/${platform}/Dockerfile", "https://raw.githubusercontent.com/hyperledger/iroha/${env.GIT_PREVIOUS_COMMIT}/docker/develop/${platform}/Dockerfile")) {
```
tyvision: what if try to write it in a manner like:
```
if ( remoteFilesDiffer( "https://raw.githubusercontent.com/hyperledger/iroha/${env.GIT_COMMIT}/docker/develop/${platform}/Dockerfile",
                        "https://raw.githubusercontent.com/hyperledger/iroha/${env.GIT_PREVIOUS_COMMIT}/docker/develop/${platform}/Dockerfile")) {
```
which is more readable.

tyvision: Also, what if there is no previous commits? `GIT_PREVIOUS_COMMIT = null`, or `GIT_PREVIOUS_COMMIT = GIT_COMMIT`?
In case of `null` - there might be errors from `curl` which might fail the whole pipeline. 

Regarding `diff` - in this case it will return non-zero exit code thus new docker image will be built and then uploaded to the dockerhub.

.jenkinsci/docker-pull-or-build.groovy
```diff
@@ -0,0 +1,49 @@
...
+  }
+  else {
+    // reuse develop branch Docker image
+    if (BRANCH_NAME == 'develop') {
+      iC = docker.image("hyperledger/iroha:${platform}-develop")
+      iC.pull()
+    }
+    else {
+      // first commit in this branch or Dockerfile modified
+      if (remoteFilesDiffer("https://raw.githubusercontent.com/hyperledger/iroha/${env.GIT_COMMIT}/docker/develop/${platform}/Dockerfile", "https://raw.githubusercontent.com/hyperledger/iroha/develop/docker/develop/${platform}/Dockerfile")) {
```
tyvision: same as above - maybe write each passing argument on different lines?

.jenkinsci/docker-pull-or-build.groovy
bakhtin: According to [Jenkins GIt plugin](https://wiki.jenkins.io/display/JENKINS/Git+Plugin#GitPlugin-Environmentvariables), `GIT_PREVIOUS_COMMIT` equals `GIT_COMMIT` in case it is the first commit to that branch

lebdron: It seems that number of conditional branches can be reduced, e.g. now it is 
```
if (differ in branch)
  if (develop)
    build image; push
  else
    build image
```
can be simplified to
```
if (differ in branch)
  build image
  if (develop)
    push
```

PR #1193, opened 2018-04-09
Use pubkey as identifier in SignatureSetType

l4l:
### Description of the Change

ed25519 can have different signature for the same public key, so it necessary to guarantee uniqueness of each public key in signature set


neewy:
This fix should contain test with text data based on modified version of ed25519

l4l:
It is not really trivial to write the test for that case, as long as:
1. requires modification of iroha-ed25519
2. or there's a need somehow skip timeout check (that is not trivial at all)

l4l:
Good idea, thanks! I will finish it soon

neewy:
Any progress? I've noticed you have added test in commit above, is it sufficient to review this fix?

l4l:
Yes, for now only that test can be done without significant projects changes and I think better review&merge it asap, then think about other components (though, they will work as well, as they have code duplication actually)

PR #1194, opened 2018-04-09
Change namespace for permissions

luckychess:
### Description of the Change

Move permissions from iroha::model to shared_model::permissions

### Benefits

Permissions.hpp was moved to shared model but namespace still was old (iroha::model). Now it's fixed.

### Possible Drawbacks 

None.

### DIFF
--- a/irohad/execution/impl/command_executor.cpp
+++ b/irohad/execution/impl/command_executor.cpp
@@ -21,8 +21,8 @@
 
 #include "execution/common_executor.hpp"
 #include "interfaces/commands/command.hpp"
-#include "validators/permissions.hpp"
 #include "utils/amount_utils.hpp"
+#include "validators/permissions.hpp"
 
 namespace iroha {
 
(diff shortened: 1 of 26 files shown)


shared_model/validators/permissions.hpp
```diff
@@ -21,8 +21,8 @@
 #include <set>
 #include <string>
 
-namespace iroha {
-  namespace model {
+namespace shared_model {
```
Alexey-N-Chernyshov: please rename safe guard as well

irohad/execution/impl/command_executor.cpp
```diff
@@ -506,15 +506,17 @@ namespace iroha {
     // any asset
     return creator_account_id == command.accountId()
         and checkAccountRolePermission(
-                creator_account_id, queries, model::can_add_asset_qty);
+                creator_account_id,
+                queries,
+                shared_model::permissions::can_add_asset_qty);
```
l4l: maybe just use using-namespace thing?

shared_model/validators/field_validator.cpp
```diff
@@ -18,8 +18,8 @@
 #include "validators/field_validator.hpp"
 #include <boost/algorithm/string_regex.hpp>
 #include <boost/format.hpp>
-#include "permissions.hpp"
 #include "cryptography/crypto_provider/crypto_verifier.hpp"
+#include "permissions.hpp"
```
l4l: Use full path

PR #1196, opened 2018-04-10
Use rxcpp for OrderingServiceImpl proposal timeout

l4l:
### Description of the Change

Previously OrderingService and OrderingGate tests was using timeouts for handling its cases. That way is pretty bad as long as tests may occasionally fail.
That pr reworks OrderingServiceImpl to use rxcpp::observable for handling proposal timeouts.

### Benefits

More stable (and probably faster) ordering tests, simpler OrderingServiceImpl structure

### DIFF
--- a/irohad/ordering/impl/ordering_service_impl.cpp
+++ b/irohad/ordering/impl/ordering_service_impl.cpp
@@ -30,21 +30,21 @@ namespace iroha {
     OrderingServiceImpl::OrderingServiceImpl(
         std::shared_ptr<ametsuchi::PeerQuery> wsv,
         size_t max_size,
-        size_t delay_milliseconds,
+        rxcpp::observable<Rep> proposal_timeout,
         std::shared_ptr<network::OrderingServiceTransport> transport,
         std::shared_ptr<ametsuchi::OrderingServicePersistentState>
             persistent_state)
         : wsv_(wsv),
           max_size_(max_size),
-          delay_milliseconds_(delay_milliseconds),
           transport_(transport),
           persistent_state_(persistent_state),
           is_finished(false) {
-      updateTimer();
       log_ = logger::log("OrderingServiceImpl");
-
       // restore state of ordering service from persistent storage
       proposal_height = persistent_state_->loadProposalHeight().value();
+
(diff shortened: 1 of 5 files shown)


l4l:
Does that behavior work before? Seems that there's a mutex deadlock in OS' dtor introduced previously

x3medima17:
@l4l On develop branch I could not reproduce this behaviour

l4l:
@x3medima17 Seems that there was another, that the unsubscribe was called in dtor and thus the lock happened, fixed

irohad/ordering/impl/ordering_service_impl.cpp
```diff
@@ -114,9 +115,8 @@ namespace iroha {
     }
 
     OrderingServiceImpl::~OrderingServiceImpl() {
-      std::lock_guard<std::mutex> lock(m);
       is_finished = true;
-      handle.unsubscribe();
+      std::lock_guard<std::mutex> lock(m);
```
x3medima17: Why do you lock mutex after changing variable state?

l4l: Consider the following case:
1. updateTimer's thread is at
https://github.com/hyperledger/iroha/blob/526d5360547cf879a0e603df8d68eb00823b3c58/irohad/ordering/impl/ordering_service_impl.cpp#L109
2. Dtor's thread is at
https://github.com/hyperledger/iroha/blob/526d5360547cf879a0e603df8d68eb00823b3c58/irohad/ordering/impl/ordering_service_impl.cpp#L118
3. Then the updateTimer comes to the play and skips the if-condition and paused at
https://github.com/hyperledger/iroha/blob/526d5360547cf879a0e603df8d68eb00823b3c58/irohad/ordering/impl/ordering_service_impl.cpp#L112
4. Then thread is switched and dtor proceed to execution flip `is_finished` and perform `queue_` destruction
5. Thread switch to the updateTimer and we get UB here

neewy:
Since you have added "candidate for closing tag" it is not clear why. Is there any impediment? Does this change still make sense to add?

l4l:
Decided to rework that fixes and add them after #1268

PR #1199, opened 2018-04-10
GetAccount returns kv only with keys

victordrobny:

### Description of the Change
GetAccount query returns only keys in kv storage.

### Benefits
No value returns. It allows making query responses smaller.


victordrobny:
Docs are fixed. Python bindings works.

PR #1200, opened 2018-04-10
Remove txCounter from Iroha

victordrobny:

### Description of the Change
Remove tx counter from Iroha


neewy:
Next time do not forget to put whole description for PR

neewy:
@victordrobny @lebdron 

> docs directory should also be updated, but probably in a separate pull request.

If you can do it here I see no problem updating information related to txCounter

lebdron:
@neewy the idea was to separate changes to code and documentation, but it should be OK since the changes to documentation are not large. 
https://github.com/hyperledger/iroha-ios also needs to be updated

PR #1204, opened 2018-04-11
Feature/amount different precision

victordrobny:
### Description of the Change
Command executor and amount + and - operators were reworked in order in order to process amounts with different precisions. 

### Benefits
It is possible to operate with amounts with different precision.


l4l:
That may lead to uint256 overflow, but not sure
May you add integration test as well? (similar to `Uint256DestOverflow`)

victordrobny:
Amount overflowing is checked in wsv_command_validate_execute_test. Why is it necessary to make integration test?

l4l:
They are similar cases, but not the same. Case with different precision has additional multiplication, thus much easier to get overflow

irohad/execution/impl/command_executor.cpp
```diff
@@ -62,16 +62,18 @@ namespace iroha {
...
           command_name);
     }
-    auto precision = asset.value()->precision();
 
-    if (command->amount().precision() != precision) {
+    auto precision = asset.value()->precision();
+    if (command->amount().precision() > precision) {
       return makeExecutionError(
-          (boost::format("precision mismatch: expected %d, but got %d")
+          (boost::format("command pecision is greater than asset precision: "
```
nickaleks: typo "pecision"

irohad/execution/impl/command_executor.cpp
```diff
@@ -354,14 +358,16 @@ namespace iroha {
           command_name);
     }
     auto precision = asset.value()->precision();
-
-    if (command->amount().precision() != precision) {
+    if (command->amount().precision() > precision) {
       return makeExecutionError(
-          (boost::format("precision mismatch: expected %d, but got %d")
+          (boost::format("command pecision is greater than asset precision: "
```
nickaleks: Same typo

irohad/execution/impl/command_executor.cpp
```diff
@@ -417,23 +425,29 @@ namespace iroha {
               .str(),
           command_name);
     }
-    // Precision for both wallets
     auto precision = asset.value()->precision();
-    if (command->amount().precision() != precision) {
+    if (command->amount().precision() > precision) {
       return makeExecutionError(
-          (boost::format("precision %d is wrong") % precision).str(),
+          (boost::format("command pecision is greater than asset precision: "
```
nickaleks: Same typo

shared_model/utils/amount_utils.hpp
```diff
@@ -31,19 +33,17 @@ iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
...
           const shared_model::interface::Amount &b) {
-  // check precisions
-  if (a.precision() != b.precision()) {
-    return iroha::expected::makeError(
-        std::make_shared<std::string>("precision mismatch"));
-  }
-  if (a.intValue() + b.intValue() < a.intValue()
-      || a.intValue() + b.intValue() < b.intValue()) {
+  auto max_precision = std::max(a.precision(), b.precision());
+  auto val_a = a.intValue() * (int)std::pow(10, max_precision - a.precision());
```
nickaleks: Why cast to int?

shared_model/utils/amount_utils.hpp
```diff
@@ -31,19 +33,17 @@ iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
...
+  auto val_b = b.intValue() * (int)std::pow(10, max_precision - b.precision());
```
nickaleks: One of these multiplications is not needed since max_precision is either a.precision() or b.precision()

shared_model/utils/amount_utils.hpp
```diff
@@ -31,19 +33,17 @@ iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
...
+  if (val_a < a.intValue() || val_b < b.intValue() || val_a + val_b < val_a
+      || val_a + val_b < val_b) {
     return iroha::expected::makeError(
         std::make_shared<std::string>("addition overflows"));
   }
   return shared_model::builder::AmountBuilderWithoutValidator()
```
nickaleks: Why use builder without validator?

shared_model/utils/amount_utils.hpp
```diff
@@ -58,22 +58,50 @@ iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
...
-      .precision(a.precision())
-      .intValue(a.intValue() - b.intValue())
+      .precision(max_precision)
+      .intValue(val_a - val_b)
       .build();
 }
 
+/**
+ * Make amount with bigger precision
+ * Otherwise nullopt is returned
```
nickaleks: Result is returned here, not optional

shared_model/utils/amount_utils.hpp
```diff
@@ -58,22 +58,50 @@ iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
...
+ * @param a amount
+ * @param b right term
+ */
+iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
```
nickaleks: You can use PolymorphicResult to avoid writing shared_ptr

shared_model/utils/amount_utils.hpp
```diff
@@ -58,22 +58,50 @@ iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
...
+                        std::shared_ptr<std::string>>
+makePrecision(const shared_model::interface::Amount &amount,
```
nickaleks: Why is this function called makePrecision when it makes amount?

shared_model/utils/amount_utils.hpp
```diff
@@ -58,22 +58,50 @@ iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
...
+          const int new_precision) {
```
nickaleks: do we need const here?

test/module/irohad/model/command_validate_execute_test.cpp
```diff
@@ -1395,6 +1409,47 @@ TEST_F(TransferAssetTest, InvalidWhenNoAssetDuringValidation) {
   ASSERT_NO_THROW(checkErrorCase(validateAndExecute()));
 }
 
+TEST_F(TransferAssetTest, InvalidWhenWrongPrecision) {
```
nickaleks: Please add given, when, then comments

test/module/shared_model/amount_utils_test.cpp
```diff
@@ -0,0 +1,147 @@
...
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+
+#include "utils/amount_utils.hpp"
+
+class AmountTest : public testing::Test {};
+
+TEST_F(AmountTest, PlusTest) {
```
nickaleks: Please add comments to tests

test/module/shared_model/amount_utils_test.cpp
```diff
@@ -0,0 +1,147 @@
...
+TEST_F(AmountTest, MinusTest) {
+  iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
+                          std::shared_ptr<std::string>>
+      a = shared_model::builder::DefaultAmountBuilder::fromString("1234.567");
+
+  iroha::expected::Result<std::shared_ptr<shared_model::interface::Amount>,
+                          std::shared_ptr<std::string>>
+      b = shared_model::builder::DefaultAmountBuilder::fromString("100");
+
+  a.match(
```
nickaleks: Can you use | operator here?

shared_model/utils/amount_utils.hpp
victordrobny: There is no operator with double

shared_model/utils/amount_utils.hpp
victordrobny: Because of it can create invalid amount in terms of stateless validation(which is equal to 0)

shared_model/utils/amount_utils.hpp
victordrobny: Well, what is the problem with const here?

shared_model/utils/amount_utils.hpp
nickaleks: Can you make a function which does static cast on std::pow result to remove duplication?

test/module/shared_model/amount_utils_test.cpp
nickaleks: I believe there exists overload of | for functions which do not return anything

victordrobny: Yes, but you don't have an option to get values from results(if I get it correct)

nickaleks: In result_fixture.hpp we have some convenience functions for tests. Maybe something from there can be used to reduce nesting

nickaleks:
Please move split amount_utils to header and cpp files.

shared_model/utils/amount_utils.hpp
```diff
@@ -18,62 +18,92 @@
...
           const shared_model::interface::Amount &b) {
-  // check precisions
-  if (a.precision() != b.precision()) {
-    return iroha::expected::makeError(
-        std::make_shared<std::string>("precision mismatch"));
-  }
-  if (a.intValue() + b.intValue() < a.intValue()
-      || a.intValue() + b.intValue() < b.intValue()) {
+  auto max_precision = std::max(a.precision(), b.precision());
+  auto val_a = a.intValue() * pow(10, max_precision - a.precision());
```
nickaleks: pow will overflow here if the difference is greater than 9

shared_model/utils/amount_utils.hpp
```diff
@@ -18,62 +18,92 @@
...
 operator+(const shared_model::interface::Amount &a,
           const shared_model::interface::Amount &b) {
-  // check precisions
-  if (a.precision() != b.precision()) {
-    return iroha::expected::makeError(
-        std::make_shared<std::string>("precision mismatch"));
-  }
-  if (a.intValue() + b.intValue() < a.intValue()
-      || a.intValue() + b.intValue() < b.intValue()) {
+  auto max_precision = std::max(a.precision(), b.precision());
```
nickaleks: Consider making the code block which computes values for the same precisions into a separate function, since you do the same thing below

shared_model/utils/impl/amount_utils_impl.cpp
```diff
@@ -0,0 +1,121 @@
+/**
```
nickaleks: Not really sure amount_utils_impl is a good name here. Amount utils does define any virtual methods

shared_model/utils/amount_utils.hpp
```diff
@@ -18,74 +18,46 @@
 #ifndef IROHA_AMOUNT_UTILS_HPP
 #define IROHA_AMOUNT_UTILS_HPP
 
+#include <boost/multiprecision/cpp_int.hpp>
+
+#include "builders/default_builders.hpp"
 #include "builders/protobuf/common_objects/proto_amount_builder.hpp"
+#include "common/result.hpp"
+
+boost::multiprecision::uint256_t increaseValuePrecision(
```
nickaleks: Should it be in a global namespace?

shared_model/utils/amount_utils.hpp
```diff
@@ -18,74 +18,46 @@
...
+
+/**
+ * Make amount with bigger precision
+ * Result is returned
+ * @param a amount
+ * @param b right term
+ */
+iroha::expected::PolymorphicResult<shared_model::interface::Amount, std::string>
+makeAmountWithPrecision(const shared_model::interface::Amount &amount,
+                        const int new_precision);
```
nickaleks: why const here?

shared_model/utils/impl/amount_utils_impl.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * @param a left term
+ * @param b right term
+ */
+iroha::expected::PolymorphicResult<shared_model::interface::Amount, std::string>
+operator-(const shared_model::interface::Amount &a,
+          const shared_model::interface::Amount &b) {
+  // check if a greater than b
+  if (a.intValue() < b.intValue()) {
+    return iroha::expected::makeError(
+        std::make_shared<std::string>("minuend is smaller than subtrahend"));
```
nickaleks: Consider printing a & b as well

shared_model/utils/impl/amount_utils_impl.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "utils/amount_utils.hpp"
+
+const boost::multiprecision::uint256_t ten = 10;
```
nickaleks: Do we need this constant? It only used once in increaseValuePrecision

victordrobny: Yes, it is used only once

shared_model/utils/amount_utils.hpp
victordrobny: It is immutable anyway

shared_model/utils/amount_utils.hpp
victordrobny: It is immutable anyway

test/module/irohad/model/command_validate_execute_test.cpp
```diff
@@ -497,36 +505,42 @@ TEST_F(SubtractAssetQuantityTest, InvalidWhenNoRoles) {
...
  * @given SubtractAssetQuantity
- * @when arguments amount is zero
+ * @when arguments amount precision is invalid
  * @then executor will be failed
  */
-TEST_F(SubtractAssetQuantityTest, InvalidWhenZeroAmount) {
-  // Amount is zero
-  Amount amount(0);
+TEST_F(SubtractAssetQuantityTest, InvalidWhenWrongPrecision) {
+  // Amount is with wrong precision (must not greater than 2)
```
nickaleks: must not be

PR #1206, opened 2018-04-11
Feature/tests ordering service impl

luckychess:
### Description of the Change

Add 2 tests for OrderingServiceImpl.

### Benefits

More tests, more coverage, more existing and potential problems revealed.

### Possible Drawbacks 

ConcurrentGenerateProposal is not passing at the moment and disabled at the moment. Its fix will be done in another task.

### DIFF
--- a/irohad/ordering/impl/ordering_service_impl.cpp
+++ b/irohad/ordering/impl/ordering_service_impl.cpp
@@ -39,12 +39,12 @@ namespace iroha {
           delay_milliseconds_(delay_milliseconds),
           transport_(transport),
           persistent_state_(persistent_state),
-          is_finished(false) {
+          is_finished_(false) {
       updateTimer();
       log_ = logger::log("OrderingServiceImpl");
 
       // restore state of ordering service from persistent storage
-      proposal_height = persistent_state_->loadProposalHeight().value();
+      proposal_height_ = persistent_state_->loadProposalHeight().value();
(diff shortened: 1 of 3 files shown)


l4l:
doesn't seem to be a pr goal

luckychess:
@x3medima17 Good point! It works without problems on my machine. I will disable it too.
@l4l What do you mean?

irohad/ordering/impl/ordering_service_impl.cpp
```diff
@@ -104,23 +103,24 @@ namespace iroha {
     }
 
     void OrderingServiceImpl::updateTimer() {
-      std::lock_guard<std::mutex> lock(m);
-      if (is_finished) {
+      //log_->warn("updateTimer called");
```
l4l: Remove plz, or leave with `->debug`

irohad/ordering/impl/ordering_service_impl.hpp
```diff
@@ -122,17 +122,17 @@ namespace iroha {
        * Proposal counter of expected proposal. Should be number of blocks in
        * the ledger + 1.
        */
-      size_t proposal_height;
+      std::atomic<size_t> proposal_height_;
```
l4l: Don't need to be atomic, already handled by mutex in `updateTimer`

PR #1209, opened 2018-04-12
add transaction creator check

Solonets:

### Description of the Change

Adds check whether a user can query a transaction
https://soramitsu.atlassian.net/browse/IR-1039

### Benefits

Don't return transaction which user doesn't have access to

### Possible Drawbacks 

-

### DIFF
--- a/irohad/model/impl/query_execution.cpp
+++ b/irohad/model/impl/query_execution.cpp
@@ -314,17 +314,21 @@ QueryProcessingFactory::executeGetAccountTransactions(
 
 QueryProcessingFactory::QueryResponseBuilderDone
 iroha::model::QueryProcessingFactory::executeGetTransactions(
-    const shared_model::interface::GetTransactions &query) {
-  const std::vector<shared_model::crypto::Hash> &hashes =
-      query.transactionHashes();
+    const shared_model::interface::GetTransactions &q,
+    const shared_model::interface::types::AccountIdType &accountId) {
+  const std::vector<shared_model::crypto::Hash> &hashes = q.transactionHashes();
 
   auto transactions = _blockQuery->getTransactions(hashes);
 
   std::vector<shared_model::proto::Transaction> txs;
+  bool can_get_all =
+      checkAccountRolePermission(accountId, *_wsvQuery, can_get_all_txs);
   transactions.subscribe([&](const auto &tx) {
     if (tx) {
-      txs.push_back(
-          *std::static_pointer_cast<shared_model::proto::Transaction>(*tx));
+      auto proto_tx =
+          *std::static_pointer_cast<shared_model::proto::Transaction>(*tx);
+      if (can_get_all or proto_tx.creatorAccountId() == accountId)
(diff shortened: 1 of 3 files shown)


irohad/model/query_execution.hpp
```diff
@@ -125,7 +125,8 @@ namespace iroha {
           const shared_model::interface::GetAccountTransactions &query);
 
       QueryResponseBuilderDone executeGetTransactions(
-          const shared_model::interface::GetTransactions &query);
+          const shared_model::interface::Query &query,
```
l4l: Strange parameter in signature

irohad/model/impl/query_execution.cpp
```diff
@@ -313,17 +313,21 @@ QueryProcessingFactory::executeGetAccountTransactions(
...
   std::vector<shared_model::proto::Transaction> txs;
+  bool can_get_all = checkAccountRolePermission(
+      query.creatorAccountId(), *_wsvQuery, can_get_all_txs);
   transactions.subscribe([&](const auto &tx) {
     if (tx) {
-      txs.push_back(
-          *std::static_pointer_cast<shared_model::proto::Transaction>(*tx));
+      auto proto_tx = *std::static_pointer_cast<shared_model::proto::Transaction>(*tx);
+      if (can_get_all or proto_tx.creatorAccountId() == query.creatorAccountId()  )
+        txs.push_back(proto_tx);
```
l4l: I'm not sure it is an issue, but prefer use `clone` explicitly (cloneable.hpp) and declare var with ref aka `auto &proto_tx`

test/integration/acceptance/get_transactions_test.cpp
```diff
@@ -221,7 +221,7 @@ TEST_F(GetTransactions, InexistentHash) {
  * @then TransactionsResponse with no transactions
  * TODO(@l4l) 02/01/18 Should be enabled after resolving IR-1039
```
l4l: remove todo plz

irohad/model/impl/query_execution.cpp
```diff
@@ -313,17 +313,21 @@ QueryProcessingFactory::executeGetAccountTransactions(
...
+  bool can_get_all =
+      checkAccountRolePermission(accountId, *_wsvQuery, can_get_all_txs);
   transactions.subscribe([&](const auto &tx) {
     if (tx) {
-      txs.push_back(
-          *std::static_pointer_cast<shared_model::proto::Transaction>(*tx));
+      auto proto_tx =
+          *std::static_pointer_cast<shared_model::proto::Transaction>(*tx);
+      if (can_get_all or proto_tx.creatorAccountId() == accountId)
+        txs.push_back(proto_tx);
```
l4l: Still concerned about that line, ping @lebdron

lebdron: `proto_tx` is pushed with lvalue, so copy ctor is called, looks fine to me.

PR #1224, opened 2018-04-15
Feature/grpc logging

l4l:
### Description of the Change

RPC errors are not logged, the pr changes that by moving logger from children of `AsyncGrpcClient` to the parent. Also adds some additional logging.

Note, that grpc errors can be observed via **GRPC_TRACE**/**GRPC_VERBOSITY** env variables, but they mostly are not human-friendly

### Benefits

More logging

### Possible Drawbacks 

None

### DIFF
--- a/irohad/consensus/yac/transport/impl/network_impl.cpp
+++ b/irohad/consensus/yac/transport/impl/network_impl.cpp
@@ -30,9 +30,9 @@ namespace iroha {
     namespace yac {
       // ----------| Public API |----------
 
-      NetworkImpl::NetworkImpl() {
-        log_ = logger::log("YacNetwork");
-      }
+      NetworkImpl::NetworkImpl()
+          : network::AsyncGrpcClient<google::protobuf::Empty>(
+                logger::log("YacNetwork")) {}
 
       void NetworkImpl::subscribe(
           std::shared_ptr<YacNetworkNotifications> handler) {
--- a/irohad/network/impl/async_grpc_client.hpp
+++ b/irohad/network/impl/async_grpc_client.hpp
@@ -32,7 +32,9 @@ namespace iroha {
     template <typename Response>
     class AsyncGrpcClient {
      public:
-      AsyncGrpcClient() : thread_(&AsyncGrpcClient::asyncCompleteRpc, this) {}
+      AsyncGrpcClient(logger::Logger &&log)
+          : thread_(&AsyncGrpcClient::asyncCompleteRpc, this),
+            log_(std::move(log)) {}
 
       /**
        * Listen to gRPC server responses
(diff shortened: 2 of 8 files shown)


x3medima17:
PR description does not follow the template :(

irohad/network/impl/async_grpc_client.hpp
```diff
@@ -42,7 +44,9 @@ namespace iroha {
         auto ok = false;
         while (cq_.Next(&got_tag, &ok)) {
           auto call = static_cast<AsyncClientCall *>(got_tag);
-
+          if (!call->status.ok()) {
```
x3medima17: Use `not` instead of `!`

irohad/network/impl/async_grpc_client.hpp
```diff
@@ -32,7 +32,9 @@ namespace iroha {
     template <typename Response>
     class AsyncGrpcClient {
      public:
-      AsyncGrpcClient() : thread_(&AsyncGrpcClient::asyncCompleteRpc, this) {}
+      AsyncGrpcClient(logger::Logger &&log)
+          : thread_(&AsyncGrpcClient::asyncCompleteRpc, this),
+            log_(std::move(log)) {}
```
x3medima17: Why do you call `std::move` on rvalue ref?

l4l: because of [c++ typing magic](https://ideone.com/31oVDm)

irohad/consensus/yac/transport/impl/network_impl.cpp
```diff
@@ -30,9 +30,9 @@ namespace iroha {
     namespace yac {
       // ----------| Public API |----------
 
-      NetworkImpl::NetworkImpl() {
-        log_ = logger::log("YacNetwork");
-      }
+      NetworkImpl::NetworkImpl()
+          : network::AsyncGrpcClient<google::protobuf::Empty>(
```
luckychess: Looks a bit unclear. Why do we need to change this line?

irohad/network/impl/async_grpc_client.hpp
```diff
@@ -32,7 +32,9 @@ namespace iroha {
     template <typename Response>
     class AsyncGrpcClient {
      public:
-      AsyncGrpcClient() : thread_(&AsyncGrpcClient::asyncCompleteRpc, this) {}
+      explicit AsyncGrpcClient(logger::Logger &&log)
+          : thread_(&AsyncGrpcClient::asyncCompleteRpc, this),
+            log_(std::move(log)) {}
```
luckychess: Same as before - it's not clear why not to create logger just here.

l4l: Related classes are children of `AsyncGrpcClient`, so they share exactly same logger. But now `AsyncGrpcClient` issues are logged related to the child, e.g issue at `asyncCompleteRpc` may have prefix of child (for example `NetworkImpl`) or something more meaningful, than just `AsyncGrpcClient`

PR #1229, opened 2018-04-16
Fix permission bug

kamilsa:

### Description of the Change

Fixes problem of missed grantable permissions in proto file. 

Bug example:
User "userone@domain" wants to grant user "usertwo@domain" to add key value information. To do so "userone@domaiun" needed to grant "can_set_my_account_detail" permission to "usertwo@permission". But to do so user "userone@domain" needs permission "can_grant_can_set_my_quorum" which was missing before in proto files although existed in iroha.

### Benefits

Includes missing permissions to proto file.

### Possible Drawbacks 

None

### Usage Examples or Tests

tx-example.py has been updated to demonstrate granting permission to set account detail to another user, set it, and check if detail has been set.


example/python/tx-example.py
```diff
@@ -188,11 +210,26 @@ def get_account_asset():
 
     print(query_response)
 
+def get_user_detail():
+    # get detail of the user
```
nickaleks: Not very informative comment :)

example/python/tx-example.py
```diff
@@ -188,11 +210,26 @@ def get_account_asset():
...
+    query = query_builder.creatorAccountId(creator) \
+        .createdTime(current_time) \
+        .queryCounter(11) \
```
nickaleks: Why is it 11?

muratovv: If you want to get unique query counter, maybe use a static variable with increment instead?

example/python/tx-example.py
```diff
@@ -156,6 +156,28 @@ def tx4():
     send_tx(tx, key_pair)
     print_status_streaming(tx)
 
+def tx5():
```
muratovv: Better to add documentation string with a description of use case. The best choice here use `given, when, then` notation. Same with the next example.
Also, the name looks doubtful. Maybe change it with more suitable and nice for a user?

kamilsa: It is not a test. We don't have checks on expected behaviour

PR #1233, opened 2018-04-16
Fix artifacts uploading on unsuccessful build

bakhtin:
Signed-off-by: bakhtin <a@bakhtin.net>

### Description of the Change
This PR fixes the case, when CI build artifacts could have been uploaded on unsuccessful build or not uploaded at all (e.g., build was aborted and artifacts were not generated)

### Benefits
More stable and consistent process of uploading artifacts

### Possible Drawbacks 
None


bakhtin:
`currentBuild.result` could potentially be `null`. Consider using `currentBuild.currentResult` instead

PR #1240, opened 2018-04-17
Fix/on absent genesis block

muratovv:
### Description of the Change
Fix segfault when the ledger doesn't contain blocks.

### Benefits
Minus one segfault in the project.
Also, bring little fix plus improvement for `itf::config_helper`: remove redundant include and logging for pg credentials.

### Possible Drawbacks 
Absent test for checking the new behavior. And I don't see a theoretical opportunity to write it.
So, reviewers should check it manually.

### DIFF
--- a/test/framework/config_helper.cpp
+++ b/test/framework/config_helper.cpp
@@ -17,8 +17,8 @@
 
 #include "framework/config_helper.hpp"
 
-#include <cstdlib>
 #include <sstream>
+#include "logger/logger.hpp"
 
 namespace integration_framework {
   std::string getPostgresCredsOrDefault(const std::string &default_conn) {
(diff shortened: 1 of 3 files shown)


test/framework/config_helper.cpp
```diff
@@ -32,6 +32,7 @@ namespace integration_framework {
       std::stringstream ss;
       ss << "host=" << pg_host << " port=" << pg_port << " user=" << pg_user
          << " password=" << pg_pass;
+      logger::log("ITF")->info("Postgres credentials: {}", ss.str());
```
Alexey-N-Chernyshov: mb use logger format instead of string stream?
`logger::log("ITF")->info("Postgres credentials: host={}, port={}", pg_host, pg_port);` and so on

PR #1243, opened 2018-04-17
Ordering gate discards outdated proposals and keeps new ones

nickaleks:
### Description of the Change

This PR changes the logic of the ordering gate. Now proposals which are older than the last committed block will be discarded from the queue.

### Benefits

Old proposals are not passed down the pipeline when newer commit arrives

### Possible Drawbacks 

None

### DIFF
--- a/irohad/main/impl/ordering_init.hpp
+++ b/irohad/main/impl/ordering_init.hpp
@@ -43,6 +43,7 @@ namespace iroha {
        * Init effective realisation of ordering gate (client of ordering
        * service)
        * @param network_address - address of ordering service
+       * @param block_query - block store to get last block height
        */
       auto createGate(std::shared_ptr<OrderingGateTransport>,
                       std::shared_ptr<ametsuchi::BlockQuery>);
@@ -69,7 +70,8 @@ namespace iroha {
        * @param loop - handler of async events
        * @param max_size - limitation of proposal size
        * @param delay_milliseconds - delay before emitting proposal
-       * @return effective realisation of OrderingGate
+       * @param block_query - block store to get last block height
+       * @return efficient implementation of OrderingGate
        */
       std::shared_ptr<ordering::OrderingGateImpl> initOrderingGate(
           std::shared_ptr<ametsuchi::PeerQuery> wsv,
--- a/irohad/main/impl/ordering_init.hpp
+++ b/irohad/main/impl/ordering_init.hpp
@@ -43,7 +43,7 @@ namespace iroha {
        * Init effective realisation of ordering gate (client of ordering
        * service)
        * @param transport - object which will be notified
-       * about incoming proposals
+       * about incoming proposals and send transactions
        * @param block_query - block store to get last block height
        */
       auto createGate(std::shared_ptr<OrderingGateTransport> transport,
--- a/irohad/main/impl/ordering_init.hpp
+++ b/irohad/main/impl/ordering_init.hpp
@@ -42,11 +42,12 @@ namespace iroha {
       /**
        * Init effective realisation of ordering gate (client of ordering
        * service)
-       * @param network_address - address of ordering service
+       * @param transport - object which will be notified
+       * about incoming proposals
        * @param block_query - block store to get last block height
        */
-      auto createGate(std::shared_ptr<OrderingGateTransport>,
-                      std::shared_ptr<ametsuchi::BlockQuery>);
+      auto createGate(std::shared_ptr<OrderingGateTransport> transport,
+                      std::shared_ptr<ametsuchi::BlockQuery> block_query);
 
       /**
        * Init ordering service
--- a/irohad/main/impl/ordering_init.hpp
+++ b/irohad/main/impl/ordering_init.hpp
@@ -18,6 +18,7 @@
 #ifndef IROHA_ORDERING_INIT_HPP
 #define IROHA_ORDERING_INIT_HPP
 
+#include "ametsuchi/block_query.hpp"
 #include "ametsuchi/peer_query.hpp"
 #include "logger/logger.hpp"
 #include "ordering/impl/ordering_gate_impl.hpp"
@@ -43,7 +44,8 @@ namespace iroha {
        * service)
        * @param network_address - address of ordering service
        */
-      auto createGate(std::shared_ptr<OrderingGateTransport>);
+      auto createGate(std::shared_ptr<OrderingGateTransport>,
+                      std::shared_ptr<ametsuchi::BlockQuery>);
 
       /**
        * Init ordering service
@@ -74,7 +76,8 @@ namespace iroha {
           size_t max_size,
           std::chrono::milliseconds delay_milliseconds,
           std::shared_ptr<ametsuchi::OrderingServicePersistentState>
-              persistent_state);
+              persistent_state,
+          std::shared_ptr<ametsuchi::BlockQuery> block_query);
 
       std::shared_ptr<ordering::OrderingServiceImpl> ordering_service;
       std::shared_ptr<ordering::OrderingGateImpl> ordering_gate;
--- a/irohad/ordering/impl/ordering_gate_impl.cpp
+++ b/irohad/ordering/impl/ordering_gate_impl.cpp
@@ -67,19 +67,22 @@ namespace iroha {
                                   })
                                   .start_with(last_block_height_);
 
+      auto subscribe = [&](auto merge_strategy) {
+        pcs_subscriber_ = merge_strategy(net_proposals_.get_observable())
+                              .subscribe([this](const auto &t) {
+                                this->tryNextRound(std::get<1>(t));
+                              });
+      };
+
       if (run_async_) {
-        pcs_subscriber_ = net_proposals_.get_observable()
-                              .combine_latest(rxcpp::synchronize_new_thread(),
-                                              top_block_height)
-                              .subscribe([this](const auto &t) {
-                                this->tryNextRound(std::get<1>(t));
-                              });
+       subscribe([&top_block_height](auto observable) {
+         return observable.combine_latest(rxcpp::synchronize_new_thread(),
+                         top_block_height);
+       });
       } else {
-        pcs_subscriber_ = net_proposals_.get_observable()
-                              .combine_latest(top_block_height)
-                              .subscribe([this](const auto &t) {
-                                this->tryNextRound(std::get<1>(t));
-                              });
+        subscribe([&top_block_height](auto observable) {
+          return observable.combine_latest(top_block_height);
+        });
       }
     }
 
--- a/irohad/ordering/impl/ordering_gate_impl.cpp
+++ b/irohad/ordering/impl/ordering_gate_impl.cpp
@@ -60,12 +60,13 @@ namespace iroha {
         const iroha::network::PeerCommunicationService &pcs) {
       log_->info("setPcs");
 
-      auto top_block_height = pcs.on_commit()
-                                  .transform([](const Commit &block) {
-                                    // find height of last commited block
-                                    return block.as_blocking().last()->height();
-                                  })
-                                  .start_with(last_block_height_);
+      auto top_block_height =
+          pcs.on_commit()
+              .transform([](const Commit &commit) {
+                // find height of last commited block
+                return commit.as_blocking().last()->height();
+              })
+              .start_with(last_block_height_);
 
       auto subscribe = [&](auto merge_strategy) {
         pcs_subscriber_ = merge_strategy(net_proposals_.get_observable())
@@ -75,10 +76,10 @@ namespace iroha {
       };
 
       if (run_async_) {
-       subscribe([&top_block_height](auto observable) {
-         return observable.combine_latest(rxcpp::synchronize_new_thread(),
-                         top_block_height);
-       });
+        subscribe([&top_block_height](auto observable) {
+          return observable.combine_latest(rxcpp::synchronize_new_thread(),
+                                           top_block_height);
+        });
       } else {
         subscribe([&top_block_height](auto observable) {
           return observable.combine_latest(top_block_height);
@@ -95,18 +96,18 @@ namespace iroha {
     }
 
     void OrderingGateImpl::tryNextRound(
-        shared_model::interface::types::HeightType last_block_height_) {
-      log_->info("TryNextRound");
+        shared_model::interface::types::HeightType last_block_height) {
+      log_->debug("TryNextRound");
       std::shared_ptr<shared_model::interface::Proposal> next_proposal;
       while (proposal_queue_.try_pop(next_proposal)) {
         // check for old proposal
-        if (next_proposal->height() < last_block_height_ + 1) {
-          log_->info("Old proposal, discarding");
+        if (next_proposal->height() < last_block_height + 1) {
+          log_->debug("Old proposal, discarding");
           continue;
         }
         // check for new proposal
-        if (next_proposal->height() > last_block_height_ + 1) {
-          log_->info("Proposal newer than last block, keeping in queue");
+        if (next_proposal->height() > last_block_height + 1) {
+          log_->debug("Proposal newer than last block, keeping in queue");
           proposal_queue_.push(next_proposal);
           break;
         }
--- a/irohad/ordering/impl/ordering_gate_impl.cpp
+++ b/irohad/ordering/impl/ordering_gate_impl.cpp
@@ -16,6 +16,7 @@
  */
 
 #include <utility>
+#include <tuple>
 
 #include "ordering/impl/ordering_gate_impl.hpp"
 
@@ -34,10 +35,11 @@ namespace iroha {
 
     OrderingGateImpl::OrderingGateImpl(
         std::shared_ptr<iroha::network::OrderingGateTransport> transport,
-        shared_model::interface::types::HeightType initial_height)
+        shared_model::interface::types::HeightType initial_height,
+        bool run_async)
         : transport_(std::move(transport)),
           last_block_height_(initial_height),
-          log_(logger::log("OrderingGate")) {}
+          log_(logger::log("OrderingGate")), run_async_(run_async) {}
 
     void OrderingGateImpl::propagateTransaction(
         std::shared_ptr<const shared_model::interface::Transaction>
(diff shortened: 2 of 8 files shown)


l4l:
`ordering_gate_service_test (SEGFAULT)` :(

irohad/ordering/impl/ordering_gate_impl.cpp
```diff
@@ -51,27 +57,57 @@ namespace iroha {
...
     void OrderingGateImpl::setPcs(
         const iroha::network::PeerCommunicationService &pcs) {
-      pcs_subscriber_ = pcs.on_commit().subscribe([this](auto) {
-        // TODO: 05/03/2018 @muratovv rework behavior of queue with respect to
-        // block height IR-1042
-        unlock_next_.store(true);
-        this->tryNextRound();
+      log_->info("setPcs");
+
+      auto tap = pcs.on_commit().transform([](const Commit &block) {
```
lebdron: tap -> top_block_height?

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -61,7 +62,6 @@ class OrderingGateTest : public ::testing::Test {
   }
 
   void SetUp() override {
-    thread = std::thread([this] {
       grpc::ServerBuilder builder;
```
lebdron: This should be clang-formatted.

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -73,32 +73,22 @@ class OrderingGateTest : public ::testing::Test {
       auto address = "0.0.0.0:" + std::to_string(port);
       // Initialize components after port has been bind
       transport = std::make_shared<OrderingGateTransportGrpc>(address);
-      gate_impl = std::make_shared<OrderingGateImpl>(transport);
+      gate_impl = std::make_shared<OrderingGateImpl>(transport, 1, false);
```
lebdron: Maybe define `1` and `false` as some constants here, because they are used multiple times in this file?

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -16,6 +16,7 @@
  */
 #include <grpc++/grpc++.h>
 #include <gtest/gtest.h>
+#include <iostream>
```
lebdron: This include should not be required now.

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -239,37 +237,146 @@ TEST(OrderingGateQueueBehaviour, ReceiveUnordered) {
   EXPECT_CALL(*pcs, on_commit())
       .WillOnce(Return(commit_subject.get_observable()));
 
-  auto pushCommit = [&] {
+  auto pushCommit = [&](auto height) {
+    std::cout << "Commiting\n";
```
lebdron: Can be removed now, as test assertions cover the behaviour.

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -239,37 +237,146 @@ TEST(OrderingGateQueueBehaviour, ReceiveUnordered) {
...
-  OrderingGateImpl ordering_gate(transport);
+  OrderingGateImpl ordering_gate(transport, 1, false);
   ordering_gate.setPcs(*pcs);
 
   auto pushProposal = [&](auto height) {
     ordering_gate.onProposal(std::make_shared<shared_model::proto::Proposal>(
         TestProposalBuilder().height(height).build()));
   };
 
+  std::condition_variable cv;
```
lebdron: Not required because sync ordering gate is used.

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -239,37 +237,146 @@ TEST(OrderingGateQueueBehaviour, ReceiveUnordered) {
...
+  std::shared_ptr<OrderingGateTransport> transport =
+      std::make_shared<MockOrderingGateTransport>();
+
+  std::shared_ptr<MockPeerCommunicationService> pcs =
+      std::make_shared<MockPeerCommunicationService>();
+  rxcpp::subjects::subject<Commit> commit_subject;
+  EXPECT_CALL(*pcs, on_commit())
+      .WillOnce(Return(commit_subject.get_observable()));
+
+  auto pushCommit = [&](auto height) {
```
lebdron: It is the same as in previous test, maybe create a fixture?

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -239,37 +237,146 @@ TEST(OrderingGateQueueBehaviour, ReceiveUnordered) {
...
+    commit_subject.get_subscriber().on_next(rxcpp::observable<>::just(
+        std::static_pointer_cast<shared_model::interface::Block>(
+            std::make_shared<shared_model::proto::Block>(
+                TestBlockBuilder().height(height).build()))));
+  };
+
+  OrderingGateImpl ordering_gate(transport, 1, false);
+  ordering_gate.setPcs(*pcs);
+
+  auto pushProposal = [&](auto height) {
```
lebdron: Same for code duplication.

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -239,37 +237,146 @@ TEST(OrderingGateQueueBehaviour, ReceiveUnordered) {
...
+  std::shared_ptr<OrderingGateTransport> transport =
+      std::make_shared<MockOrderingGateTransport>();
+
+  std::shared_ptr<MockPeerCommunicationService> pcs =
+      std::make_shared<MockPeerCommunicationService>();
+  rxcpp::subjects::subject<Commit> commit_subject;
+  EXPECT_CALL(*pcs, on_commit())
+      .WillOnce(Return(commit_subject.get_observable()));
+
+  auto pushCommit = [&](auto height) {
```
lebdron: Same for code duplication, please make a fixture.

test/module/irohad/ordering/ordering_gate_service_test.cpp
```diff
@@ -16,6 +16,7 @@
  */
 
 #include <gtest/gtest.h>
+#include <iostream>
```
lebdron: Not required now because behaviour is covered by assertions.

test/module/irohad/ordering/ordering_gate_service_test.cpp
```diff
@@ -54,7 +55,7 @@ class OrderingGateServiceTest : public ::testing::Test {
     EXPECT_CALL(*pcs_, on_commit())
         .WillRepeatedly(Return(commit_subject_.get_observable()));
     gate_transport = std::make_shared<OrderingGateTransportGrpc>(address);
-    gate = std::make_shared<OrderingGateImpl>(gate_transport);
+    gate = std::make_shared<OrderingGateImpl>(gate_transport, 1);
```
lebdron: Async ordering gate should not be required here.

test/module/irohad/ordering/ordering_gate_service_test.cpp
```diff
@@ -105,19 +106,24 @@ class OrderingGateServiceTest : public ::testing::Test {
       size_t times) {
     auto wrapper = make_test_subscriber<CallExact>(gate->on_proposal(), times);
     gate->on_proposal().subscribe([this](auto) {
+      std::cout << "on proposal subscriber 1\n";
```
lebdron: `std::cout` can be removed.

irohad/ordering/impl/ordering_gate_impl.hpp
```diff
@@ -76,26 +80,34 @@ namespace iroha {
       /**
        * Try to push proposal for next consensus round
```
lebdron: Add description for the new parameter.

irohad/ordering/impl/ordering_gate_impl.hpp
```diff
@@ -55,8 +57,10 @@ namespace iroha {
     class OrderingGateImpl : public network::OrderingGate,
                              public network::OrderingGateNotification {
      public:
-      explicit OrderingGateImpl(
-          std::shared_ptr<iroha::network::OrderingGateTransport> transport);
+      OrderingGateImpl(
+          std::shared_ptr<iroha::network::OrderingGateTransport> transport,
```
lebdron: Parameter description will be useful here.

irohad/ordering/impl/ordering_gate_impl.cpp
```diff
@@ -51,27 +57,57 @@ namespace iroha {
...
+        // find height of last commited block
+        return block.as_blocking().last()->height();
+      }).start_with(last_block_height_);
+
+      if (run_async_) {
+        pcs_subscriber_ =
+            net_proposals_.get_observable()
+                .combine_latest(rxcpp::synchronize_new_thread(), tap)
+                .subscribe([this](auto t) {
```
lebdron: `const auto &` should be used to avoid tuple copy. same below.

irohad/main/impl/ordering_init.hpp
```diff
@@ -43,7 +44,8 @@ namespace iroha {
        * service)
        * @param network_address - address of ordering service
```
lebdron: Documentation does not reflect method signature.

irohad/main/impl/ordering_init.hpp
```diff
@@ -74,7 +76,8 @@ namespace iroha {
           size_t max_size,
           std::chrono::milliseconds delay_milliseconds,
           std::shared_ptr<ametsuchi::OrderingServicePersistentState>
-              persistent_state);
+              persistent_state,
+          std::shared_ptr<ametsuchi::BlockQuery> block_query);
```
lebdron: Documentation does not reflect method signature.

irohad/ordering/impl/ordering_gate_impl.hpp
```diff
@@ -76,26 +80,34 @@ namespace iroha {
...
-      std::atomic_bool unlock_next_{true};
+      std::mutex proposal_mutex_;
 
       /// queue with all proposals received from ordering service
       tbb::concurrent_priority_queue<
           std::shared_ptr<shared_model::interface::Proposal>,
           ProposalComparator>
           proposal_queue_;
 
+      // last commited block height
```
lebdron: `///` should be used, because `//` is simple comment while `///` is documentation, so that doxygen parses it correctly.

irohad/ordering/impl/ordering_gate_impl.cpp
```diff
@@ -51,27 +58,60 @@ namespace iroha {
...
-      pcs_subscriber_ = pcs.on_commit().subscribe([this](auto) {
-        // TODO: 05/03/2018 @muratovv rework behavior of queue with respect to
-        // block height IR-1042
-        unlock_next_.store(true);
-        this->tryNextRound();
+      log_->info("setPcs");
 
-      });
+      auto top_block_height = pcs.on_commit()
+                                  .transform([](const Commit &block) {
```
lebdron: block -> commit ? because argument is not a single block

test/module/irohad/ordering/ordering_gate_service_test.cpp
```diff
@@ -109,15 +109,17 @@ class OrderingGateServiceTest : public ::testing::Test {
       cv.notify_one();
     });
     gate->on_proposal().subscribe([this](auto proposal) {
+      std::cout << "on proposal subscriber 2\n";
```
lebdron: Not required now.

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -161,26 +159,50 @@ TEST_F(OrderingGateTest, ProposalReceivedByGateWhenSent) {
   ASSERT_TRUE(wrapper.validate());
 }
 
+class QueueBehaviorTest : public ::testing::Test {
+ public:
+  QueueBehaviorTest(): ordering_gate(transport, 1, false) {};
+
+  void SetUp() override {
+    std::shared_ptr<OrderingGateTransport> transport =
```
lebdron: It will not be possible to set any expects on this `transport` pointer because gmock type is erased, so OrderingGateTransport -> MockOrderingGateTransport

lebdron: Also fixture member is not used, so it also will not be possible to set any expects in tests. Please reuse fixture member.

irohad/ordering/impl/ordering_gate_impl.cpp
```diff
@@ -51,27 +58,60 @@ namespace iroha {
...
-        std::shared_ptr<shared_model::interface::Proposal> next_proposal;
-        proposal_queue_.try_pop(next_proposal);
-        log_->info("Pass the proposal to pipeline");
+    void OrderingGateImpl::tryNextRound(
+        shared_model::interface::types::HeightType last_block_height_) {
+      log_->info("TryNextRound");
+      std::shared_ptr<shared_model::interface::Proposal> next_proposal;
+      while (proposal_queue_.try_pop(next_proposal)) {
+        // check for old proposal
+        if (next_proposal->height() < last_block_height_ + 1) {
```
l4l: `(next_proposal->height() <= last_block_height_)`?

irohad/ordering/impl/ordering_gate_impl.cpp
```diff
@@ -51,27 +58,60 @@ namespace iroha {
...
+          log_->info("Old proposal, discarding");
+          continue;
+        }
+        // check for new proposal
+        if (next_proposal->height() > last_block_height_ + 1) {
+          log_->info("Proposal newer than last block, keeping in queue");
+          proposal_queue_.push(next_proposal);
```
l4l: Should it be pushed again? Seems that sending to observable is enough, doesn't it?

irohad/ordering/impl/ordering_gate_impl.hpp
```diff
@@ -20,9 +20,11 @@
 
 #include "network/ordering_gate.hpp"
 
-#include <atomic>
 #include <tbb/concurrent_priority_queue.h>
+#include <atomic>
```
l4l: Unordered headers :(

irohad/simulator/impl/simulator.cpp
```diff
@@ -65,8 +66,10 @@ namespace iroha {
         const shared_model::interface::Proposal &proposal) {
       log_->info("process proposal");
       // Get last block from local ledger
-      block_queries_->getTopBlocks(1).as_blocking().subscribe(
-          [this](auto block) { last_block = block; });
+      block_queries_->getTopBlocks(1)
+          .subscribe_on(rxcpp::observe_on_new_thread())
```
l4l: May you explain why should it be on a separate thread?

test/module/irohad/ordering/ordering_gate_test.cpp
```diff
@@ -137,6 +128,13 @@ TEST_F(OrderingGateTest, ProposalReceivedByGateWhenSent) {
   auto wrapper = make_test_subscriber<CallExact>(gate_impl->on_proposal(), 1);
   wrapper.subscribe();
 
+  std::shared_ptr<MockPeerCommunicationService> pcs =
```
l4l: `auto pcs = ..`?

irohad/ordering/impl/ordering_gate_impl.hpp
```diff
@@ -55,8 +57,16 @@ namespace iroha {
     class OrderingGateImpl : public network::OrderingGate,
                              public network::OrderingGateNotification {
      public:
-      explicit OrderingGateImpl(
-          std::shared_ptr<iroha::network::OrderingGateTransport> transport);
+      /**
+       * @param transport - network communication layer
+       * @param initial_height - height of the last block stored on this peer
+       * @param run_async - whether proposals should be handled on
```
l4l: handled <s>on</s>

irohad/ordering/impl/ordering_gate_impl.hpp
l4l: Also that header seems unused

irohad/ordering/impl/ordering_gate_impl.cpp
```diff
@@ -51,27 +58,60 @@ namespace iroha {
...
+      log_->info("TryNextRound");
+      std::shared_ptr<shared_model::interface::Proposal> next_proposal;
+      while (proposal_queue_.try_pop(next_proposal)) {
+        // check for old proposal
+        if (next_proposal->height() < last_block_height_ + 1) {
+          log_->info("Old proposal, discarding");
+          continue;
+        }
+        // check for new proposal
+        if (next_proposal->height() > last_block_height_ + 1) {
```
l4l: why +1?
peer has: 1->2->3
last_block_height_ is 3
next_proposal->height() == 4 should be considered as a new one, shouldn't it?

irohad/ordering/impl/ordering_gate_impl.cpp
```diff
@@ -51,27 +58,60 @@ namespace iroha {
...
     }
 
-    void OrderingGateImpl::tryNextRound() {
-      if (not proposal_queue_.empty() and unlock_next_.exchange(false)) {
-        std::shared_ptr<shared_model::interface::Proposal> next_proposal;
-        proposal_queue_.try_pop(next_proposal);
-        log_->info("Pass the proposal to pipeline");
+    void OrderingGateImpl::tryNextRound(
+        shared_model::interface::types::HeightType last_block_height_) {
+      log_->info("TryNextRound");
```
l4l: debug maybe? it's triggered quite often

irohad/simulator/impl/simulator.cpp
nickaleks: It is a bug, related to calling simulator from the same thread which waits on simulator, which leads to hanging since we are blocked. You can ask Andrei for details

irohad/ordering/impl/ordering_gate_impl.cpp
nickaleks: No, because if we had commit for block 3, then it means that proposal 4 should be passed to the pipeline since it can be applied.

irohad/simulator/impl/simulator.cpp
nickaleks: Otherwise, we need to synchronize access from separate threads.

irohad/ordering/impl/ordering_gate_impl.cpp
nickaleks: I think it is more clear that we need proposal which is one older than last commited block

irohad/ordering/impl/ordering_gate_impl.cpp
nickaleks: It should not be sent to since it cannot be applied yet. That's why we have break after.

lebdron:
@l4l lock you have mentioned is required because network -- grpc server, is used in the test

PR #1247, opened 2018-04-18
fix synchronization issues

Solonets:

### Description of the Change

Retry to synchronize storage if the previous attempt was unsuccessful

### Benefits

Before, if the block query failed (nodes hadn't applied block yet), the polling node doesn't make any attempt to synchronize again. In this PR it is fixed. 

### Possible Drawbacks 

None

### DIFF
--- a/irohad/consensus/yac/impl/yac_gate_impl.hpp
+++ b/irohad/consensus/yac/impl/yac_gate_impl.hpp
@@ -49,6 +49,11 @@ namespace iroha {
                     std::shared_ptr<network::BlockLoader> block_loader,
                     uint64_t delay);
         void vote(const shared_model::interface::Block &) override;
+        /**
+         * method called when commit recived
+         * assumes to retrieve a block eventually
+         * @return observable with the Block commited
+         */
         rxcpp::observable<std::shared_ptr<shared_model::interface::Block>>
         on_commit() override;
 
--- a/test/module/irohad/consensus/yac/yac_gate_test.cpp
+++ b/test/module/irohad/consensus/yac/yac_gate_test.cpp
@@ -208,3 +208,57 @@ TEST_F(YacGateTest, LoadBlockWhenDifferentCommit) {
 
   ASSERT_TRUE(gate_wrapper.validate());
 }
+
+/**
+ * @given yac gate
+ * @when recives new commit different to the one it voted for
+ * @then polls nodes for the block with corresponding hash until it succeed,
+ * (reciving none on the first poll)
+ */
+TEST_F(YacGateTest, LoadBlockWhenDifferentCommitFailFirst) {
+  // Vote for block => receive different block => load committed block
+
+  // make blocks
+  EXPECT_CALL(*block_creator, on_block())
+      .WillOnce(Return(rxcpp::observable<>::just(expected_block)));
+
+  // make hash from block
+  EXPECT_CALL(*hash_provider, makeHash(_)).WillOnce(Return(expected_hash));
+
+  // generate order of peers
+  EXPECT_CALL(*peer_orderer, getOrdering(_))
+      .WillOnce(Return(ClusterOrdering::create({mk_peer("fake_node")})));
+
+  EXPECT_CALL(*hash_gate, vote(expected_hash, _)).Times(1);
+
+  // expected values
+  expected_hash = YacHash("actual_proposal", "actual_block");
+
+  message.hash = expected_hash;
+
+  commit_message = CommitMessage({message});
+  expected_commit = rxcpp::observable<>::just(commit_message);
+
+  // yac consensus
+  EXPECT_CALL(*hash_gate, on_commit()).WillOnce(Return(expected_commit));
+
+  // convert yac hash to model hash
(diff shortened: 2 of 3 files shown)


lebdron:
Please write a test which will cover new behaviour, and a method description which states the assumption that the block will be eventually received from the network.

irohad/consensus/yac/impl/yac_gate_impl.hpp
```diff
@@ -50,6 +50,7 @@ namespace iroha {
                     uint64_t delay);
         void vote(const shared_model::interface::Block &) override;
         rxcpp::observable<std::shared_ptr<shared_model::interface::Block>>
+        //method assumes to retrieve a block eventually
```
l4l: Quite strange to find a comment in the middle of fn signature

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -208,3 +208,50 @@ TEST_F(YacGateTest, LoadBlockWhenDifferentCommit) {
 
   ASSERT_TRUE(gate_wrapper.validate());
 }
+TEST_F(YacGateTest, LoadBlockWhenDifferentCommitFailFirst) {
+  // Vote for block => receive different block => load committed block
```
l4l: Use gwt comment style, like here:
https://github.com/hyperledger/iroha/blob/ed665deb84aba285e2dfc217188bba8a9cc5ce2e/test/module/irohad/consensus/yac/supermajority_checker_test.cpp#L33-L37

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -208,3 +208,50 @@ TEST_F(YacGateTest, LoadBlockWhenDifferentCommit) {
 
   ASSERT_TRUE(gate_wrapper.validate());
 }
+TEST_F(YacGateTest, LoadBlockWhenDifferentCommitFailFirst) {
```
l4l: The test looks really similar to the previous, maybe it's better refactor those (e.g extract same routines) and simplify the tests code?

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -208,3 +208,55 @@ TEST_F(YacGateTest, LoadBlockWhenDifferentCommit) {
 
   ASSERT_TRUE(gate_wrapper.validate());
 }
+/**
+ * @given yac gate
+ * @when recives new commit different to the one it voted for
+ * @then polls nodes for the block with corresponding hash untill it succeed
+ */
+TEST_F(YacGateTest, LoadBlockWhenDifferentCommitFailFirst) {
```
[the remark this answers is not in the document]
Solonets: oh, yeah, mb

irohad/consensus/yac/impl/yac_gate_impl.hpp
```diff
@@ -49,6 +49,10 @@ namespace iroha {
                     std::shared_ptr<network::BlockLoader> block_loader,
                     uint64_t delay);
         void vote(const shared_model::interface::Block &) override;
+        /**
```
lebdron: Please format this as it is done https://github.com/hyperledger/iroha/pull/1252/files#diff-602f45921f2f767d49042cbc8221d796R27 here, with `@return`.

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -208,3 +208,55 @@ TEST_F(YacGateTest, LoadBlockWhenDifferentCommit) {
 
   ASSERT_TRUE(gate_wrapper.validate());
 }
+/**
+ * @given yac gate
+ * @when recives new commit different to the one it voted for
+ * @then polls nodes for the block with corresponding hash untill it succeed
```
lebdron: This description also applies to the previous test. It should be stated that it does not succeed with download on the first try.

lebdron: untill -> until

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -208,3 +208,55 @@ TEST_F(YacGateTest, LoadBlockWhenDifferentCommit) {
 
   ASSERT_TRUE(gate_wrapper.validate());
 }
+/**
```
lebdron: Can be separated from previous test with new line.

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -208,3 +208,55 @@ TEST_F(YacGateTest, LoadBlockWhenDifferentCommit) {
...
+ * @given yac gate
+ * @when recives new commit different to the one it voted for
+ * @then polls nodes for the block with corresponding hash untill it succeed
+ */
+TEST_F(YacGateTest, LoadBlockWhenDifferentCommitFailFirst) {
+  // Vote for block => receive different block => load committed block
```
lebdron: This comment should not be required, because there is given, when, then description.

PR #1252, opened 2018-04-19
One small step for Iroha — one giant transaction for a client

neewy:
### Description of the Change

Recently we had a problem with message limit in Iroha, but this PR might fix it. Following suggestion from @lebdron here http://nanxiao.me/en/message-length-setting-in-grpc/ helped a lot.

### Benefits

Now Iroha can handle message size of 2^31 bytes.

### Possible Drawbacks 

No configurable size

tx_heavy_data VeryLargeTxWithManyCommands


l4l:
May it led to some problems, like bandwidth increasing or dos vector? Would be nice to double check all drawbacks that can appear

neewy:
@l4l biggest drawback is that right now limit is not reasonable. We might consider limiting size of message later, but so far keeping only 4MB is really not so much

luckychess:
@neewy Nice pull request! It's completely my fault that I missed this approach before.
Two notes:
1) FieldValidator::value_size still equals to 4MB. I think it should be bumped, what do you think? And if you agree please also bump value passed to generateData() in OneLargeTx (or it will fail).
2) There are still 2 tests fail (you left them disabled). We should to create a task to fix them.

test/integration/acceptance/tx_heavy_data.cpp
```diff
@@ -162,34 +160,36 @@ TEST_F(HeavyTransactionTest, DISABLED_ManyLargeTxes) {
...
+  // TODO: IR-1264 neewy 19/04/2018
+  std::this_thread::sleep_for(std::chrono::seconds(10));
+  itf.done();
 }
 
 /**
- * NOTE: test is disabled until fix of
- * https://soramitsu.atlassian.net/browse/IR-1205 will not be completed.
+ * TODO: disabled until proposal process time in simulator is not optimized
+ * and now it takes more than 5 minutes 19/04/2018 neewy
```
lebdron: Is it also IR-1264?

irohad/torii/query_client.cpp
```diff
@@ -11,6 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#include "network/impl/grpc_channel_builder.hpp"
 #include "torii/query_client.hpp"
```
x3medima17: Reorder please, first include in .cpp file should be the .hpp that you implement

irohad/torii/query_client.cpp
```diff
@@ -19,11 +20,8 @@ namespace torii_utils {
   using iroha::protocol::QueryResponse;
 
   QuerySyncClient::QuerySyncClient(const std::string &ip, size_t port)
-      : ip_(ip),
-        port_(port),
-        stub_(iroha::protocol::QueryService::NewStub(
-            grpc::CreateChannel(ip + ":" + std::to_string(port),
-                                grpc::InsecureChannelCredentials()))) {}
+      : ip_(ip), port_(port), stub_(iroha::network::createClient<iroha::protocol::QueryService>(
```
x3medima17: Broken clang-format

PR #1253, opened 2018-04-19
Feature/rework wsv query command test with shm

Alexey-N-Chernyshov:
### Description of the Change
Rework wsv_query_command_test with shared_model.

### Benefits
One of the last in the tests.

### Possible Drawbacks 
no


test/module/irohad/ametsuchi/wsv_query_command_test.cpp
```diff
@@ -33,12 +31,15 @@ namespace iroha {
...
-        account.quorum = 1;
-        account.json_data = R"({"id@domain": {"key": "value"}})";
+        domain = clone(
+            TestDomainBuilder().domainId("domain").defaultRole(role).build());
+
+        account = clone(TestAccountBuilder()
+                            .domainId(domain->domainId())
+                            .accountId("id@" + domain->domainId())
+                            .quorum(1)
+                            .jsonData(R"({"id@domain": {"key": "value"}})")
```
x3medima17: You use `id@domain` 11 times, I would suggest to define a constant, same for the whole string, you wrote it 3 times

test/module/irohad/ametsuchi/wsv_query_command_test.cpp
```diff
@@ -59,8 +60,8 @@ namespace iroha {
       }
 
       std::string role = "role", permission = "permission";
-      model::Account account;
-      model::Domain domain;
+      std::shared_ptr<shared_model::interface::Account> account;
```
x3medima17: I think unique_ptr would be enough

test/module/irohad/ametsuchi/wsv_query_command_test.cpp
Alexey-N-Chernyshov: Yes, but it is a part of JSON. If I use a constant, it would reduce JSON readability.

PR #1254, opened 2018-04-19
[Fix] Add wrapped exception to ModelCrypto::fromPrivateKey

stinger112:
[IR-1223](https://soramitsu.atlassian.net/browse/IR-1223)

### Description of the Change

This PR restricts the length of  ModelCrypto::fromPrivateKey function to 32 and throws wrapped exception to SWIG.

### Benefits

An exception can be caught by client's code.

### DIFF
--- a/shared_model/bindings/model_crypto.cpp
+++ b/shared_model/bindings/model_crypto.cpp
@@ -28,6 +28,9 @@ namespace shared_model {
 
     crypto::Keypair ModelCrypto::fromPrivateKey(
         const std::string &private_key) {
+      if (private_key.length() != 32)  {
+        throw iroha::BadFormatException("blob_t: input string has incorrect length " + std::to_string(private_key.length()));
+      }
       auto byte_string = iroha::hexstringToBytestring(private_key);
       if (not byte_string) {
         throw std::runtime_error("invalid seed");
--- a/shared_model/bindings/model_crypto.hpp
+++ b/shared_model/bindings/model_crypto.hpp
@@ -37,7 +37,7 @@ namespace shared_model {
 
       /**
        * Creates keypair (ed25519) from provided private key
-       * @param private_key - ed25519 hex-encoded private key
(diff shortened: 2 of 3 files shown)


l4l:
I think it should be done for native model, not for particular bindings

stinger112:
> I think it should be done for native model, not for particular bindings

@l4l, `iroha::hexstringToBytestring` can take strings with random length (it needs to be just even number), but `ModelCrypto::fromPrivateKey` can take only 32 length (as @luckychess said).

l4l:
I mean the wrapper over model should contain as least logic as possible. I think crypto will be refactored soon and will verify all of the sizes internally

shared_model/bindings/model_crypto.cpp
```diff
@@ -28,6 +28,9 @@ namespace shared_model {
 
     crypto::Keypair ModelCrypto::fromPrivateKey(
         const std::string &private_key) {
+      if (private_key.length() != 32)  {
+        throw iroha::BadFormatException("blob_t: input string has incorrect length " + std::to_string(private_key.length()));
+      }
       auto byte_string = iroha::hexstringToBytestring(private_key);
       if (not byte_string) {
         throw std::runtime_error("invalid seed");
```
luckychess: Please change exception type to `std::invalid_argument`. Test for this case will be also nice to see.

shared_model/bindings/model_crypto.cpp
```diff
@@ -28,6 +28,9 @@ namespace shared_model {
 
     crypto::Keypair ModelCrypto::fromPrivateKey(
         const std::string &private_key) {
+      if (private_key.length() != 32)  {
```
luckychess: My bad: size must be equal to 64 obviously (32 bytes means 64-length string when hex-encoded). Please change here and in other places.

lebdron:
Fixed in #1429

PR #1255, opened 2018-04-20
HasTxWithHash for Block Query

nickaleks:
### Description of the Change

Add method to check for the existence of transaction in a blockchain. Apply it in command service to verify that transaction was successfully saved.

### Benefits

Previously transaction was retrieved to check that it was saved, which was doing a lot of unnecessary work. Checking for existence is simpler

### Possible Drawbacks 

None


test/module/irohad/ametsuchi/block_query_test.cpp
```diff
@@ -366,3 +366,26 @@ TEST_F(BlockQueryTest, GetTop2Blocks) {
 
   ASSERT_TRUE(wrapper.validate());
 }
+
+/**
+ * @given block store with 2 blocks totally containing 3 txs created by
```
l4l: I think unrelated info in test comments can be omitted

PR #1179, opened 2018-04-03
[WIP] CI bindings

bakhtin:
Signed-off-by: Artyom Bakhtin <a@bakhtin.net>

### Description of the Change
Bindings for different languages and plarforms:

- [x] Java
- [x] Python

Bindings are available on [artifact server](https://artifact.soramitsu.co.jp/iroha/bindings/)

PR #1258, opened 2018-04-22
Python/Java/Android bindings

bakhtin:

### Description of the Change
CI:
- added Android bindings generation. Based on [this script](https://github.com/hyperledger/iroha/blob/e70649f7288499cd27a66788b06b3cc1706cb416/example/Android/android-build.sh). Dockerfile in `android/Dockerfile` contains instructions for building base image. Iroha itself then checked out into Docker container with necessary changes in `CMakeLists.txt`. Dockerfile supports building for multiple platforms, ABIs and build types (`PLATFORM`, `VERSION` and `BUILD_TYPE_A` variables respectively). Allowed parameters are defined in the beginning of Jenkinsfile as properties.
- rewrite Python and Java bindings generation. Added support for Python 2.7 as suggested by @l4l [here](https://github.com/hyperledger/iroha/pull/1170#pullrequestreview-109064226)
- rewrite [`dockerPullOrUpdate`](https://github.com/hyperledger/iroha/blob/fe0dd0ec77f871a778864fe904fd3ae6ac5695bc/.jenkinsci/docker-pull-or-build.groovy) function to support Docker images required to build bindings. Fix the function exit with an error in case Docker images is unavailable in Dockerhub. It now builds an image instead.

Regarding [forced dependencies resolving](https://github.com/hyperledger/iroha/pull/1170#issuecomment-378348753) by @l4l. [As explained](https://github.com/hyperledger/iroha/pull/1170#issuecomment-378573268) by @luckychess, appropriate protobuf version must be installed.
Regarding [ed25519 packaging](https://github.com/hyperledger/iroha/pull/1170#issuecomment-378348753). `libed25519`  is linked statically and not shown in the `ldd` output

#1179 is no more relevant

### Benefits
Bindings for multiple languages and platforms will now be available without struggling of building it themselves. 

### Possible Drawbacks 
None

### Usage Examples
Bindings building is switched off by default. In order to build it, one should log into [Jenkins](https://jenkins.soramitsu.co.jp), navigate to `develop` branch and click `Build with parameters` on the left. Setup the platform you wish to build the bindings for and start the build.
Upon successful build, bindings will be uploaded to [artifact server](https://artifact.soramitsu.co.jp) into `bindings` folder


.jenkinsci/bindings.groovy
```diff
@@ -1,27 +1,69 @@
...
+  def commit = env.GIT_COMMIT
+  def artifactsPath = sprintf('%1$s/android-bindings-%2$s-%3$s-%4$s-%5$s-%6$s.zip', 
+    [currentPath, "\$PLATFORM", abiVersion, "\$BUILD_TYPE_A", sh(script: 'date "+%Y%m%d"', returnStdout: true).trim(), commit.substring(0,6)])
+  sh """
+    (cd /iroha; git init; git remote add origin https://github.com/hyperledger/iroha.git; \
+    git fetch --depth 1 origin develop; git checkout -t origin/develop)
+  """
+  sh """
+    . /entrypoint.sh; \
+    sed -i.bak "s~find_package(JNI REQUIRED)~#find_package(JNI REQUIRED)~" /iroha/shared_model/bindings/CMakeLists.txt; \
```
l4l: That better be done as separate file, like a patch or so. Otherwise changes in `android-build.sh` require changes here

.jenkinsci/bindings.groovy
```diff
@@ -1,27 +1,69 @@
...
+    sed -i.bak "/target_include_directories(\\\${SWIG_MODULE_irohajava_REAL_NAME} PUBLIC/,+3d" /iroha/shared_model/bindings/CMakeLists.txt; \
+    sed -i.bak "s~swig_link_libraries(irohajava~swig_link_libraries(irohajava \"/protobuf/.build/lib\${PROTOBUF_LIB_NAME}.a\" \"\${NDK_PATH}/platforms/android-$abiVersion/\${ARCH}/usr/\${LIBP}/liblog.so\"~" /iroha/shared_model/bindings/CMakeLists.txt; \
+    sed -i.bak "s~find_library(protobuf_LIBRARY protobuf)~find_library(protobuf_LIBRARY \${PROTOBUF_LIB_NAME})~" /iroha/cmake/Modules/Findprotobuf.cmake; \
+    sed -i.bak "s~find_program(protoc_EXECUTABLE protoc~set(protoc_EXECUTABLE \"/protobuf/host_build/protoc\"~" /iroha/cmake/Modules/Findprotobuf.cmake; \
+    cmake -H/iroha/shared_model -B/iroha/shared_model/build -DCMAKE_SYSTEM_NAME=Android -DCMAKE_SYSTEM_VERSION=$abiVersion -DCMAKE_ANDROID_ARCH_ABI=\$PLATFORM \
+      -DANDROID_NDK=\$NDK_PATH -DCMAKE_ANDROID_STL_TYPE=c++_static -DCMAKE_BUILD_TYPE=\$BUILD_TYPE_A -DTESTING=OFF \
+      -DSHARED_MODEL_DISABLE_COMPATIBILITY=ON -DSWIG_JAVA=ON -DCMAKE_PREFIX_PATH=\$DEPS_DIR
+    """
+  sh "cmake --build /iroha/shared_model/build --target irohajava -- -j${params.PARALLELISM}"
```
l4l: That lines below are the same to `doJavaBindings` ending

.jenkinsci/remote-files-differ.groovy
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env groovy
+
+def remoteFilesDiffer(f1, f2) {
+  sh "curl -L -o /tmp/${env.GIT_COMMIT}/f1 --create-dirs ${f1}"
```
l4l: `-sSL`?
same below

.jenkinsci/debug-build.groovy
```diff
@@ -20,8 +21,11 @@ def doDebugBuild(coverageEnabled=false) {
     + " -e POSTGRES_PASSWORD=${env.IROHA_POSTGRES_PASSWORD}"
     + " --name ${env.IROHA_POSTGRES_HOST}"
     + " --network=${env.IROHA_NETWORK}")
-
-  def iC = dPullOrBuild.dockerPullOrUpdate()
+  def iC = dPullOrBuild.dockerPullOrUpdate("${platform}-develop",
+                                           "https://raw.githubusercontent.com/hyperledger/iroha/${env.GIT_COMMIT}/docker/develop/${platform}/Dockerfile",
```
l4l: Maybe it's possible to use some global var for that?

.jenkinsci/bindings.groovy
```diff
@@ -1,27 +1,69 @@
...
+  def commit = env.GIT_COMMIT
+  def artifactsPath = sprintf('%1$s/python-bindings-%2$s-%3$s-%4$s.zip', 
+    [currentPath, buildType, sh(script: 'date "+%Y%m%d"', returnStdout: true).trim(), commit.substring(0,6)])
+  sh """
+    cmake \
+      -H. \
+      -Bbuild \
+      -DCMAKE_BUILD_TYPE=$buildType \
+      -DSWIG_PYTHON=ON \
+      -DSUPPORT_PYTHON2=ON
```
l4l: The function relates only to the python2 bindings, so maybe rename it accordingly?

.jenkinsci/bindings.groovy
bakhtin: Agree. I think it should look like a single patch that is applicable both for `android-build.sh` and pipeline script. @luckychess what do you think?

.jenkinsci/debug-build.groovy
bakhtin: `iC` is actually "not-so-global". It is used in Linux builds, but not required for MacOS build. `dPullOrBuild.dockerPullOrUpdate()` returns Groovy object, although global variable can only be defined in `environment` section of Jenkinsfile and only of type `string`

.jenkinsci/bindings.groovy
bakhtin: OK, as far as I see current version of CMakeLists file allows to specify only one version of Python that will be supported with generated bindings:
https://github.com/hyperledger/iroha/blob/61969d20d1e8d71c60f4417fe271c30cf2d6dce4/shared_model/bindings/CMakeLists.txt#L47-L51
Am I right? Can we modify it to support Python3 by default, while `SUPPORT_PYTHON2` would really enable support for Python2 and not just switching Python version?

.jenkinsci/debug-build.groovy
l4l: I mean `https://raw.githubusercontent.com/hyperledger/iroha/` substring

.jenkinsci/bindings.groovy
l4l: Yeah, I think it can be done, though directories would change

luckychess: We can run cmake twice: one time with ```-DSUPPORT_PYTHON2=ON``` and another with ```-DSUPPORT_PYTHON2=OFF```.

.jenkinsci/bindings.groovy
luckychess: Sounds good. But I do not like idea of the patch because every minor change of original CMakeLists.txt will break everything. May be a separate bash script - but can it will be called from here?

docker/develop/x86_64/Dockerfile
```diff
@@ -25,15 +25,16 @@ RUN set -e; \
...
         wget curl file unzip gdb iputils-ping vim ccache \
-        gcovr cppcheck doxygen graphviz graphviz-dev; \
+        gcovr cppcheck doxygen graphviz graphviz-dev zip; \
     apt-get -y clean
 
-# install cmake 3.7.2
-RUN set -e; \
-    git clone https://gitlab.kitware.com/cmake/cmake.git /tmp/cmake; \
-    (cd /tmp/cmake ; git checkout 35413bf2c1b33980afd418030af27f184872af6b); \
+# install cmake 3.8.2
```
luckychess: May be 3.10.3 would be better?

.jenkinsci/bindings.groovy
```diff
@@ -1,27 +1,69 @@
...
+  def artifactsPath = sprintf('%1$s/android-bindings-%2$s-%3$s-%4$s-%5$s-%6$s.zip', 
+    [currentPath, "\$PLATFORM", abiVersion, "\$BUILD_TYPE_A", sh(script: 'date "+%Y%m%d"', returnStdout: true).trim(), commit.substring(0,6)])
+  sh """
+    (cd /iroha; git init; git remote add origin https://github.com/hyperledger/iroha.git; \
+    git fetch --depth 1 origin develop; git checkout -t origin/develop)
+  """
+  sh """
+    . /entrypoint.sh; \
+    sed -i.bak "s~find_package(JNI REQUIRED)~#find_package(JNI REQUIRED)~" /iroha/shared_model/bindings/CMakeLists.txt; \
+    sed -i.bak "s~# the include path to jni.h and jni_md.h~SET(CMAKE_SWIG_FLAGS \\\${CMAKE_SWIG_FLAGS} -package \${PACKAGE})~" /iroha/shared_model/bindings/CMakeLists.txt; \
```
luckychess: This doesn't work - ```-package``` should be set before ```swig_add_library``` call (that's actually a bug in the current development branch). My suggestion is to use command
```
sed -i.bak "s~find_package(JNI REQUIRED)~SET(CMAKE_SWIG_FLAGS \${CMAKE_SWIG_FLAGS} -package ${PACKAGE})~" /iroha/shared_model/bindings/CMakeLists.txt
```
instead of this one and previous.

.jenkinsci/bindings.groovy
bakhtin: It could be a bash script. But the thing is that variables used in that script must be properly escaped. Otherwise, Jenkins interprets them as Groovy variables (see `entrypoint.sh` in this commit). With that said, it would require to maintain two separate versions of the script: one for self-building and one for CI. The idea with the patch is more tangible solution as for me.

.jenkinsci/docker-pull-or-build.groovy
```diff
@@ -10,37 +10,51 @@ def remoteFilesDiffer(f1, f2) {
...
-      }
+def buildOptionsString(options) {
+  def s = ''
+  if (options) {
+    options.each { k, v ->
+      s += "--build-arg ${k}=${v} "
     }
+    return s
+  }
+  return ''
```
l4l: Why not just one `return s` here

.jenkinsci/remote-files-differ.groovy
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env groovy
+
+def remoteFilesDiffer(f1, f2) {
+  sh "curl -sSL -o /tmp/${env.GIT_COMMIT}/f1 --create-dirs ${f1}"
+  sh "curl -sSL -o /tmp/${env.GIT_COMMIT}/f2 ${f2}"
+  diffExitCode = sh(script: "diff -q /tmp/${env.GIT_COMMIT}/f1 /tmp/${env.GIT_COMMIT}/f2", returnStatus: true)
+  if (diffExitCode == 0) {
+    return false
+  }
+  return true
```
l4l: D:
`return diffExitCode != 0`?

docker/develop/armv7l/Dockerfile
```diff
@@ -160,6 +164,13 @@ RUN set -e; \
     ldconfig; \
     rm -rf /tmp/rapidjson
 
+# install optional
+RUN set -e; \
+    git clone https://github.com/martinmoene/optional-lite /tmp/optional-lite; \
```
l4l: Ehm, that isn't really used now

PR #1262, opened 2018-04-23
Fix artifacts uploading

bakhtin:

### Description of the Change

- "Debug" build type build followed by "Release" build in CI pipeline generates two sets of artifacts for each build type respectively. This PR disables artifacts generation for "Debug" build type so that they would not be uploaded to artifact server. We only want to have "Release" artifacts published.
- Include TAR packaged artifacts previously generated but not uploaded for Linux builds.
- Replace `cp` with `mv` to save some disk space used by a single build.

### Benefits
- Only "Release" artifacts are uploaded
- TAR packaged binaries for non-Debian OS flavors
- Reduce disk space usage during a CI build

### Possible Drawbacks 
None

### DIFF
--- a/.jenkinsci/release-build.groovy
+++ b/.jenkinsci/release-build.groovy
@@ -46,12 +46,13 @@ def doReleaseBuild() {
     sh "ccache --show-stats"
 
     // copy build package to the volume
-    sh "cp ./build/iroha-*.deb /tmp/${GIT_COMMIT}/iroha.deb"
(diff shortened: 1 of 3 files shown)


.jenkinsci/release-build.groovy
```diff
@@ -46,12 +46,13 @@ def doReleaseBuild() {
     sh "ccache --show-stats"
 
     // copy build package to the volume
```
x3medima17: You say copy, when you actually move

PR #1268, opened 2018-04-24
Rework ordering service event handling

lebdron:
### Description of the Change
Fix issues with heap-use-after-free and `generateProposal` thread working after destructor.
 - `unsubscribe` call now blocks until a running method is completed
 - no more mutex for destructor
 - mutex is used when a transaction is received from the network, so that subject calls are not concurrent
 - queue size is verified when subscriber is called, and not when event is received

### Benefits
More stable implementation

### Possible Drawbacks 
Code duplication in constructor caused by templates in rxcpp, which are only removed after `subscribe`. A possible solution is to pass coordination as template parameter, but this will require some type rework and renaming.

### DIFF
--- a/irohad/ordering/impl/ordering_service_impl.cpp
+++ b/irohad/ordering/impl/ordering_service_impl.cpp
@@ -51,44 +51,32 @@ namespace iroha {
               rxcpp::observe_on_new_thread())
               .map([](auto) { return ProposalEvent::kTimerEvent; });
 
+      auto subscribe = [&](auto merge_strategy) {
+        handle_ = merge_strategy(rxcpp::observable<>::from(
+                                     timer, transactions_.get_observable()))
+                      .subscribe([this](auto &&v) {
+                        auto check_queue = [&] {
+                          switch (v) {
+                            case ProposalEvent::kTimerEvent:
+                              return not queue_.empty();
+                            case ProposalEvent::kTransactionEvent:
+                              return queue_.unsafe_size() >= max_size_;
+                            default:
+                              BOOST_ASSERT_MSG(false, "Unknown value");
+                          }
+                        };
+                        if (check_queue()) {
+                          this->generateProposal();
+                        }
+                      });
+      };
+
       if (is_async) {
-        handle_ =
-            rxcpp::observable<>::from(timer, transactions_.get_observable())
-                .merge(rxcpp::synchronize_new_thread())
-                .subscribe([this](auto &&v) {
-                  auto check_queue = [&] {
-                    switch (v) {
-                      case ProposalEvent::kTimerEvent:
-                        return not queue_.empty();
-                      case ProposalEvent::kTransactionEvent:
-                        return queue_.unsafe_size() >= max_size_;
-                      default:
-                        BOOST_ASSERT_MSG(false, "Unknown value");
-                    }
-                  };
-                  if (check_queue()) {
-                    this->generateProposal();
-                  }
-                });
+        subscribe([](auto observable) {
+          return observable.merge(rxcpp::synchronize_new_thread());
+        });
       } else {
-        handle_ =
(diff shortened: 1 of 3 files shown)


irohad/ordering/impl/ordering_service_impl.hpp
```diff
@@ -122,17 +122,18 @@ namespace iroha {
...
+
+      /// Internal event observable handle
+      rxcpp::composite_subscription handle_;
 
       /**
-       * Set after destruction
+       * Mutex for incoming transactions
        */
-      bool is_finished;
+      std::mutex m_;
```
Alexey-N-Chernyshov: mb rename to mutex_

irohad/ordering/impl/ordering_service_impl.hpp
```diff
@@ -42,20 +41,26 @@ namespace iroha {
...
-     * @param persistent_state - storage for persistent state of ordering
-     * service
      */
     class OrderingServiceImpl : public network::OrderingService {
      public:
+      /**
+       * Constructor
+       * @param wsv interface for fetching peers from world state view
+       * @param max_size maximum size of proposal
+       * @param delay_milliseconds time
```
l4l: timeout for proposal generation

irohad/ordering/impl/ordering_service_impl.cpp
```diff
@@ -33,55 +33,99 @@ namespace iroha {
...
+                .merge(rxcpp::synchronize_new_thread())
+                .subscribe([this](auto &&v) {
+                  auto check_queue = [&] {
+                    switch (v) {
+                      case ProposalEvent::kTimerEvent:
+                        return not queue_.empty();
+                      case ProposalEvent::kTransactionEvent:
+                        return queue_.unsafe_size() >= max_size_;
+                      default:
+                        BOOST_ASSERT_MSG(false, "Unknown value");
```
l4l: I think better say smth to log & return false
Also does the compilers enforce `default` (e.g like not all cases handled or so)? so maybe just remove it

irohad/ordering/impl/ordering_service_impl.hpp
```diff
@@ -77,22 +82,16 @@ namespace iroha {
...
-       * Collect transactions from queue
-       * Passes the generated proposal to publishProposal
-       */
-      void generateProposal() override;
-
-      /**
-       * Method update peers for sending proposal
+       * Events for queue check strategy
        */
+      enum class ProposalEvent { kTransactionEvent, kTimerEvent };
```
l4l: Can be moved to cpp I think

irohad/ordering/impl/ordering_service_impl.cpp
```diff
@@ -33,55 +33,99 @@ namespace iroha {
...
+
+      rxcpp::observable<ProposalEvent> timer =
+          rxcpp::observable<>::interval(
+              std::chrono::milliseconds(delay_milliseconds_),
+              rxcpp::observe_on_new_thread())
+              .map([](auto) { return ProposalEvent::kTimerEvent; });
+
+      if (is_async) {
+        handle_ =
+            rxcpp::observable<>::from(timer, transactions_.get_observable())
```
l4l: I guess it can be moved to function that accepts callback, so might look like:
```cpp
if (is_async)
  subs([](auto s) {
    return s.merge(rxcpp::synchronize_new_thread())
  })
else
  subs([](auto s) {
    return s.merge()
  };
```

irohad/ordering/impl/ordering_service_impl.cpp
lebdron: Why? If a wrong value of enum class somehow appeared in observable, then someone has manually casted a value to enum, which should not happen.

PR #1277, opened 2018-04-27
Add dbnname param

kamilsa:
### Description of the Change

Postgres database can have dbname parame. We can use it to run every test with database in parallel by creating a new instance of database for each test

### Benefits

More flexible database creation

### Possible Drawbacks 

Drop database is not implemented, so after all tests are executed an empty instance of database might be left.

### DIFF
--- a/irohad/ametsuchi/impl/postgres_options.cpp
+++ b/irohad/ametsuchi/impl/postgres_options.cpp
@@ -25,12 +25,12 @@ namespace iroha {
         std::vector<std::string> key_value;
         boost::split(key_value, s, boost::is_any_of("="));
         if (key_value.size() != 2) {
-          return expected::makeError("postgres options parse error");
+          return expected::makeError("postgres options parse error: cannot get param name and value from " + s);
         }
         std::string key = key_value.at(0);
         std::string value = key_value.at(1);
         if (key.empty() or value.empty()) {
-          return expected::makeError("postgres options parse error");
+          return expected::makeError("postgres options parse error: param name or value is empty");
         }
         options.insert({key, value});
         if (key != "dbname") {
--- a/irohad/ametsuchi/impl/postgres_options.cpp
+++ b/irohad/ametsuchi/impl/postgres_options.cpp
@@ -1,0 +1,80 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2018 All Rights Reserved.
+ * http://soramitsu.co.jp
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ametsuchi/impl/postgres_options.hpp"
+#include <boost/algorithm/string.hpp>
+#include <regex>
+
+namespace iroha {
+  namespace ametsuchi {
+
+    expected::Result<PostgresOptions, std::string> PostgresOptions::create(
+        std::string postgres_options) {
+      // parse options
+      std::regex pattern{R"([^ ]+)"};
+      std::set<std::string> tokens{
+          std::sregex_token_iterator{std::begin(postgres_options),
+                                     std::end(postgres_options),
+                                     pattern},
+          std::sregex_token_iterator{}};
+      std::unordered_map<std::string, std::string> options;
+      std::string pg_opt_without_db_name;
+      for (const auto &s : tokens) {
+        std::vector<std::string> key_value;
+        boost::split(key_value, s, boost::is_any_of("="));
+        if (key_value.size() != 2) {
+          return expected::makeError("postgres options parse error");
+        }
+        std::string key = key_value.at(0);
+        std::string value = key_value.at(1);
+        if (key.empty() or value.empty()) {
+          return expected::makeError("postgres options parse error");
+        }
+        options.insert({key, value});
+        if (key != "dbname") {
+          pg_opt_without_db_name += s + " ";
+        }
+      }
+      return expected::makeValue(
+          PostgresOptions(postgres_options, pg_opt_without_db_name, options));
+    }
+
+    PostgresOptions::PostgresOptions(
+        const std::string &pg_opt,
+        const std::string &pg_opt_without_db_name,
+        const std::unordered_map<std::string, std::string> &options_map)
+        : pg_opt_(pg_opt),
+          pg_opt_without_db_name_(pg_opt_without_db_name),
+          options_map_(options_map) {}
+
+    std::string PostgresOptions::optionsString() {
+      return pg_opt_;
+    }
+
+    std::string PostgresOptions::optionsStringWithoutDbName() {
+      return pg_opt_without_db_name_;
+    }
+
+    boost::optional<std::string> PostgresOptions::getOption(
+        const std::string option) {
+      return options_map_.find(option) != options_map_.end()
+          ? options_map_.at(option)
+          : boost::none;
+    }
+
+  }  // namespace ametsuchi
+}  // namespace iroha
--- a/irohad/ametsuchi/impl/postgres_options.cpp
+++ b/irohad/ametsuchi/impl/postgres_options.cpp
@@ -1,18 +1,6 @@
 /**
- * Copyright Soramitsu Co., Ltd. 2018 All Rights Reserved.
- * http://soramitsu.co.jp
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *        http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
  */
 
 #include "ametsuchi/impl/postgres_options.hpp"
--- a/irohad/ametsuchi/impl/postgres_options.cpp
+++ b/irohad/ametsuchi/impl/postgres_options.cpp
@@ -61,18 +61,18 @@ namespace iroha {
           pg_opt_without_db_name_(pg_opt_without_db_name),
           options_map_(options_map) {}
 
-    std::string PostgresOptions::optionsString() {
+    std::string PostgresOptions::optionsString() const {
       return pg_opt_;
     }
 
-    std::string PostgresOptions::optionsStringWithoutDbName() {
+    std::string PostgresOptions::optionsStringWithoutDbName() const {
       return pg_opt_without_db_name_;
     }
 
     boost::optional<std::string> PostgresOptions::getOption(
-        const std::string option) {
+        const std::string &option) const {
       return options_map_.find(option) != options_map_.end()
-          ? options_map_.at(option)
+          ? boost::make_optional(options_map_.at(option))
           : boost::none;
     }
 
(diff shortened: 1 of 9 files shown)


irohad/ametsuchi/impl/postgres_options.cpp
```diff
@@ -0,0 +1,80 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2018 All Rights Reserved.
```
l4l: Use the following instead plz:
```
/**
 * Copyright Soramitsu Co., Ltd. All Rights Reserved.
 */
```

irohad/ametsuchi/impl/postgres_options.cpp
```diff
@@ -0,0 +1,80 @@
...
+      for (const auto &s : tokens) {
+        std::vector<std::string> key_value;
+        boost::split(key_value, s, boost::is_any_of("="));
+        if (key_value.size() != 2) {
+          return expected::makeError("postgres options parse error");
+        }
+        std::string key = key_value.at(0);
+        std::string value = key_value.at(1);
+        if (key.empty() or value.empty()) {
+          return expected::makeError("postgres options parse error");
```
l4l: There's two exactly same messages for different cases. Would be nice to specify both of errors more precisely

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -237,19 +253,30 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
...
+                                   options.optionsStringWithoutDbName());
+        };
+
+        auto ctx_result =
+            initConnections(block_store_dir, options.optionsString());
+        expected::Result<std::shared_ptr<StorageImpl>, std::string> storage;
+        ctx_result.match(
+            [&](expected::Value<ConnectionContext> &ctx) {
+              storage = expected::makeValue(std::shared_ptr<StorageImpl>(
+                  new StorageImpl(block_store_dir,
```
l4l: make_shared

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -237,19 +253,30 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
...
+                                     std::string> {
+        options.getOption("dbname") | [&options](const auto &dbname) {
+          createDatabaseIfNotExist(dbname,
+                                   options.optionsStringWithoutDbName());
+        };
+
+        auto ctx_result =
+            initConnections(block_store_dir, options.optionsString());
+        expected::Result<std::shared_ptr<StorageImpl>, std::string> storage;
+        ctx_result.match(
```
l4l: Update from develop branch and use:
```js
return initConnections(block_store_dir, options.optionsString())
    .and_res(expected::makeValue(..));
```

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -237,19 +253,30 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
...
-                                std::move(ctx.value.pg_lazy),
-                                std::move(ctx.value.pg_nontx))));
-          },
-          [&](expected::Error<std::string> &error) { storage = error; });
-      return storage;
+      return PostgresOptions::create(postgres_options) |
+                 [&block_store_dir](const PostgresOptions &options)
+                 -> expected::Result<std::shared_ptr<StorageImpl>,
+                                     std::string> {
+        options.getOption("dbname") | [&options](const auto &dbname) {
```
l4l: On error case should be a error I guess, so use match

l4l: Or maybe just use in a chain of results
get_dbname (bind)-> create db (and_res)-> initConnection (and_res)-> call ctor

kamilsa: getOption returns optional. If it is empty, i.e. dbname param does not exist, we do not need to create the database

irohad/ametsuchi/impl/storage_impl.cpp
lebdron: I think it does not work because of private constructor.

irohad/ametsuchi/impl/postgres_options.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <boost/algorithm/string.hpp>
+#include <regex>
+
+namespace iroha {
+  namespace ametsuchi {
+
+    expected::Result<PostgresOptions, std::string> PostgresOptions::create(
+        std::string postgres_options) {
+      // parse options
+      std::regex pattern{R"([^ ]+)"};
```
muratovv: This code doesn't look obvious by itself. Provide comments with some example and parsing on it. And maybe we can use boost::spirit instead of this code?

irohad/ametsuchi/impl/storage_impl.hpp
```diff
@@ -44,6 +45,10 @@ namespace iroha {
 
     class StorageImpl : public Storage {
      protected:
+      static void createDatabaseIfNotExist(
```
muratovv: I think this function has to return a result with a status of the operation. Such as:
```
Result<bool, exeption>
``` 
or something like this.

test/module/irohad/ametsuchi/CMakeLists.txt
```diff
@@ -39,6 +39,18 @@ target_link_libraries(kv_storage_test
...
 
+addtest(storage_test storage_test.cpp)
+target_link_libraries(storage_test
+    ametsuchi
+    libs_common
+    )
+
+addtest(postgres_options_test postgres_options_test.cpp)
+target_link_libraries(postgres_options_test
+    ametsuchi
```
muratovv: Looks like tests of parsing shouldn't depend on storage, isn't it?

irohad/ametsuchi/impl/storage_impl.cpp
l4l: Indeed, missed it

test/module/irohad/ametsuchi/postgres_options_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+        ASSERT_TRUE(param2);
+        ASSERT_EQ(param2.value(), "val2");
+      },
+      [](auto) { FAIL() << "Creation of PostgresOptions object failed"; });
+}
+
+/**
+ * @given pg_opt string without non_existing_param
+ * @when PostgresOptions object is created from given pg_opt string
+ * @then PostgresOptions object successfully created @and doesn't contain
```
muratovv: The little issue about `@and`.
Better to separate and statements for readability:
```
@then PostgresOptions object successfully created
AND doesn't contain non_existing_param in it
```

test/module/irohad/ametsuchi/postgres_options_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+      [](iroha::expected::Value<PostgresOptions> options) {
+        ASSERT_FALSE(options.value.getOption("non_existing_param"));
+      },
+      [](auto) { FAIL() << "Creation of PostgresOptions object failed"; });
+}
+
+/**
+ * @given pg_opt string with dbname param
+ * @when PostgresOptions object is created from given pg_opt string
+ * @then PostgresOptions object successfully created @and doesn't contain
```
muratovv: sab

test/module/irohad/ametsuchi/storage_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "ametsuchi/impl/storage_impl.hpp"
+
+using namespace iroha::ametsuchi;
+using namespace iroha::expected;
+
+class StorageTest : public ::testing::Test {
+ protected:
+  std::string block_store_path =
+      (boost::filesystem::temp_directory_path() / "block_store").string();
+  std::string dbname_ = "irohadb";
```
muratovv: I guess this name better to make random. For example pick UUID. It can be generated by boost.

irohad/ametsuchi/impl/postgres_options.cpp
kamilsa: spirit would be less readable for those who are not familiar with it (me for example)

irohad/ametsuchi/impl/storage_impl.hpp
kamilsa: Might be Result<void, message_type>

test/module/irohad/ametsuchi/CMakeLists.txt
kamilsa: postres options is part of ametsuchi target

irohad/ametsuchi/impl/storage_impl.hpp
muratovv: True for the new one and false for existed db.

irohad/ametsuchi/impl/postgres_options.cpp
lebdron: It should take less time for people to understand several rules, such as `username=`, `password=`, `dbname=`, accompanied with comments, than the parsing code referenced.

By the way, why not just to use regex to find `dbname=[^ ]+`, and then exclude it from original string?

kamilsa: The goal here is to parse all params, not just dbname

kamilsa: If it is really critical, I can spend some time to learn spirit and apply it here

irohad/ametsuchi/impl/postgres_options.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <boost/algorithm/string.hpp>
+#include <regex>
+
+namespace iroha {
+  namespace ametsuchi {
+
+    expected::Result<PostgresOptions, std::string> PostgresOptions::create(
+        std::string postgres_options) {
+      // Split into param=value list (split by space)
+      std::regex pattern{R"([^ ]+)"};
```
muratovv: I guess there are two options:

muratovv: I think there are two options:
* easy way - make parsing only for one required db field

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -256,39 +256,37 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
...
+        return expected::makeError(string_res.value());
+      }
+
+      auto ctx_result =
+          initConnections(block_store_dir, options.optionsString());
+      expected::Result<std::shared_ptr<StorageImpl>, std::string> storage;
+      ctx_result.match(
+          [&](expected::Value<ConnectionContext> &ctx) {
+            storage = expected::makeValue(std::shared_ptr<StorageImpl>(
+                new StorageImpl(block_store_dir,
```
muratovv: Why doesn't change it with `std::make_shared`?

muratovv: Why doesn't change it with `std::make_shared`?

irohad/ametsuchi/impl/postgres_options.hpp
```diff
@@ -0,0 +1,63 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2018 All Rights Reserved.
```
l4l: use new copyright header plz

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -196,6 +196,29 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
       block_store_->dropAll();
     }
 
+    expected::Result<bool, std::string> StorageImpl::createDatabaseIfNotExist(
+        const std::string &dbname,
+        const std::string &options_str_without_dbname) {
+      auto temp_connection =
+          std::make_unique<pqxx::lazyconnection>(options_str_without_dbname);
```
l4l: Why via pointer? What's the problem with
`pqxx::lazyconnection temp_connection(options_str_without_dbname)`?

test/module/irohad/ametsuchi/ametsuchi_fixture.hpp
```diff
@@ -97,8 +99,14 @@ namespace iroha {
 
       std::shared_ptr<StorageImpl> storage;
 
+      // generate random valid dbname
+      std::string dbname_ = "d"
+          + boost::uuids::to_string(boost::uuids::random_generator()())
+                .substr(0, 8);
```
l4l: why?

test/module/irohad/ametsuchi/storage_test.cpp
```diff
@@ -0,0 +1,80 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2018 All Rights Reserved.
```
l4l: outdated copyright header

test/module/irohad/ametsuchi/ametsuchi_fixture.hpp
l4l: sorry, i mean the size

kamilsa: first 8 characters of uuid are letters. 9th -- dash

PR #1280, opened 2018-04-27
YAC Rx timer dependency

lebdron:
### Description of the Change
Add external rx observable to yac timer, so that it is possible to write it in a testable way -- events are fired at expected time points.

### Benefits
Timer testability without real time assumptions

### Possible Drawbacks 
Not as flexible as before -- it's not possible to reuse the same timer with different delays. But since changes are local -- only consensus is involved -- it's fine to change interface later when something more flexible will be required.

### DIFF
--- a/irohad/consensus/yac/impl/timer_impl.cpp
+++ b/irohad/consensus/yac/impl/timer_impl.cpp
@@ -21,17 +21,17 @@
 namespace iroha {
   namespace consensus {
     namespace yac {
-      void TimerImpl::invokeAfterDelay(uint64_t millis,
-                                       std::function<void()> handler) {
+      TimerImpl::TimerImpl(rxcpp::observable<Rep> invoke_delay)
+          : invoke_delay_(invoke_delay) {}
+
+      void TimerImpl::invokeAfterDelay(std::function<void()> handler) {
         deny();
-        handler_ = std::move(handler);
-        timer = rxcpp::observable<>::timer(std::chrono::milliseconds(millis));
-        handle = timer.subscribe_on(rxcpp::observe_on_new_thread())
-                     .subscribe([this](auto) { handler_(); });
+        handle_ = invoke_delay_.subscribe(
+            [handler{std::move(handler)}](auto) { handler(); });
       }
 
       void TimerImpl::deny() {
-        handle.unsubscribe();
+        handle_.unsubscribe();
       }
 
       TimerImpl::~TimerImpl() {
--- a/irohad/consensus/yac/impl/timer_impl.cpp
+++ b/irohad/consensus/yac/impl/timer_impl.cpp
@@ -21,12 +21,13 @@
 namespace iroha {
   namespace consensus {
     namespace yac {
-      TimerImpl::TimerImpl(rxcpp::observable<Rep> invoke_delay)
-          : invoke_delay_(invoke_delay) {}
+      TimerImpl::TimerImpl(
+          std::function<rxcpp::observable<TimeoutType>()> invoke_delay)
+          : invoke_delay_(std::move(invoke_delay)) {}
 
       void TimerImpl::invokeAfterDelay(std::function<void()> handler) {
         deny();
-        handle_ = invoke_delay_.subscribe(
+        handle_ = invoke_delay_().subscribe(
             [handler{std::move(handler)}](auto) { handler(); });
       }
 
--- a/irohad/consensus/yac/impl/timer_impl.hpp
+++ b/irohad/consensus/yac/impl/timer_impl.hpp
@@ -27,13 +27,14 @@ namespace iroha {
       class TimerImpl : public Timer {
        public:
         /// Delay observable type
-        using Rep = long;
+        using TimeoutType = long;
 
         /**
          * Constructor
          * @param invoke_delay cold observable which specifies invoke strategy
          */
-        TimerImpl(rxcpp::observable<Rep> invoke_delay);
+        explicit TimerImpl(
+            std::function<rxcpp::observable<TimeoutType>()> invoke_delay);
         TimerImpl(const TimerImpl &) = delete;
         TimerImpl &operator=(const TimerImpl &) = delete;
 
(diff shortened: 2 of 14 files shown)


irohad/consensus/yac/impl/timer_impl.cpp
```diff
@@ -21,17 +21,17 @@
...
+          : invoke_delay_(invoke_delay) {}
+
+      void TimerImpl::invokeAfterDelay(std::function<void()> handler) {
         deny();
-        handler_ = std::move(handler);
-        timer = rxcpp::observable<>::timer(std::chrono::milliseconds(millis));
-        handle = timer.subscribe_on(rxcpp::observe_on_new_thread())
-                     .subscribe([this](auto) { handler_(); });
+        handle_ = invoke_delay_.subscribe(
+            [handler{std::move(handler)}](auto) { handler(); });
```
muratovv: what is the point to move pointer? It is the same as a copy, isn't it?

irohad/consensus/yac/impl/timer_impl.hpp
```diff
@@ -26,21 +26,25 @@ namespace iroha {
     namespace yac {
       class TimerImpl : public Timer {
        public:
-        TimerImpl() = default;
+        /// Delay observable type
+        using Rep = long;
```
muratovv: `Rep` is not so meaningful name, maybe assign some more intuitive alias?

irohad/main/impl/consensus_init.cpp
```diff
@@ -45,8 +45,9 @@ namespace iroha {
         return crypto;
       }
 
-      auto YacInit::createTimer() {
-        return std::make_shared<TimerImpl>();
+      auto YacInit::createTimer(std::chrono::milliseconds delay_milliseconds) {
+        return std::make_shared<TimerImpl>(rxcpp::observable<>::timer(
+            delay_milliseconds, rxcpp::observe_on_new_thread()));
```
muratovv: Each timer with emit new thread? So, Is it possible to rework it with a scheduler that handles only one thread?

test/integration/consensus/consensus_sunny_day.cpp
```diff
@@ -42,7 +42,8 @@ class FixedCryptoProvider : public MockYacCryptoProvider {
     // TODO 15.04.2018 x3medima17 IR-1189: move to separate class
     auto size =
         shared_model::crypto::DefaultCryptoAlgorithmType::generateKeypair()
-            .publicKey().size();
+            .publicKey()
+            .size();
```
muratovv: This change looks redundant. Is it intentional?

test/integration/consensus/consensus_sunny_day.cpp
```diff
@@ -122,7 +108,7 @@ class ConsensusSunnyDayTest : public ::testing::Test {
     }
     if (num_peers == 1) {
       delay_before = 0;
-      delay_after = 50;
+      delay_after = 5 * 1000;
```
muratovv: Why was delay changed? So, in this case, we also use cv and this is kinda of time assumption.

test/integration/consensus/consensus_sunny_day.cpp
```diff
@@ -138,9 +124,13 @@ std::vector<std::shared_ptr<shared_model::interface::Peer>>
     ConsensusSunnyDayTest::default_peers;
 
 TEST_F(ConsensusSunnyDayTest, SunnyDayTest) {
+  std::condition_variable cv;
+  std::mutex m;
   auto wrapper = make_test_subscriber<CallExact>(yac->on_commit(), 1);
-  wrapper.subscribe(
-      [](auto hash) { std::cout << "^_^ COMMITTED!!!" << std::endl; });
+  wrapper.subscribe([&cv](auto hash) {
+    std::cout << "^_^ COMMITTED!!!" << std::endl;
```
muratovv: So ancient and cool.
This is not an issue. Just a warm recall.

test/module/irohad/consensus/yac/yac_simple_cold_case_test.cpp
```diff
@@ -41,27 +41,7 @@ using namespace std;
...
-
-  auto yac_ = Yac::create(YacVoteStorage(),
-                          std::make_shared<MockYacNetwork>(network_),
-                          std::make_shared<MockYacCryptoProvider>(crypto_),
-                          std::make_shared<MockTimer>(timer_),
-                          *order,
-                          fake_delay_);
-
-  network_.subscribe(yac_);
+  // covered by setup
```
muratovv: Maybe remove this test, in this case?

irohad/consensus/yac/impl/timer_impl.cpp
l4l: Function stores erased types (can be arbitrary big object), so It might be faster in some strange cases

test/integration/consensus/consensus_sunny_day.cpp
l4l: plz do not delete
it is the only happiness in debugging

irohad/consensus/yac/impl/timer_impl.cpp
kamilsa: It might make sense if change handler's type to rvalue. In general why don't use perfect forwarding here?

test/integration/consensus/consensus_sunny_day.cpp
lebdron: clang-format

muratovv: So, the issue is resolved.

irohad/consensus/yac/impl/timer_impl.cpp
lebdron: Update: mocking template function is a hassle, do we really need it?

l4l: let it live like now, there's no real problem there

bakhtin:
test comment

test/integration/consensus/consensus_sunny_day.cpp
lebdron: Time assumption is required because gRPC is used in the test. Seems like 50 milliseconds was not enough and caused the test to fail from time to time.

lebdron:
@l4l The idea is that this timer interface can have different implementations, and rx is one of them. Other consensus components don't have to know about rx.

PR #1288, opened 2018-05-02
add can_get_blocks permission

Solonets:

### Description of the Change

add can_get_blocks permission to the system, to the read_all_group and implicitly to genesis blocks
 
### Benefits

needed for block query in future

### Possible Drawbacks 

--


l4l:
Maybe create full pr that entirely introduces the feature? Not only the permission name in the list

Solonets:
no

PR #1290, opened 2018-05-02
CI fixes

bakhtin:
### Description of the Change
- Rework GIT_PREVIOUS_COMMIT variable null-value check
- Fix Coverage stage run even it is supposed to be disabled
- Fix `parallelism` variable not set properly for ARMv7 platform builds
- Force checking `when` conditional prior occupying an agent 

### Benefits
- Speed up builds for the first branch commit as we download Dockerfile from a correct URL the image for which is usually present on a FS
- Speed up builds by skipping over Coverage stage when not needed
- Speed up builds by not waiting in stages to check for `when` step results. E.g., ARMv7 is usually skipped but still requires to allocate an agent to execute `when` step. Not needed anymore.

### Possible Drawbacks 
None

### DIFF
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -2,7 +2,7 @@
 
 def doDebugBuild(coverageEnabled=false) {
   def dPullOrBuild = load ".jenkinsci/docker-pull-or-build.groovy"
-  def parallelism = env.PARALLELISM
+  def parallelism = params.PARALLELISM
   def platform = sh(script: 'uname -m', returnStdout: true).trim()
   def previousCommit = !env.GIT_PREVIOUS_COMMIT ? env.GIT_COMMIT : env.GIT_PREVIOUS_COMMIT
   // params are always null unless job is started
@@ -12,7 +12,7 @@ def doDebugBuild(coverageEnabled=false) {
   if (!parallelism) {
     parallelism = 4
   }
-  if ("arm7" in env.NODE_NAME) {
+  if (env.NODE_NAME.contains('arm7')) {
     parallelism = 1
   }
   sh "docker network create ${env.IROHA_NETWORK}"
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -2,9 +2,10 @@
 
 def doDebugBuild(coverageEnabled=false) {
   def dPullOrBuild = load ".jenkinsci/docker-pull-or-build.groovy"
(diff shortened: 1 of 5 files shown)


.jenkinsci/debug-build.groovy
```diff
@@ -4,22 +4,23 @@ def doDebugBuild(coverageEnabled=false) {
   def dPullOrBuild = load ".jenkinsci/docker-pull-or-build.groovy"
   def parallelism = params.PARALLELISM
   def platform = sh(script: 'uname -m', returnStdout: true).trim()
+  def previousCommit = !env.GIT_PREVIOUS_COMMIT ? env.GIT_COMMIT : env.GIT_PREVIOUS_COMMIT
```
lebdron: Please add comment that `GIT_PREVIOUS_COMMIT` is null when PR is opened.

Jenkinsfile
```diff
@@ -376,10 +402,11 @@ pipeline {
           def bindings = load ".jenkinsci/bindings.groovy"
           def dPullOrBuild = load ".jenkinsci/docker-pull-or-build.groovy"
           def platform = sh(script: 'uname -m', returnStdout: true).trim()
+          def previousCommit = !env.GIT_PREVIOUS_COMMIT ? env.GIT_COMMIT : env.GIT_PREVIOUS_COMMIT
```
lebdron: Same as above for comment. Looks like this has to be in a separate function.

Jenkinsfile
```diff
@@ -399,7 +426,7 @@ pipeline {
           if (params.AndroidBindings) {
             iC = dPullOrBuild.dockerPullOrUpdate("android-${params.ABPlatform}-${params.ABBuildType}",
                                                  "${env.GIT_RAW_BASE_URL}/${env.GIT_COMMIT}/docker/android/Dockerfile",
-                                                 "${env.GIT_RAW_BASE_URL}/${env.GIT_PREVIOUS_COMMIT}/docker/android/Dockerfile",
+                                                 "${env.GIT_RAW_BASE_URL}/${previousCommit}/docker/android/Dockerfile",
```
lebdron: Similar change was in 4 places, any ideas how to reduce this?

bakhtin: We could introduce a separate variable containing Dockerfile's URL. But this generalization would work unless we are using same URLs for all different types of bindings. In case at some point there is a separate Dockerfile for e.g., Python bindings, we would have to use URL itself and not a variable.

PR #1303, opened 2018-05-05
Streaming fixes

luckychess:
### Description of the Change
Pull request fixes several race conditions around status streaming call.

### Benefits
Get rid of streaming crashes and wrong behaviour.

### Possible Drawbacks
~~None.~~ It's still possible that something is not fixed. I was able to successfully run torii_streaming_test 100000 times under address sanitizer and same amount of times without it - but you can't prove a bug absence, only bug presence.


irohad/torii/impl/command_service.cpp
```diff
@@ -59,13 +59,19 @@ namespace torii {
                             .txHash(tx_hash)
                             .notReceived()
                             .build();
+        log_->info(
```
nickaleks: Maybe make it debug? Same for log below.

irohad/torii/impl/command_service.cpp
```diff
@@ -59,13 +59,19 @@ namespace torii {
...
+            "Response not found: adding item to cache: {}, status NOT "
+            "RECEIVED",
+            tx_hash.hex());
         cache_->addItem(tx_hash, response.getTransport());
```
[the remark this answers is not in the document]
lebdron: Suggestion is to just do:
```C++
tx_processor_->transactionNotifier().subscribe([this](auto iroha_response) {
  auto proto_response =
      std::static_pointer_cast<shared_model::proto::TransactionResponse>(
          iroha_response);
  auto tx_hash = proto_response->transactionHash();
  cache_->addItem(tx_hash, proto_response->getTransport());
});
```

irohad/torii/impl/command_service.cpp
```diff
@@ -92,7 +98,6 @@ namespace torii {
               response.set_tx_status(
                   iroha::protocol::TxStatus::STATELESS_VALIDATION_SUCCESS);
 
-              cache_->addItem(tx_hash, response);
```
lebdron: I would really suggest to add logging in conditional before line 87.

PR #1297, opened 2018-05-04
Experimental MSVC shared model build

lebdron:
### Description of the Change
Changes allowing MSVC build of shared model. Only installed protobuf and gtest are supported due to more complicated shared library linking on Windows.

### Benefits
Develop shared model on Windows!

### Possible Drawbacks 
"Redundant" `ciso646` includes.

### Usage Examples or Tests
- `choco install visualstudio2017-workload-vctools` - Microsoft compilers
- `choco install cmake` - CMake
- Install [vcpkg](https://github.com/Microsoft/vcpkg) as described
- Install the packages: `vcpkg install protobuf:x86-windows gtest:x86-windows boost-multiprecision:x86-windows boost-variant:x86-windows boost-format:x86-windows boost-serialization:x86-windows boost-algorithm:x86-windows boost-conversion:x86-windows`
- Open toolchain console: Start-Visual Studio 2017-x86 Native Tools Command Prompt for VS 2017
- `cmake -HC:\path\to\shared_model -BC:\path\to\build -DCMAKE_TOOLCHAIN_FILE=C:\path\to\vcpkg\scripts\buildsystems\vcpkg.cmake`
- `cmake --build C:\path\to\build --config Debug`
- Run tests with `ctest -C Debug` in build directory
Expected output:
```
Test project C:\path\to\build
      Start  1: lazy_test
 1/33 Test  #1: lazy_test ..................................   Passed    0.03 sec
      Start  2: reference_holder_test
 2/33 Test  #2: reference_holder_test ......................   Passed    0.03 sec
      Start  3: shared_proto_commands_test
 3/33 Test  #3: shared_proto_commands_test .................   Passed    0.09 sec
      Start  4: shared_proto_tx_response_test
 4/33 Test  #4: shared_proto_tx_response_test ..............   Passed    0.09 sec
      Start  5: shared_proto_transaction_test
 5/33 Test  #5: shared_proto_transaction_test ..............   Passed    0.11 sec
      Start  6: shared_proto_queries_test
 6/33 Test  #6: shared_proto_queries_test ..................   Passed    0.10 sec
      Start  7: shared_proto_query_responses_test
 7/33 Test  #7: shared_proto_query_responses_test ..........   Passed    0.10 sec
      Start  8: shared_proto_util_test
 8/33 Test  #8: shared_proto_util_test .....................   Passed    0.07 sec
      Start  9: shared_proto_amount_test
 9/33 Test  #9: shared_proto_amount_test ...................   Passed    0.10 sec
[...]

PR #1308, opened 2018-05-07
fixed boost linking 

sjashwin:
### Description of the Change
 Added target links for libraries that could not find the boost include files. Tested with CMake version 3.11.1 and boost version 1.67.0_1

### Benefits
All the packages find boost files.

### Possible Drawbacks
None

### Usage Example or Tests
`mkdir build; cd build; cmake ..; make -j$(nproc)`
`cmake --build build --target test`

### DIFF
--- a/libs/common/CMakeLists.txt
+++ b/libs/common/CMakeLists.txt
@@ -2,12 +2,17 @@ add_library(libs_common
     files.cpp
     )
 
+add_library(common INTERFACE)
+
+target_include_directories(common INTERFACE
+        ${PROJECT_SOURCE_DIR}/libs/common
+        )
+
 target_link_libraries(libs_common
     logger
+    boost
     )
-target_link_libraries(libs_common boost)
-add_library(common INTERFACE)
-target_include_directories(common INTERFACE
-    ${PROJECT_SOURCE_DIR}/libs/common
-    )
+target_link_libraries(common INTERFACE
+        boost
+        )
 
--- a/libs/parser/CMakeLists.txt
+++ b/libs/parser/CMakeLists.txt
@@ -1,2 +1,4 @@
 add_library(parser STATIC parser.cpp)
-target_link_libraries(parser boost)
+target_link_libraries(parser 
+    boost
+    )
--- a/test/module/irohad/common/CMakeLists.txt
+++ b/test/module/irohad/common/CMakeLists.txt
@@ -1,8 +1,8 @@
 AddTest(blob_converter_test blob_converter_test.cpp)
+
 target_link_libraries(blob_converter_test
-    common
-        boost
-    )
+        common
+        )
 
 AddTest(raw_block_loader_test raw_block_loader_test.cpp)
 target_link_libraries(raw_block_loader_test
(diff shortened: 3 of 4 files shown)


laSinteZ:
@sjashwin thank you for you work!
However, we have some requirements:
1. Please make a little bit more detailed PR description (like here: https://github.com/hyperledger/iroha/pull/1297)
2. Please, sign-off your commit with DCO (https://github.com/apps/dco). It's Hyperledger's requirement. You can `git reset --soft HEAD~` to ammend previous commit and after it recommit again with `-s` option. After it, you need to force push a commit, so we can see it here.

sjashwin:
@laSinteZ I've made the required changes.

libs/common/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ add_library(libs_common
 target_link_libraries(libs_common
     logger
     )
-
+target_link_libraries(libs_common boost)
```
lebdron: This can be moved to statement above, so it will look like:
```CMake
target_link_libraries(libs_common
    logger
    boost
    )
```

libs/parser/CMakeLists.txt
```diff
@@ -1 +1,2 @@
 add_library(parser STATIC parser.cpp)
+target_link_libraries(parser boost)
```
lebdron: We have a style guide to write it as
```CMake
target_link_libraries(parser 
    boost
    )
```
so that adding libraries does not affect the whole statement in diff. Please change the formatting.

test/module/irohad/common/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 AddTest(blob_converter_test blob_converter_test.cpp)
 target_link_libraries(blob_converter_test
     common
+        boost
```
lebdron: Please move this `boost` to `common` target, in `libs/common/CMakeLists.txt`, because test is made for that library. It will look like:
```CMake
target_link_libraries(common
    boost
    )
```

test/module/libs/converter/CMakeLists.txt
```diff
@@ -1 +1,3 @@
 addtest(string_converter_test string_converter_test.cpp)
+
+target_link_libraries(string_converter_test boost)
```
lebdron: This test is also done for `common` library, so it should be linked to `common` instead of `boost` directly. Also consider the style guide mentioned above.

test/module/irohad/common/CMakeLists.txt
sjashwin:
```
target_link_libraries(common INTERFACE
    boost
    )
```
Adding this to libs/common/CMAKELists.txt does not work

sjashwin: Sorry I understood it wrong. I've made the changes and it works!!

laSinteZ:
Oh @sjashwin I see that you've accidentally pushed code related to Segmentation fault on negative precision into this PR. Could you please remove unrelated to this PR commits, since we can't merge it otherwise? 😔

PR #1330, opened 2018-05-11
Release notes DRAFT

Solonets:
# DRAFT

### Assets

In this release following changes were made:

## Uploading of the build artifacts
Artifacts will now be available on https://artifact.soramitsu.co.jp for downloading
## Features
- genesis block generation from a list of transactions
#### How to use
`iroha-cli --genesis_block --genesis_transaction transaction.json` @l4l 
- Added: error message to response, if stateless validation has failed
#### How to use
`ToriiResponse` now has field `error_message` with the error description of transaction
`ErrorResponse` now has field `message` with  the error description of query
- QueryResponse now have hex-encoded query hashe instead of binary hashe
## Client libraries & shared model
  - Transaction counter is no longer used in the transaction structure
  - Full transition to shared_model 
### Performance
- A Regular queue is replaced with a priority queue in ordering gate so that proposals are propagated in order to the system if they are received in random order from the network. That causes fewer issues with simulator warnings about last block height being inconsistent with proposal height.
### Other
- Python and Javascript bindings can be compiled at the same time

### Bugfixes
- Fixed: signature comparing
- Fixed: boost linking
- Fixed: numerous synchronization issues
- Fixed: inconsistent genesis block generation
- Fixed: possible fork attacks
- Fixed: numerous permission issues
- Fixed: segfaults when the ledger doesn't contain blocks
- Fixed: numerous memory management issues
- Other minor fixes


neewy:
Please expand features section with "how to use" section and "known issues" if there are any 🙏

PR #1336, opened 2018-05-14
Remove polymorphic wrapper from tx response, specified visitor

lebdron:
### Description of the Change
Nobody likes polymorphic wrappers, right? This is the first pull request in a series which will remove polymorphic wrapper from shared model, making clear `const iface &` return types.
There are several helper methods included for deserialization and visitation, so that polymorphic wrapper does not have to be removed from the whole project in one pull request, making a smoother transition.

### Benefits
Simpler code.

### Possible Drawbacks 
None.

### DIFF
--- a/shared_model/utils/variant_deserializer.hpp
+++ b/shared_model/utils/variant_deserializer.hpp
@@ -68,9 +68,9 @@ namespace shared_model {
          * @param v result variant
          */
         template <class V, class T = typename V::types, class Archive>
-        static auto invoke(Archive &&ar, int which) -> typename std::enable_if<
+        static auto invoke(Archive &&ar, int which) -> std::enable_if_t<
             not std::is_same<FrontType<S>, FrontType<T>>::value,
-            V>::type {
+            V> {
           if (which == 0) {
             using head_type = FrontType<S>;
             using variant_head_type = FrontType<T>;
@@ -106,9 +106,9 @@ namespace shared_model {
          * @param v result variant
          */
         template <class V, class T = typename V::types, class Archive>
-        static auto invoke(Archive &&ar, int which) -> typename std::enable_if<
-            std::is_same<FrontType<S>, FrontType<T>>::value,
-            V>::type {
+        static auto invoke(Archive &&ar, int which)
+            -> std::enable_if_t<std::is_same<FrontType<S>, FrontType<T>>::value,
+                                V> {
           if (which == 0) {
             using head_type = FrontType<S>;
             return V(head_type(std::forward<Archive>(ar)));


shared_model/interfaces/utils/specified_visitor.hpp
```diff
@@ -26,21 +26,30 @@ namespace shared_model {
   namespace interface {
     template <typename Type>
     class SpecifiedVisitor
-        : public boost::static_visitor<
-              boost::optional<shared_model::detail::PolymorphicWrapper<Type>>> {
+        : public boost::static_visitor<boost::optional<const Type &>> {
      private:
-      using Y = shared_model::detail::PolymorphicWrapper<Type>;
+      using Y = shared_model::detail::PolymorphicWrapper<
+          std::remove_cv_t<std::remove_reference_t<Type>>>;
```
l4l: maybe remove only const?

shared_model/interfaces/utils/specified_visitor.hpp
```diff
@@ -26,21 +26,30 @@ namespace shared_model {
...
 
       template <typename T>
-      boost::optional<Y> operator()(
+      boost::optional<const Type &> operator()(
           const shared_model::detail::PolymorphicWrapper<T> &t) const {
         return boost::none;
       }
+
+      template <typename T>
+      boost::optional<const Type &> operator()(const T &t) const {
```
l4l: Comment might be useful here (or even class level doc)

shared_model/utils/visitor_apply_for_all.hpp
```diff
@@ -30,7 +30,14 @@ namespace shared_model {
     class ToStringVisitor : public boost::static_visitor<std::string> {
      public:
       template <typename InputType>
-      std::string operator()(const InputType &operand) const {
+      auto operator()(const InputType &operand) const
+          -> decltype(operand.toString(), std::string()) {
```
l4l: That really hard to read and understand, what's that for? Inheritance from static_visitor is already provide the type.

shared_model/interfaces/utils/specified_visitor.hpp
```diff
@@ -26,21 +26,30 @@ namespace shared_model {
...
+        return *t;
+      }
+
+      boost::optional<const Type &> operator()(const Type &t) const {
         return t;
       }
 
       template <typename T>
-      boost::optional<Y> operator()(
+      boost::optional<const Type &> operator()(
```
kamilsa: I think this and next operators can be merged into a single one like:
```
boost::optional<const Type&> operator()(const auto&) const {return boost::none;}
```

shared_model/interfaces/transaction_responses/tx_response.hpp
```diff
@@ -65,7 +65,7 @@ namespace shared_model {
       /**
        * @return attached concrete tx response
        */
-      virtual const ResponseVariantType &get() const = 0;
+      virtual ResponseVariantType get() const = 0;
```
kamilsa: why remove const here?

shared_model/interfaces/utils/specified_visitor.hpp
kamilsa: why remove volitile?

test/module/shared_model/builders/transaction_responses/transaction_builders_common.hpp
```diff
@@ -18,21 +18,17 @@
...
 auto verifyType() {
-  return [](auto val) {
-    if (std::is_same<decltype(val), T>::value) {
-      SUCCEED();
-    } else {
-      FAIL() << "obtained: " << typeid(decltype(val)).name() << std::endl
-             << "expected: " << typeid(T).name() << std::endl;
-    }
-  };
+  return iroha::make_visitor([](const T &val) { SUCCEED(); },
```
kamilsa: Not an issue, just nice

shared_model/utils/visitor_apply_for_all.hpp
lebdron: It does not help because we should do SFINAE for `InputType` which has `operator->`, and reference `InputType`.

shared_model/interfaces/utils/specified_visitor.hpp
lebdron: This using will be removed after polymorphic wrapper is removed, and now polymorphic wrapper is only used with non-const non-reference types.
@l4l Since polymorphic wrapper stores shared pointer, having a pointer to a reference seems fragile.
@kamilsa yes, `std::remove_const` can be used here

shared_model/utils/visitor_apply_for_all.hpp
kamilsa: ok, clear

shared_model/interfaces/transaction_responses/tx_response.hpp
lebdron: https://isocpp.org/blog/2013/06/quick-q-should-i-return-a-const-value-stackoverflow

> The answer is now a definite “No” because it prevents move from returned values.

shared_model/utils/variant_deserializer.hpp
```diff
@@ -65,10 +68,12 @@ namespace shared_model {
          * @param v result variant
          */
         template <class V, class T = typename V::types, class Archive>
-        static V invoke(Archive &&ar, int which) {
+        static auto invoke(Archive &&ar, int which) -> typename std::enable_if<
```
nickaleks: please use std::enable_if_t

nickaleks:
Build fails on Linux :(

shared_model/utils/variant_deserializer.hpp
```diff
@@ -68,9 +68,10 @@ namespace shared_model {
          * @param v result variant
          */
         template <class V, class T = typename V::types, class Archive>
-        static auto invoke(Archive &&ar, int which) -> typename std::enable_if<
-            not std::is_same<FrontType<S>, FrontType<T>>::value,
-            V>::type {
+        static auto invoke(Archive &&ar, int which) ->
+            typename std::enable_if_t<
```
nickaleks: if you use enable_if_t you do not need typename

PR #1338, opened 2018-05-14
Reduce integration test execution time

lebdron:
### Description of the Change
Before: Total Test time (real) = 838.55 sec
After: Total Test time (real) =  53.59 sec

Added logging to postgres_block_query.cpp because nonexistent hash was giving `JSON deserialization failed` message, which is misleading.

### Benefits
Reduced test execution time.

### Possible Drawbacks 
None.


test/integration/acceptance/acceptance_fixture.hpp
```diff
@@ -0,0 +1,101 @@
...
+#include "module/shared_model/builders/protobuf/test_query_builder.hpp"
+#include "module/shared_model/builders/protobuf/test_transaction_builder.hpp"
+
+namespace shared_model {
+  namespace proto {
+    class TransactionResponse;
+  }
+}  // namespace shared_model
+
+class AcceptanceFixture : public ::testing::Test {
```
muratovv: Add a description to the class.

test/integration/acceptance/acceptance_fixture.hpp
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Creates the transaction with the user creation commands
+   * @param perms are the permissions of the user
+   * @return built tx and a hash of its payload
+   */
+  shared_model::proto::Transaction makeUserWithPerms(
+      const std::vector<std::string> &perms);
+
+  template <typename Builder>
+  auto base(Builder builder) -> decltype(
```
muratovv: Add a description to the method.

test/integration/acceptance/acceptance_fixture.hpp
```diff
@@ -0,0 +1,101 @@
...
+      builder.creatorAccountId(std::string()).createdTime(uint64_t()));
+
+  /**
+   * Create valid base pre-built transaction
+   * @return pre-built tx
+   */
+  auto baseTx() -> decltype(base(TestUnsignedTransactionBuilder()));
+
+  auto baseQry() -> decltype(base(TestUnsignedQueryBuilder()));
```
muratovv: Same as above

test/integration/acceptance/create_role_test.cpp
```diff
@@ -189,31 +141,30 @@ TEST_F(CreateRole, MaxLenRoleName) {
...
  * @given some user with can_create_role permission
- * @when execute tx with CreateRole command with inexistent permission name
+ * @when execute tx with CreateRole command with nonexistent permission name
  * @then the tx hasn't passed stateless validation
  *       (aka skipProposal throws)
  */
-TEST_F(CreateRole, DISABLED_InexistentPerm) {
-  IntegrationTestFramework itf;
-  itf.setInitialState(kAdminKeypair)
+TEST_F(CreateRole, DISABLED_NonexistentPerm) {
```
muratovv: I think here we have to add a note about why this test fails and add a task which fixes it.

test/integration/acceptance/create_role_test.cpp
```diff
@@ -189,31 +141,30 @@ TEST_F(CreateRole, MaxLenRoleName) {
...
+              checkStatelessInvalid);
 }
 
 /**
  * @given some user with can_create_role permission
- * @when execute tx with CreateRole command with existent role name
+ * @when execute tx with CreateRole command with existing role name
  * @then there is an empty verified proposal
  */
 TEST_F(CreateRole, DISABLED_ExistingRole) {
```
muratovv: same as above

test/integration/acceptance/transfer_asset_test.cpp
```diff
@@ -159,13 +141,17 @@ TEST_F(TransferAsset, WithOnlyCanTransferPerm) {
  * @then there is an empty proposal
  */
 TEST_F(TransferAsset, WithOnlyCanReceivePerm) {
-  IntegrationTestFramework()
+  IntegrationTestFramework(1)
```
muratovv: Why don't set `3` and omit to skip proposals and block?

test/integration/acceptance/transfer_asset_test.cpp
```diff
@@ -202,22 +190,27 @@ TEST_F(TransferAsset, InexistentDest) {
...
- * @when execute tx with TransferAsset command with inexistent asset
+ * @when execute tx with TransferAsset command with nonexistent asset
  * @then there is an empty proposal
  */
-TEST_F(TransferAsset, InexistentAsset) {
-  const std::string &inexistent = "inexist#test"s;
-  IntegrationTestFramework()
+TEST_F(TransferAsset, NonexistentAsset) {
+  const std::string &nonexistent = "inexist#test"s;
+  IntegrationTestFramework(1)
```
muratovv: Same as about omitting skip proposal and block.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -57,14 +57,14 @@ namespace integration_framework {
    public:
     /**
      * Construct test framework instance
-     * @param maximum_proposal_size - (default = 10) Maximum amount of
-     * transactions per proposal
+     * @param maximum_proposal_size - Maximum amount of transactions per
```
nickaleks: Since transaction is a countable noun, correct form will be "maximum **number** of transactions"

test/integration/acceptance/acceptance_fixture.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "datetime/time.hpp"
+#include "framework/integration_framework/integration_test_framework.hpp"
+#include "interfaces/utils/specified_visitor.hpp"
+
+using namespace std::string_literals;
+
+AcceptanceFixture::AcceptanceFixture()
+    : kUser("user"s),
+      kRole("role"s),
```
nickaleks: "s not needed. Using namespace can also be removed

test/integration/acceptance/acceptance_fixture.cpp
```diff
@@ -0,0 +1,111 @@
...
+  return makeUserWithPerms(kRole, perms);
+}
+
+template <typename Builder>
+auto AcceptanceFixture::base(Builder builder) -> decltype(
+    builder.creatorAccountId(std::string()).createdTime(uint64_t())) {
+  return builder.creatorAccountId(kUserId).createdTime(iroha::time::now());
+}
+
+template auto AcceptanceFixture::base<TestUnsignedTransactionBuilder>(
```
nickaleks: Nice!

test/integration/acceptance/add_asset_qty_test.cpp
```diff
@@ -174,18 +123,17 @@ TEST_F(AddAssetQuantity, Uint256DestOverflow) {
...
  * @given some user with all required permissions
- * @when execute tx with AddAssetQuantity command with inexistent account
+ * @when execute tx with AddAssetQuantity command with nonexistent account
  * @then there is an empty proposal
  */
-TEST_F(AddAssetQuantity, InexistentAccount) {
-  const std::string &inexistent = "inexist@test"s;
-  IntegrationTestFramework()
+TEST_F(AddAssetQuantity, NonexistentAccount) {
+  const std::string &nonexistent = "inexist@test"s;
```
nickaleks: Not really sure this convoluted form grants us any benefits. IMO it better to write
`std::string nonexistent = "blabla"`. We still perform one construction, but a reader of the code no longer has to be aware of things such as lifetime extensions etc.

test/integration/acceptance/transfer_asset_test.cpp
lebdron: If we set it to 3, all proposals should have 3 transactions. Since 4 transactions are sent in the test, we will not be able to easily verify that 4th transaction is stateful invalid without sending another two transactions. 
A possible solution is to set `4`, send all transactions, and verify the status of bad expected transaction by doing query and verifying transaction status.
Unfortunately, there is a number (more than 2 tests mentioned in the review) of tests which may be rewritten by applying logic above. I would consider making a separate task to simplify the assertions in acceptance tests.

test/integration/acceptance/add_asset_qty_test.cpp
lebdron: Regarding `operator""s`, it has a profit of compile-time introspection of string length, which does not occur in `std::string("str")`, so we can experiment with it in methods which take `std::string` object, but we pass `const char*` in it.

PR #1339, opened 2018-05-14
Extract signaturesSubset function

l4l:
### Description of the Change

Decrease redundancy in public key check in signature range.

### Benefits

Less self-repeating code

### Possible Drawbacks 

There's additional alloc+copy in SupermajorityCheckerImpl, that probably might be optimized

### DIFF
--- a/irohad/torii/processor/impl/query_processor_impl.cpp
+++ b/irohad/torii/processor/impl/query_processor_impl.cpp
@@ -16,6 +16,8 @@
  */
 
 #include "torii/processor/query_processor_impl.hpp"
+#include "boost/range/size.hpp"
+#include "validation/utils.hpp"
 
 namespace iroha {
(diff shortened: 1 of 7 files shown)


irohad/torii/processor/impl/query_processor_impl.cpp
```diff
@@ -40,17 +42,15 @@ namespace iroha {
...
       auto signatories = wsv_query->getSignatories(qry.creatorAccountId());
-      if (not signatories) {
-        return false;
-      }
-      return std::find(
-                 signatories->begin(), signatories->end(), sig.publicKey())
-          != signatories->end();
+      const auto &sig = qry.signatures();
+
+      return boost::range_detail::range_calculate_size(sig) == 1
```
kamilsa: use boost::size instead

PR #1345, opened 2018-05-17
Static regex objects in field validator

nickaleks:
### Description of the Change

Change regular expressions in field validator from class member to static variables.

### Benefits

10x Performance improvements

### Possible Drawbacks 

Cannot specify regexes when creating validator


l4l:
Any updates on that?

nickaleks:
Waited for streaming fixes. Need to resolve some bugs in the tests, then will merge

test/integration/acceptance/acceptance_fixture.cpp
```diff
@@ -27,7 +27,9 @@ AcceptanceFixture::AcceptanceFixture()
             framework::SpecifiedVisitor<
                 shared_model::interface::StatelessFailedTxResponse>(),
             status.get()));
-      }) {}
+      }),
+      initial_time(iroha::time::now()),
+      tx_counter(0){}
```
l4l: that's strange to be here

test/integration/acceptance/acceptance_fixture.cpp
```diff
@@ -114,3 +116,7 @@ template auto AcceptanceFixture::complete<TestUnsignedQueryBuilder>(
         builder.build()
             .signAndAddSignature(std::declval<shared_model::crypto::Keypair>())
             .finish());
+
+iroha::time::time_t AcceptanceFixture::getUniqueTime() {
+  return initial_time + tx_counter++;
```
l4l: race condition

test/integration/acceptance/get_transactions_test.cpp
```diff
@@ -66,12 +66,14 @@ TEST_F(GetTransactions, HaveNoGetPerms) {
   };
 
   auto dummy_tx = dummyTx();
-  IntegrationTestFramework(2)
+  IntegrationTestFramework(1)
```
l4l: Why is that fixes in this pr?

test/integration/acceptance/acceptance_fixture.cpp
nickaleks: It should not be called concurrently

test/integration/acceptance/get_transactions_test.cpp
nickaleks: because speed up causes tests to fail more often

test/integration/acceptance/acceptance_fixture.cpp
nickaleks: It is used to calculate unique time

PR #1349, opened 2018-05-17
Fix build issues related to changes in "amount utils"

igor-egorov:
### Description of the Change

It was unable to build Iroha on some systems. The fix arranges dependencies as it should be.

Thanks to @lebdron for the fix!

### Benefits

Iroha can be built on systems without google protobuf header files installed.

### Possible Drawbacks 

None

### Usage Examples or Tests

`cmake -H. -Bbuild`
`cmake --build build`

### DIFF
--- a/shared_model/utils/amount_utils.hpp
+++ b/shared_model/utils/amount_utils.hpp
@@ -20,11 +20,12 @@
 
 #include <boost/multiprecision/cpp_int.hpp>
 
-#include "builders/default_builders.hpp"
-#include "builders/protobuf/common_objects/proto_amount_builder.hpp"
 #include "common/result.hpp"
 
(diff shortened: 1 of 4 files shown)


shared_model/utils/amount_utils.hpp
```diff
@@ -20,11 +20,12 @@
...
 #include <boost/multiprecision/cpp_int.hpp>
 
-#include "builders/default_builders.hpp"
-#include "builders/protobuf/common_objects/proto_amount_builder.hpp"
 #include "common/result.hpp"
 
 namespace shared_model {
+  namespace interface {
+    class Amount;
+  }  // namespace interface
```
nickaleks: Add empty line pls

PR #1360, opened 2018-05-21
Feature/block creator height

victordrobny:
### Description of the Change
Simulator gets height from blockstore, not from proposal


l4l:
any updates on that?

PR #1363, opened 2018-05-21
Fix build of stateful_validator target

igor-egorov:

### Description of the Change

Build was broken on systems without protobuf headers installed.

### Benefits

Build is possible again.

### Possible Drawbacks 

None

### Usage Examples or Tests 

```
git clean -ffdx
cmake -H. -Bbuild
cmake --build build
```

### DIFF
--- a/irohad/validation/CMakeLists.txt
+++ b/irohad/validation/CMakeLists.txt
@@ -22,6 +22,7 @@ target_link_libraries(stateful_validator
     rxcpp
     shared_model_interfaces
     logger
+    shared_model_proto_backend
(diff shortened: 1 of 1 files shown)


muratovv:
What about tests? What is the way to prevent those type of issues in the future?

igor-egorov:
The only way to detect such issues on CI is to remove google.protobuf headers from system paths.

lebdron:
@muratovv If libraries are not installed in system path, then they will not be included by default, and it will be possible to track down such issues.

PR #1370, opened 2018-05-23
Add validation rule for precision field

igor-egorov:

### Description of the Change

Add missing validation rule for precision field. Add cases for precision field in field validator test.

### Benefits

Improved coverage.

### Possible Drawbacks 

Now the validation rule is useless because of used type for precision value in c++. It will become a relevant check when the underlying type of PrecisionType will be changed.

### Usage Examples or Tests 

compile & test

### DIFF
--- a/shared_model/validators/field_validator.cpp
+++ b/shared_model/validators/field_validator.cpp
@@ -16,8 +16,11 @@
  */
 
 #include "validators/field_validator.hpp"
+
 #include <boost/algorithm/string_regex.hpp>
 #include <boost/format.hpp>
+#include <limits>
+
 #include "cryptography/crypto_provider/crypto_verifier.hpp"
 #include "permissions.hpp"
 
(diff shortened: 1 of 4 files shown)


shared_model/validators/field_validator.cpp
```diff
@@ -208,7 +211,20 @@ namespace shared_model {
     void FieldValidator::validatePrecision(
         ReasonsGroupType &reason,
         const interface::types::PrecisionType &precision) const {
-      // define precision constraints
+      /* The following validation is pointless since PrecisionType is already
+       * uint8_t, but it is going to be changed and the validation will become
+       * meaningful.
+       */
+      auto min = std::numeric_limits<uint8_t>::min();
```
lebdron: `auto` will not work here because of the following signature `static constexpr T min();`, same with `max`. The `precision < min or precision > max` will probably perform incorrect cast.

Please check the following page and confirm if there will be no issues if `PrecisionType` will be unsigned, or have greater size than `uint8_t`. http://en.cppreference.com/w/cpp/language/operator_comparison

test/module/shared_model/validators/field_validator_test.cpp
```diff
@@ -130,6 +130,11 @@ class FieldValidatorTest : public ValidatorsTest {
...
                                           permission_test_cases));
 
+    field_validators.insert(makeValidator("precision",
+                                          &FieldValidator::validatePrecision,
+                                          &FieldValidatorTest::precision,
+                                          precision_test_cases));
+
     // TODO: add validation to all fields
     for (const auto &field : {"value",
                               "signature",
```
lebdron: Precision can be removed from line 144.

PR #1374, opened 2018-05-24
Storage init test

kamilsa:
### Description of the Change

Storage init test did not take into account environment variable which led to falling test on CI, this PR fixes that issue. 
### Benefits

No test is falling anymore

### Possible Drawbacks 

None

### DIFF
--- a/test/module/irohad/ametsuchi/storage_init_test.cpp
+++ b/test/module/irohad/ametsuchi/storage_init_test.cpp
@@ -13,6 +13,24 @@ using namespace iroha::ametsuchi;
 using namespace iroha::expected;
 
 class StorageInitTest : public ::testing::Test {
+ public:
+  StorageInitTest() {
+    auto pg_host = std::getenv("IROHA_POSTGRES_HOST");
+    auto pg_port = std::getenv("IROHA_POSTGRES_PORT");
+    auto pg_user = std::getenv("IROHA_POSTGRES_USER");
+    auto pg_pass = std::getenv("IROHA_POSTGRES_PASSWORD");
+    if (not pg_host) {
+      pg_opt_without_dbname_ =
+          "host=localhost port=5432 user=postgres password=mysecretpassword ";
+    } else {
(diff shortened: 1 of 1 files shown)


test/module/irohad/ametsuchi/storage_init_test.cpp
```diff
@@ -13,6 +13,24 @@ using namespace iroha::ametsuchi;
...
+    auto pg_pass = std::getenv("IROHA_POSTGRES_PASSWORD");
+    if (not pg_host) {
+      pg_opt_without_dbname_ =
+          "host=localhost port=5432 user=postgres password=mysecretpassword ";
+    } else {
+      std::stringstream ss;
+      ss << "host=" << pg_host << " port=" << pg_port << " user=" << pg_user
+         << " password=" << pg_pass;
+      pg_opt_without_dbname_ = ss.str();
+    }
```
igor-egorov: The code in constructor does not look good to me. :( 
Specifically the fact that initialization of `pg_port`, `pg_user` and `pg_pass` with default values depends on the presence of `IROHA_POSTGRES_HOST` environment variable. 

I suggest independent initialization of mentioned variables.

If you are comfortable with it, please use the code below. Or you can use your own implementation.

```
  StorageInitTest() {
    std::string env_var;
    std::stringstream ss;
    ss << "host=" << (env_var = std::getenv("IROHA_POSTGRES_HOST") ? env_var : "localhost");
    ss << " port=" << (env_var = std::getenv("IROHA_POSTGRES_PORT") ? env_var : "5432");
    ss << " user=" << (env_var = std::getenv("IROHA_POSTGRES_USER") ? env_var : "postgres");
    ss << " password=" << (env_var = std::getenv("IROHA_POSTGRES_PASSWORD") ? env_var : "mysecretpassword");
    pg_opt_without_dbname_ = ss.str();
    pgopt_ = pg_opt_without_dbname_ + " dbname=" + dbname_;
  }
```

PR #1375, opened 2018-05-24
Add emscripted bindings

stinger112:
### Description of the Change

This PR introduces a new way to build JS library for all platforms by Emscripten.

### Benefits

The first and the most important feature of this PR: we can use compiled library on almost **any** JS platform, particularly on modern **browsers**.
The second, but very important too: we can **sufficiently** decrease efforts to maintain the native code of the NPM package because we no need SWIG anymore!

### Possible Drawbacks 

Maintainers should spend some time to create and maintain *Embind* C++ bindings and JS wrapper code to properly build the lib for Emscripten.
Also, we should create some unit test to ensure binding correctness between C++ and JS types.
The third one drawback is a C++ type system. Not all types and semantic structures (e.g. template-based classes) can be easily turned to JS equivalents. It needs to be previously discussed before introduce such classes and methods to the Shared Library API. And, obviously, developers of Shared Library shouldn't overuse polymorphism.

### Usage Examples or Tests

This library is a drop-in replacement of the previous one. No additional changes in existing application code needed.

### Alternate Designs

Alternate design can be implemented using WebIDL bind system instead Embind. More information you can find in the [**report**](https://gist.github.com/stinger112/c0ec155a0f4083dd7013d1dd027a6678).

### Additional

Working on this PR still in progress, but most of the functionality (including tests) already working.
Please, don't remove this PR. Thank you :)

### How to build

1. Pull docker image with build environment
```
docker pull stinger112/soramitsu:iroha-emscripten
```
2. Go to the Iroha folder and run image with mounted iroha source directory
```
docker run -it -v $(pwd):/opt/iroha iroha-emscripten
```

Then do actions below **inside a container**:

3. Set Emscripten env vars
```
source /opt/emsdk/emsdk_set_env.sh
```
4. Go to the folder
```
cd /opt/iroha/shared_model
```
5. Create build configuration
```
emcmake cmake -H. \
              -Bbuild \
              -DTESTING=OFF \
              -DCMAKE_BUILD_TYPE=Release \
              -DCMAKE_PREFIX_PATH=$EMSCRIPTEN/system \
              -DEMSCRIPTEN=ON
```
6. Build the project
```
cmake --build build --target irohalib -- -j$(nproc)
```

7. Test the package:
```
cd /opt/iroha/shared_model/packages/javascript
npm install
npm test
```

### TODOs

- [ ] Create clean docker image with Emscripten's compilation environment
- [ ] Add correct memory management with **shared_ptr** and **deleteLater**
- [ ] Add helper functions and classes to JS wrapper
- [x] Enable compiler optimizations
- [x] Actualize the code according actual **Shared Library** in develop
- [x] Cleanup code


shared_model/bindings/model_query_builder_embind.cpp
```diff
@@ -0,0 +1,66 @@
...
+  class_<Query>("Query")
+  .function("creatorAccountId", &Query::creatorAccountId)
+  .function("queryCounter", &Query::getQueryCounter)
+  .function("blob", &Query::blob)
+  .function("payload", &Query::payload)
+  .function("addSignature", &Query::addSignature)
+  .function("createdTime", &Query::getCreatedTime)
+  ;
+
+  typedef UnsignedWrapper<Query> UnsignedWrapper;
```
l4l: Will it be visible to the client code? It's not the best name, if so

shared_model/bindings/model_query_builder_embind.cpp
```diff
@@ -0,0 +1,66 @@
...
+  ;
+
+  class_<Query>("Query")
+  .function("creatorAccountId", &Query::creatorAccountId)
+  .function("queryCounter", &Query::getQueryCounter)
+  .function("blob", &Query::blob)
+  .function("payload", &Query::payload)
+  .function("addSignature", &Query::addSignature)
+  .function("createdTime", &Query::getCreatedTime)
+  ;
```
l4l: signatures? :(
I guess it's not that trivial so add a todo plz (also for the other proto model types)

shared_model/bindings/model_transaction_builder_embind.cpp
```diff
@@ -0,0 +1,143 @@
...
+  typedef ModelProto<UnsignedWrapper> ModelProto;
+  class_<ModelProto>("ModelProtoTransaction")
+  .constructor<UnsignedWrapper&>()
+  .function("signAndAddSignature", &ModelProto::signAndAddSignature)
+  .function("finish", &ModelProto::finish)
+  ;
+
+  using Role = interface::permissions::Role;
+  enum_<Role>("Role")
+  .value("kAppendRole", Role::kAppendRole)
```
l4l: Can initialization and type adding be splitted? e.g:
```cpp
auto e = enum_<Role>("Role");
e.value(..., ...);
```

If so, than it's possible to use smth like:
```cpp
auto e = enum_<Role>("Role");
for (auto i = 0; i < Role::COUNT; ++i) {
  auto p = static_cast<Role>(i);
  e = e.value(toString(p), p);
}
```

shared_model/bindings/model_transaction_builder_embind.cpp
```diff
@@ -0,0 +1,143 @@
...
+  class_<RolePermissionSet>("RolePermissionSet")
+  .constructor<>()
+  .function("size", &RolePermissionSet::size)
+  .function("reset", &RolePermissionSet::reset)
+  .function("set", &RolePermissionSet::set)
+  .function("unset", &RolePermissionSet::unset)
+  .function("test", &RolePermissionSet::test)
+  .function("none", &RolePermissionSet::none)
+  .function("isSubsetOf", &RolePermissionSet::isSubsetOf)
+  ;
```
l4l: What about operators?

shared_model/packages/javascript/README.md
```diff
@@ -2,6 +2,8 @@
 
 Official Iroha JavaScript Library. https://github.com/hyperledger/iroha
 
+This package fully compatible with [Iroha 1.0 beta-3](https://github.com/hyperledger/iroha/releases/tag/v1.0.0_beta-2)
```
l4l: Title - `beta-3`
link - `beta-2`
Some of them definitely should be fixed

shared_model/packages/javascript/index.js
```diff
@@ -1,6 +1,80 @@
-var binary = require('node-pre-gyp')
-var path = require('path')
-var bindingPath = binary.find(path.resolve(path.join(__dirname, './package.json')))
-var binding = require(bindingPath)
+/**
```
l4l: Use instead:
```
/**
 * Copyright Soramitsu Co., Ltd. All Rights Reserved.
 */
```

shared_model/packages/javascript/index.js
```diff
@@ -1,6 +1,80 @@
...
+        }
+      }
+    })
+  } else {
+    return obj
+  }
+}
+
+// Export proxyfied embinded classes
+module.exports = Object.entries(Module)
```
l4l: something is really painful with formatting here

shared_model/packages/javascript/scripts/clean.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+echo "Cleaning all generated directories and files..."
```
l4l: Without confirmation looks redundant

shared_model/packages/javascript/package.json
```diff
@@ -35,40 +31,18 @@
...
-    "x64"
-  ],
   "dependencies": {
     "google-protobuf": "^3.5.0",
-    "grpc": "^1.9.1",
-    "node-pre-gyp": "^0.6.39",
-    "npm": "^5.8.0",
-    "run": "^1.4.0",
-    "test": "^0.6.0"
+    "grpc": "^1.12.2"
```
l4l: 1.11 is used

shared_model/packages/javascript/tests/common.js
```diff
@@ -0,0 +1,66 @@
...
+test('Permissions tests', function (t) {
+  t.plan(6)
+  
+  t.comment('RolePermissionSet tests')
+  t.doesNotThrow(() => new RolePermissionSet(), null, 'Should be presented as a class')
+
+  const perms = new RolePermissionSet()
+  t.equals(perms.size(), 43, 'Should have size 43')
+  t.throws(() => perms.set(), ERROR_MESSAGES.ILLEGAL_NUMBER_OF_ARGUMENTS, 'Should throw ILLEGAL_NUMBER_OF_ARGUMENTS')
+  // TODO: Type checking doesn't work with enums, so we need to add checkers in library wrapper
```
l4l: jira id?

shared_model/packages/javascript/tests/common.js
```diff
@@ -0,0 +1,66 @@
+const test = require('tape')
```
l4l: copyright header?

shared_model/utils/emscripten_helpers.hpp
```diff
@@ -0,0 +1,39 @@
+/**
```
l4l: again

shared_model/utils/emscripten_helpers.hpp
```diff
@@ -0,0 +1,39 @@
...
+      throw std::invalid_argument("Cannot pass argument as a Number");
+
+  // Get string with a number by Number.toString() JS method
+  // It prevents automatic conversion in val.as treats val as JS String
+  std::string str_with_ll_uint = input.call<std::string>("toString");
+
+  return std::stoull(str_with_ll_uint);
+}
+
+#endif  // IROHA_EMBIND_HELPERS
```
l4l: missed newline

l4l:
Also, maybe consider leaving the existing node library as is and just adding a new one?

shared_model/packages/javascript/tests/common.js
stinger112: It's not a preferred task now.

shared_model/packages/javascript/scripts/clean.sh
stinger112: No error during script invocation is a confirmation.

shared_model/bindings/model_query_builder_embind.cpp
stinger112: No, it's not visible. 
It's just for internal purposes: to have less the same template code across modules.

shared_model/backend/protobuf/queries/proto_query.hpp
```diff
@@ -81,6 +81,17 @@ namespace shared_model {
 
       interface::types::TimestampType createdTime() const override;
 
+    #ifdef EMSCRIPTEN
```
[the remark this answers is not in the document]
stinger112: I tried for days to solve this issue. 
My first principle during implementation was to completely separate emscripten library logic from the bindings, but I didn't find a solution for this.

In short, the problem is:
We have the methods return simple types like uint64_t. This types cannot be cast to js types automatically by Embind. So, we can't create some kind of js wrapper for such types before they will be converted on the C++ side to something Embind can consume.
Also, we can't simply create child class in C++ side and bind it instead of a binding class. Because our API design, in any case, will return an instance of a binding class. To create an appropriate descendant we at least have to overload all the methods.

So this is the less painful solution then others I found. Adding this conditional inclusion was briefly discussed with @muratovv .

Also, I'm not a good C++ specialist, so if you can suggest some tricky trick to solve this it will be awesome.

shared_model/bindings/model_crypto_embind.cpp
```diff
@@ -0,0 +1,51 @@
...
+using namespace shared_model::bindings;
+using namespace shared_model::crypto;
+
+EMSCRIPTEN_BINDINGS(model_crypto)
+{
+  register_vector<uint8_t>("ByteVector");
+  register_vector<std::string>("StringVector");
+
+  class_<Blob>("Blob")
+  .constructor<const std::string&>()
```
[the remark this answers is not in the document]
stinger112: This is exactly like in constructor of Blob object from __shared_model/cryptography/blob.hpp__.

ASAIK, when some methods in Embind (including constructor) get value from JS, it holds a reference to the object. It leads original object will not be garbage collected until the reference is used inside emscripted code. Probably :D

shared_model/bindings/model_transaction_builder_embind.cpp
stinger112: It's not a preferred or planned task now.
In the next iteration.

stinger112:
@l4l 

> Maybe consider leaving the existing node library as is (e.g moving to node folder_) and just adding a new one?

Impossible. Presence of **bindings.gyp** inside package brokes pipeline and enable SWIG compilation on `npm install`. So I had to remove it.

 > Also stuff like -DCMAKE_PREFIX_PATH=$EMSCRIPTEN/system shouldn't be used for building

It has to point to Emscripten's alternative search path because this is cross-compilation. Please take a look to the Emscripten guides for more details.

shared_model/bindings/model_query_builder_embind.cpp
```diff
@@ -11,56 +11,58 @@ using namespace shared_model::proto;
...
-  
-  class_<Hash, base<Blob>>("Hash")
-  .constructor<const std::string&>()
-  .function("toString", &Hash::toString)
-  ;
 
+  class_<Hash, base<Blob>>("Hash").constructor<const std::string &>().function(
+      "toString", &Hash::toString);
+
+  // TODO 20.06.18 stinger112: Add underlaying types for commented methods
```
l4l: jira id is missing

stinger112: It's not a preferred or planned task now. 
We don't know where we will do it. 
So, before I deleted this comments with todos, then return it after your review. It can not be even implemented at all if we decide so.

What should I do?

shared_model/packages/javascript/tests/helpers.js
```diff
@@ -0,0 +1,21 @@
+/**
```
luckychess: Copyright plz.

shared_model/bindings/model_query_builder_embind.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+using namespace shared_model::crypto;
+
+EMSCRIPTEN_BINDINGS(model_query_builder) {
+  register_vector<Hash>("HashVector");
+
+  class_<Hash, base<Blob>>("Hash").constructor<const std::string &>().function(
+      "toString", &Hash::toString);
+
+  // TODO 20.06.18 stinger112: Add underlaying types for commented methods
```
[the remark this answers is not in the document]
stinger112: It will be implemented in the next iteration of **iroha-lib**, because it's not so trivial and needs additional efforts. It's not possible (and not required) to implement all of the functionality at once,
so we with @laSinteZ decided to do it later.

shared_model/bindings/model_transaction_builder_embind.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "model_proto.hpp"
+#include "model_transaction_builder.hpp"
+
+using namespace emscripten;
+using namespace shared_model;
+using namespace shared_model::bindings;
+using namespace shared_model::proto;
+
+EMSCRIPTEN_BINDINGS(model_transaction_builder) {
+  // TODO 20.06.18 stinger112: Add underlaying types for commented methods
```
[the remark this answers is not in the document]
stinger112: The same as above.

laSinteZ:
@l4l @luckychess I suggest merging this PR with TODO's for now and consider Emscripten as an experimental feature of JS library with not all functionality. Especialy knowhing that our client libraries might be _slightly changed_ in future.

For "all functionality" we have a fallback SWIG version still available: you just need to rename `bindings.gyp.disabled` to `bindings.gyp` (maybe we will have a script for it later)

neewy:
@luckychess was removed from the review list due to his involvement with research-based tasks at the moment

stinger112:
As I see nobody wants to review this PR.

neewy:
@stinger112 @laSinteZ this is not a matter of wish, but I think that people lack skills related to this code. What if we do something like:
1) live code demo? and short explanation of the problem
2) walkthrough reviewers through the code

Usually this helps.

PR #1381, opened 2018-05-26
Remove polymorphic wrapper from query response

lebdron:
### Description of the Change
Continuing #1366 
Replace `boost::get` with `SpecifiedVisitor` because `boost::get` is also used for `boost::optional`, and it cannot be applied to interface.

### Benefits
Even less polymorphic wrapper!

### Possible Drawbacks 
None.

### DIFF
--- a/shared_model/backend/protobuf/query_responses/proto_query_response.hpp
+++ b/shared_model/backend/protobuf/query_responses/proto_query_response.hpp
@@ -67,8 +67,8 @@ namespace shared_model {
       QueryResponse(QueryResponse &&o) noexcept
           : QueryResponse(std::move(o.proto_)) {}
 
-      QueryResponseVariantType get() const override {
-        return *variant_;
+      const QueryResponseVariantType &get() const override {
+        return *ivariant_;
       }
 
       const interface::types::HashType &queryHash() const override {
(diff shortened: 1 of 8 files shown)


shared_model/backend/protobuf/query_responses/proto_query_response.hpp
```diff
@@ -85,18 +68,32 @@ namespace shared_model {
...
-        return loadQueryResponse<ProtoQueryResponseListType>(*proto_);
+        auto &&ar = *proto_;
+        int which =
+            ar.GetDescriptor()->FindFieldByNumber(ar.response_case())->index();
+        return shared_model::detail::variant_impl<ProtoQueryResponseListType>::
+            template load<ProtoQueryResponseVariantType>(
+                std::forward<decltype(ar)>(ar), which);
       }};
 
+      const Lazy<QueryResponseVariantType> ivariant_{
```
l4l: Is there any point in that additional variant?

lebdron: The issue is return type of `get()` method. Boost 1.65.1 `boost::apply_visitor` and `boost::variant::apply` don't accept rvalues, so we need a reference. This can be done by storing the variant in class with minimal changes to the project.

lebdron:
@l4l I think that unsafe methods can be used in tests, because errors result in test failures -- the expected behavior. I would more likely remove the optional from specified visitor to minimize the number of dereferences in tests => simpler code.

PR #1382, opened 2018-05-26
include package name declarations in generated java bindings

onsjkm:
Include package name declarations inside generated java bindings 

Classes inside the generated lib can now be imported from the package. 

Don't foresee any drawbacks

### DIFF
--- a/shared_model/bindings/CMakeLists.txt
+++ b/shared_model/bindings/CMakeLists.txt
@@ -33,7 +33,7 @@ if (SWIG_PYTHON OR SWIG_JAVA OR SWIG_CSHARP OR SWIG_NODE)
     find_package(swig REQUIRED)
     include(${SWIG_USE_FILE})
 
-    if (SWIG_JAVA)
+    if (SWIG_JAVA AND SWIG_JAVA_PKG)
         set(CMAKE_SWIG_FLAGS -package ${SWIG_JAVA_PKG})
         string(REPLACE "." "/" CMAKE_SWIG_OUTDIR  ${SWIG_JAVA_PKG})   
     else()
--- a/shared_model/bindings/CMakeLists.txt
+++ b/shared_model/bindings/CMakeLists.txt
@@ -33,7 +33,13 @@ if (SWIG_PYTHON OR SWIG_JAVA OR SWIG_CSHARP OR SWIG_NODE)
     find_package(swig REQUIRED)
     include(${SWIG_USE_FILE})
 
-    set(CMAKE_SWIG_FLAGS "")
+    if (SWIG_JAVA)
+        set(CMAKE_SWIG_FLAGS -package ${SWIG_JAVA_PKG})
+        set(CMAKE_SWIG_OUTDIR ${SWIG_JAVA_OUTDIR})     
+    else()
+        set(CMAKE_SWIG_FLAGS "")
+    endif()
+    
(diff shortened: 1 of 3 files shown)


shared_model/bindings/CMakeLists.txt
```diff
@@ -33,7 +33,13 @@ if (SWIG_PYTHON OR SWIG_JAVA OR SWIG_CSHARP OR SWIG_NODE)
     find_package(swig REQUIRED)
     include(${SWIG_USE_FILE})
 
-    set(CMAKE_SWIG_FLAGS "")
+    if (SWIG_JAVA)
+        set(CMAKE_SWIG_FLAGS -package ${SWIG_JAVA_PKG})
+        set(CMAKE_SWIG_OUTDIR ${SWIG_JAVA_OUTDIR})     
+    else()
+        set(CMAKE_SWIG_FLAGS "")
```
l4l: Why are you clear out flags here? Doesn't seem necessary for me

shared_model/bindings/CMakeLists.txt
```diff
@@ -33,7 +33,13 @@ if (SWIG_PYTHON OR SWIG_JAVA OR SWIG_CSHARP OR SWIG_NODE)
     find_package(swig REQUIRED)
     include(${SWIG_USE_FILE})
 
-    set(CMAKE_SWIG_FLAGS "")
+    if (SWIG_JAVA)
+        set(CMAKE_SWIG_FLAGS -package ${SWIG_JAVA_PKG})
+        set(CMAKE_SWIG_OUTDIR ${SWIG_JAVA_OUTDIR})
```
l4l: That should be set based on java package, I guess `string(REGEX REPLACE)` might do the thing

shared_model/bindings/CMakeLists.txt
onsjkm: Clearing out flags was the original code. So I left it there to avoid unknown side effects for non java sdks

l4l:
Building irohad (`cmake .. -DSWIG_JAVA=ON`) causes:
```
CMake Error at shared_model/bindings/CMakeLists.txt:38 (string):
  string sub-command REPLACE requires at least four arguments.
```

Btw, examples work fine for me

onsjkm:
@l4l can you try irohad build again? Now pkg name is set only if it's defined and should not have unintended side effects from other builds.

luckychess:
Looks good, thank you @onsjkm for contribution.
I was afraid about possible break of android build script but fortunately this pr doesn't affect it.

bakhtin:
@onsjkm we require all commits to be signed-off (with `git --signoff` option). Seems that your commits do not have this signature. Could you please verify that? `git` should be configured with author and email that matches one in your GitHub profile

l4l:
@onsjkm Actually there's [no need](https://github.com/probot/dco#how-it-works) in "real" signing, the only requirement is in "Signed-off-by: <name> <email>" message, that can be appended by `git commit -s`
Also you may try [shell script](https://gist.github.com/l4l/694ca936ef538a881cdc3604cc12b2e0), though I haven't used it for some time

You will see the a green tick near the DCO pr status instead of this:
screenshot_20180601_011024

onsjkm:
@l4l FYI, the signer script adds message only to the first commit in the list.

PR #1383, opened 2018-05-28
Fix/storage init test

kamilsa:
### Description of the Change

Fixes the storage init test, which fails when environment variables for pg options are not set

### Benefits

No tests are falling anymore

### Possible Drawbacks 

NOne

### DIFF
--- a/test/framework/CMakeLists.txt
+++ b/test/framework/CMakeLists.txt
@@ -27,6 +27,7 @@ add_library(integration_framework
     )
 target_link_libraries(integration_framework
     application
+    integration_framework_config_helper
     )
 
 target_include_directories(integration_framework PUBLIC ${PROJECT_SOURCE_DIR}/test)
--- a/test/framework/CMakeLists.txt
+++ b/test/framework/CMakeLists.txt
@@ -33,7 +33,4 @@ target_link_libraries(integration_framework
 target_include_directories(integration_framework PUBLIC ${PROJECT_SOURCE_DIR}/test)
 
 add_library(integration_framework_config_helper config_helper.cpp)
-target_link_libraries(integration_framework_config_helper
-    logger
-    )
 target_include_directories(integration_framework_config_helper PUBLIC ${PROJECT_SOURCE_DIR}/test)
--- a/test/framework/config_helper.cpp
+++ b/test/framework/config_helper.cpp
@@ -21,19 +21,21 @@
 #include "logger/logger.hpp"
 
 namespace integration_framework {
+
   std::string getPostgresCredsOrDefault(const std::string &default_conn) {
     auto pg_host = std::getenv("IROHA_POSTGRES_HOST");
     auto pg_port = std::getenv("IROHA_POSTGRES_PORT");
     auto pg_user = std::getenv("IROHA_POSTGRES_USER");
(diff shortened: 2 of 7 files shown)


test/module/irohad/ametsuchi/storage_init_test.cpp
```diff
@@ -15,13 +15,17 @@ using namespace iroha::expected;
 class StorageInitTest : public ::testing::Test {
  public:
   StorageInitTest() {
-    std::string env_var;
+    auto pg_host = std::getenv("IROHA_POSTGRES_HOST");
```
lebdron: Please use the existing method instead
https://github.com/hyperledger/iroha/blob/25c90357a796034cbee688a45cb55d1b2163bd08/test/framework/config_helper.cpp#L24-L37

muratovv:
Also, pls, fix DCO and pass CI.

test/module/irohad/ametsuchi/storage_init_test.cpp
kamilsa: how it is better?

lebdron: Because the logic is the same, so no need to maintain it in two different methods.

test/framework/config_helper.cpp
```diff
@@ -21,19 +21,19 @@
...
   std::string getPostgresCredsOrDefault(const std::string &default_conn) {
     auto pg_host = std::getenv("IROHA_POSTGRES_HOST");
     auto pg_port = std::getenv("IROHA_POSTGRES_PORT");
     auto pg_user = std::getenv("IROHA_POSTGRES_USER");
     auto pg_pass = std::getenv("IROHA_POSTGRES_PASSWORD");
-    if (not pg_host) {
-      return default_conn;
-    } else {
+
+    if (pg_host and pg_port and pg_user and pg_pass) {
```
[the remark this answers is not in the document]
kamilsa: Do you have any suggestions how to so it?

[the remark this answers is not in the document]
lebdron: I think we should create a separate task for @l4l suggestion.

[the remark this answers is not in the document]
l4l: Actually that was already done before (though with a bugs):
https://github.com/hyperledger/iroha/pull/1383/files#diff-bd25f78bbc537db41bd7eb3f8a189e4fL21

I don't think that it should be done in other pr

test/framework/config_helper.cpp
```diff
@@ -21,19 +21,19 @@
 #include "logger/logger.hpp"
```
lebdron: Can be removed as it is unused now.

test/framework/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ add_library(integration_framework
     )
 target_link_libraries(integration_framework
     application
+    integration_framework_config_helper
     )
 
 target_include_directories(integration_framework PUBLIC ${PROJECT_SOURCE_DIR}/test)
```
lebdron: Also `integration_framework_config_helper` does not have to be linked with `logger` anymore, so `target_link_libraries` for it can be removed.

PR #1384, opened 2018-05-28
Shared model permissions

### LINKED ISSUE #1045: Permission Implementation Improvement
# Permissions Implementation Improvement

This document provides an overview of a proposal to improve implementation of permissions in shared model.

_Note_: This document does not discuss current permission model, its scope is pure technical and does not affect behavior of the system.

## Permission Model

Each role has a set of permissions, associated with it. Additionally, each account can have so called _grantable permissions_, which are permissions given to it by some other account. Grantable permissions allow account to perform some actions on other account.

In order to execute a command, account needs to have certain permission. Each command has unique permissions for its execution. Also, some commands require additional permissions, e.g _transferAsset_ command requires destination account to have _can_receive_asset_ permission.

Each query has 3 distinct permissions associated with it. They allow executing query on a specific account, domain or globally.

For example account A can have permission to get balance of account A, so he will not have permission to get balance of account B, even if they belong to the same domain. However, if account A has _domain_ permission, to get balance, it would be possible to perform query on account B.

## Current Permission Implementation

Currently, all permissions in Iroha are implemented as strings. This has exactly one benefit - ease of use for developers. Strings are easy to debug and code working with strings is straightforward.

However, strings have several disadvantages. First of all it is easy to mistype them, for clients and developers. They need to be allocated and take up a lot of space. They are not type-safe and thus error-prone.

This proposal tries to solve given issues while maintaining ease of use.

## Enum Based implementation

Enum is a good choice for permissions, since they are just set of constants. It provides type-safety, efficiency, and clarity.

I propose we declare permission Enum - either one, or several for command, query, and grantable permissions, each command and query will have an associated permission as a static const variable of a class.

We would need to define mapping for protobuf and postgres enums. This could be done by assigning same integer values to corresponding permissions in all three enums and then simply performing static_cast. This way we can avoid defining giant switch-cases / maps.

One big issue with enums in C++ is that there is no idiomatic way to get string representation. There are several possible solutions to this issue:

1. Brute force - define mapping from each permission to string

2. Rely on implementation, such as protobuf to provide mapping. Define abstract class like PermissionPrinter, which will have virtual method toString(permission) and implement it for each backend of shared model. Since we can map protobuf enums to our enums, we can use protobuf reflection as one possible implementation.

3. Rely on things such as Boost Preprocessor or other code-generation solutions.

This issue is open for discussion and would like to hear an opinion of other developers.

l4l:
### Description of the Change

Introduction new shared model permissions and `PermissionSet` to work with

### Benefits

### Possible Drawbacks 

There still problem of lacking compile time check, that mapping proto <=> sm is 1 to 1

Test located at `permissions_test.cpp`, that show usage example

### Alternate Designs

Were presented at #1045


shared_model/backend/protobuf/permissions.hpp
```diff
@@ -0,0 +1,59 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IROHA_SHARED_MODEL_PROTO_PERMISSIONS_HPP
+#define IROHA_SHARED_MODEL_PROTO_PERMISSIONS_HPP
+
+#include "interfaces/permissions.hpp"
+
+#include <boost/optional.hpp>
+#include <boost/variant.hpp>
```
nickaleks: Do you need this include?

nickaleks: Is this include needed?

shared_model/backend/protobuf/permissions.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace shared_model {
+  namespace proto {
+    namespace permissions {
+
+      /**
+       * @param perm protocol object for conversion
+       * @return sm object if conversion can be done
+       */
+      boost::optional<interface::permissions::Role> fromTransport(
```
nickaleks: Since you return optional, functions can be made noexcept

shared_model/backend/protobuf/permissions.hpp
```diff
@@ -0,0 +1,59 @@
...
+       * @param sm object for conversion
+       * @return protobuf object
+       */
+      iroha::protocol::GrantablePermission toTransport(
+          interface::permissions::Grantable r);
+      /**
+       * @param sm object for conversion
+       * @return its string representation
+       */
+      std::string toString(interface::permissions::Role r);
```
nickaleks: Should be Grantable here

shared_model/interfaces/permissions.hpp
```diff
@@ -0,0 +1,129 @@
...
+        kSetMyAccountDetail,
+        kTransferMyAssets,
+
+        COUNT
+      };
+    }  // namespace permissions
+
+    template <typename Perm>
+    class PermissionSet
+        : protected std::bitset<static_cast<size_t>(Perm::COUNT)> {
```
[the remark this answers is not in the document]
l4l: `for i in boost::irange(0, set.size()) ...`
Nevertheless afair there is no cases for iterating over the permission set

shared_model/backend/protobuf/permissions.hpp
```diff
@@ -0,0 +1,64 @@
...
+      iroha::protocol::GrantablePermission toTransport(
+          interface::permissions::Grantable r);
+      /**
+       * @param sm object for conversion
+       * @return its string representation
+       */
+      std::string toString(interface::permissions::Grantable r);
+
+      std::vector<std::string> toString(
+          interface::PermissionSet<interface::permissions::Role> set);
```
nickaleks: Pass by reference?

Also add empty line between two methods please.

It would also be nice to add comment, with string example

shared_model/backend/protobuf/permissions.hpp
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+#ifndef IROHA_SHARED_MODEL_PROTO_PERMISSIONS_HPP
+#define IROHA_SHARED_MODEL_PROTO_PERMISSIONS_HPP
+
+#include "interfaces/permissions.hpp"
+
+#include <boost/optional.hpp>
+#include <string>
+#include <vector>
```
nickaleks: Separate includes according to google style guide please

shared_model/backend/protobuf/impl/permissions.cpp
```diff
@@ -0,0 +1,72 @@
...
+      std::string toString(interface::permissions::Grantable r) {
+        return iroha::protocol::GrantablePermission_Name(toTransport(r));
+      }
+
+      std::vector<std::string> toString(
+          const interface::PermissionSet<interface::permissions::Role> &set) {
+        std::vector<std::string> v;
+        for (size_t i = 0; i < set.size(); ++i) {
+          auto perm = static_cast<interface::permissions::Role>(i);
+          if (set[perm]) {
```
grimadas: Can you use `at`?

test/module/shared_model/backend_proto/permissions_test.cpp
```diff
@@ -0,0 +1,161 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "backend/protobuf/permissions.hpp"
+#include <gtest/gtest.h>
+#include <boost/range/algorithm/for_each.hpp>
```
grimadas: is clang applied?

shared_model/backend/protobuf/impl/permissions.cpp
l4l: There's no point, it's iterating over set.size(), so basically there is no overflow, thus no need in check

test/module/shared_model/backend_proto/permissions_test.cpp
l4l: actually it was, dunno why that isn't handled

shared_model/backend/protobuf/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 add_library(shared_model_proto_backend
+    impl.cpp
```
nickaleks: why is impl.cpp here?

shared_model/backend/protobuf/impl/permissions.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "backend/protobuf/permissions.hpp"
+
+namespace shared_model {
+  namespace proto {
+    namespace permissions {
+
+      boost::optional<interface::permissions::Role> fromTransport(
+          iroha::protocol::RolePermission perm) noexcept {
+        return iroha::protocol::RolePermission_IsValid(perm)
+            ? boost::optional<interface::permissions::Role>(
```
nickaleks: do we need explicit constructor here?

l4l: yeah, because of ternary operator

l4l:
Few comments about last commit: static_cast is not allowed for private base classes, so `operator==`/`operator!=` cannot be called, here's the examples:
https://ideone.com/ZgTWmd
https://ideone.com/qotIAr

PR #1388, opened 2018-05-28
Remove polymorphic wrapper from query

lebdron:
### Description of the Change
Continuing #1381 
Remove polymorphic wrapper from query processing factory and query validator

### Benefits
Less polymorphic wrapper.

### Possible Drawbacks 
None.

### DIFF
--- a/shared_model/backend/protobuf/queries/proto_query.hpp
+++ b/shared_model/backend/protobuf/queries/proto_query.hpp
@@ -38,53 +38,23 @@
 #include "backend/protobuf/queries/proto_get_transactions.hpp"
 #include "backend/protobuf/util.hpp"
 
-template <typename... T, typename Archive>
-shared_model::interface::Query::QueryVariantType loadQuery(Archive &&ar) {
-  if (not ar.has_payload()) {
-    throw std::invalid_argument("Query missing payload");
-  }
-  if (ar.payload().query_case()
-      == iroha::protocol::Query_Payload::QueryCase::QUERY_NOT_SET) {
-    throw std::invalid_argument("Missing concrete query");
-  }
-  int which = ar.payload()
-                  .GetDescriptor()
-                  ->FindFieldByNumber(ar.payload().query_case())
-                  ->index_in_oneof();
-  return shared_model::detail::variant_impl<T...>::template load<
-      shared_model::interface::Query::QueryVariantType>(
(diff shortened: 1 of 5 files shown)


shared_model/backend/protobuf/queries/proto_query.hpp
```diff
@@ -139,10 +109,33 @@ namespace shared_model {
...
+
+      using LazyVariantType = Lazy<ProtoQueryVariantType>;
       // ------------------------------| fields |-------------------------------
       // lazy
-      const LazyVariantType variant_{
-          [this] { return loadQuery<ProtoQueryListType>(*proto_); }};
+      const LazyVariantType variant_{[this] {
+        auto &&ar = *proto_;
+        if (not ar.has_payload()) {
+          throw std::invalid_argument("Query missing payload");
```
x3medima17: Is it necessary to use exceptions?

lebdron: It is possible to use assertions here, since these exceptions are for debugging purposes.

x3medima17: We have to handle it somehow

luckychess: Actually not only for debug. Check ```ModelQueryBuilderTest``` - it handles case for possible missing payload. It can be triggered from client libraries so exception is only possible solution here.

shared_model/builders/protobuf/builder_templates/query_template.hpp
```diff
@@ -204,6 +204,13 @@ namespace shared_model {
 
       auto build() const {
         static_assert(S == (1 << TOTAL) - 1, "Required fields are not set");
+        if (not query_.has_payload()) {
+          throw std::invalid_argument("Query missing payload");
```
x3medima17: Do we catch this exception in all places where we use TemplateQueryBuilder?

lebdron: Yes, it is mapped to optional in server code, and caught manually by client library users.

shared_model/backend/protobuf/queries/proto_query.hpp
lebdron: @luckychess Yes, you are correct, my mistake. Since I moved these checks to builder, it should still be OK.

PR #1389, opened 2018-05-28
Split signAndAddSignature of UnsignedWrapper

l4l:
### Description of the Change

Separate logic signAndAddSignature finishing, so that possible to add multiple signatures to built object

### Benefits

SM object can be build with multiple signatures

### Possible Drawbacks 

None


shared_model/builders/protobuf/unsigned_proto.hpp
```diff
@@ -47,11 +47,15 @@ namespace shared_model {
...
+      UnsignedWrapper &signAndAddSignature(const crypto::Keypair &keypair) {
         auto signedBlob = shared_model::crypto::CryptoSigner<>::sign(
             shared_model::crypto::Blob(unsigned_.payload()), keypair);
         unsigned_.addSignature(signedBlob, keypair.publicKey());
         // TODO: 05.12.2017 luckychess think about false case
+        return *this;
+      }
+
+      T finish() {
         return unsigned_;
```
x3medima17: Shouldn't it return a signed object?

l4l: Indeed, wrong name actually. It gets signatures in method above
Though it might be meaningful somehow prohibit getting the object before signing

PR #1396, opened 2018-05-30
Fix block query race condition

lebdron:
### Description of the Change
Same connection was reused from different gRPC threads causing errors.
 - Return block query with unique connection

 - Use mutex when writing to query notifier

 - Remove unused connections from storage

 - Use unique connection for each status query

### Benefits
Stable block query in terms of postgres connections.

### Possible Drawbacks 
A new connection is created on each query 🙈 

### Usage Examples or Tests 
https://github.com/hyperledger/iroha/blob/1443c2611ccc2eb3e14e93e10e8d3cdf2da2fdf2/test/integration/acceptance/query_test.cpp#L74-L91


l4l:
Fix codacy issue plz, don't make it feeling useless :(

PR #1400, opened 2018-05-31
Remove polymorphic wrapper from transaction and transaction collection

lebdron:
### Description of the Change
Continuing #1397 
Replaced intermediate transaction collection in stateful validator with `boost::filtered`
Hopefully only one more PR required to remove polymorphic wrapper from the project
GCC 5.4 required a workaround for set initialization because it used `insert` instead of `emplace` in iterator constructor

### Benefits
Even less polymorphic wrapper

### Possible Drawbacks 
None

### DIFF
--- a/shared_model/backend/protobuf/block.hpp
+++ b/shared_model/backend/protobuf/block.hpp
@@ -119,8 +119,12 @@ namespace shared_model {
       }};
 
       const Lazy<SignatureSetType<proto::Signature>> signatures_{[this] {
-        return SignatureSetType<proto::Signature>(proto_->signatures().begin(),
-                                                  proto_->signatures().end());
+        auto signatures = proto_->signatures()
+            | boost::adaptors::transformed([](const auto &x) {
+                            return proto::Signature(x);
+                          });
+        return SignatureSetType<proto::Signature>(signatures.begin(),
+                                                  signatures.end());
       }};
 
       const Lazy<interface::types::BlobType> payload_blob_{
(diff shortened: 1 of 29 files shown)


shared_model/backend/protobuf/block.hpp
```diff
@@ -127,11 +119,12 @@ namespace shared_model {
...
-        SignatureSetType<proto::Signature> sigs;
-        for (const auto &sig : proto_->signatures()) {
-          sigs.emplace(sig);
-        }
-        return sigs;
+        auto signatures = proto_->signatures()
+            | boost::adaptors::transformed([](const auto &x) {
+                            return proto::Signature(x);
+                          });
+        return SignatureSetType<proto::Signature>(signatures.begin(),
```
kamilsa: Why not just return signatures?

shared_model/backend/protobuf/transaction.hpp
```diff
@@ -112,12 +109,12 @@ namespace shared_model {
...
-                                 SignatureSetType<proto::Signature>{},
-                                 [](auto &&acc, const auto &sig) {
-                                   acc.emplace(sig);
-                                   return std::forward<decltype(acc)>(acc);
-                                 });
+        auto signatures = proto_->signatures()
+            | boost::adaptors::transformed([](const auto &x) {
+                            return proto::Signature(x);
+                          });
+        return SignatureSetType<proto::Signature>(signatures.begin(),
```
kamilsa: same as above

shared_model/backend/protobuf/block.hpp
lebdron: `signatures` type is `boost::range`, and `std::unordered_set` cannot be constructed from it directly.

PR #1401, opened 2018-05-31
Remove polymorphic wrapper

lebdron:
### Description of the Change
Continuing #1400 
The last pull request in the series.
Removed old typedefs from crypto types.

### Benefits
Polymorphic wrapper is no more.

### Possible Drawbacks 
None.


irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
```diff
@@ -20,6 +20,7 @@
 #include <boost/range/irange.hpp>
 #include <numeric>
 #include <random>
+#include "common/types.hpp"
```
x3medima17: No empty line

shared_model/interfaces/common_objects/types.hpp
```diff
@@ -18,12 +18,12 @@
 #ifndef IROHA_SHARED_MODEL_TYPES_HPP
 #define IROHA_SHARED_MODEL_TYPES_HPP
 
-#include <boost/range/any_range.hpp>
 #include <cstdint>
 #include <set>
 #include <string>
 #include <vector>
 
+#include <boost/range/any_range.hpp>
```
x3medima17: Not the right place for this include

lebdron: We have resolved that it is OK according to google codestyle.

PR #1407, opened 2018-06-02
New permissions in validation phase

l4l:
### Description of the Change

New permissions in validation phase

### Benefits

### Possible Drawbacks 

Quite slower, because require some additional allocation (mostly for set)


shared_model/backend/protobuf/from_old.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <map>
+#include <set>
+#include <string>
+#include "backend/protobuf/permissions.hpp"
+
+namespace shared_model {
+  namespace interface {
+    namespace permissions {
+      static inline Role fromOldR(const std::string &s) {
```
luckychess: I suggest not to use such shortening - this one and `fromOldG` just below are quite confusing imo.

test/module/shared_model/backend_proto/permissions_test.cpp
```diff
@@ -172,3 +172,16 @@ TEST(ProtoPermission, PermissionSubset) {
   ASSERT_FALSE(nonsub.isSubsetOf(sub));
   ASSERT_FALSE(nonsub.isSubsetOf(big));
 }
+
+TEST(ProtoPermission, Strings) {
```
luckychess: Given/when/then plz.

test/module/shared_model/backend_proto/permissions_test.cpp
```diff
@@ -172,3 +172,16 @@ TEST(ProtoPermission, PermissionSubset) {
...
+  using Role = shared_model::interface::permissions::Role;
+  using PermSet = shared_model::interface::PermissionSet<Role>;
+  PermSet set{};
+  for (size_t i = 0; i < set.size(); ++i) {
+    set.set(static_cast<Role>(i));
+  }
+  for (const auto &v : toString(set)) {
+    std::cout << v << " ";
```
luckychess: Could we use asserts somehow instead of printing lines?

shared_model/backend/protobuf/from_old.hpp
l4l: they're temporary functions and should be removed after

test/module/shared_model/backend_proto/permissions_test.cpp
l4l: that functions should be here, sry

PR #1415, opened 2018-06-05
Entity-relationship model (RTD)

neewy:
### Description of the Change

Have added description of entity-relationship model in read the docs section. It is related to world state view.

### Benefits

More clear explanation for users who want to figure our what is the state of the chain.

### Possible Drawbacks 

None


docs/source/core_concepts/index.rst
```diff
@@ -8,10 +8,12 @@ We will look through the basics of Iroha in this section.
 .. WARNING:: Docs are constantly updated and this section is going to be improved.
```
Akvinikym: Just a stylish note: maybe, it's better to highlight the word **warning**?

docs/source/core_concepts/er_model.rst
```diff
@@ -0,0 +1,83 @@
+Entity-relationship model
+=========================
+
+Each Hyperledger Iroha peer has a state, called "World State View",
+which is represented by a set of entities and relations between them.
+To explain you more which entities exist in the system and what are the relations,
+this sections includes ER diagram and an explanation for its components.
```
Akvinikym: https://www.italki.com/question/267722
Looks like, it will be correct to say "an explanation of" in this case

docs/source/core_concepts/er_model.rst
```diff
@@ -0,0 +1,83 @@
...
+Signatory
+---------
+
+ - public_key — a public key
+
+Domain
+------
+
+ - domain_id — identifier of a domain
+ - default_role — a default role per user created in the domain, references existing role
```
Akvinikym: Again, a stylish issue - this is the only entry in ER-diagram, which is written with a smaller font. Extending the Domain rectangle by a couple of pixels will make it perfect

docs/source/core_concepts/index.rst
neewy: it is syntax from https://sphinx-rtd-theme.readthedocs.io/en/latest/demo/demo.html#admonitions

docs/source/core_concepts/er_model.rst
neewy: 🙆

docs/source/core_concepts/er_model.rst
neewy: Default export from workbench…

PR #1417, opened 2018-06-05
Fix/docs config sample

luckychess:
### Description of the Change

Add docs for mst_enable parameter in Iroha config.

### Benefits

Docs good, no docs bad.

### Possible Drawbacks 

None.

### DIFF
--- a/docs/source/guides/configuration.rst
+++ b/docs/source/guides/configuration.rst
@@ -15,7 +15,8 @@ at ``example/config.sample``
     "max_proposal_size": 10,
     "proposal_delay": 5000,
     "vote_delay": 5000,
-    "load_delay": 5000
+    "load_delay": 5000,
+    "mst_enable" : false
   }
 
 As you can see, configuration file is a valid ``json`` structure. Let's go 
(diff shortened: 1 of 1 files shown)


docs/source/guides/configuration.rst
```diff
@@ -54,3 +55,6 @@ Environment-specific parameters
 - ``load_delay`` is a waiting time in milliseconds before loading committed 
   block from next peer. We recommend setting this number the same value as 
   ``proposal_delay`` or even higher.
+- ``mst_enable`` enables or disables multisignature transaction support in
+  Iroha. We recommend to set this parameter to ``false`` at the moment until
```
igor-egorov: We recommend *setting* this parameter ...

laSinteZ: >`unless` you really need it maybe?

docs/source/guides/configuration.rst
```diff
@@ -54,3 +55,6 @@ Environment-specific parameters
 - ``load_delay`` is a waiting time in milliseconds before loading committed 
   block from next peer. We recommend setting this number the same value as 
   ``proposal_delay`` or even higher.
+- ``mst_enable`` enables or disables multisignature transaction support in
+  Iroha. We recommend setting this parameter to ``false`` at the moment until
```
l4l: I would suggest removing the last sentence, as long it might be outdated soon and this point might be forgotten

PR #1418, opened 2018-06-06
Fix getting started docs

bakhtin:
### Description of the Change
- Getting started guide did not reflect changes required to run Iroha with Postgres. `config.sample` file from `example` directory uses `localhost` as a Postgres host Iroha should connect to. It actually should connect to the host matching Postgres container name.
- Removed block storage creation step as Docker creates it automatically on container start up.
- The other section suggested to override entrypoint and run Iroha manually. Rewrite it so Iroha uses entrypoint defined on Docker image creation. Supplement `docker run` command with `KEY` environment variable. That is to instruct Iroha container using keys from `example` directory.

### Benefits
Streamlined Iroha launching process

### Possible Drawbacks 
None

### DIFF
--- a/docs/source/getting_started/index.rst
+++ b/docs/source/getting_started/index.rst
@@ -44,29 +44,17 @@ Starting PostgreSQL Container
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Now we need to run ``PostgreSQL`` in a container, attach it to the network you
-have created before, and expose ports for communication:
+have created before:
 
 .. code-block:: shell
 
   docker run --name some-postgres \
   -e POSTGRES_USER=postgres \
   -e POSTGRES_PASSWORD=mysecretpassword \
-  -p 5432:5432 \
   --network=iroha-network \
   -d postgres:9.5
 
-.. note:: If you already have Postgres running on a host system on default port
-  (5432), then you should pick another free port that will be occupied. For
-  example, 5433: ``-p 5433:5432 \``
-
-Creating Blockstore
-^^^^^^^^^^^^^^^^^^^
-Before we run Iroha container, we should create persistent volume to store
-files, storing blocks for the chain. It is done via the following command:
-
-.. code-block:: shell
-
-  docker volume create blockstore
+.. note:: Postgres port can be exposed for debugging. If you need this option, add ``-p 127.0.0.1:5432:5432`` to ``docker run`` command. It binds port 5432 to IP ``127.0.0.1`` on your local machine. You can then use Postgres client to connect and query database.
 
 Configuring Iroha Network
 ^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -90,6 +78,13 @@ In order to get those files, you need to clone the
   save some time and bandwidth. If you want to get a full commit history, you
   can omit this option.
 
+We should adjust Iroha config to match Postgres host it will connect to. Edit ``host`` parameter of ``config.sample`` file in ``./iroha/example`` directory. Set it to ``some-postgres``.
+
+.. Attention:: In the usual situation, you need to provide a config file, generate
+  genesis block and keypair. However, as a part of this guide, we provide an
+  example configuration for you. Please do not use these settings in a
+  production.
+
 Starting Iroha Container
 ^^^^^^^^^^^^^^^^^^^^^^^^
 We are ready to launch our Iroha container. Let's do it with the following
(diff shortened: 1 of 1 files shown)


docs/source/getting_started/index.rst
```diff
@@ -90,58 +78,44 @@ In order to get those files, you need to clone the
   save some time and bandwidth. If you want to get a full commit history, you
   can omit this option.
 
+We should adjust Iroha config to match Postgres host it will connect to. Edit ``host`` parameter of ``config.sample`` file in ``./iroha/example`` directory. Set it to ``some-postgres``.
```
igor-egorov: We can eliminate this extra step if `docker/release/entrypoint.sh` will use `config.docker` instead of `config.sample`. 

I suggest to modify `entrypoint.sh` and left here a note about `iroha:latest` till master branch would not be updated.

docs/source/getting_started/index.rst
```diff
@@ -160,14 +134,13 @@ let's get familiar with ``iroha-cli``
 
   <script src="https://asciinema.org/a/6dFA3CWHQOgaYbKfQXtzApDob.js" id="asciicast-6dFA3CWHQOgaYbKfQXtzApDob" async></script>
```
igor-egorov: Please update ascii cinema "video".

docs/source/getting_started/index.rst
laSinteZ: To be honest, when I was writing this dock NOT using an entrypoint.sh and entering `irohad` command was made on purpose, so newcomers have a chance to understand what's going on in the container (no autostarting everything)

docs/source/getting_started/index.rst
laSinteZ: I can provide source files and we can modify them :)

docs/source/getting_started/index.rst
```diff
@@ -44,29 +44,17 @@ Starting PostgreSQL Container
...
 Now we need to run ``PostgreSQL`` in a container, attach it to the network you
-have created before, and expose ports for communication:
+have created before:
 
 .. code-block:: shell
 
   docker run --name some-postgres \
   -e POSTGRES_USER=postgres \
   -e POSTGRES_PASSWORD=mysecretpassword \
-  -p 5432:5432 \
```
laSinteZ: Will it work w/o exposing the port? Also, what if user has something on this port?

docs/source/getting_started/index.rst
```diff
@@ -44,29 +44,17 @@ Starting PostgreSQL Container
...
   --network=iroha-network \
   -d postgres:9.5
 
-.. note:: If you already have Postgres running on a host system on default port
-  (5432), then you should pick another free port that will be occupied. For
-  example, 5433: ``-p 5433:5432 \``
-
-Creating Blockstore
```
laSinteZ: Please explain why don't we need a blockstore anymore 🙏

docs/source/getting_started/index.rst
```diff
@@ -90,58 +78,44 @@ In order to get those files, you need to clone the
...
+- ``-e KEY=node0 \`` instructs Iroha to use keys from ``example`` directory (``node0.priv`` and ``node0.pub``).
 - ``--network=iroha-network \`` adds our container to previously created
-  ``iroha-network``, so Iroha and Postgres could see each other.
-- ``--entrypoint=/bin/bash \`` Because ``hyperledger/iroha`` has
-  the custom script which runs after starting the container, we want to
-  override it so we can start Iroha Daemon manually.
-- ``hyperledger/iroha:develop`` is the image which has the ``develop``
-  branch.
-
-Launching Iroha Daemon
```
laSinteZ: To be honest, when I was writing this dock NOT using an entrypoint.sh and entering `irohad` command was made on purpose, so newcomers have a chance to understand what's going on in the container (no autostarting everything), so I want people to suffer :D

laSinteZ: So I am not sure if we really need to remove this part, but I guess it will be easier for people (if fe adjust our image)

docs/source/getting_started/index.rst
bakhtin: Since both Iroha and Postgres containers in the same `iroha-network`, it will work without exposing ports on the host. I also put a note about exposing Postgres port to 127.0.0.1 for debugging purposes. In this case it should not be used by any other program

docs/source/getting_started/index.rst
bakhtin: Docker creates volume automatically when running `docker run` if it does not exist.

docs/source/getting_started/index.rst
bakhtin: I will open a PR to use `config.docker` instead of 'config.sample'

bakhtin:
I suggest accepting PR #1426 first. I will then adjust this PR based on comments

igor-egorov:
Contacted @bakhtin: Artem will check the relevance of the PR and will let me know if my actions will be required.

bakhtin:
Closing PR as replacing `config.sample` with `config.docker` is sufficient to make current docs work

PR #1422, opened 2018-06-06
Refactoring/os interfaces

muratovv:
### Description of the Change
Little improvement of OS interface before large addition of BFT OS.
This PR is just clean up before the work.

### Benefits
* remove a redundant method from OS interface
* rename current implementation of OS

### Possible Drawbacks 
The review is required.


irohad/ordering/impl/on_demand_ordering_service.cpp
```diff
@@ -0,0 +1,6 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "on_demand_ordering_service.h"
```
lebdron: Include path should be relative to irohad directory 🙈

l4l: also should be hpp

muratovv: Sorry about that. I an new one in Iroha community.

PR #1427, opened 2018-06-07
Fix jar build script

luckychess:
### Description of the Change
After changing of default java package path to *.java files was also changed but it wasn't reflected in jar building script. This pr fixes this problem.

### Benefits
Without fix jar package is useless.

### Possible Drawbacks 
None.

### DIFF
--- a/example/java/build_library.sh
+++ b/example/java/build_library.sh
@@ -22,7 +22,7 @@ gradle javaDocJar
 cp build/shared_model/bindings/libs/* dist/
 
 cd build/shared_model/bindings
-jar -cvf iroha_lib.jar *.java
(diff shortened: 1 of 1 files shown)


example/java/build_library.sh
```diff
@@ -22,7 +22,7 @@ gradle javaDocJar
 cp build/shared_model/bindings/libs/* dist/
 
 cd build/shared_model/bindings
-jar -cvf iroha_lib.jar *.java
+jar -cvf iroha_lib.jar jp/co/soramitsu/iroha/*.java
```
l4l: Can be `**/*.java`, though not sure that is working for all of the shells

luckychess: Well, let's not try :)

PR #1434, opened 2018-06-07
Use new permissions in builders/bindings

l4l:
### Description of the Change

Builders migrated to use new permissions
Fixed permission/permission set interface a bit
FieldValidator now has 4 validation methods a pair for permission enums and a pair for their sets. Tests added only for used ones (two aren't used actually)
RolePermissionsResponse uses shared model permission

### Benefits

### Possible Drawbacks 

None

### DIFF
--- a/shared_model/bindings/bindings.i
+++ b/shared_model/bindings/bindings.i
@@ -35,6 +35,7 @@ namespace std {
   %template(StringVector) vector<string>;
   %ignore vector<shared_model::crypto::Hash>::vector(size_type);
   %template(HashVector) vector<shared_model::crypto::Hash>;
+  %template(IntVector) vector<int>;
 };
 
 %exception {
@@ -42,6 +43,8 @@ namespace std {
     $action
   } catch (const std::invalid_argument &e) {
     SWIG_exception(SWIG_ValueError, e.what());
+  } catch (const std::out_of_range &e) {
+    SWIG_exception(SWIG_OverflowError, e.what());
   }
 }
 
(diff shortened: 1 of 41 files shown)


shared_model/bindings/model_transaction_builder.hpp
```diff
@@ -175,7 +175,7 @@ namespace shared_model {
        */
       ModelTransactionBuilder grantPermission(
           const interface::types::AccountIdType &account_id,
-          const interface::types::PermissionNameType &permission);
+          interface::permissions::Grantable permission);
```
Akvinikym: const ref is removed for purpose?

l4l: `Grantable` is just a enum, const does nothing special here

shared_model/bindings/bindings.i
```diff
@@ -51,6 +54,19 @@ namespace std {
 %rename(kp_equal) shared_model::crypto::Keypair::operator==;
 %rename(hash_invoke) shared_model::crypto::Hash::Hasher::operator();
 
+%ignore shared_model::interface::PermissionSet::PermissionSet(std::initializer_list<shared_model::interface::permissions::Role>);
```
luckychess: To prevent 503 Swig warnings you can add lines like
```
%rename(opx) shared_model::interface::PermissionSet<shared_model::interface::permissions::Role>::operator[];
%rename(opy) shared_model::interface::PermissionSet<shared_model::interface::permissions::Grantable>::operator[];
```
just above.

shared_model/bindings/bindings.i
```diff
@@ -51,6 +54,19 @@ namespace std {
...
+%ignore shared_model::interface::PermissionSet::PermissionSet(std::initializer_list<shared_model::interface::permissions::Grantable>);
+
+%extend shared_model::interface::PermissionSet {
+  PermissionSet(const std::vector<int> &perms) {
+    shared_model::interface::PermissionSet<Perm> *set = new shared_model::interface::PermissionSet<Perm>;
```
luckychess: I'm a bit worried about this constructor. Does garbage collector in client languages able to correctly remove such objects?

shared_model/interfaces/permissions.hpp
```diff
@@ -100,11 +112,6 @@ namespace shared_model {
...
       PermissionSet<Perm> &operator|=(const PermissionSet<Perm> &r);
       PermissionSet<Perm> &operator^=(const PermissionSet<Perm> &r);
-
-     private:
-      constexpr auto bit(Perm p) const {
-        return static_cast<size_t>(p);
-      }
     };
 
     extern template class PermissionSet<permissions::Role>;
```
luckychess: This lines produce warnings in Swig (320) a probably redundant here and can be removed safely, what do you think?

shared_model/bindings/bindings.i
l4l: That will create unnecessary names that can be used in bindings. Not really a desired behavior

shared_model/bindings/bindings.i
l4l: Sure, swig only support managed languages (afair)

shared_model/interfaces/permissions.hpp
l4l: Yes it causes warning, though cannot be removed:
1. it's **significantly** optimize template instantiations (once, not at each include)
2. it cannot be removed, just because template must be instantiated somewhere (definitions from cpp file must be moved here)

PR #1436, opened 2018-06-07
Use new perms in Queries

l4l:
### Description of the Change

RolePermissionsResponse uses shared model permission

### Benefits

### Possible Drawbacks 

None

### DIFF
--- a/shared_model/backend/protobuf/query_responses/proto_role_permissions_response.hpp
+++ b/shared_model/backend/protobuf/query_responses/proto_role_permissions_response.hpp
@@ -19,6 +19,8 @@
 #define IROHA_SHARED_MODEL_PROTO_ROLE_PERMISSIONS_RESPONSE_HPP
 
 #include "backend/protobuf/common_objects/trivial_proto.hpp"
+#include "backend/protobuf/from_old.hpp"
+#include "backend/protobuf/permissions.hpp"
 #include "interfaces/query_responses/role_permissions.hpp"
 #include "responses.pb.h"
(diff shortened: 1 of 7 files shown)


shared_model/backend/protobuf/query_responses/proto_role_permissions_response.hpp
```diff
@@ -40,24 +42,32 @@ namespace shared_model {
...
+      const Lazy<interface::RolePermissionSet> rolePermissions_{[this] {
         return boost::accumulate(
             rolePermissionsResponse_.permissions(),
-            PermissionNameCollectionType{},
+            interface::RolePermissionSet{},
             [](auto &&permissions, const auto &permission) {
-              permissions.emplace_back(permission);
-              return std::move(permissions);
+              permissions.set(interface::permissions::fromOldR(permission));
+              return std::forward<decltype(permissions)>(permissions);
```
Akvinikym: Is forward here really necessary? Why not keeping move?

l4l:
That's pr is highly depends on #1434, so I will merge them

shared_model/backend/protobuf/query_responses/proto_role_permissions_response.hpp
l4l: Just similarly to that line:
https://github.com/hyperledger/iroha/blob/632f14e0a305e7cde3f14683a6f7d9660e6c36c4/shared_model/backend/protobuf/commands/impl/proto_create_role.cpp#L25

PR #1402, opened 2018-06-01
Feature/get account assets

Solonets:
### Description of the Change

changes getAccountAssets to return all of the assets of account instead of one.

### Possible Drawbacks 

None

PR #1437, opened 2018-06-07
Fix/bindings get account assets

l4l:
### Description of the Change

Binding tests fixes for GetAccountAssets query that was missed in #1402 

### Benefits

Tests passes

### Possible Drawbacks 

None

### Running Tests

`cmake .. -DSWIG_JAVA=ON -DSWIG_PYTHON=ON -DSWIG_NODE=ON`
`make`
`ctest -R (python|java)`

### DIFF
--- a/shared_model/packages/javascript/tests/queryBuilder.js
+++ b/shared_model/packages/javascript/tests/queryBuilder.js
@@ -5,7 +5,7 @@ const accountId = 'admin@test'
 const assetId = 'coin#test'
 
 test('ModelQueryBuilder tests', function (t) {
-  t.plan(50)
+  t.plan(47)
 
   let queryBuilder = new iroha.ModelQueryBuilder()
   const time = (new Date()).getTime()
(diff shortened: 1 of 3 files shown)


shared_model/packages/javascript/tests/queryBuilder.js
```diff
@@ -5,7 +5,7 @@ const accountId = 'admin@test'
 const assetId = 'coin#test'
 
 test('ModelQueryBuilder tests', function (t) {
-  t.plan(50)
+  t.plan(47)
```
nickaleks: What does this line mean?

l4l: number of tests :\

lebdron:
Well, GetAccountAssets tests work in docker develop image after some actions, but other tests fail.
```
  ..................................................x............................................................................terminate called after throwing an instance of 'std:
:out_of_range'
  what():  bitset::set: __position (which is 18446744072523542336) >= _Nb (which is 43)
Aborted (core dumped)



  ---
    operator: throws
    expected: '/Hash set should contain at least one hash/'
    actual:   undefined
    at: Test.<anonymous> (/opt/iroha/iroha/shared_model/packages/javascript/tests/queryBuilder.js:92:5)
    stack: |-
  ...



  24 tests
  126 passed
  1 failed
```

l4l:
javascripts are indeed failing, but that behavior was before the pr and requires shared model permissions (and maybe smth else)

PR #1442, opened 2018-06-08
fix linking error, add protobuf dependency, fix bug

victordrobny:

### Description of the Change
fix linking error on linux(clang 6), add protobuf dependency, fix bug

### Benefits
Iroha now compiles on linux with clang 6


irohad/execution/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ add_library(common_execution
 target_link_libraries(common_execution
         rxcpp
         boost
```
l4l: backend should have boost

PR #1450, opened 2018-06-10
Prepare YAC for BFT OS and gossip vote propagation

lebdron:
### Description of the Change
Change YAC transport and interfaces:
 - Transport Commit and Reject messages replaced with State message
 - Single endpoint used to send state
 - Single methods to process state instead of 3 methods for vote, commit, and reject in YAC related classes
 - YAC interface returns outcome -- commit or reject, so it can be appropriately handled by consensus users
 - Commit/reject is passed to YAC subscribers only after receiving an appropriate message instead of a single vote [IR-1401](https://soramitsu.atlassian.net/browse/IR-1401)

### Benefits
 - Unified network interface which can be used by multiple propagation strategies like multicast or gossip
 - Simplified logic in transport, storage, crypto verifier, message handlers

### Possible Drawbacks 
 - More complex logic in YAC message handler -- probably it is possible to move it to separate class

### Note to reviewers
Since the pull request affects ~22~34 files, it may be difficult to review as a single change/pull request.
Each commit in this branch passes the tests, so it can be safely separated into multiple pull requests by request of reviewers.
Also you can consider reviewing separate commits, as they localize the changes to different entities, such as network notifications, crypto provider, storage, and YAC itself.

### DIFF
--- a/irohad/consensus/yac/storage/yac_vote_storage.hpp
+++ b/irohad/consensus/yac/storage/yac_vote_storage.hpp
@@ -18,8 +18,8 @@
 #ifndef IROHA_YAC_VOTE_STORAGE_HPP
 #define IROHA_YAC_VOTE_STORAGE_HPP
 
-#include <memory>
 #include <boost/optional.hpp>
+#include <memory>
 #include <unordered_map>
 #include <vector>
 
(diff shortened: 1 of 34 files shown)


irohad/consensus/yac/storage/yac_vote_storage.hpp
```diff
@@ -61,38 +67,18 @@ namespace iroha {
         // --------| public api |--------
 
         /**
-         * Insert vote in storage
-         * @param msg - current vote message
+         * Insert votes in storage
+         * @param state - current message with votes
          * @param peers_in_round - number of peers participated in round
-         * @return structure with result of inserting. Nullopt if mgs not valid.
+         * @return structure with result of inserting. Nullopt if msg not valid.
```
muratovv: Maybe rename nullopt with boost::None?

irohad/consensus/yac/storage/yac_vote_storage.hpp
```diff
@@ -61,38 +67,18 @@ namespace iroha {
...
-         * Insert reject message in storage
-         * @param reject - message with votes
-         * @param peers_in_round - number of peers in current consensus round
-         * @return structure with result of inserting.
-         * Nullopt if reject not valid.
-         */
-        boost::optional<Answer> store(RejectMessage reject,
-                                       uint64_t peers_in_round);
+        boost::optional<Answer> store(std::vector<VoteMessage> state,
+                                      uint64_t peers_in_round);
```
muratovv: I have an unnecessary suggestion. Add a typedef for peers_in_round?

PR #1453, opened 2018-06-12
[feature] Flat file interface

Akvinikym:
[IR-1047](https://soramitsu.atlassian.net/browse/IR-1047)

### Description of the Change

Flat file previously existed only in form of header and cpp files, which did not allow to create a test mock out of it. So, interface, from which implementation of flat file is derived, was added, together with related test mock.

### Benefits

Flat file now can be mocked; also, possibility for other implementations of FlatFile interface is opened.

### Possible Drawbacks 

None

### DIFF
--- a/irohad/ametsuchi/CMakeLists.txt
+++ b/irohad/ametsuchi/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_library(ametsuchi
-    impl/flat_file/impl/flat_file_impl.cpp
+        impl/flat_file/flat_file.cpp
     impl/storage_impl.cpp
     impl/temporary_wsv_impl.cpp
     impl/mutable_storage_impl.cpp
--- a/irohad/ametsuchi/CMakeLists.txt
+++ b/irohad/ametsuchi/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_library(ametsuchi
-    impl/flat_file/flat_file.cpp
+        impl/flat_file/impl/flat_file_impl.cpp
     impl/storage_impl.cpp
     impl/temporary_wsv_impl.cpp
     impl/mutable_storage_impl.cpp
@@ -11,7 +11,7 @@ add_library(ametsuchi
     impl/postgres_ordering_service_persistent_state.cpp
     impl/wsv_restorer_impl.cpp
     impl/postgres_options.cpp
-    )
+        impl/flat_file/flat_file.hpp)
 
 target_link_libraries(ametsuchi
     logger
--- a/irohad/ametsuchi/CMakeLists.txt
+++ b/irohad/ametsuchi/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_library(ametsuchi
-        impl/flat_file/flat_file.cpp
+    impl/flat_file/flat_file.cpp
     impl/storage_impl.cpp
     impl/temporary_wsv_impl.cpp
     impl/mutable_storage_impl.cpp
--- a/irohad/ametsuchi/CMakeLists.txt
+++ b/irohad/ametsuchi/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_library(ametsuchi
-        impl/flat_file/impl/flat_file_impl.cpp
+    impl/flat_file/impl/flat_file_impl.cpp
     impl/storage_impl.cpp
     impl/temporary_wsv_impl.cpp
     impl/mutable_storage_impl.cpp
@@ -11,7 +11,7 @@ add_library(ametsuchi
     impl/postgres_ordering_service_persistent_state.cpp
     impl/wsv_restorer_impl.cpp
     impl/postgres_options.cpp
-        impl/flat_file/flat_file.hpp)
+    )
 
 target_link_libraries(ametsuchi
     logger
--- a/irohad/ametsuchi/impl/flat_file/flat_file.cpp
+++ b/irohad/ametsuchi/impl/flat_file/flat_file.cpp
@@ -39,7 +39,7 @@ boost::optional<std::unique_ptr<FlatFile>> FlatFile::create(
   return std::make_unique<FlatFile>(*res, path, private_tag{});
 }
 
-bool FlatFile::add(Identifier id, const std::vector<uint8_t> &block) {
+bool FlatFile::add(Identifier id, const Bytes &block) {
   // TODO(x3medima17): Change bool to generic Result return type
 
   if (id != current_id_ + 1) {
--- a/irohad/ametsuchi/impl/flat_file/flat_file.cpp
+++ b/irohad/ametsuchi/impl/flat_file/flat_file.cpp
@@ -74,7 +74,7 @@ bool FlatFile::add(Identifier id, const std::vector<uint8_t> &block) {
   return true;
 }
 
-boost::optional<std::vector<uint8_t>> FlatFile::get(Identifier id) const {
+boost::optional<Bytes> FlatFile::get(Identifier id) const {
   const auto filename =
       boost::filesystem::path{dump_dir_} / FlatFile::id_to_name(id);
   if (not boost::filesystem::exists(filename)) {
@@ -82,7 +82,7 @@ boost::optional<std::vector<uint8_t>> FlatFile::get(Identifier id) const {
     return boost::none;
   }
   const auto fileSize = boost::filesystem::file_size(filename);
-  std::vector<uint8_t> buf;
+  Bytes buf;
   buf.resize(fileSize);
   boost::filesystem::ifstream file(filename, std::ifstream::binary);
   if (not file.is_open()) {
--- a/irohad/ametsuchi/impl/flat_file/flat_file.cpp
+++ b/irohad/ametsuchi/impl/flat_file/flat_file.cpp
@@ -74,7 +74,7 @@ bool FlatFile::add(Identifier id, const std::vector<uint8_t> &block) {
   return true;
 }
 
-boost::optional<Bytes> FlatFile::get(Identifier id) const {
+boost::optional<FlatFile::Bytes> FlatFile::get(Identifier id) const {
   const auto filename =
       boost::filesystem::path{dump_dir_} / FlatFile::id_to_name(id);
   if (not boost::filesystem::exists(filename)) {
--- a/irohad/ametsuchi/impl/flat_file/flat_file.cpp
+++ b/irohad/ametsuchi/impl/flat_file/flat_file.cpp
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include "flat_file.hpp"
+#include "ametsuchi/impl/flat_file/flat_file.hpp"
 
 #include <boost/filesystem.hpp>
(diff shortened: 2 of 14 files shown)


irohad/ametsuchi/impl/flat_file/impl/flat_file_impl.hpp
```diff
@@ -0,0 +1,128 @@
+/**
+ * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
+ * http://soramitsu.co.jp
```
lebdron: Please use new header as in `flat_file.hpp`

irohad/ametsuchi/impl/postgres_block_query.hpp
```diff
@@ -23,25 +23,25 @@
...
+    class FlatFileImpl;
 
     /**
      * Class which implements BlockQuery with a Postgres backend.
      */
     class PostgresBlockQuery : public BlockQuery {
      public:
       PostgresBlockQuery(pqxx::nontransaction &transaction_,
-                         FlatFile &file_store);
+                         FlatFileImpl &file_store);
```
lebdron: Original idea was to test `PostgresBlockQuery` without concrete implementation as I understand. Using FlatFileImpl here will not allow passing a mock to constructor. So `FlatFile` should be used in this file instead if I am not mistaken.

irohad/ametsuchi/impl/storage_impl.hpp
```diff
@@ -99,7 +99,7 @@ namespace iroha {
       const PostgresOptions postgres_options_;
 
      private:
-      std::unique_ptr<FlatFile> block_store_;
+      std::unique_ptr<FlatFileImpl> block_store_;
```
lebdron: Maybe use `FlatFile` here so that it is possible to use mock in the future? Because now it is not possible due to static methods hardcoded with `FlatFileImpl`.

irohad/ametsuchi/impl/storage_impl.hpp
```diff
@@ -30,12 +30,12 @@
...
 
-    class FlatFile;
+    class FlatFileImpl;
 
     struct ConnectionContext {
-      explicit ConnectionContext(std::unique_ptr<FlatFile> block_store);
+      explicit ConnectionContext(std::unique_ptr<FlatFileImpl> block_store);
 
-      std::unique_ptr<FlatFile> block_store;
+      std::unique_ptr<FlatFileImpl> block_store;
```
lebdron: Same for `FlatFile` here.

test/module/irohad/ametsuchi/ametsuchi_mocks.hpp
```diff
@@ -246,6 +247,16 @@ namespace iroha {
       }
     };
 
+    class MockFlatFile : public FlatFile {
+     public:
+      MOCK_METHOD2(add, bool(Identifier id, const std::vector<uint8_t> &blob));
+      MOCK_CONST_METHOD1(get,
+                         boost::optional<std::vector<uint8_t>>(Identifier id));
+      MOCK_CONST_METHOD0(directory, std::string(void));
```
l4l: void is redundant, similar to function args name (e.g `id` above)

irohad/ametsuchi/impl/postgres_block_query.hpp
Akvinikym: You are right, auto-refactoring changed that, I did not see

test/module/irohad/ametsuchi/ametsuchi_mocks.hpp
Akvinikym: Void is used everywhere upwards, so I thought it was a good idea to stick to existing code style?

irohad/ametsuchi/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 add_library(ametsuchi
-    impl/flat_file/flat_file.cpp
+        impl/flat_file/flat_file.cpp
```
lebdron: Indent is not consistent :( Please check your IDE settings.

irohad/ametsuchi/impl/flat_file/flat_file.cpp
```diff
@@ -1,21 +1,10 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
  */
 
-#include "ametsuchi/impl/flat_file/flat_file.hpp"
+#include "flat_file.hpp"
```
lebdron: Please use file path relative to irohad directory. `ametsuchi/impl/flat_file/flat_file.cpp`

irohad/ametsuchi/impl/flat_file/flat_file.cpp
```diff
@@ -37,7 +26,7 @@ std::string FlatFile::id_to_name(Identifier id) {
 
 boost::optional<std::unique_ptr<FlatFile>> FlatFile::create(
     const std::string &path) {
-  auto log_ = logger::log("FlatFile::create()");
+  auto log_ = logger::log("KeyValueStorage::create()");
```
lebdron: `FlatFile` seemed more appropriate, because this method is not called from `KeyValueStorage`.

irohad/ametsuchi/impl/flat_file/flat_file.cpp
```diff
@@ -121,10 +110,10 @@ void FlatFile::dropAll() {
 // ----------| private API |----------
 
 FlatFile::FlatFile(Identifier current_id,
-                   const std::string &path,
-                   FlatFile::private_tag)
+                           const std::string &path,
```
lebdron: Is this the right indentation? Previous seemed more consistent.

irohad/ametsuchi/impl/flat_file/flat_file.cpp
```diff
@@ -121,10 +110,10 @@ void FlatFile::dropAll() {
...
+                           FlatFile::private_tag)
     : dump_dir_(path) {
-  log_ = logger::log("FlatFile");
+  log_ = logger::log("KeyValueStorage");
```
lebdron: Same as above, `FlatFile` seems more appropriate to me.

irohad/ametsuchi/impl/flat_file/flat_file.hpp
```diff
@@ -125,8 +91,8 @@ namespace iroha {
        * @param path - folder of storage
        */
       FlatFile(Identifier last_id,
-               const std::string &path,
-               FlatFile::private_tag);
+                   const std::string &path,
```
lebdron: Same as above, is the indentation OK?

irohad/ametsuchi/impl/flat_file/flat_file.hpp
```diff
@@ -1,28 +1,15 @@
...
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
  */
 
-#ifndef IROHA_FLAT_FILE_HPP
-#define IROHA_FLAT_FILE_HPP
+#ifndef IROHA_FLAT_FILE_IMPL_HPP
```
lebdron: Please rename to `IROHA_FLAT_FILE_HPP`, so that it is consistent with filename. Same in the end of this file.

irohad/ametsuchi/key_value_storage.hpp
```diff
@@ -0,0 +1,59 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IROHA_FLAT_FILE_HPP
```
lebdron: `IROHA_FLAT_FILE_HPP` -> `IROHA_KEY_VALUE_STORAGE_HPP`

irohad/ametsuchi/CMakeLists.txt
Akvinikym: Hmm, very strange issues with indentation, when I push I see perfectly fine picture, but after that such things happen... I'll check it

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -17,7 +17,7 @@
 
 #include "ametsuchi/impl/storage_impl.hpp"
 #include <boost/format.hpp>
-#include "ametsuchi/impl/flat_file/flat_file.hpp"  // for FlatFile
+#include "ametsuchi/impl/flat_file/flat_file.hpp"  // for FlatFiles
```
nickaleks: Why FlatFiles?

nickaleks: Also don't think this comment is needed at all

irohad/ametsuchi/key_value_storage.hpp
```diff
@@ -0,0 +1,59 @@
...
+       * Type of storage key
+       */
+      using Identifier = uint32_t;
+
+      /**
+       * Add entity with binary data
+       * @param id - reference key
+       * @param blob - data associated with key
+       */
+      virtual bool add(Identifier id, const std::vector<uint8_t> &blob) = 0;
```
nickaleks: In shared_model/cryptography/blob.hpp we have Bytes type, which is std::vector<uint8_t>. Maybe use it here? I am not sure about adding a dependency to shared model, but we can make it a separate declaration.

irohad/ametsuchi/key_value_storage.hpp
```diff
@@ -0,0 +1,59 @@
...
+      virtual boost::optional<std::vector<uint8_t>> get(
+          Identifier id) const = 0;
+
+      /**
+       * @return folder of storage
+       */
+      virtual std::string directory() const = 0;
+
+      /**
+       * @return maximal not null key
```
nickaleks: Comment is not well worded. It implies a total ordering of the keys, but the name of the function last_id implies chronological ordering. Thus interface contradicts documentation.

For example if we add keys in this order: 1 -> 3 -> 2. Name of the function suggests that key 2 is returned, but comment tells us that we should expect 3.

Akvinikym: @return chronologically last not null key?

test/module/irohad/ametsuchi/ametsuchi_mocks.hpp
```diff
@@ -246,6 +247,16 @@ namespace iroha {
       }
     };
 
+    class MockKeyValueStorage : public KeyValueStorage {
+     public:
+      MOCK_METHOD2(add, bool(Identifier, const std::vector<uint8_t> &));
```
nickaleks: Change to bytes please

PR #1454, opened 2018-06-12
Fix variable forward for variadic in Windows build

lebdron:
### Description of the Change
Windows build fails due to "`Args` are not unpacked" error, but it works with `decltype(args)`, which serves the same purpose.

### Benefits
Successful Windows build

### Possible Drawbacks 
None

### Note
Build of python bindings fails due to CI update, so there is an error displayed. The code itself builds fine.


lebdron:
jenkins/branch fails due to missing python 3 in CI, which should be fixed in another pull request, so that it will be possible to test python bindings on Windows. Shared model build works.
jenkins/pr-head fails due to the following nondeterministic test:
https://github.com/hyperledger/iroha/blob/aa96e9eb6079586cf775fb87279dabd9206a1fb7/test/integration/acceptance/transfer_asset_test.cpp#L285-L292
Transactions can be received by OS in any order, so this should be also fix in another pull request.

PR #1460, opened 2018-06-14
Blocks streaming

kamilsa:
### Description of the Change

Adds block streaming endpoint to query service. 
### Benefits

Allows to subscribe on committed blocks in iroha

### Possible Drawbacks 

No integration test is provided 

### Usage Examples or Tests

Query service test has tests on BlocksQuery.
block-query.py added in example/python. Run iroha, execute this script to subscribe on blocks and send several transactions to check blocks-query


example/python/blocks-query.py
```diff
@@ -0,0 +1,49 @@
...
+    query_blob = iroha.ModelProtoBlocksQuery(query).signAndAddSignature(key_pair).finish().blob()
+    proto_query = queries_pb2.BlocksQuery()
+
+    if sys.version_info[0] == 2:
+        tmp = ''.join(map(chr, query_blob))
+    else:
+        tmp = bytes(query_blob)
+
+    proto_query.ParseFromString(tmp)
+    # print(proto_query.DebugString())
```
x3medima17: I think we don't need it

irohad/torii/impl/query_service.cpp
```diff
@@ -25,6 +26,7 @@ namespace torii {
   QueryService::QueryService(
       std::shared_ptr<iroha::torii::QueryProcessor> query_processor)
       : query_processor_(query_processor) {
+    log_ = logger::log("Query Service");
```
x3medima17: We can do it using delegation

irohad/torii/impl/query_service.cpp
```diff
@@ -88,4 +90,71 @@ namespace torii {
...
+                                    proto_block_error_response.getTransport(),
+                                    grpc::WriteOptions());
+                              });
+                        }
+                      });
+            },
+            [this, writer](const auto &error) {
+              log_->info("Stateless invalid: {}", error.error);
+              iroha::protocol::BlockQueryResponse response;
+              auto error_response = response.mutable_error_response();
```
x3medima17: Don't we have builders for errors?

irohad/torii/impl/query_service.cpp
```diff
@@ -88,4 +90,71 @@ namespace torii {
...
+                                        block_error_response);
+                                writer->WriteLast(
+                                    proto_block_error_response.getTransport(),
+                                    grpc::WriteOptions());
+                              });
+                        }
+                      });
+            },
+            [this, writer](const auto &error) {
+              log_->info("Stateless invalid: {}", error.error);
```
x3medima17: It is either a warning, or an error

test/module/irohad/torii/query_service_test.cpp
```diff
@@ -164,3 +176,121 @@ TEST_F(QueryServiceTest, InvalidWhenDuplicateHash) {
...
+
+  auto client = torii_utils::QuerySyncClient(ip, port);
+  std::vector<iroha::protocol::BlockQueryResponse> responses;
+  auto proto_blocks_query =
+      std::static_pointer_cast<shared_model::proto::BlocksQuery>(blocks_query);
+  client.FetchCommits(proto_blocks_query->getTransport(), responses);
+
+  ASSERT_EQ(responses.size(), 1);
+  auto response = responses.at(0);
+  ASSERT_TRUE(response.has_error_response());
```
x3medima17: Why don't we check value of the error?

irohad/torii/impl/query_service.cpp
kamilsa: I think in query service this is unnecessary, since this is transport layer and we know about protobuf and grpc here

irohad/torii/impl/query_service.cpp
kamilsa: Should iroha log a warning if it received stateless invalid query? I don't think so, this is legal behaviour for iroha

test/module/irohad/torii/query_service_test.cpp
kamilsa: You right. Now block error response is very dumb. We should update it in future.

irohad/torii/impl/query_service.cpp
x3medima17: Summon @luckychess

irohad/torii/impl/query_service.cpp
```diff
@@ -88,4 +89,71 @@ namespace torii {
...
+                      sub,
+                      [this, context, &sub, request, writer](
+                          const std::shared_ptr<
+                              shared_model::interface::BlockQueryResponse>
+                              response) {
+                        if (context->IsCancelled()) {
+                          log_->info("Unsubscribed");
+                          sub.unsubscribe();
+                        } else {
+                          log_->info("{} receives committed block",
```
x3medima17: Actually it might receive an error, like stateful invalid

irohad/torii/impl/query_service.cpp
```diff
@@ -88,4 +89,76 @@ namespace torii {
     return grpc::Status::OK;
   }
 
+  grpc::Status QueryService::FetchCommits(
+      grpc::ServerContext *context,
+      const iroha::protocol::BlocksQuery *request,
+      grpc::ServerWriter<iroha::protocol::BlockQueryResponse> *writer) {
+    log_->info("Fetching commits");
```
lebdron: These are rather debug statements, so `debug` can be used instead as it is done in `CommandService::StatusStream`. Same applies to logs below.

test/module/irohad/torii/query_service_test.cpp
```diff
@@ -164,3 +176,121 @@ TEST_F(QueryServiceTest, InvalidWhenDuplicateHash) {
           shared_model::interface::StatelessFailedErrorResponse>(),
       resp.get()));
 }
+
+class BlocksQueryServiceTest : public ::testing::Test {
```
lebdron: It seems that this test is rather similar to `ToriiQueriesTest`. Consider moving it to `torii_queries_test.cpp` and reusing the fixture, because it contains mostly the same code.

test/module/irohad/torii/query_service_test.cpp
```diff
@@ -17,10 +17,14 @@
 
 #include "torii/query_service.hpp"
 #include "backend/protobuf/query_responses/proto_query_response.hpp"
+#include "builders/default_builders.hpp"
```
lebdron: Please revert changes in this file.

test/module/irohad/torii/torii_service_query_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "backend/protobuf/query_responses/proto_query_response.hpp"
+#include "builders/default_builders.hpp"
+#include "builders/protobuf/queries.hpp"
+#include "main/server_runner.hpp"
+#include "module/irohad/torii/torii_mocks.hpp"
+#include "module/shared_model/builders/protobuf/test_query_builder.hpp"
+#include "module/shared_model/builders/protobuf/test_query_response_builder.hpp"
+#include "torii/query_client.hpp"
+#include "torii/query_service.hpp"
+#include "utils/query_error_response_visitor.hpp"
```
lebdron: `QueryErrorResponseChecker` is not used in this file, please verify that no unused includes are present.

test/module/irohad/torii/CMakeLists.txt
```diff
@@ -37,4 +37,13 @@ target_link_libraries(torii_queries_test
 addtest(query_service_test query_service_test.cpp)
 target_link_libraries(query_service_test
     torii_service
+    server_runner
```
lebdron: Query service test is not changed in this pull request, so these libraries should not be present here.

PR #1461, opened 2018-06-15
Feature/hash from hex

Alexey-N-Chernyshov:
### Description of the Change
Add Hash constructor from Blob to be able to create Hash from hex string in Java bindings.

### Benefits
Can create Hash from hex string in Java bindings. It is used for query getTransactions().

### Possible Drawbacks 
no

### DIFF
--- a/docs/source/guides/deployment.rst
+++ b/docs/source/guides/deployment.rst
@@ -54,6 +54,8 @@ In case of valid assumptions, the only thing that remains is to launch the daemo
 |               | used by peer to sign the blocks                                 |
 +---------------+-----------------------------------------------------------------+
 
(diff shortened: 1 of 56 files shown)


l4l:
What's the problem with: `Hash(blob.hex())`?

Alexey-N-Chernyshov:
@l4l, 'blob.hex()' returns hex representation while `Hash(string)` constructor requires binary string. There was no way to create `Hash` from hex string in java bindings.

l4l:
Ah, I see. What about `toBinaryString`:
https://github.com/hyperledger/iroha/blob/f138d79a3fc17b329b61bc31a7463649d11b6978/shared_model/cryptography/blob.hpp#L32

Works in python for me:
screenshot_20180615_145430

Alexey-N-Chernyshov:
It doesn't work in Java. I think it is because of encoding. Moreover the binary string shouldn't appear in client sources.

PR #1469, opened 2018-06-17
Transaction sequence interface

kamilsa:

### Description of the Change

In order to implement transaction batches inside iroha we might need to operate with transaction's sequences, where transactions from the same batch should be sequentially placed inside collection. This is not fair implementation, this is just interface. To have fair implementation less operator should be implemented

### Benefits

Transaction sequence can be used, without fair logic at the moment

### Possible Drawbacks 

No fair implementation

### DIFF
--- a/shared_model/interfaces/common_objects/types.hpp
+++ b/shared_model/interfaces/common_objects/types.hpp
@@ -92,6 +92,10 @@ namespace shared_model {
           boost::any_range<Transaction,
                            boost::random_access_traversal_tag,
                            const Transaction &>;
+      using TransactionForwardCollectionType =
+          boost::any_range<Transaction,
+                           boost::forward_traversal_tag,
+                           const Transaction &>;
       using AccountAssetCollectionType =
           boost::any_range<AccountAsset,
                            boost::random_access_traversal_tag,
--- a/shared_model/interfaces/common_objects/types.hpp
+++ b/shared_model/interfaces/common_objects/types.hpp
@@ -92,10 +92,6 @@ namespace shared_model {
           boost::any_range<Transaction,
                            boost::random_access_traversal_tag,
                            const Transaction &>;
-      using TransactionForwardCollectionType =
-          boost::any_range<Transaction,
-                           boost::forward_traversal_tag,
-                           const Transaction &>;
       using AccountAssetCollectionType =
           boost::any_range<AccountAsset,
                            boost::random_access_traversal_tag,
--- a/shared_model/interfaces/iroha_internal/transaction_sequence.cpp
+++ b/shared_model/interfaces/iroha_internal/transaction_sequence.cpp
@@ -4,16 +4,32 @@
  */
 
 #include "interfaces/iroha_internal/transaction_sequence.hpp"
+#include "validators/transaction_sequence_validator.hpp"
 
 namespace shared_model {
   namespace interface {
 
-    bool TransactionComparator::operator()(
-        const std::shared_ptr<Transaction> &lhs,
-        const std::shared_ptr<Transaction> &rhs) const {
-      // implement fair logic here
-      return lhs->hash().hex() < rhs->hash().hex();
+    TransactionSequence::TransactionSequence(
+        const boost::any_range<Transaction, boost::forward_traversal_tag>
+            &transactions)
+        : transactions_(transactions) {}
+
+    template <typename Validator>
+    iroha::expected::Result<TransactionSequence, std::string>
+    TransactionSequence::createTransactionSequence(
+        const types::TransactionsCollectionType &transactions) {
+      Validator validator{};
+      auto answer = validator.validate(transactions);
+      if (answer.hasErrors()) {
+        return iroha::expected::makeError(answer.reason());
+      }
+      return iroha::expected::makeValue(TransactionSequence(transactions));
     }
+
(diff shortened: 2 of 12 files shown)


shared_model/interfaces/iroha_internal/transaction_sequence.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include "interfaces/iroha_internal/transaction_sequence.hpp"
+
+namespace shared_model {
+  namespace interface {
+
+    bool TransactionComparator::operator()(
+        const std::shared_ptr<Transaction> &lhs,
+        const std::shared_ptr<Transaction> &rhs) const {
+      // implement fair logic here
```
muratovv: Implementation of this method will be not obvious because the operator must stay in butches in one order. I can't imagine such implementation. Do you have an algorithm for that?

shared_model/interfaces/iroha_internal/transaction_sequence.hpp
```diff
@@ -0,0 +1,32 @@
...
+      bool operator()(const std::shared_ptr<Transaction> &lhs,
+                      const std::shared_ptr<Transaction> &rhs) const;
+    };
+
+    /**
+     * Transaction sequence is the collection of transactions, where
+     * transactions from the same batch are placed contiguously
+     */
+    using TransactionSequence =
+        std::set<std::shared_ptr<Transaction>, TransactionComparator>;
```
muratovv: I am not sure about `set` here. Why is it not a vector?

shared_model/interfaces/iroha_internal/transaction_sequence.hpp
```diff
@@ -0,0 +1,32 @@
...
+    struct TransactionComparator {
+      bool operator()(const std::shared_ptr<Transaction> &lhs,
+                      const std::shared_ptr<Transaction> &rhs) const;
+    };
+
+    /**
+     * Transaction sequence is the collection of transactions, where
+     * transactions from the same batch are placed contiguously
+     */
+    using TransactionSequence =
```
muratovv: Using not guarantee that we store our collection in a proper way. Maybe rework it with a delegation? I mean, create a class with will hold collection with objects and getter for the collection.

shared_model/interfaces/iroha_internal/transaction_sequence.cpp
kamilsa: 1. Every transaction without batch should be less than any transaction without batch.
2. Between two transactions from different batches, the transaction from the batch where tx_hashes is less than tx_hashes from another batch will be less
3. Two transactions from the same batch are sorted using tx_hashes array

shared_model/interfaces/iroha_internal/transaction_sequence.hpp
kamilsa: Set guarantees the order and uniqueness

shared_model/interfaces/iroha_internal/transaction_sequence.hpp
kamilsa: What do you mean by proper way? I think we should have iteratable interface with insert method which stores transactions in proper order. Set suits here. Any other requirements you see there?

shared_model/interfaces/common_objects/types.hpp
```diff
@@ -92,6 +92,10 @@ namespace shared_model {
           boost::any_range<Transaction,
                            boost::random_access_traversal_tag,
                            const Transaction &>;
+      using TransactionForwardCollectionType =
```
lebdron: Please don't add types to this file. Any modification of this file triggers the rebuild of almost the whole project, because of transitive includes. This definition is better to be locally defined in transaction_sequence.hpp

shared_model/interfaces/iroha_internal/transaction_sequence.hpp
```diff
@@ -0,0 +1,57 @@
...
+     public:
+      /**
+       * Creator of transaction sequence
+       * @tparam Validator checks validation logic
+       * @param transactions collection of transactions
+       * @return valid transaction sequence
+       */
+      template <typename Validator>
+      iroha::expected::Result<TransactionSequence, std::string>
+      createTransactionSequence(
```
muratovv: There is required to pass an instance of `Validator` type

shared_model/interfaces/iroha_internal/transaction_sequence.hpp
```diff
@@ -0,0 +1,57 @@
...
+    namespace types {
+      using TransactionForwardCollectionType =
+          boost::any_range<Transaction,
+                           boost::forward_traversal_tag,
+                           const Transaction &>;
+    }
+
+    /**
+     * Transaction sequence is the collection of transactions where:
+     * 1. All transactions from the same batch are place contiguously
```
muratovv: Those rules are related to validation. Maybe better to move it to validation class?

shared_model/interfaces/iroha_internal/transaction_sequence.hpp
muratovv: Function, also, should be static.

neewy:
Why such feature is not implemented in a trunk? @kamilsa @muratovv

PR #1472, opened 2018-06-18
Get top block function

Akvinikym:
### Description of the Change

Previously only method getTopBlocks(int) existed, but all of its calls were getTopBlocks(1) to get the top-most block, so an additional function was added, getTopBlock(), and all calls were substituted.

### Benefits

A new, more specific function, which is more convenient to use in such cases.

### Possible Drawbacks 

None.


irohad/ametsuchi/impl/postgres_block_query.cpp
```diff
@@ -242,5 +242,19 @@ namespace iroha {
       return block_store_.last_id();
     }
 
+    expected::Result<BlockQuery::wBlock, std::string>
+    PostgresBlockQuery::getTopBlock() {
+      auto block =
+          block_store_.get(block_store_.last_id()) | [](const auto &bytes) {
+            return shared_model::converters::protobuf::jsonToModel<
```
nickaleks: We have protobuf leaking here. We need a task to implement converters which return interface objects. Could you add a task for that?

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -74,17 +74,17 @@ namespace iroha {
...
+      auto block_result = getBlockQuery()->getTopBlock();
       return expected::makeValue<std::unique_ptr<MutableStorage>>(
           std::make_unique<MutableStorageImpl>(
-              top_hash.value_or(shared_model::interface::types::HashType("")),
+              block_result.match(
+                  [](expected::Value<
+                      std::shared_ptr<shared_model::interface::Block>> block) {
+                    return block.value->hash();
+                  },
+                  [](expected::Error<std::string> error) {
```
nickaleks: You can omit the name of the parameter if it is unused. Also pass by reference to avoid extra copies

irohad/main/impl/ordering_init.cpp
```diff
@@ -6,13 +6,24 @@
...
 namespace iroha {
   namespace network {
     auto OrderingInit::createGate(
         std::shared_ptr<OrderingGateTransport> transport,
         std::shared_ptr<ametsuchi::BlockQuery> block_query) {
-      auto height = block_query->getTopBlocks(1).as_blocking().last()->height();
+      shared_model::interface::types::HeightType height;
+      block_query->getTopBlock().match(
+          [&height](
+              expected::Value<std::shared_ptr<shared_model::interface::Block>>
```
nickaleks: Same here, prefer pass by reference

irohad/main/impl/ordering_init.cpp
```diff
@@ -6,13 +6,24 @@
...
+                  block) { height = block.value->height(); },
```
nickaleks: To avoid declaring height out of the scope, we can write createGate like this:
image

irohad/main/irohad.cpp
```diff
@@ -173,8 +173,12 @@ int main(int argc, char *argv[]) {
...
   // check if at least one block is available in the ledger
   auto blocks_exist = false;
-  irohad.storage->getBlockQuery()->getTopBlocks(1).subscribe(
-      [&blocks_exist](auto block) { blocks_exist = true; });
+  irohad.storage->getBlockQuery()->getTopBlock().match(
+      [&blocks_exist](iroha::expected::Value<
+                      std::shared_ptr<shared_model::interface::Block>> block) {
+        blocks_exist = true;
+      },
+      [](iroha::expected::Error<std::string> error) {});
```
nickaleks: Can be simplified to:
image

irohad/simulator/impl/simulator.cpp
```diff
@@ -66,17 +66,25 @@ namespace iroha {
         const shared_model::interface::Proposal &proposal) {
       log_->info("process proposal");
       // Get last block from local ledger
-      block_queries_->getTopBlocks(1)
-          .subscribe_on(rxcpp::observe_on_new_thread())
-          .as_blocking()
-          .subscribe([this](auto block) { last_block = block; });
-      if (not last_block) {
-        log_->warn("Could not fetch last block");
+      auto block_fetched = false;
```
nickaleks: You can use the same scheme as with block_loader_impl to avoid extra bool. It is still not very pretty, but we need some additional refactoring to make it prettier, so this is out of scope of this pr

test/module/irohad/ametsuchi/block_query_test.cpp
```diff
@@ -389,3 +397,47 @@ TEST_F(BlockQueryTest, HasTxWithInvalidHash) {
...
 }
+
+/**
+ * @given block store with preinserted blocks
+ * @when getTopBlock is invoked on this block store
+ * @then returned top block's height is equal to the inserted one's
+ */
+TEST_F(BlockQueryTest, GetTopBlockSuccess) {
+  boost::optional<shared_model::interface::types::HeightType> top_block_height;
+  blocks->getTopBlock().match(
```
nickaleks: Please use methods from result_fixture to simplify the code

nickaleks: Also we need to verify that returned block is indeed the top one

irohad/ametsuchi/impl/postgres_block_query.cpp
lebdron: These tasks should be sufficient:
https://soramitsu.atlassian.net/browse/IR-937 Rework shared_model interface with getTransport method with default type
https://soramitsu.atlassian.net/browse/IR-1040 Add transport retrieval method for interfaces 
They both cover the issue with dependency injection

irohad/simulator/impl/simulator.cpp
Akvinikym: To perform he same schema here, I should nullify a class variable last_block, which does not sound very good... another variant is to place the rest of function under lambda "good" branch, but this is even worse

test/module/irohad/ametsuchi/block_query_test.cpp
Akvinikym: The top-most attribute of the block is verified in 418 line via height comparison, which is unique for both blocks in this store, and only the top-most is of height 2

irohad/simulator/impl/simulator.cpp
nickaleks: I see, then we can return boolean from match and initialize block_fetched there

test/module/irohad/ametsuchi/block_query_test.cpp
```diff
@@ -389,3 +398,29 @@ TEST_F(BlockQueryTest, HasTxWithInvalidHash) {
...
+ * @then returned top block's height is equal to the inserted one's
+ */
+TEST_F(BlockQueryTest, GetTopBlockSuccess) {
+  auto top_block_opt = framework::expected::val(blocks->getTopBlock());
+  ASSERT_TRUE(top_block_opt);
+  ASSERT_EQ(top_block_opt.value().value->height(), 2);
+}
+
+/**
+ * @given block store with mocked flat file
```
nickaleks: Not very good formulation, the point here is that block store is empty

irohad/main/impl/ordering_init.cpp
```diff
@@ -6,18 +6,31 @@
...
-      auto height = block_query->getTopBlocks(1).as_blocking().last()->height();
-      auto gate =
-          std::make_shared<ordering::OrderingGateImpl>(transport, height);
-      log_->info("Creating Ordering Gate with initial height {}", height);
-      transport->subscribe(gate);
-      return gate;
+      return block_query->getTopBlock().match(
+          [this, &transport](
+              expected::Value<std::shared_ptr<shared_model::interface::Block>>
+                  &block) -> std::shared_ptr<ordering::OrderingGateImpl> {
```
nickaleks: Please change return type of the lambda to std::shared_ptr<ordering::OrderingGate> to preserve polymorphism

PR #1477, opened 2018-06-18
Make scripts fail if any of bash cmds fail

Warchant:
### Description of the Change

If one of internal commands inside any of bash scripts fail (exit with non-0 code), the whole script continues, giving a false hope that bindings were built.

### Benefits

This change prevents this behavior, and make the whole script fail if any subcommand fails.

### Possible Drawbacks 

None.

### DIFF
--- a/example/java/build_library.sh
+++ b/example/java/build_library.sh
@@ -1,8 +1,8 @@
-#!/bin/bash
+#!/bin/bash -e
 cd "$(dirname "$0")"
 
 # folder with bindings and native library
-mkdir dist
+mkdir dist || true
 
 # build native library
 ./prepare.sh
--- a/example/java/build_library.sh
+++ b/example/java/build_library.sh
@@ -2,7 +2,7 @@
 cd "$(dirname "$0")"
 
 # folder with bindings and native library
-mkdir dist || true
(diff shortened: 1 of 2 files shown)


example/java/build_library.sh
```diff
@@ -1,8 +1,8 @@
-#!/bin/bash
+#!/bin/bash -e
 cd "$(dirname "$0")"
 
 # folder with bindings and native library
-mkdir dist
+mkdir dist || true
```
Alexey-N-Chernyshov: Useful PR.
Please, consider `mkdir -p` option instead `|| true` before merge.

Warchant: how is that better than this?

Alexey-N-Chernyshov: `mkdir -p` creates dir if it doesn't exist and will not return fail in case it is exists, so there is no need to `|| true`

Alexey-N-Chernyshov: also `mkdir -p` fails in case it is actually impossible to create dir while `mkdir dist || true` will continue script

Alexey-N-Chernyshov:
please, check if CI succeded

PR #1480, opened 2018-06-19
Fix on commit

kamilsa:

### Description of the Change

Until fix blocks were emitted only in insert_block method, which is not right, since it should emit every block when commit on storage is happening.
### Benefits

Proper behavior of ametsuchi

### Possible Drawbacks 

None
### Usage Examples or Tests

Run ametsuchi test

### DIFF
--- a/test/module/irohad/ametsuchi/ametsuchi_test.cpp
+++ b/test/module/irohad/ametsuchi/ametsuchi_test.cpp
@@ -638,7 +638,6 @@ TEST_F(AmetsuchiTest, TestingStorageWhenInsertBlock) {
  * @then committed block is emitted to observable
  */
 TEST_F(AmetsuchiTest, TestingStorageWhenCommitBlock) {
-  auto log = logger::testLog("TestStorage");
   ASSERT_TRUE(storage);
 
   auto expected_block = getBlock();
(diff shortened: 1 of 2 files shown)


test/module/irohad/ametsuchi/ametsuchi_test.cpp
```diff
@@ -632,6 +632,46 @@ TEST_F(AmetsuchiTest, TestingStorageWhenInsertBlock) {
...
+
+  // create test subscriber to check if committed block is as expected
+  auto wrapper = make_test_subscriber<CallExact>(storage->on_commit(), 1);
+  wrapper.subscribe([&expected_block](const auto &block) {
+    ASSERT_EQ(*block, expected_block);
+  });
+
+  log->info("Try commit the block");
+
+  std::unique_ptr<MutableStorage> mutable_storage;
```
kamilsa: I know that we have framework::expected. But it does not work for mutable storage, because its copies unique ptr in that case

test/module/irohad/ametsuchi/ametsuchi_test.cpp
```diff
@@ -632,6 +632,46 @@ TEST_F(AmetsuchiTest, TestingStorageWhenInsertBlock) {
...
+  storage->createMutableStorage().match(
+      [&mutable_storage](
+          iroha::expected::Value<std::unique_ptr<MutableStorage>> &_storage) {
```
lebdron: `_storage` is not a valid variable name according to guidelines https://google.github.io/styleguide/cppguide.html#Variable_Names

test/module/irohad/ametsuchi/ametsuchi_test.cpp
lebdron: Maybe check if it's possible to modify the methods and move the value?

test/module/irohad/ametsuchi/ametsuchi_test.cpp
x3medima17: I don't think that passing pointer by reference is a good idea

kamilsa: why?

PR #1484, opened 2018-06-20
Add implementation of on_demand ordering service

muratovv:
### Description of the Change
Add implementation of ordering service that managed by proposals by request.
Method of the work:
Each round we:
 * pack new proposal into unordered_map for sharing in the network;
 * erase old proposals;
 * collect transactions in the concurrent queue for further packing to the proposal in a next round.

Execution model:
Each network method may be called from concurrently, updates of invariant are covered by a write lock, rest by a read lock. `OnTransactions` may be called in different threads simultaneously.

Rounds:
We form proposal round as a pair [BlockRound, RejectRound], where the BlockRound is a number of blocks in the ledger + 1, RejectRound is a number of rejects + 1.
For example, first round in the network will be [2, 1].

### Benefits

One small step to BFT Iroha

### Possible Drawbacks 
The implementation doesn't contain tests.

### Reviewers notes

Pls, keep focusing on the following issues:
* redundant copies, especially, in the proposal generation
* race conditions
* imperative style - I hope, that you may suggest a more elegant solution.


irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
+
+  insertLast(outcome, round);
+  erase();
 }
 
-// -------------------- | OdOsNotification |------------------------------
+// ----------------------------| OdOsNotification |-----------------------------
 
 void OnDemandOrderingServiceImpl::onTransactions(
     std::vector<std::shared_ptr<shared_model::interface::Transaction>>
```
nickaleks: const ref?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
         transactions) {
-  throw std::logic_error("Function not yet implemented");
+  // read lock
+  boost::shared_lock<boost::shared_mutex> guard(lock_);
+
+  std::for_each(
+      transactions.begin(), transactions.end(), [this](const auto &obj) {
```
nickaleks: Concurrent insertion for batches?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
 }
 
 boost::optional<std::shared_ptr<shared_model::interface::Proposal>>
-OnDemandOrderingServiceImpl::onRequestProposal(uint64_t round) {
-  throw std::logic_error("Function not yet implemented");
+OnDemandOrderingServiceImpl::onRequestProposal(transport::RoundType round) {
+  // read lock
+  boost::shared_lock<boost::shared_mutex> guard(lock_);
+  try {
+    return proposal_map_.at(round);
```
nickaleks: since this section is protected by mutex, we can check if key is present and then request key. This avoids exception

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
+  } catch (std::out_of_range &ex) {
+    return boost::none;
+  }
+}
+
+// ---------------------------------| Private |---------------------------------
+
+void OnDemandOrderingServiceImpl::insertLast(
```
nickaleks: Maybe better naming?

nickaleks: Also needs to specify that it is not thread-safe

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
+      break;
+    case RoundOutput::REJECT:
+      next_round =
+          std::make_pair(current_round.first, current_round.second + 1);
+      break;
+  }
+}
+
+OnDemandOrderingServiceImpl::ProposalType
+OnDemandOrderingServiceImpl::emitProposal() {
```
nickaleks: Maybe make it a function which accepts queue of transactions?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
+  auto proto_txes = collection | boost::adaptors::transformed([](auto &tx) {
+                      return static_cast<const ProtoTxType &>(*tx);
+                    });
+
+  return std::make_shared<decltype(mutable_proposal.build())>(
+      mutable_proposal.transactions(proto_txes).build());
+}
+
+void OnDemandOrderingServiceImpl::erase() {
+  if (round_queue_.empty() and round_queue_.front().second == 1) {
```
nickaleks: Need to keep at least 2 successful proposals

nickaleks: Also not empty

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
+  // read lock
+  boost::shared_lock<boost::shared_mutex> guard(lock_);
+
+  std::for_each(
+      transactions.begin(), transactions.end(), [this](const auto &obj) {
+        current_proposal_.second.push(obj);
+      });
 }
 
 boost::optional<std::shared_ptr<shared_model::interface::Proposal>>
```
nickaleks: Can use ProposalType

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -24,7 +31,61 @@ namespace iroha {
...
+       */
+      void erase();
+
+      /**
+       * @return packed proposal from current round queue
+       */
+      ProposalType emitProposal();
+
+      /**
+       * Queue which holds all round in linear sequential
```
nickaleks: missing "order"

irohad/ordering/on_demand_os_transport.hpp
```diff
@@ -23,9 +26,26 @@ namespace iroha {
...
        */
-      using RoundType = uint64_t;
+      using RejectRoundType = uint32_t;
+
+      /**
+       * Type of proposal round
+       */
+      using RoundType = std::pair<BlockRoundType, RejectRoundType>;
+
+      class RoundTypeHasher {
```
nickaleks: Please add documentation why it is needed

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -8,6 +8,13 @@
 
 #include "ordering/on_demand_ordering_service.hpp"
 
+#include <tbb/concurrent_queue.h>
+#include <boost/thread.hpp>
+
+#include <mutex>
```
nickaleks: Unused include

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
+          std::make_pair(current_round.first, current_round.second + 1);
+      break;
+  }
+}
+
+OnDemandOrderingServiceImpl::ProposalType
+OnDemandOrderingServiceImpl::emitProposal() {
+  auto mutable_proposal = shared_model::proto::ProposalBuilder()
+                              .height(current_proposal_.first.first)
+                              .createdTime(0);  // TODO insert correct time
```
l4l: just a reminder

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
+      mutable_proposal.transactions(proto_txes).build());
+}
+
+void OnDemandOrderingServiceImpl::erase() {
+  if (round_queue_.empty() and round_queue_.front().second == 1) {
+    proposal_map_.erase(round_queue_.front());
+    round_queue_.pop();
+    return;
+  } else {
+    while (not round_queue_.empty() or round_queue_.front().second != 1) {
```
l4l: and?

muratovv: Nice issue, thx.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,104 @@
...
 using namespace iroha::ordering;
 
-// ------------------ | OnDemandOrderingService |-------------------------
+// -------------------------| OnDemandOrderingService |-------------------------
 
 void OnDemandOrderingServiceImpl::onCollaborationOutcome(
     RoundOutput outcome, transport::RoundType round) {
-  throw std::logic_error("Function not yet implemented");
+  // exclusive write lock
+  boost::upgrade_lock<boost::shared_mutex> upgrade_lock(lock_);
```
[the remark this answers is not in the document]
muratovv: Because we need read\write lock idiom.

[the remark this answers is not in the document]
l4l: afair it can be achieved with unique_lock (thus with only STL)

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
muratovv: I have thanked for it. The solution is the following: we will rework the interface of `onTransaction` with a list of batches and insert batches concurrently.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,135 @@
...
 void OnDemandOrderingServiceImpl::onTransactions(
-    std::vector<std::shared_ptr<shared_model::interface::Transaction>>
-        transactions) {
-  throw std::logic_error("Function not yet implemented");
+    const CollectionType &transactions) {
+  // read lock
+  std::shared_lock<std::shared_timed_mutex> guard(lock_);
+  log_->info("onTransactions => collections size = {}", transactions.size());
+
+  std::for_each(
```
nickaleks: Maybe use insert or similar function instead of for_each?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,135 @@
...
+  log_->info("onTransactions => collection is inserted");
+}
+
+boost::optional<OnDemandOrderingServiceImpl::ProposalType>
+OnDemandOrderingServiceImpl::onRequestProposal(transport::RoundType round) {
+  // read lock
+  std::shared_lock<std::shared_timed_mutex> guard(lock_);
+  auto proposal = proposal_map_.find(round);
+  if (proposal != proposal_map_.end()) {
+    return (*proposal).second;
```
nickaleks: proposal->second?

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -8,23 +8,100 @@
...
+       */
+      size_t transaction_limit_;
+
+      /**
+       * Max number of available proposals in one OS`1234567890'
+       */
+      size_t number_of_proposals_;
+
+      /**
+       * Queue which holds all round in linear order
```
nickaleks: rounds?

irohad/ordering/on_demand_os_transport.hpp
```diff
@@ -23,32 +26,65 @@ namespace iroha {
...
-       * Type of round indexing
+       * Type of round indexing by blocks
+       */
+      using BlockRoundType = uint64_t;
+
+      /**
+       * Type of round indexing by reject before new block commit
        */
-      using RoundType = uint64_t;
+      using RejectRoundType = uint32_t;
```
nickaleks: Why uint32_t?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
muratovv: What are you imagine?

irohad/ordering/on_demand_os_transport.hpp
muratovv: Instead of what?

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -8,23 +8,100 @@
...
+#include "logger/logger.hpp"
+
 namespace iroha {
   namespace ordering {
     class OnDemandOrderingServiceImpl : public OnDemandOrderingService {
      public:
-      // --------------------- | OnDemandOrderingService | ---------------------
+      OnDemandOrderingServiceImpl(
+          size_t transaction_limit,
+          size_t number_of_proposals = 3,
```
lebdron: Please add a description for default values to make the unfamiliar readers more comfortable.

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -8,23 +8,100 @@
...
+      void onTransactions(const CollectionType &transactions) override;
+
+      boost::optional<ProposalType> onRequestProposal(
+          transport::RoundType round) override;
+
+     private:
+      /**
+       * Type of stored transactions
+       */
+      using TransactionType =
```
lebdron: Is it possible to use `unique_ptr` here? The object is created in transport, and then moved to a proposal. It seems to express the ownership more clearly.

irohad/ordering/on_demand_os_transport.hpp
```diff
@@ -23,32 +26,65 @@ namespace iroha {
...
        public:
+        /**
+         * Type of stored proposals
+         */
+        using ProposalType = std::shared_ptr<shared_model::interface::Proposal>;
+
+        /**
+         * Type of stored transactions
+         */
+        using TransactionType =
```
lebdron: Same as in implementation.

irohad/ordering/on_demand_os_transport.hpp
```diff
@@ -23,32 +26,65 @@ namespace iroha {
...
 
       /**
        * Notification interface of on demand ordering service.
        */
       class OdOsNotification {
        public:
+        /**
+         * Type of stored proposals
+         */
+        using ProposalType = std::shared_ptr<shared_model::interface::Proposal>;
```
lebdron: Please state why shared is required here in the doc. It suggests that implementation will give a copy of the pointer, which then in turn should be copied into transport object. Maybe create a copy inside ordering service and give a `unique_ptr` instead?

irohad/ordering/on_demand_os_transport.hpp
```diff
@@ -23,32 +26,65 @@ namespace iroha {
...
+        using CollectionType = std::vector<TransactionType>;
+
         /**
          * Callback on receiving transactions
          * @param transactions - vector of passed transactions
          */
-        virtual void onTransactions(
-            std::vector<std::shared_ptr<shared_model::interface::Transaction>>
-                transactions) = 0;
+        virtual void onTransactions(const CollectionType &transactions) = 0;
```
lebdron: This interface suggests that `transactions` ownership is not passed to receiver, which does not seem to be the case, because they are effectively moved to a proposal. Maybe use rvalue reference explicitly instead of ref-to-const?

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
lebdron: Maybe also use `CollectionType::value_type` to share the type with notification?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,135 @@
...
+  current_proposal_.first = next_round;
+  current_proposal_.second.clear();  // operator = of tbb::queue is closed
 }
 
-boost::optional<std::shared_ptr<shared_model::interface::Proposal>>
-OnDemandOrderingServiceImpl::onRequestProposal(uint64_t round) {
-  throw std::logic_error("Function not yet implemented");
+OnDemandOrderingServiceImpl::ProposalType
+OnDemandOrderingServiceImpl::emitProposal() {
+  auto mutable_proposal = shared_model::proto::ProposalBuilder()
```
lebdron: Performance impact of using builder can be significant https://github.com/hyperledger/iroha/pull/1037 , consider removing?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,135 @@
...
+
+  TransactionType current_tx;
+  using ProtoTxType = shared_model::proto::Transaction;
+  std::vector<TransactionType> collection;
+
+  // outer method should guarantee availability of at least one transaction in
+  // queue
+  while (current_proposal_.second.try_pop(current_tx)
+         and collection.size() < transaction_limit_) {
+    collection.emplace_back(current_tx);
```
lebdron: `onTransactions` does not enforce the size of the queue, so `current_proposal_` can contain more transactions than `transaction_limit`. Please comment if this is wrong, or the case should be handled to avoid missing transactions.

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -8,23 +8,100 @@
...
+       */
+      ProposalType emitProposal();
+
+      /**
+       * Max number of transaction in one proposal
+       */
+      size_t transaction_limit_;
+
+      /**
+       * Max number of available proposals in one OS`1234567890'
```
lebdron:
```
`1234567890'
``` 
looks like it does not belong here.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,135 @@
...
+  return std::make_shared<decltype(mutable_proposal.build())>(
+      mutable_proposal.transactions(proto_txes).build());
+}
+
+void OnDemandOrderingServiceImpl::tryErase() {
+  if (round_queue_.size() < number_of_proposals_) {
+    return;
+  }
+
+  if (round_queue_.empty() and round_queue_.front().second == 1) {
```
lebdron: Seems that `1` should be defined as some constant with description, because it is used multiple times with a special meaning of first round.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -5,24 +5,135 @@
...
+  }
+
+  round_queue_.push(current_proposal_.first);
+  log_->info("packNextProposal: pushed in queue");
+
+  auto current_round = current_proposal_.first;
+  decltype(current_round) next_round;
+  switch (outcome) {
+    case RoundOutput::SUCCESSFUL:
+      next_round = std::make_pair(current_round.first + 1, 1);
```
lebdron: Same issue with `1`, maybe define a constant?

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
l4l: tbb won't allow :(

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -4,25 +4,147 @@
...
+  return std::make_unique<shared_model::proto::Proposal>(
+      std::move(proto_proposal));
+}
+
+void OnDemandOrderingServiceImpl::tryErase() {
+  if (round_queue_.size() < number_of_proposals_) {
+    return;
+  }
+
+  if (round_queue_.empty() and round_queue_.front().second == kFirstRound) {
```
lebdron: This will lead to undefined behavior if `round_queue_` is empty, because it will start evaluating the second expression which accesses `front`. Please fix, add comments for conditionals and simplify if possible.

PR #1485, opened 2018-06-20
Binary Testing Framework

igor-egorov:
### Description of the Change

The PR introduces Binary Testing Framework.
BTF allows performing stateful validation for transactions and queries built using client libraries. 
For now, only Python is supported. Java support will be added later on.

This PR comes with python scripts that cover all the set of Iroha permissions.

### Benefits

Now here is the tool for stateful validation of transactions and queries created via client libraries.

Also, here is a set of python examples that shows how to use Iroha permissions model. (Scripts' listings will be included into "permissions" page at ReadTheDocs).

### Possible Drawbacks

No drawbacks were observed.

### Usage Examples or Tests

```
cmake -H. -Bbuild -DSWIG_PYTHON=ON
cmake --build build --target binary_test
cd build
ctest -R binary
```

### Possible Improvements

The existing set of python example scripts can be extended with tests which validate the correctness of resulting state after applying transactions to Iroha (that approach was not the point of current tests set).

Such improvement will require changes in the implementation of Python and Java launchers.


igor-egorov:
Note about SonarQube report - that is just a false positive error.

> launchers.cpp#L57: The function 'initialized' is never used.

https://github.com/hyperledger/iroha/blob/53f1894ed2b8ba703f6fccb9b939b9c9a76762f7/test/integration/binary/binaries_test_fixture.hpp#L117

test/integration/binary/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  endif ()
+
+  foreach (item "block" "commands" "primitive" "queries")
+    compile_proto_to_python("${item}.proto")
+    list(APPEND PROTO_SWIG_DEPS "${SWIG_BUILD_DIR}/${item}_pb2.py")
+  endforeach (item)
+
+  add_executable(binary_test
+      launchers.cpp
+      binaries_test.cpp)
```
muratovv: Better to move `)` to the new line.

test/integration/binary/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  target_link_libraries(binary_test
+      gtest
+      gmock
+      boost
+      schema
+      shared_model_proto_backend
+      shared_model_interfaces
```
muratovv: Looks redundant, but it is not a required issue.

test/integration/binary/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+      integration_framework
+      bindings
+      )
+  target_include_directories(binary_test PUBLIC ${PROJECT_SOURCE_DIR}/test)
+
+  add_dependencies(binary_test irohapy)
+  add_test(
```
muratovv: Use `addtest` instead. https://github.com/hyperledger/iroha/blob/e8cdcc1f0b87058814e8d44c1b5f7b1839af3e92/cmake/functions.cmake#L23

Pls, note that function assumes "type" of test which is required for CI.

test/integration/binary/binaries_test.cpp
```diff
@@ -0,0 +1,202 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "binaries_test_fixture.hpp"
```
muratovv: Use full path instead.

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+  }  // namespace interface
+
+}  // namespace shared_model
+
+namespace query_validation {
+
+  namespace internal {
+
+    class Void {};
```
muratovv: Add purpose of usage.

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+  }  // namespace internal
+
+  using QueryIterator = std::vector<proto::Query>::iterator;
+
+  template <typename ResponseType>
+  inline void checkQueryResponse(
```
muratovv: I think naming is misleading. Better to reflect type checking here.

muratovv: Also, pls add documentation for the function and functions below.

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -0,0 +1,139 @@
...
+  template <typename... Responses>
+  inline void validateQueries(QueryIterator it,
+                              QueryIterator end,
+                              IntegrationTestFramework &itf) {
+    _validateQueries<Responses..., internal::Void>(it, end, itf);
+  }
+
+}  // namespace query_validation
+
+template <typename T>
```
muratovv: Add explanation of T.

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -0,0 +1,139 @@
...
+    return makeGenesis(launcher.transactions[0], *launcher.admin_key);
+  }
+
+  static void blockWithTransactionValidation(
+      const std::shared_ptr<shared_model::interface::Block> &result) {
+    ASSERT_EQ(result->transactions().size(), 1);
+  }
+
+  template <typename... QueryResponses>
+  void doStupidTest(const unsigned &transactions_expected = 0,
```
muratovv: Naming is ambiguous. Pls, rework with a more convenient name.

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "launchers.hpp"
```
muratovv: Rework with full path

test/integration/binary/launchers.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <boost/process.hpp>
+#include <chrono>
+#include <string>
+#include "backend/protobuf/queries/proto_query.hpp"
+#include "backend/protobuf/transaction.hpp"
+#include "cryptography/keypair.hpp"
+
+namespace binary_test {
+
+  class Launcher {
```
muratovv: Add some documentation with purpose

test/integration/binary/launchers.hpp
```diff
@@ -0,0 +1,48 @@
...
+   public:
+    virtual std::string launchCommand(const std::string &test_case) = 0;
+
+    void operator()(const std::string &example);
```
muratovv: Add doc, pls

test/integration/binary/launchers.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+    bool initialized(const unsigned &transactions_expected = 0,
+                     const unsigned &queries_expected = 0);
+
+    std::shared_ptr<shared_model::crypto::Keypair> admin_key;
+    std::vector<shared_model::proto::Transaction> transactions;
+    std::vector<shared_model::proto::Query> queries;
+
+   protected:
+    void readBinaries(boost::process::ipstream &stream);
```
muratovv: Add contract for the stream in a documentation

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
...
+    std::stringstream s;
+    s << "bash -c \"${PYTHON_INTERPRETER} "
+         "${ROOT_DIR}/example/python/permissions/"
+      << example << ".py\"";
+    return s.str();
+  }
+
+  std::string JavaLauncher::launchCommand(const std::string &example) {
+    return "";
+    // tbd, igor-egorov, 2018-06-20, IR-1389
```
muratovv: s/tbd/TODO

example/python/permissions/can_add_asset_qty.py
```diff
@@ -0,0 +1,43 @@
...
+    tx = iroha.ModelTransactionBuilder() \
+        .createdTime(commons.now()) \
+        .creatorAccountId(alice['id']) \
+        .addAssetQuantity(alice['id'], 'coin#test', '5000.99') \
+        .build()
+    return iroha.ModelProtoTransaction(tx) \
+        .signAndAddSignature(alice['key']).finish()
+
+
+print(admin['key'].privateKey().hex())
```
muratovv: Maybe move this framework-related code to separate class (and file) for reducing boilerplate?

test/integration/binary/CMakeLists.txt
lebdron: It is not possible, because a custom test is defined here, while `addtest` creates a generic test which calls the test binary without any parameters. The test call is replaces here with python interpreter call.

example/python/permissions/can_add_asset_qty.py
```diff
@@ -0,0 +1,43 @@
...
+#
+
+import iroha
+import commons
+
+admin = commons.user('admin@test')
+alice = commons.user('alice@test')
+
+
+def genesis_tx():
```
l4l: That function is pretty similar among the files, that better be moved somewhere

test/integration/binary/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+
+if (SWIG_PYTHON OR SWIG_JAVA)
```
l4l: Consider adding a variable `IS_BINDINGS` or `SWIG_ENABLED`, etc
otherwise it be duplicated in bindings folder as well

test/integration/binary/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+endif ()
+
+if (SWIG_PYTHON)
+  if (SUPPORT_PYTHON2)
+    find_package(PythonInterp 2.7 REQUIRED)
+  else ()
+    find_package(PythonInterp 3.5 REQUIRED)
+  endif ()
+
+  foreach (item "block" "commands" "primitive" "queries")
```
l4l: The dependency should be moved somewhere at separate target

test/integration/binary/CMakeLists.txt
l4l: at least mark with a todo, same for the following one

test/integration/binary/binaries_test_fixture.hpp
l4l: *class doc

test/integration/binary/binaries_test_fixture.hpp
l4l: or even better rename to `SdkLauncher`

test/integration/binary/binaries_test_fixture.hpp
l4l: and consider adding a doc

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -0,0 +1,139 @@
...
+                    const unsigned &queries_expected = 0) {
+    if (launcher.initialized(transactions_expected, queries_expected)) {
+      IntegrationTestFramework itf(1);
+
+      itf.setInitialState(*launcher.admin_key, genesis());
+      for (auto i = 1u; i < launcher.transactions.size(); ++i) {
```
l4l: `std::for_each`?

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <gtest/gtest.h>
+#include "common/byteutils.hpp"
+
+#include "bindings/model_crypto.hpp"
+
+using namespace boost::process;
+
+namespace binary_test {
+
+  constexpr auto cTimeToKill = std::chrono::minutes(15);
```
l4l: quite a lot

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
...
+    ipstream pipe;
+    const auto &command = launchCommand(example);
+    if (command.empty()) {
+      FAIL() << "Launcher provided empty command";
+    }
+    child c(command, std_out > pipe);
+    auto terminated = c.wait_for(cTimeToKill);
+    if (not terminated) {
+      c.terminate();
+      FAIL() << "Child process was terminated because execution time limit "
```
l4l: Is it always a test fail?

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+  void Launcher::readBinaries(ipstream &stream) {
+    transactions.clear();
+    queries.clear();
+    std::string line;
+    bool first_line(true);
+    while (stream and std::getline(stream, line) and not line.empty()) {
+      iroha::protocol::Transaction proto_tx;
+      iroha::protocol::Query proto_query;
+      if (auto byte_string = iroha::hexstringToBytestring(line)) {
```
l4l: monadic bind?

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  void Launcher::readBinaries(ipstream &stream) {
+    transactions.clear();
+    queries.clear();
+    std::string line;
+    bool first_line(true);
+    while (stream and std::getline(stream, line) and not line.empty()) {
+      iroha::protocol::Transaction proto_tx;
+      iroha::protocol::Query proto_query;
```
l4l: Can be moved inside the following `if`

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
...
+        }
+      }
+    }
+  }
+
+  bool Launcher::initialized(const unsigned &transactions_expected,
+                             const unsigned &queries_expected) {
+    checkAsserts(transactions_expected, queries_expected);
+    return admin_key and (transactions.size() == transactions_expected)
+        and (queries.size() == queries_expected);
```
l4l: remove redundant braces plz, otherwise syntax highlighting will consider `and` as a function

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
...
+  void Launcher::checkAsserts(const unsigned &transactions_expected,
+                              const unsigned &queries_expected) {
+    ASSERT_TRUE(admin_key);
+    ASSERT_EQ(transactions.size(), transactions_expected);
+    ASSERT_EQ(queries.size(), queries_expected);
+  }
+
+  std::string PythonLauncher::launchCommand(const std::string &example) {
+    std::stringstream s;
+    s << "bash -c \"${PYTHON_INTERPRETER} "
```
l4l: Why via bash? is there any problem with direct launch?

test/integration/binary/launchers.cpp
```diff
@@ -0,0 +1,84 @@
...
+         "${ROOT_DIR}/example/python/permissions/"
```
l4l: boost::filesystem better be used

test/integration/binary/launchers.cpp
l4l: who cares, tbd is also ok

test/integration/binary/launchers.hpp
```diff
@@ -0,0 +1,48 @@
...
+    void checkAsserts(const unsigned &transactions_expected,
+                      const unsigned &queries_expected);
+  };
+
+  class PythonLauncher : public Launcher {
+   public:
+    virtual std::string launchCommand(const std::string &example);
```
l4l: <s>virtual </s>std::string launchCommand(const std::string &example) **override**;
same below

test/integration/binary/launchers.hpp
```diff
@@ -0,0 +1,48 @@
...
+  class Launcher {
+   public:
+    virtual std::string launchCommand(const std::string &test_case) = 0;
+
+    void operator()(const std::string &example);
+
+    bool initialized(const unsigned &transactions_expected = 0,
+                     const unsigned &queries_expected = 0);
+
+    std::shared_ptr<shared_model::crypto::Keypair> admin_key;
```
l4l: boost::optional seems more suitable (or even without any wrapper & early fail)

test/integration/binary/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  endforeach (item)
+
+  add_executable(binary_test
+      launchers.cpp
+      binaries_test.cpp)
+  target_link_libraries(binary_test
+      gtest
+      gmock
+      boost
+      schema
```
l4l: also redundant, it must be used in backend

example/python/permissions/commons.py
```diff
@@ -0,0 +1,67 @@
...
+import iroha
+from time import time
+
+
+def now():
+    return int(time() * 1000)
+
+
+def all_permissions():
+    return iroha.RolePermissionSet([
```
l4l:
```python
r = iroha.RolePermissionSet()
for i in range(r.size()):
  r.set(i)
return i
```

example/python/permissions/commons.py
```diff
@@ -0,0 +1,67 @@
...
+        iroha.Role_kSetMyQuorum,
+        iroha.Role_kAddMySignatory,
+        iroha.Role_kRemoveMySignatory,
+        iroha.Role_kTransferMyAssets,
+        iroha.Role_kSetMyAccountDetail,
+        iroha.Role_kGetBlocks
+    ])
+
+
+def user(user_id):
```
l4l: That will generate a new keypair for the same user, so consider renaming the method
or [static vars hack can be used](https://code-maven.com/slides/python-programming/static-variable)

test/integration/binary/CMakeLists.txt
muratovv: Can we create a generic function for those purposes? We have to manage types of tests in CI.

test/integration/binary/binaries_test_fixture.hpp
muratovv: This may be inconvenient in the future when you put to template objects which are different from `Launcher` type.

test/integration/binary/launchers.cpp
muratovv: Tbd not ok because you can't grep it fast. Better to remember only one name instead of two.

muratovv: @igor-egorov you may create a snippet with TODO in your favorite IDE.

test/integration/binary/CMakeLists.txt
lebdron: Maybe an easier solution would be to override `add_test`, do custom logic and pass the arguments to real `add_test`. This will add types to all tests added with `add_test`.

igor-egorov: Discussed offline. Summary: there is no problem because `python_binary_test` target becomes visible only when SWIG for Python is enabled (`-DSWIG_PYTHON=ON`). Thus, CI would not be confused.

example/python/permissions/can_add_asset_qty.py
igor-egorov: Discussed with Fedor in offline.
Let's keep this code for now. We expect the protocol of communication between BTF and external scripts to be changed, so then we can reduce the amount of such boilerplate code.

test/integration/binary/launchers.cpp
igor-egorov: Such a big timeout is used intentionally in order to let binary generators pass on slow CI or fail for any reason but time out.

test/integration/binary/launchers.cpp
igor-egorov: Yes. The hang of external generator is considered as test case fail.

test/integration/binary/launchers.cpp
igor-egorov: It is possible to use monadic bind here but it is just an alternative way of implementation. Let's keep current version because suggested change does not bring visible benefits.

test/integration/binary/launchers.cpp
igor-egorov: We have to (due to tricky cmake that set environment variables which cannot be accessed without "bash layer").

example/python/permissions/commons.py
igor-egorov: Discussed offline. Since the listing of commons.py will also be included in the documentation page about permissions, it is better to preserve fully-qualified constants' names for better presentation for documentation users.

test/integration/binary/CMakeLists.txt
igor-egorov: Discussed offline. Not a problem.

test/integration/binary/launchers.cpp
l4l: Code become cleaner since you don't need to get explicitly get value from optional. Also at the codebase it's the preferred style

example/python/permissions/can_add_asset_qty.py
igor-egorov: Finally, it was decided not to modify this place. Moving of the common part will lead to hiding command of domain creation that includes the test role. Otherwise, there will be nothing to move.

Thus, in order to increase examples' demonstrativeness, genesis functions would not be modified.

example/python/permissions/commons.py
igor-egorov: Discussed offline. Decided to clarify method naming and not to touch the internals for now.

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -44,86 +44,143 @@ namespace shared_model {
...
-        framework::SpecifiedVisitor<ResponseType>(), response.get()));
-  }
+    /**
+     * Asserts that actual query response type equals to expected type.
+     *
+     * @tparam ExpectedResponseType - expected type of QueryResponse object
+     * @param response - QueryResponse object to check
+     */
+    template <typename ExpectedResponseType>
+    inline void checkQueryResponseType(
```
l4l: Would be better to return bool, instead of assertion
Otherwise fails will look like `Assertion failed at binaries_test_fixture.hpp:67`, that is pretty useless

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -44,86 +44,143 @@ namespace shared_model {
...
 
-      itf.setInitialState(*launcher.admin_key, genesis());
-      for (auto i = 1u; i < launcher.transactions.size(); ++i) {
-        itf.sendTx(launcher.transactions[i])
-            .checkBlock(blockWithTransactionValidation);
-      }
+      itf.setInitialState(launcher.admin_key.value(), genesis());
+
+      std::for_each(
+          std::next(  // first transaction was used as genesis transaction
```
l4l: could you move the comment at the separate line above? looks ugly :(

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+#include <gtest/gtest.h>
+
+#include <vector>
+#include "builders/protobuf/block.hpp"
+#include "framework/integration_framework/integration_test_framework.hpp"
+#include "framework/specified_visitor.hpp"
+#include "integration/binary/launchers.hpp"
+
+using namespace boost::process;
```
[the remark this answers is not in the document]
l4l: You can use alias instead: `namespace proc = boost::process`

test/integration/binary/binaries_test_fixture.hpp
```diff
@@ -0,0 +1,210 @@
...
+     *
+     * @tparam Head - the type of expected query response for the current query
+     * @tparam Tail - the rest of expected query response types
+     * @param it - points to query needs to be checked on current step
+     * @param end - queries' vector iterator that points to .end()
+     * @param itf - already initialized ITF instance that will be used queries
+     *    execution.
+     */
+    template <typename Head, typename... Tail>
+    inline void _validateQueries(::query_validation::QueryIterator it,
```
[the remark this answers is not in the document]
l4l: Means internal, I guess

PR #1488, opened 2018-06-20
Fix blocks query handle

kamilsa:

### Description of the Change

BlocksQueryHandle method emitted block error response to the blocks observable, which lead to getting stateful invalid responses by users who subscribed on committed blocks.
### Benefits

No one will get block error response but the caller who sent bad request

### Possible Drawbacks 

None

### Usage Examples or Tests 
New blocks query test


neewy:
@Solonets please use English only in pull request discussions, this is a mandatory thing for all maintainers

PR #1489, opened 2018-06-20
Refactor common executor

x3medima17:
Signed-off-by: Dumitru <dimasavva17@gmail.com>

### Description of the Change
Made `getAccountPermissions` and `checkAccountRolePermission` functional.

### Benefits
No side effects
No object copying, thus faster runtime,

### Possible Drawbacks 
None

### DIFF
--- a/irohad/execution/impl/common_executor.cpp
+++ b/irohad/execution/impl/common_executor.cpp
@@ -28,16 +28,16 @@ namespace iroha {
   getAccountPermissions(const std::string &account_id,
                         ametsuchi::WsvQuery &queries) {
     return queries.getAccountRoles(account_id) | [&queries](const auto &roles) {
-      return std::accumulate(roles.begin(),
-                             roles.end(),
-                             shared_model::interface::RolePermissionSet{},
-                             [&queries](auto &&permissions, const auto &role) {
-                               queries.getRolePermissions(role) |
-                                   [&](const auto &perms) {
-                                     permissions |= perms;
-                                   };
-                               return permissions;
-                             });
+      return std::accumulate(
+          roles.begin(),
+          roles.end(),
+          shared_model::interface::RolePermissionSet{},
+          [&queries](auto &&permissions, const auto &role) {
+            queries.getRolePermissions(role) |
+                [&](const auto &perms) { permissions |= perms; };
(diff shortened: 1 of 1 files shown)


irohad/execution/impl/common_executor.cpp
```diff
@@ -27,37 +27,34 @@ namespace iroha {
...
+    return queries.getAccountRoles(account_id) | [&queries](const auto &roles) {
+      return std::accumulate(roles.begin(),
+                             roles.end(),
+                             shared_model::interface::RolePermissionSet{},
+                             [&queries](auto &&permissions, const auto &role) {
+                               queries.getRolePermissions(role) |
+                                   [&](const auto &perms) {
+                                     permissions |= perms;
+                                   };
+                               return permissions;
```
kamilsa: won't it make a copy?

l4l: yeah, would be nice to add std::forward

irohad/execution/impl/common_executor.cpp
```diff
@@ -27,37 +27,34 @@ namespace iroha {
...
-      if (not rolePerms)
-        continue;
-      if (rolePerms->test(permission)) {
-        return true;
-      }
-    }
-    return false;
+    return queries.getAccountRoles(account_id) |
+        [&permission, &queries](const auto &accountRoles) {
+          return std::any_of(accountRoles.begin(),
```
l4l: :+1:

irohad/execution/impl/common_executor.cpp
kamilsa: Then decltype auto should be added as well

irohad/execution/impl/common_executor.cpp
```diff
@@ -28,16 +28,16 @@ namespace iroha {
...
-                                   };
-                               return permissions;
-                             });
+      return std::accumulate(
+          roles.begin(),
+          roles.end(),
+          shared_model::interface::RolePermissionSet{},
+          [&queries](auto &&permissions, const auto &role) {
+            queries.getRolePermissions(role) |
+                [&](const auto &perms) { permissions |= perms; };
```
l4l: formatting :(

PR #1490, opened 2018-06-21
An assortment of fixes for bindings

lebdron:
### Description of the Change
- Remove server-side objects from separate shared model build, which fixes failing Android build
- Fix Windows build on Jenkins due to invalid flag used
- Fix Android build on Jenkins, which was always using develop branch for builds
- Fix patching of ed25519 library when it is called from shared model directory, due to wrong absolute path used
- Remove `backend/protobuf` from SWIG interface which is not required. Windows build failed due to this header being present
- Add missing ciso646 header for Windows build
- Enable getTransactions tests on Javascript (__bonus__)

### Benefits
Working build for available platforms: python/java/javascript linux/windows android

### Possible Drawbacks 
None

### DIFF
--- a/.jenkinsci/bindings.groovy
+++ b/.jenkinsci/bindings.groovy
@@ -109,7 +109,7 @@ def doAndroidBindings(abiVersion) {
   sh """
     (cd /iroha; git init; git remote add origin https://github.com/hyperledger/iroha.git; \
     # TODO possible race on consequent pushes to the same branch
-    git fetch --depth 1 origin ${env.GIT_LOCAL_BRANCH}; git checkout -t ${env.GIT_BRANCH})
+    git fetch --depth 1 origin ${env.GIT_LOCAL_BRANCH}; git checkout -t origin/${env.GIT_LOCAL_BRANCH})
   """
   sh """
     . /entrypoint.sh; \
--- a/.jenkinsci/bindings.groovy
+++ b/.jenkinsci/bindings.groovy
@@ -22,7 +22,7 @@ def doJavaBindings(os, buildType=Release) {
       -DSWIG_JAVA=ON \
       ${cmakeOptions}
   """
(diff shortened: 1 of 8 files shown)


.jenkinsci/bindings.groovy
```diff
@@ -108,7 +110,8 @@ def doAndroidBindings(abiVersion) {
     [currentPath, "\$PLATFORM", abiVersion, "\$BUILD_TYPE_A", sh(script: 'date "+%Y%m%d"', returnStdout: true).trim(), commit.substring(0,6)])
   sh """
     (cd /iroha; git init; git remote add origin https://github.com/hyperledger/iroha.git; \
-    git fetch --depth 1 origin develop; git checkout -t origin/develop)
+    # TODO possible race on consequent pushes to the same branch
```
bakhtin: Could you elaborate on that?

.jenkinsci/bindings.groovy
```diff
@@ -108,7 +110,8 @@ def doAndroidBindings(abiVersion) {
...
+    git fetch --depth 1 origin ${env.GIT_LOCAL_BRANCH}; git checkout -t origin/${env.GIT_LOCAL_BRANCH})
```
bakhtin: This will fail if built from PR:
image
I suggest to change it to: `git fetch origin ${GIT_COMMIT}; git checkout FETCH_HEAD)`. 
I tested it and it works OK: https://jenkins.soramitsu.co.jp/job/iroha/job/iroha-hyperledger/job/fix%252Ffix-bindings-build/4/consoleFull
image

.jenkinsci/bindings.groovy
lebdron: What if someone pushed two commits to the same branch in a row? It seems to me that for two running CI pipelines, if the first pipeline did not reach this line with git fetch yet, it will also use the latest commit instead of previous one.

PR #1493, opened 2018-06-22
Stateful error responses to log

Akvinikym:
[IR-1455](https://soramitsu.atlassian.net/browse/IR-1455) [IR-1456](https://soramitsu.atlassian.net/browse/IR-1456) [IR-1457](https://soramitsu.atlassian.net/browse/IR-1457)

### Description of the Change

This is the second P.R. from 'Stateful Validation Error Response' series. Now, stateful validation errors are propagated up to simulator, where they are logged.

### Benefits

Stateful error is now logged, not thrown away.

### Possible Drawbacks 

None

### DIFF
--- a/irohad/ametsuchi/impl/temporary_wsv_impl.cpp
+++ b/irohad/ametsuchi/impl/temporary_wsv_impl.cpp
@@ -38,60 +38,64 @@ namespace iroha {
       transaction_->exec("BEGIN;");
     }
 
-    expected::Result<void, std::string> TemporaryWsvImpl::apply(
+    expected::Result<void, validation::CommandNameAndError>
+    TemporaryWsvImpl::apply(
         const shared_model::interface::Transaction &tx,
-        std::function<expected::Result<void, std::string>(
+        std::function<expected::Result<void, validation::CommandNameAndError>(
             const shared_model::interface::Transaction &, WsvQuery &)>
             apply_function) {
       const auto &tx_creator = tx.creatorAccountId();
       command_executor_->setCreatorAccountId(tx_creator);
       command_validator_->setCreatorAccountId(tx_creator);
-      auto execute_command =
-          [this](auto &command,
-                 size_t command_index) -> expected::Result<void, std::string> {
+      auto execute_command = [this](auto &command, size_t command_index)
+          -> expected::Result<void, validation::CommandNameAndError> {
         // Validate command
-        return expected::map_error<std::string>(
+        return expected::map_error<validation::CommandNameAndError>(
                    boost::apply_visitor(*command_validator_, command.get()),
                    [command_index](CommandError &error) {
-                     return (boost::format("stateful validation error: could "
-                                           "not validate "
-                                           "command with index %d: %s")
-                             % command_index % error.toString())
-                         .str();
+                     return validation::CommandNameAndError{
+                         error.command_name,
+                         (boost::format("stateful validation error: could "
+                                        "not validate "
+                                        "command with index %d: %s")
+                          % command_index % error.toString())
+                             .str()};
                    })
             | [this, command_index, &command] {
                 // Execute commands
-                return expected::map_error<std::string>(
+                return expected::map_error<validation::CommandNameAndError>(
                     boost::apply_visitor(*command_executor_, command.get()),
                     [command_index](CommandError &error) {
-                      return (boost::format(
-                                  "stateful validation error: could not "
-                                  "execute command with index %d: %s")
-                              % command_index % error.toString())
-                          .str();
+                      return validation::CommandNameAndError{
+                          error.command_name,
+                          (boost::format("stateful validation error: could not "
+                                         "execute command with index %d: %s")
+                           % command_index % error.toString())
+                              .str()};
                     });
               };
       };
 
       transaction_->exec("SAVEPOINT savepoint_;");
 
-      return apply_function(tx, *wsv_) |
-                 [this,
-                  &execute_command,
-                  &tx]() -> expected::Result<void, std::string> {
+      return apply_function(tx, *wsv_) | [this, &execute_command, &tx]()
+                 -> expected::Result<void, validation::CommandNameAndError> {
         // check transaction's commands validness
         const auto &commands = tx.commands();
-        std::string cmd_error;
+        validation::CommandNameAndError cmd_name_error;
         for (size_t i = 0; i < commands.size(); ++i) {
           // in case of failed command, rollback and return
           if (not execute_command(commands[i], i)
-                      .match([](expected::Value<void> &) { return true; },
-                             [&cmd_error](expected::Error<std::string> &error) {
-                               cmd_error = error.error;
-                               return false;
-                             })) {
+                      .match(
+                          [](expected::Value<void> &) { return true; },
+                          [&cmd_name_error](
+                              expected::Error<validation::CommandNameAndError>
+                                  &error) {
+                            cmd_name_error = error.error;
+                            return false;
+                          })) {
             transaction_->exec("ROLLBACK TO SAVEPOINT savepoint_;");
-            return expected::makeError(cmd_error);
+            return expected::makeError(cmd_name_error);
           }
         }
         // success
--- a/irohad/ametsuchi/impl/temporary_wsv_impl.cpp
+++ b/irohad/ametsuchi/impl/temporary_wsv_impl.cpp
@@ -75,12 +75,10 @@ namespace iroha {
 
       transaction_->exec("SAVEPOINT savepoint_;");
 
-      return expected::map_error<std::string>(
-                 apply_function(tx, *wsv_),
-                 [](std::string &error) { return error; })
-                 | [this,
-                    &execute_command,
-                    &tx]() -> expected::Result<void, std::string> {
+      return apply_function(tx, *wsv_) |
+                 [this,
+                  &execute_command,
+                  &tx]() -> expected::Result<void, std::string> {
         // check transaction's commands validness
         const auto &commands = tx.commands();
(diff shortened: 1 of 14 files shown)


irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -76,8 +134,31 @@ namespace iroha {
 
       log_->info("transactions in verified proposal: {}",
                  validated_proposal.transactions().size());
-      return std::make_shared<decltype(validated_proposal)>(
-          validated_proposal.getTransport());
+      return std::make_pair(std::make_shared<decltype(validated_proposal)>(
+                                validated_proposal.getTransport()),
+                            transactions_errors_log);
+    }  // namespace validation
```
lebdron: `// namespace validation` does not look right, it should be removed because namespace is closed below.

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -76,8 +134,31 @@ namespace iroha {
...
+
+    std::string StatefulValidatorImpl::formSignaturesErrorMsg(
+        const shared_model::interface::types::SignatureRangeType &signatures,
+        const std::vector<shared_model::interface::types::PubkeyType>
+            &signatories) {
+      using namespace std::string_literals;
```
lebdron: Is this required? Maybe use `.append(x.toString()).append('\n')` instead?

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -29,44 +32,99 @@ namespace iroha {
       log_ = logger::log("SFV");
     }
 
-    std::shared_ptr<shared_model::interface::Proposal>
+    shared_model::interface::types::VerifiedProposalAndErrors
     StatefulValidatorImpl::validate(
         const shared_model::interface::Proposal &proposal,
         ametsuchi::TemporaryWsv &temporaryWsv) {
+      using namespace std::string_literals;
```
lebdron: It doesn't look used in this method, so it can be removed.

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -29,44 +32,99 @@ namespace iroha {
...
+                    (boost::format("stateful validator error: could not fetch "
+                                   "signatories of "
+                                   "account %s")
+                     % tx.creatorAccountId())
+                        .str());
+              }
+              return expected::makeError(
+                  (boost::format(
+                       "stateful validator error: not enough "
+                       "signatures on account %s; account's quorum %d, "
```
lebdron: Not sure if I understood the message correctly, but maybe "not enough signatures in transaction" is more straightforward? Because the error appears when `tx.sigs < acc.quorum`.

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -93,37 +94,37 @@ namespace iroha {
...
-                         std::make_pair(error.error, tx_index));
-                     return false;
-                   });
+            .match(
+                [](expected::Value<void> &) { return true; },
+                [&transactions_errors_log,
+                 &proposal, &tx](expected::Error<std::vector<std::string>> &error) {
+                  auto txs = proposal.transactions();
+                  transactions_errors_log.push_back(std::make_pair(
+                      error.error, boost::range::find(txs, tx) - txs.begin()));
```
lebdron: Still, please consider using `boost::adaptors::filtered` and `boost::adaptors::indexed` to avoid linear `find`.

Akvinikym: The problem is that there's no way one can pass additional arguments to ```filtered```, I tried to use bind or lambda capture, but failed

irohad/validation/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ target_link_libraries(stateful_validator
     )
 
 add_library(chain_validator
-    impl/chain_validator_impl.cpp)
+    impl/chain_validator_impl.cpp stateful_validator_common.hpp)
```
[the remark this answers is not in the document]
lebdron: Also hpp files are not required in cmake targets
http://voices.canonical.com/jussi.pakkanen/2013/03/26/a-list-of-common-cmake-antipatterns/

irohad/validation/stateful_validator_common.hpp
```diff
@@ -0,0 +1,29 @@
...
+#define IROHA_STATEFUL_VALIDATOR_COMMON_HPP
+
+#include "common/types.hpp"
+
+namespace iroha {
+  namespace validation {
+
+    /// Type of per-transaction errors, which appeared during validation
+    /// process; contains errors themselves and transaction hashes
+    using TransactionsErrors = std::vector<
```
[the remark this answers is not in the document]
lebdron: The issue is that this structure will not be used as map in tx processor, since it has to iterate through all txs and push statuses to observable. Maybe change it to map when there will be a case for it?

irohad/validation/CMakeLists.txt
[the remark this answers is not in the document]
Akvinikym: Yeah, it's added by CLion for some reason, when a new file created, and I forgot to remove it :(

irohad/validation/stateful_validator_common.hpp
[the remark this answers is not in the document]
kamilsa: We can leave it this way. Hopefully if there will be a use case for map one won't forget to update it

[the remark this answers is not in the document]
Akvinikym: I've already done this, soon the P.R. will come

irohad/ametsuchi/impl/temporary_wsv_impl.cpp
```diff
@@ -38,49 +38,70 @@ namespace iroha {
...
+                   [command_index](CommandError &error) {
+                     return validation::CommandNameAndError{
+                         error.command_name,
+                         (boost::format("stateful validation error: could "
+                                        "not validate "
+                                        "command with index %d: %s")
+                          % command_index % error.toString())
+                             .str()};
+                   })
+            | [this, command_index, &command] {
```
l4l: I think arbitrary ordering is applicable here so maybe use `and_res`?

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -29,34 +33,88 @@ namespace iroha {
...
+                return expected::makeValue(*account);
+              }
+              return expected::makeError(validation::CommandNameAndError{
+                  "Initial transaction verification: no such account",
+                  (boost::format("stateful validator error: could not fetch "
+                                 "account with id %s")
+                   % tx.creatorAccountId())
+                      .str()});
+            }() |
+                [&](const auto &account)
```
l4l: would be nice to move out this labmdas, so the code will be more readable

irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -76,8 +135,29 @@ namespace iroha {
...
       log_->info("transactions in verified proposal: {}",
                  validated_proposal.transactions().size());
-      return std::make_shared<decltype(validated_proposal)>(
-          validated_proposal.getTransport());
+      return std::make_pair(std::make_shared<decltype(validated_proposal)>(
+                                validated_proposal.getTransport()),
+                            transactions_errors_log);
+    }
+
+    std::string StatefulValidatorImpl::formSignaturesErrorMsg(
```
l4l: shouldn't be in the class interface, declare it as a non-class static function

irohad/validation/stateful_validator_common.hpp
```diff
@@ -0,0 +1,46 @@
...
+  namespace validation {
+
+    /// Name of the failed command
+    using CommandName = std::string;
+
+    /// Error, with which the command failed
+    using CommandError = std::string;
+
+    /// Failed command's name and error
+    using CommandNameAndError = std::pair<CommandName, CommandError>;
```
l4l: Why isn't it a result?

irohad/validation/stateful_validator_common.hpp
```diff
@@ -0,0 +1,46 @@
...
+    /// process; contains names of commands, commands errors themselves and
+    /// transaction hashes
+    using TransactionsErrors =
+        std::vector<std::pair<CommandNameAndError,
+                              shared_model::interface::types::HashType>>;
+
+    /// Type of verified proposal and errors appeared in the process; first
+    /// dimension of errors vector is transaction, second is error itself with
+    /// number of transaction, where it happened
+    using VerifiedProposalAndErrors =
```
l4l: Same

irohad/validation/impl/stateful_validator_impl.cpp
Akvinikym: Isn't it? From my point of view, those lambdas form a logical pipeline, which is very nice...

irohad/validation/impl/stateful_validator_impl.cpp
Akvinikym: It is not (and I think not going to be) used anywhere, so maybe it's a bad idea to move it out of the class scope

irohad/validation/stateful_validator_common.hpp
Akvinikym: Why should it be a result? It can't be success or fail, it's just a structure for name and error itself

irohad/validation/stateful_validator_common.hpp
Akvinikym: Because we can't choose, we have both proposal with good transactions and errors for those, which did not pass stateful validation

PR #1497, opened 2018-06-25
Check block in mutable storage

kamilsa:
### Description of the Change

In chain validator we might need to check some logic on either EmptyBlock or Block types. Current interface of mutable storage does not allow that. To overcome this issue check method has been introduced which applies predicates logic on the block without actually applying the block to the storage.

### Benefits

We can validate block on storage without applying it

### Possible Drawbacks 

Probably we don't need check method to take AbstractBlock as parameter but use EmptyBlock instead.

### Usage Examples or Tests

Mutable storage test has been introduced

### DIFF
--- a/irohad/ametsuchi/mutable_storage.hpp
+++ b/irohad/ametsuchi/mutable_storage.hpp
@@ -24,7 +24,8 @@
 namespace shared_model {
   namespace interface {
     class Block;
-  }
+    class AbstractBlock;
+  }  // namespace interface
 }  // namespace shared_model
 
 namespace iroha {
@@ -38,6 +39,22 @@ namespace iroha {
      */
     class MutableStorage {
      public:
+      template <typename T>
+      using MutableStoragePredicateType =
+          std::function<bool(const T &,
+                             WsvQuery &,
+                             const shared_model::interface::types::HashType &)>;
+
+      /**
+       * Checks if block satisfies predicated
+       * @param block block to be checked
+       * @param predicate function returning true if predicate satisfied and
+       * false otherwise
+       * @return result of predicate
+       */
+      virtual bool check(const shared_model::interface::AbstractBlock &block,
+                         MutableStoragePredicateType<decltype(block)>) = 0;
+
       /**
        * Applies a block to current mutable state
        * using logic specified in function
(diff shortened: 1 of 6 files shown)


test/module/irohad/ametsuchi/mutable_storage_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+                [&expected_block, &expected_res](
+                    const auto &block, const auto &, const auto &) {
+                  EXPECT_EQ(expected_block, block);
+                  return expected_res;
+                }));
+}
+
+INSTANTIATE_TEST_CASE_P(MutableStorageParameterizedTest,
+                        MutableStorageTest,
+                        Bool(), );
```
kamilsa: comma is needed to make it compile https://github.com/google/googletest/issues/1419

l4l:
please write meaningful pr title

irohad/ametsuchi/mutable_storage.hpp
```diff
@@ -38,6 +39,22 @@ namespace iroha {
      */
     class MutableStorage {
      public:
+      template <typename T>
+      using MutableStoragePredicateType =
```
muratovv: Add documentation for using.

irohad/ametsuchi/mutable_storage.hpp
```diff
@@ -38,6 +39,22 @@ namespace iroha {
...
+                             const shared_model::interface::types::HashType &)>;
+
+      /**
+       * Checks if block satisfies predicated
+       * @param block block to be checked
+       * @param predicate function returning true if predicate satisfied and
+       * false otherwise
+       * @return result of predicate
+       */
+      virtual bool check(const shared_model::interface::AbstractBlock &block,
```
muratovv: Maybe rework abstract block with block variant?

test/module/irohad/ametsuchi/mutable_storage_test.cpp
muratovv: Maybe put a comment in the code?

muratovv: Maybe put a comment in the code?

PR #1502, opened 2018-06-26
Add oneof case bound check

l4l:
### Description of the Change

Command/Query cases aren't properly validated

### Benefits

Lesser bugs

### Possible Drawbacks 

Validators now require protobuf backend objects but should be fixed with IR-1040


shared_model/validators/transaction_validator.hpp
```diff
@@ -265,7 +266,22 @@ namespace shared_model {
...
+        constexpr auto cmd_begin = iroha::protocol::Command::kAddAssetQuantity;
+        constexpr auto cmd_end = cmd_begin
+            + boost::mpl::size<shared_model::interface::Command::
+                                   CommandListType>::type::value;
         for (const auto &command : tx.commands()) {
+          auto cmd_case =
+              static_cast<const shared_model::proto::Command &>(command)
+                  .getTransport()
+                  .command_case();
+          if (cmd_begin > cmd_case || cmd_case >= cmd_end) {
```
lebdron: Same as in query, `getTransport().has_command()` is simpler and covers the desired case.

shared_model/validators/query_validator.hpp
```diff
@@ -174,9 +174,25 @@ namespace shared_model {
...
+        constexpr auto qry_begin = iroha::protocol::Query_Payload::kGetAccount;
+        constexpr auto qry_end =
+            qry_begin
+            + boost::mpl::size<
+                  shared_model::interface::Query::QueryListType>::type::value;
+        auto qry_case = static_cast<const shared_model::proto::Query &>(qry)
+                            .getTransport()
+                            .payload()
+                            .query_case();
+        if (qry_begin > qry_case || qry_case >= qry_end) {
```
lebdron: `payload().has_query()` can be used instead, since it's simpler and covers the case.

test/module/shared_model/validators/query_validator_test.cpp
```diff
@@ -66,6 +66,22 @@ TEST_F(QueryValidatorTest, StatelessValidTest) {
...
 }
 
+/**
+ * @given Protobuf query object with unset query
+ * @when validate is called
+ * @then there is a error returned
+ */
+TEST_F(QueryValidatorTest, UnsetQuery) {
+  iroha::protocol::Query qry;
+  auto *meta = new iroha::protocol::QueryPayloadMeta();
```
lebdron: Use `mutable_meta` instead to avoid `new` call.

test/module/shared_model/validators/transaction_validator_test.cpp
```diff
@@ -101,13 +101,25 @@ TEST_F(TransactionValidatorTest, StatelessValidTest) {
...
+ * @given Protobuf transaction object with unset command
+ * @when validate is called
+ * @then there is a error returned
+ */
+TEST_F(TransactionValidatorTest, UnsetCommand) {
+  iroha::protocol::Transaction tx = generateEmptyTransaction();
+  tx.mutable_payload()->set_creator_account_id(account_id);
+  tx.mutable_payload()->set_created_time(created_time);
+  auto answer = transaction_validator.validate(proto::Transaction(tx));
+  ASSERT_TRUE(answer.hasErrors());
```
lebdron: Answer here is `Transaction: [[Transaction should contain at least one command ]]`, which is not the described case.
There should be an additional call `tx.mutable_payload()->add_commands();`

test/module/shared_model/validators/transaction_validator_test.cpp
```diff
@@ -101,13 +101,26 @@ TEST_F(TransactionValidatorTest, StatelessValidTest) {
...
+ * @given Protobuf transaction object with unset command
+ * @when validate is called
+ * @then there is a error returned
+ */
+TEST_F(TransactionValidatorTest, UnsetCommand) {
+  iroha::protocol::Transaction tx = generateEmptyTransaction();
+  tx.mutable_payload()->set_creator_account_id(account_id);
+  tx.mutable_payload()->set_created_time(created_time);
+  auto answer = transaction_validator.validate(proto::Transaction(tx));
+  tx.mutable_payload()->add_commands()->clear_add_asset_quantity();
```
lebdron: After discussion, it seems that `clear_add_asset_quantity()` is redundant, since `add_commands()` does not set `add_asset_quantity` by default, and compiler won't probably optimize any calls. So `clear_add_asset_quantity()` call can be removed.

PR #1506, opened 2018-06-28
Proposal and block txs validation

kamilsa:
### Description of the Change

Adds check on transactions collections in proposal and block using transaction collection validator
This validator will be lately used by endpoint which will receive list of transactions

### Benefits

After batch logic check is implemented it will be checked in proposal and block for transactions

### Possible Drawbacks 

Transaction collection validator still does not have


l4l:
Could you understand why is that needed? It looks like significant step towards the validator over-complication

If I understood correctly, most of the pr aimed for that diff lines:
screenshot_20180628_221201

But what's the problem with the previous code then if it did almost the same thing?

kamilsa:
We need to be able to validate transactions collections (in proposal, block or just list of transactions) using either signed or unsigned transaction collection validator. Now we don't have fair implementations of them

shared_model/validators/transactions_collection/unsigned_transactions_collection_validator.cpp
```diff
@@ -5,14 +5,42 @@
 
 #include "validators/transactions_collection/unsigned_transactions_collection_validator.hpp"
 
+#include <boost/format.hpp>
```
Akvinikym: The same with library include at the top of the file

shared_model/validators/transactions_collection/unsigned_transactions_collection_validator.hpp
```diff
@@ -15,9 +15,15 @@ namespace shared_model {
      * Unsigned transactions collection validator allows to some transaction
      * from the collection to be unsigned. Batch logic should be checked
      */
+    template <typename TransactionValidator>
     class UnsignedTransactionsCollectionValidator
-        : public TransactionsCollectionValidator {
+        : public TransactionsCollectionValidator<TransactionValidator> {
+     private:
+      using ParentType = TransactionsCollectionValidator<TransactionValidator>;
```
Akvinikym: This is used twice in the code, which can lead to mistakes in case we want to change names later... Maybe, it's better to define this type as ```protected``` in the base class? It also accepts template parameter etc, and there will be no code duplication as well

shared_model/validators/transactions_collection/signed_transactions_collection_validator.cpp
```diff
@@ -5,14 +5,42 @@
 
 #include "validators/transactions_collection/signed_transactions_collection_validator.hpp"
 
+#include <boost/format.hpp>
```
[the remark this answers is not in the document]
l4l: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
First one should be related header

PR #1510, opened 2018-06-28
Remove PermissionSet constraints

l4l:
### Description of the Change

Remove PermissionSet constraints so it's possible to perform `CreateRole` with empty permission set (in other words role doesn't have to have any permission)

### Benefits

Possible to create role without rights

### Possible Drawbacks 

None?

### Usage Examples or Tests

`make create_role_test && test_bin/create_role_test`

### DIFF
--- a/shared_model/validators/transaction_validator.hpp
+++ b/shared_model/validators/transaction_validator.hpp
@@ -22,7 +22,7 @@
 #include <boost/variant/static_visitor.hpp>
 
 #include "backend/protobuf/permissions.hpp"
-#include "interfaces/transaction.hpp"
+#include "backend/protobuf/transaction.hpp"
 #include "validators/answer.hpp"
 
 namespace shared_model {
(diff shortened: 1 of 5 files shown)


shared_model/validators/transaction_validator.hpp
```diff
@@ -118,7 +118,14 @@ namespace shared_model {
         addInvalidCommand(reason, "CreateRole");
 
         validator_.validateRoleId(reason, cr.roleName());
-        validator_.validateRolePermissions(reason, cr.rolePermissions());
+        for (auto i : static_cast<const shared_model::proto::CreateRole &>(cr)
```
victordrobny: Why can't you use rolePermissions()? As far as I understood it returns set of Role permissions and you can iterate over it

l4l: That may trigger out of bounds for bitset, if the proto is invalid

PR #1521, opened 2018-06-30
Fix issue with synchronization after time gap

l4l:
### Description of the Change

This pr introduces 3 main prerequisite fixes:
- Reordering of *Validator ctor parameters, so that `FieldValidator` is the first one
- Reusing existing `FieldValidator` in SignableModelValidator
- `FieldValidator` can now has time provider other than `iroha::time::now`

And the main fix is for BlockValidator. It now performs time validation based on the block field, so it's possible to synchronize blocks after a time gap passed.

### Benefits

Probably faster `SignableModelValidator`, since it don't need to allocate field validator
Peers can synchronize "expired" (in terms of current time) blocks from the network

### Possible Drawbacks 

`BlockValidator` is allocated on each `BlockLoaderImpl::retrieveBlock` (should it be a class field?)
`BlockLoaderImpl` cannot be parametrized with `BlockValidator` (though it wasn't used)

### Usage Examples or Tests

Test hasn't been added, but the existing one was modified and can verify the behavior:
`test_bin/block_validator_test`


irohad/network/impl/block_loader_impl.cpp
```diff
@@ -45,6 +42,18 @@ const char *kTopBlockRetrieveFail = "Failed to retrieve top block";
 const char *kPeerRetrieveFail = "Failed to retrieve peers";
 const char *kPeerFindFail = "Failed to find requested peer";
 
+struct TimerWrapper : public val::FieldValidator {
+  TimerWrapper(iroha::ts64_t t)
```
Alexey-N-Chernyshov: Codacy asks to make the constructor `explicit` and I agree with it.

PR #1536, opened 2018-07-04
Check the signatures for its size

l4l:
Signatures are validated properly, but there is no check for their existence. The pr fixes that.
Also adds test for the Torii (test for the FieldValidator is pretty tedious) for that case

### DIFF
--- a/shared_model/validators/field_validator.cpp
+++ b/shared_model/validators/field_validator.cpp
@@ -309,6 +309,9 @@ namespace shared_model {
         ReasonsGroupType &reason,
         const interface::types::SignatureRangeType &signatures,
         const crypto::Blob &source) const {
+      if (boost::size(signatures) == 0) {
+        reason.second.push_back("Signatures cannot be empty");
+      }
       for (const auto &signature : signatures) {
         const auto &sign = signature.signedData();
         const auto &pkey = signature.publicKey();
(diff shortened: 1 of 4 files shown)


test/module/irohad/torii/torii_service_test.cpp
```diff
@@ -432,3 +432,34 @@ TEST_F(ToriiServiceTest, StreamingNoTx) {
   ASSERT_EQ(torii_response.at(0).tx_status(),
             iroha::protocol::TxStatus::NOT_RECEIVED);
 }
+
+/**
+ * @given Torii service and the valid tx without signature
+ * @when sending tx to torii
+ * @then the response is STATELESS_VALIDATION_FAILED
+ */
+TEST_F(ToriiServiceTest, EmptySignatures) {
```
lebdron: Please move to `acceptance/tx_acceptance_test.cpp` and use
```
shared_model::proto::Transaction tx =
      baseTx<TestTransactionBuilder>().build();
```

shared_model/validators/field_validator.cpp
```diff
@@ -309,6 +309,9 @@ namespace shared_model {
         ReasonsGroupType &reason,
         const interface::types::SignatureRangeType &signatures,
         const crypto::Blob &source) const {
+      if (boost::size(signatures) == 0) {
```
lebdron: Please use `boost::empty`
https://github.com/hyperledger/iroha/pull/1543#discussion_r201229007

PR #1540, opened 2018-07-05
GetAccountDetail: high interface

Akvinikym:
### Description of the Change

Second and last part of changing getAccountDetail interface. Here, the interfaces and Proto message are extended with new optional fields.

### Benefits

New interface is working and can be used.

### Possible Drawbacks 

None


shared_model/interfaces/queries/get_account_detail.hpp
```diff
@@ -33,6 +35,16 @@ namespace shared_model {
        */
       virtual const types::AccountIdType &accountId() const = 0;
 
+      /**
+       * @return key from key-value storage
+       */
+      virtual boost::optional<types::AccountDetailKeyType> key() const = 0;
```
nickaleks: Please add a comment about why the fields are optional and what is the behavior

shared_model/bindings/model_query_builder.hpp
```diff
@@ -135,11 +135,14 @@ namespace shared_model {
...
        * Retrieves details for a given account
        * @param account_id - account to retrieve details from
-       * @param detail - key to retrieve
+       * @param key - under which keys data should be returned
+       * @param writer - from which writers details should be returned
        * @return builder with getAccountDetail query inside
        */
       ModelQueryBuilder getAccountDetail(
-          const interface::types::AccountIdType &account_id);
+          const interface::types::AccountIdType &account_id = "",
```
nickaleks: Can the account be empty?

shared_model/builders/protobuf/builder_templates/query_template.hpp
```diff
@@ -155,10 +155,20 @@ namespace shared_model {
         });
       }
 
-      auto getAccountDetail(const interface::types::AccountIdType &account_id) {
+      auto getAccountDetail(const interface::types::AccountIdType &account_id = "",
+                            const interface::types::AccountDetailKeyType &key = "",
+                            const interface::types::AccountIdType &writer = "") {
         return queryField([&](auto proto_query) {
```
nickaleks: pass by reference?

shared_model/backend/protobuf/queries/impl/proto_get_account_detail.cpp
```diff
@@ -27,7 +27,23 @@ namespace shared_model {
         : GetAccountDetail(std::move(o.proto_)) {}
 
     const interface::types::AccountIdType &GetAccountDetail::accountId() const {
-      return account_detail_.account_id();
+      return account_detail_.opt_account_id_case()
```
kamilsa: I think `account_detail_.has_account_id()` would be more readable

Akvinikym: ```has_account_id()``` is generated as ```private```, so the other variant to check, if it's empty, is to use switch-case, but here it will be awful IMHO

PR #1543, opened 2018-07-05
New client api

l4l:
### Description of the Change

Step toward migration to the native SDK. This pr introduces 3 main function for client code: `hash`, `validate` and `sign`. Related test for java and python have been added.

Also requires #1536 for test passing

### Possible Drawbacks 

A lot of code duplication, feel free to propose ideas for fixes

### Usage Examples or Tests

Added `python_client_test` and java one, as previously in `java_builders_test`

### Alternate Designs

SWIG can accept pair `char*, size_t` and remain using native stuff (like byte[] in java), although it's still requires non-trivial (e.g memory management) for return types. So for being consistent interface use std::vector both for return value and input ones


shared_model/validators/field_validator.cpp
```diff
@@ -310,6 +310,9 @@ namespace shared_model {
         ReasonsGroupType &reason,
         const interface::types::SignatureRangeType &signatures,
         const crypto::Blob &source) const {
+      if (boost::size(signatures) == 0) {
```
kamilsa: Use `boost::empty` instead. `boost::size` might iterate over the collection, whereas `boost::empty` does not

test/module/irohad/network/block_loader_test.cpp
```diff
@@ -125,12 +132,18 @@ TEST_F(BlockLoaderTest, ValidWhenOneBlock) {
 
   // time validation should work based on the block field
   // so it should pass statless BlockLoader validation
```
kamilsa: please fix type with word stateless here and in next comments

shared_model/bindings/client_api.cpp
```diff
@@ -0,0 +1,122 @@
...
+        return boost::make_optional(
+            val.validate(proto::Transaction(tx)).reason());
+      };
+      if (s) {
+        return s.value();
+      }
+      return std::string("unknown object");
+    }
+
+    using namespace iroha;
```
kamilsa: You have same thing on line 31. You can remove this line

shared_model/bindings/client_api.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace shared_model {
+  namespace bindings {
+
+    std::string convert(const Blob &blob) {
+      return std::string(blob.begin(), blob.end());
+    }
+
+    template <typename Proto>
+    boost::optional<Proto> get(const std::string &blob) {
```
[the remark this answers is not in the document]
l4l: that's internal function
don't really think it should be that descriptive

shared_model/bindings/client_api.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+        auto sig = tx.add_signatures();
+        sig->set_signature(crypto::toBinaryString(signature));
+        sig->set_pubkey(crypto::toBinaryString(key.publicKey()));
+        return boost::make_optional(tx);
+      };
+      if (s) {
+        return proto::makeBlob(s.value()).blob();
+      }
+      logger::log("client_api")->warn("Unknown protobuf object");
```
Warchant: what if client does not want to use stdout logger? how to disable it?

I guess there is no way.

IMO all errors should be explicitly mentioned in exceptions

test/module/shared_model/bindings/ClientTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @Test
+  void sign() {
+    Transaction tx = Transaction.newBuilder()
+                         .setPayload(Transaction.Payload.newBuilder().setReducedPayload(
+                             defaultPayload().addCommands(validAddPeer())))
+                         .build();
+    assertEquals(tx.getSignaturesList().size(), 0);
+    Transaction signed_tx = deserialize(iroha.signTransaction(serialize(tx), keys));
+    assertEquals(signed_tx.getSignaturesList().size(), 1);
```
Warchant: it is not enough to check that signature has size 1, this signature may be wrong

add signature verification and comparison with hardcoded value

test/module/shared_model/bindings/ClientTest.java
```diff
@@ -0,0 +1,142 @@
...
+    return Command.newBuilder().setAddPeer(Commands.AddPeer.newBuilder().setPeer(
+        Primitive.Peer.newBuilder()
+            .setAddress("127.0.0.1:50500")
+            .setPeerKey(ByteString.copyFrom(String.format("%32.32s", " ").getBytes()))));
+  }
+
+  Transaction.Payload.ReducedPayload.Builder defaultPayload() {
+    return Transaction.Payload.ReducedPayload.newBuilder()
+        .setCreatorAccountId("admin@test")
+        .setCreatedTime(System.currentTimeMillis())
```
Warchant: you should fix time

test/module/shared_model/bindings/ClientTest.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  @Test
+  void hash() {
+    Transaction tx = Transaction.newBuilder()
+                         .setPayload(Transaction.Payload.newBuilder().setReducedPayload(
+                             defaultPayload().addCommands(validAddPeer())))
+                         .build();
+    ByteVector h = iroha.hashTransaction(serialize(tx));
+    assertEquals(h.size(), 32);
```
Warchant: when you fix time, you should also add assert for hash content

test/module/shared_model/bindings/client-test.py
```diff
@@ -0,0 +1,67 @@
...
+  def valid_add_peer_command(self):
+    command = cmd.Command()
+    command.add_peer.peer.address = "127.0.0.1:50500"
+    command.add_peer.peer.peer_key = b'A' * 32
+    return command
+
+  def unsigned_tx(self):
+    tx = blk.Transaction()
+    tx.payload.reduced_payload.creator_account_id = "admin@test"
+    tx.payload.reduced_payload.created_time = int(time() * 1000)
```
Warchant: same here, you should fix all values in test

shared_model/bindings/client_api.cpp
l4l: 2>/dev/null

What kind of exception do you mean?

Warchant: 2>/dev/null is not applicable.

Just imagine this situation - you are poor rust developer, you need to use some native library. You write bindings for this library and execute your favorite function. But it starts to write data to STDOUT. 

No good

test/module/shared_model/bindings/ClientTest.java
Warchant: by "fix" I mean, you should hardcode the value

shared_model/bindings/client_api.cpp
stinger112: ...or you want to disable any logging at all and just catch errors in exceptions.

test/module/shared_model/bindings/ClientTest.java
Warchant: Exactly. Bindings do not allow to fix values in texts. Talk about this with team

shared_model/bindings/client_api.cpp
l4l: let it be then, I'll just remove

PR #1548, opened 2018-07-09
Update transaction sequence

kamilsa:
### Description of the Change

Updates transaction sequence to have batches method which allows to iterate over the batches in transaction sequence

### Benefits

Batches method is introduced, tests on transaction sequence are included

### Possible Drawbacks 

To be able to validate collection of shared pointers on transactions new method validatePointers was introduced in collection validator

### Usage Examples or Tests

Tests are included


shared_model/validators/transactions_collection/transactions_collection_validator.hpp
```diff
@@ -27,14 +28,27 @@ namespace shared_model {
...
           : transaction_validator_(transactions_validator) {}
 
+      // TODO: IR-1505, igor-egorov, 2018-07-05 Remove method below when
+      // proposal and block will return collection of shared transactions
       /**
        * Validates collection of transactions
        * @param transactions collection of transactions
        * @return Answer containing errors if any
        */
       virtual Answer validate(
```
Akvinikym: Maybe, method's implementation should be either moved back to .cpp file, or it should not be declared as ```virtual```?

shared_model/validators/transactions_collection/transactions_collection_validator.hpp
```diff
@@ -27,14 +28,27 @@ namespace shared_model {
...
           const interface::types::TransactionsForwardCollectionType
-              &transactions) const = 0;
+              &transactions) const {
+        interface::types::SharedTxsCollectionType res;
+        std::for_each(
```
Akvinikym: I think, it will be better if written as follows:
```
std::transform(std::begin(transactions),
               std::end(transactions),
               std::back_inserter(res),
               [](const auto &tx) { return clone(tx); }); 
```
And also, for the sake of performance, before transforming the space for ```res``` should be reserved, either in vector constructor, or with separate command, such as ```res.reserve(transactions.size());```

shared_model/validators/transactions_collection/unsigned_transactions_collection_validator.cpp
```diff
@@ -5,27 +5,25 @@
 
 #include "validators/transactions_collection/unsigned_transactions_collection_validator.hpp"
 
-#include <boost/format.hpp>
```
Akvinikym: Why? It's used in file's method

test/module/shared_model/backend_proto/proto_transaction_sequence_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+auto prepareTransactionBuilder(const std::string &creator) {
+  return TestTransactionBuilder().creatorAccountId(creator);
+}
+
+/**
+ * Creates atomic batch from provided creator accounts
+ * @param creators vector of creator account ids
+ * @return atomic batch of the same size as the size of creator account ids
+ */
+auto createAtomicBatch(std::vector<std::string> creators) {
```
Akvinikym: I assume, this function has a potential to be heavily used in future tests, so maybe move it to some kind of common files or something like that?

Akvinikym: And also, it would wise to parametrize it with an argument of type ```types::BatchType```, and create the corresponding batch type, so we wouldn't have two separate functions for ordered and atomic batches, with a lot of redundant code

test/module/shared_model/backend_proto/proto_transaction_sequence_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+  EXPECT_CALL(res, validatePointers(_)).WillOnce(Return(answer));
+
+  std::shared_ptr<interface::Transaction> tx(
+      clone(prepareTransactionBuilder("account@domain").build()));
+
+  auto tx_sequence = interface::TransactionSequence::createTransactionSequence(
+      std::vector<decltype(tx)>{tx, tx, tx}, res);
+
+  ASSERT_FALSE(framework::expected::val(tx_sequence));
```
Akvinikym: Maybe ```ASSERT_TRUE(framework::expected::err(tx_sequence));``` is better?

test/module/shared_model/backend_proto/proto_transaction_sequence_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+  EXPECT_CALL(txs_validator, validatePointers(_))
+      .WillOnce(Return(validation::Answer()));
+
+  size_t batches_number = 3;
+  size_t txs_in_batch = 2;
+  size_t single_transactions = 1;
+
+  interface::types::SharedTxsCollectionType tx_collection;
+  for (size_t i = 0; i < batches_number; i++) {
+    std::vector<std::string> creators;
```
Akvinikym: Again, maybe you should reserve space for vectors in constructors?

test/module/shared_model/backend_proto/proto_transaction_sequence_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    auto batch = createAtomicBatch(creators);
+    tx_collection.insert(tx_collection.begin(), batch.begin(), batch.end());
+  }
+  for (size_t i = 0; i < single_transactions; i++) {
+    tx_collection.emplace_back(
+        clone(prepareTransactionBuilder("single_tx_account@domain"
+                                        + std::to_string(i))
+                  .build()));
```
Akvinikym: I'm not sure, but is it according to ```clang-format```?

shared_model/interfaces/iroha_internal/transaction_sequence.cpp
```diff
@@ -25,21 +25,54 @@ namespace shared_model {
...
+          batches.emplace_back(std::vector<std::shared_ptr<Transaction>>{tx});
+        }
+      }
+      for (auto it : extracted_batches) {
+        batches.emplace_back(it.second);
+      }
+
+      batches_.emplace(batches);
+      return batches;
+    }
```
Solonets: Why not Lazy?

kamilsa: I've tried. But it resulted into runtime error. I don't know why

test/module/shared_model/backend_proto/proto_transaction_sequence_test.cpp
kamilsa: Does not matter, but ok

test/module/shared_model/backend_proto/proto_transaction_sequence_test.cpp
kamilsa: It should be put into the framework namespace. As you mentioned we can parameterise it with BatchType. But we might also need to set different creators. Or we might want to set transactions with certain commands. I would like to define possible use cases for such function and then rework it and put into separate file

test/module/shared_model/backend_proto/proto_transaction_sequence_test.cpp
kamilsa: Surprisingly, yes

PR #1552, opened 2018-07-10
Transaction Batches Stateful Validation

Akvinikym:
### Description of the Change

Adds stateful validation checks for transaction batches:

1. If the transaction does not belong to any batch, proceed as usual
2. If the transaction belongs to ordered batch, in case of failure just move to the next transaction again
3. If the transaction belongs to atomic batch, failure of it leads to failure of all transactions in that batch

Also, tests for stateful validator and transaction batch validation were added.

### Benefits

Logic of transaction batches is improved.

### Possible Drawbacks 

Additional overhead for validation and WSV save points management.

### Alternative designs

Savepoint wrapper's design can be a theme of discussion. Shortly, I have decided to implement it as a nested structure of TemporaryWsv, because:
1. It cannot be used without temp WSV - there's just no sense in it
2. It must "see" some of private fields of temp WSV in order to function

There are two alternatives I see: declare it in some kind of temp WSV common file, and then either parametrize it with a template, or make necessary field of temp WSV friends to the wrapper. Both solutions, in my opinion, looks worse, than the one implemented


irohad/ametsuchi/impl/temporary_wsv_impl.cpp
```diff
@@ -78,16 +78,28 @@ namespace iroha {
...
           }
         }
         // success
-        transaction_->exec("RELEASE SAVEPOINT savepoint_;");
+        releaseSavepoint("savepoint_");
         return {};
       };
     }
 
+    void TemporaryWsvImpl::createSavepoint(const std::string &name) {
```
l4l: It's barely safe. Return please new object with RAII behavior, such that dtor calls rollback, unless specified explicit release

irohad/ametsuchi/impl/temporary_wsv_impl.hpp
```diff
@@ -30,6 +30,20 @@ namespace iroha {
   namespace ametsuchi {
     class TemporaryWsvImpl : public TemporaryWsv {
      public:
+      struct SavepointWrapper : public TemporaryWsv::SavepointWrapper {
```
kamilsa: Rename it to `SavepointWrapperImpl`

irohad/ametsuchi/impl/temporary_wsv_impl.hpp
```diff
@@ -39,17 +53,14 @@ namespace iroha {
               const shared_model::interface::Transaction &, WsvQuery &)>
               function) override;
 
-      void createSavepoint(const std::string &name) override ;
-
-      void releaseSavepoint(const std::string &name) override ;
-
-      void rollbackToSavepoint(const std::string &name) override;
+      std::shared_ptr<TemporaryWsv::SavepointWrapper> createSavepoint(
```
kamilsa: Better to return `unique_ptr`

irohad/ametsuchi/temporary_wsv.hpp
```diff
@@ -39,6 +39,19 @@ namespace iroha {
...
      public:
+      /**
+       * Wrapper for savepoints in wsv state; rollbacks to savepoint, if
+       * destroyed without explicit release, releases it otherwise
+       */
+      struct SavepointWrapper {
+        /**
+         * Release the savepoint
+         */
+        virtual void release() = 0;
```
kamilsa: Can we put release logic into destructor?

irohad/ametsuchi/temporary_wsv.hpp
```diff
@@ -56,24 +69,14 @@ namespace iroha {
...
-       * @param name of savepoint to be rollbacked to
-       */
-      virtual void rollbackToSavepoint(const std::string &name) = 0;
+      virtual std::shared_ptr<SavepointWrapper> createSavepoint(const std::string &name) = 0;
 
       virtual ~TemporaryWsv() = default;
-    };
-  }  // namespace ametsuchi
+
+    };  // namespace ametsuchi
```
kamilsa: Fix namespace duplication

irohad/ametsuchi/temporary_wsv.hpp
l4l: I don't think that's a good idea. It completely implementation specific stuff

Akvinikym: Yeah, we've already spoken about this with Kamil, decided to leave it as-is

PR #1555, opened 2018-07-11
Use amount as a string in schema

l4l:
### Description of the Change

protobuf now for the amount `string` type
Removed legacy amount/amount

### Benefits

Easier to write code w/o bindings, code become simplier imo

### Possible Drawbacks 

String might have bad impact on performance


schema/primitive.proto
```diff
@@ -86,17 +85,18 @@ message Signature {
   bytes signature = 2;
 }
 
-message uint256 {
-  uint64 first = 1;
-  uint64 second = 2;
-  uint64 third = 3;
-  uint64 fourth = 4;
-}
+// message uint256 {
```
nickaleks: remove?

shared_model/interfaces/common_objects/amount.hpp
```diff
@@ -1,84 +1,72 @@
...
+
+      Amount *clone() const override;
+
+     private:
+      // lazy
+      template <typename T>
+      using Lazy = detail::LazyInitializer<T>;
+      const std::string amount_;
+
+      const Lazy<boost::multiprecision::uint256_t> multiprecision_repr_;
```
nickaleks: Why lazy?

shared_model/interfaces/common_objects/impl/amount.cpp
```diff
@@ -0,0 +1,70 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "interfaces/common_objects/amount.hpp"
+#include <regex>
```
nickaleks: Please add space

shared_model/interfaces/common_objects/impl/amount.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include <boost/multiprecision/cpp_dec_float.hpp>
+#include "utils/string_builder.hpp"
+
+namespace shared_model {
+  namespace interface {
+    Amount::Amount(const std::string &amount) : Amount(std::string(amount)) {}
```
nickaleks: Lazy is not initialized here

shared_model/interfaces/common_objects/impl/amount.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    std::string Amount::toString() const {
+      return detail::PrettyStringBuilder()
+          .init("Amount")
+          .append("value", amount_)
+          .finalize();
+    }
+
+    Amount *Amount::clone() const {
```
nickaleks: Since amount is not polymorphic, we can remove clone

shared_model/utils/amount_utils.cpp
```diff
@@ -53,10 +37,12 @@ namespace shared_model {
              % b.intValue().str())
                 .str()));
       }
-      return shared_model::builder::AmountBuilderWithoutValidator()
-          .precision(max_precision)
-          .intValue(val_a + val_b)
-          .build();
+      std::string val = (val_a + val_b).str();
+      if (max_precision) {
```
nickaleks: make condition explicit please

shared_model/utils/amount_utils.cpp
```diff
@@ -113,10 +101,12 @@ namespace shared_model {
         return iroha::expected::makeError(
             std::make_shared<std::string>("operation overflows number"));
       }
-      return shared_model::builder::AmountBuilderWithoutValidator()
-          .precision(new_precision)
-          .intValue(val_amount)
-          .build();
+      std::string val = val_amount.str();
+      if (new_precision) {
```
nickaleks: Same about explicit condition

test/module/irohad/execution/command_validate_execute_test.cpp
```diff
@@ -105,20 +105,7 @@ class CommandValidateExecuteTest : public ::testing::Test {
...
-        .build()
-        .match(
-            [&](expected::Value<
-                std::shared_ptr<shared_model::interface::Amount>> &v) {
-              balance = v.value;
-            },
-            [](expected::Error<std::shared_ptr<std::string>> &e) {
-              FAIL() << *e.error;
-            });
+    balance = std::make_shared<shared_model::interface::Amount>("1.50");
```
nickaleks: Why use pointer here?

shared_model/interfaces/common_objects/amount.hpp
l4l: Because it's deserialization is not that fast

shared_model/interfaces/common_objects/impl/amount.cpp
l4l: ehm, why do you think so?

test/module/irohad/execution/command_validate_execute_test.cpp
l4l: It's the minimal possible fix there

shared_model/interfaces/common_objects/impl/amount.cpp
l4l: Need to be implemented as a `ModelPrimitive` deriviative

igor-egorov:
Bindings and BTF fails. Investigation in progress.

shared_model/interfaces/common_objects/amount.hpp
```diff
@@ -1,84 +1,72 @@
...
     /**
      * Representation of fixed point number
      */
-    class Amount : public ModelPrimitive<Amount> {
+    class Amount final : public ModelPrimitive<Amount> {
      public:
+      explicit Amount(const std::string &amount);
+      explicit Amount(std::string &&amount);
+
+      Amount(const Amount &o);
```
nickaleks: If you add move and copy constructors please add assignment operators as well

shared_model/interfaces/common_objects/impl/amount.cpp
```diff
@@ -0,0 +1,71 @@
...
+            static const std::regex r("([0-9]+)(\\.([0-9]+))?");
+            // 123.456 will have the following groups:
+            //   [0] -> 123.456, [1] -> 123
+            //   [2] -> .456,    [3] -> 456
+            std::smatch match;
+            if (std::regex_match(this->amount_, match, r)
+                && match.size() == 4) {
+              this->precision_ = match[3].length();
+              auto str = match[0].str();
+              decltype(str.size()) pos = match[1].length();
```
nickaleks: auto?

shared_model/interfaces/common_objects/impl/amount.cpp
```diff
@@ -0,0 +1,71 @@
...
+              if (pos < str.size()) {
+                str.erase(str.begin() + pos);
```
nickaleks: please add comment to indicate what this does

shared_model/interfaces/common_objects/impl/amount.cpp
```diff
@@ -0,0 +1,71 @@
...
+              }
+              return boost::multiprecision::uint256_t(str);
+            }
+            return std::numeric_limits<boost::multiprecision::uint256_t>::min();
```
nickaleks: We return 0 in case of failure. So we don't know if we have invalid or zero amount in case something goes wrong

nickaleks: I think we should validate amount string in constructor, rather than when we try to get its fields

shared_model/interfaces/common_objects/impl/amount.cpp
l4l: for some reason it deduce signed type which leads to warning :(

shared_model/interfaces/common_objects/impl/amount.cpp
l4l: Amount have stateless validation for being positive

nickaleks: Only for commands. We can have 0 balance on our account

shared_model/interfaces/common_objects/impl/amount.cpp
nickaleks: size_t maybe?

l4l: yeah, it can be

shared_model/interfaces/common_objects/impl/amount.cpp
l4l: WSV is assumed to be correct, so there's no checks there

shared_model/interfaces/common_objects/amount.hpp
l4l: With const member it's impossible

PR #1557, opened 2018-07-12
Feature/sql executors

victordrobny:
### Description of the Change
Sql command executor. 

### Benefits
Less db communication, better command execution performance.


test/integration/acceptance/transfer_asset_test.cpp
```diff
@@ -328,9 +328,9 @@ TEST_F(TransferAsset, BigPrecision) {
   const std::string kNewAssetId =
       kNewAsset + "#" + IntegrationTestFramework::kDefaultDomain;
   const auto kPrecision = 5;
-  const std::string kInitial = "500";
-  const std::string kForTransfer = "1";
-  const std::string kLeft = "499";
+  const std::string kInitial = "500.00000";
```
lebdron: Please use `db.precision >= command.precision` instead of `=` so that previous value would be valid.

irohad/ametsuchi/command_executor.hpp
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Error for command execution or validation
+     * Contains command name, as well as an error message
+     */
+    struct CommandError {
+      std::string command_name;
+      std::string error_message;
+
+      std::string toString() const {
+        return (boost::format("%s: %s") % command_name % error_message).str();
```
nickaleks: Can we remove boost::format here? It will allow us to remove include.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -0,0 +1,770 @@
...
+#include "ametsuchi/impl/postgres_command_executor.hpp"
+
+#include <boost/format.hpp>
+
+#include "ametsuchi/impl/postgres_wsv_common.hpp"
+
+namespace iroha {
+  namespace ametsuchi {
+
+    expected::Error<CommandError> makeCommandError(
```
nickaleks: Please put all local functions in an anonymous namespace. This will make them link internally and not pollute ametsuchi namespace.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -0,0 +1,770 @@
...
+    CommandResult makeCommandResultByReturnedValue(
+        soci::statement &st,
+        std::string command_name,
+        std::vector<std::function<std::string()>> &error_generator) noexcept {
+      uint32_t result;
+      st.exchange(soci::into(result));
+      st.define_and_bind();
+      try {
+        st.execute(true);
+        for (size_t i = 0; i < error_generator.size(); i++) {
```
nickaleks: Please remove this loop (use result + 1 as index)

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -0,0 +1,770 @@
...
+            &creator_account_id) {
+      creator_account_id_ = creator_account_id;
+    }
+
+    CommandResult PostgresCommandExecutor::operator()(
+        const shared_model::interface::AddAssetQuantity &command) {
+      auto &account_id = creator_account_id_;
+      auto &asset_id = command.assetId();
+      auto amount = command.amount().toStringRepr();
+      uint32_t precision = command.amount().precision();
```
nickaleks: auto?

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -0,0 +1,770 @@
...
+              ELSE 6
+          END AS result;)";
+      // clang-format on
+      st.exchange(soci::use(src_account_id, "src_account_id"));
+      st.exchange(soci::use(dest_account_id, "dest_account_id"));
+      st.exchange(soci::use(asset_id, "asset_id"));
+      st.exchange(soci::use(amount, "value"));
+      st.exchange(soci::use(precision, "precision"));
+      std::vector<std::function<std::string()>> message_gen = {
+          [&] { return "Destanation account does not exist"; },
```
nickaleks: Typo

irohad/ametsuchi/impl/postgres_command_executor.cpp
victordrobny: Facepalm

irohad/ametsuchi/impl/postgres_command_executor.cpp
victordrobny: No. Precision is uint8_t, but soci has a bug with uint8_t.

irohad/ametsuchi/command_executor.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+#ifndef IROHA_AMETSUCHI_COMMAND_EXECUTOR_HPP
+#define IROHA_AMETSUCHI_COMMAND_EXECUTOR_HPP
+
+#include <boost/format.hpp>
+
+#include "common/result.hpp"
+#include "interfaces/common_objects/types.hpp"
+
+namespace shared_model {
```
nickaleks: I think in the future it would be beneficial to move such declarations to separate header inside shared_model (not related to this pr, just a thought)

irohad/ametsuchi/command_executor.hpp
nickaleks: Just concatenate the strings?

nickaleks: Or just move the definition to cpp file and remove include here

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -0,0 +1,792 @@
...
+ * to be used as a return error.
+ * Function is passed instead of string to avoid overhead of string
+ * construction in successful case.
+ * @return CommandResult with combined error message
+ * in case of result contains error
+ */
+template <typename Function>
+iroha::ametsuchi::CommandResult makeCommandResult(
+    soci::statement &st,
+    std::string command_name,
```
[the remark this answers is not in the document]
victordrobny: const char * is passed here. It is not possible to convert char * to std::string &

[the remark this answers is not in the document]
victordrobny: To make reference here, it is necessary to construct std::string outside this method. But since char * is always passed, I decided to construct string here

irohad/ametsuchi/command_executor.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Error for command execution or validation
+     * Contains command name, as well as an error message
+     */
+    struct CommandError {
+      std::string command_name;
+      std::string error_message;
+
+      std::string toString() const;
```
nickaleks: Nice

PR #1558, opened 2018-07-12
Fix build of Binary Testing Framework

igor-egorov:

### Description of the Change

Build of `binary_test` target was broken. Now the target can be built.

### Benefits

BTF can be built.

### Usage Examples or Tests

```
cmake -Bbuild -H. -DSWIG_PYTHON=ON
cmake --build build --target binary_test
```


igor-egorov:
Won't wait for CI, since it does not build python bindings anyway.

PR #1561, opened 2018-07-13
OAdd thread-safe collection set

muratovv:
### Description of the Change
* Add thread-safe set which operates with collections.
* Add tests for it

### Benefits
One step closer to BFT-Iroha

### Possible Drawbacks 
Absent concurrent tests


libs/cache/collection_set.hpp
```diff
@@ -0,0 +1,72 @@
...
+        }
+      }
+
+      /**
+       * Blocking walk through the collection
+       * @tparam Callable - type of functor
+       * @param callable - functor for invocation on each element
+       */
+      template <typename Callable>
+      void foreach (Callable &&callable) const {
```
l4l: Redundant space after fn name

test/module/libs/cache/CMakeLists.txt
```diff
@@ -16,3 +16,5 @@ addtest(cache_test cache_test.cpp)
 target_link_libraries(cache_test
         torii_service
         )
+
+addtest(transaction_cache_test transaction_cache_test.cpp)
```
l4l: please wrap

libs/cache/collection_set.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+      /**
+       * Remove all elements which are occured in the passed collection
+       * @tparam Collection - type of passed collection
+       * @param collection - elements to remove
+       */
+      template <typename Collection>
+      void removeValues(Collection &&collection) {
+        std::lock_guard<std::shared_timed_mutex> lock(mutex_);
+        for (auto val : collection) {
```
l4l: `auto &&val`

test/module/libs/cache/transaction_cache_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @given empty set
+ * @when  check that empty set doesn't contain elements
+ * AND insert some collection
+ * @then  check that elements are appeared
+ */
+TEST(TransactionCacheTest, insert) {
+  auto number_of_calls = 0;
+  CollectionSet<int> set;
+  set.foreach ([&number_of_calls](const auto &val) { number_of_calls++; });
```
l4l: plz use formatting

test/module/libs/cache/transaction_cache_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * @given empty set
+ * @when insert some collection
+ * AND insert duplicated
+ * @then  check that duplicates are not appeared
+ */
+TEST(TransactionCacheTest, insertDuplicates) {
+  auto number_of_calls = 0;
```
l4l: here is the number of elements, not calls

test/module/libs/cache/transaction_cache_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+  set.insertValues(std::vector<int>({1, 2}));
+
+  set.foreach ([&number_of_calls](const auto &val) { number_of_calls++; });
+  ASSERT_EQ(2, number_of_calls);
+}
+
+/**
+ * @given empty set
+ * @when insert some collection
+ * AND insert duplicated
```
l4l: with repeated elements

test/module/libs/cache/transaction_cache_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * @given empty set
+ * @when insert some collection
+ * AND remove another collection with same and different elements
+ * @then  check that duplicates and removed elements are not appeared
+ */
+TEST(TransactionCacheTest, remove) {
+  auto number_of_calls = 0;
```
l4l: Same here

test/module/libs/cache/transaction_cache_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * @given empty set
+ * @when  check that empty set doesn't contain elements
+ * AND insert some collection
+ * @then  check that elements are appeared
+ */
+TEST(TransactionCacheTest, insert) {
+  auto number_of_calls = 0;
+  CollectionSet<int> set;
```
l4l: That probably can be moved to a fixture

test/module/libs/cache/transaction_cache_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+  set.foreach ([&number_of_calls](const auto &val) { number_of_calls++; });
+  ASSERT_EQ(0, number_of_calls);
+
+  set.insertValues(std::vector<int>({1, 2}));
```
l4l: Why vector? Won't it work for just `{1, 2}`?

l4l:
Would be nice to add a test with concurrency emulation (e.g waits on cv responses)

test/module/libs/cache/transaction_cache_test.cpp
muratovv: Number of calls, because we count lambda invocations on `foreach`. But the numbers are the same.

test/module/libs/cache/transaction_cache_test.cpp
muratovv: Because compiler don't know a type

libs/cache/collection_set.hpp
muratovv: Shtoshe, it is a bug of clang-format. I think `foreach` is a reserved word. I've checked with `for_each` name and it is working well.

muratovv: I'll rename the function.

PR #1562, opened 2018-07-13
Increase presision lifetime in insertAsset

l4l:
### Description of the Change

`soci::use` uses the lvalue variables. The precision in insertAsset was passed as temporary an thus used after its scope ending. The pr is extends the precision lifetime

### Benefits

Minus one bug

### Possible Drawbacks 

None


neewy:
@l4l presision is a nice meme

PR #1567, opened 2018-07-16
Noncopyable proposal

nickaleks:
### Description of the Change

Implement non-copyable proposal.

### Benefits

Avoid unnecessary copying of proposal. Separate code to header and cpp results in faster compilation.

### Possible Drawbacks 

None


test/benchmark/bm_proto_creation.cpp
```diff
@@ -11,8 +11,8 @@
  *
  * The purpose of this benchmark is to keep track of performance costs related
  * to blocks and proposals copying/moving.
- *
- * Each benchmark runs transaction() and commands() call to
+ *
```
muratovv: Looks like redundant space.

test/benchmark/bm_proto_creation.cpp
```diff
@@ -11,8 +11,8 @@
  *
  * The purpose of this benchmark is to keep track of performance costs related
  * to blocks and proposals copying/moving.
- *
- * Each benchmark runs transaction() and commands() call to
+ * 
+ * Each benchmark runs transaction() and commands() call to
```
muratovv: sab

shared_model/backend/protobuf/proposal.hpp
```diff
@@ -21,48 +9,38 @@
...
 
      private:
-      // lazy
       template <typename T>
       using Lazy = detail::LazyInitializer<T>;
 
       const Lazy<std::vector<proto::Transaction>> transactions_{[this] {
-        return std::vector<proto::Transaction>(proto_->transactions().begin(),
-                                               proto_->transactions().end());
+        return std::vector<proto::Transaction>(
```
Akvinikym: Isn't it possible to move this initialisation to constructor in cpp file as well?

nickaleks: Unfortunately no

PR #1568, opened 2018-07-16
Endpoint Receive Transaction List

Akvinikym:
### Description of the Change

Before the P.R. Torii was able to only handle transactions one-by-one, which is not very convenient. Now, such thing as Transaction List was introduced in Protobuf and Torii, which allows to process transactions in a sequence.

### Benefits

One now can send transactions through Protobuf in a list, not one-by-one.

### Possible Drawbacks 

None


test/module/shared_model/builders/protobuf/transport_builder_test.cpp
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <endpoint.pb.h>
```
kamilsa: Please use "" instead of <>

irohad/torii/impl/command_service.cpp
```diff
@@ -19,18 +19,21 @@
...
-#include "backend/protobuf/transaction_responses/proto_tx_response.hpp"
-
 #include "ametsuchi/block_query.hpp"
 #include "backend/protobuf/transaction.hpp"
-#include "builders/protobuf/transaction_responses/proto_transaction_status_builder.hpp"
+#include "backend/protobuf/transaction_responses/proto_tx_response.hpp"
+// breaking sequence of this two includes will break the build
+// clang-format off
 #include "builders/protobuf/transport_builder.hpp"
+#include "builders/protobuf/transaction_sequence_builder.hpp"
```
kamilsa: How about to add `#include "builders/protobuf/transport_builder.hpp"` to `transaction_sequence_builder.hpp`, so that no one ever experience that issue (I suspect developers might spend too much time to figure out why same problem arises)

irohad/torii/impl/command_service.cpp
```diff
@@ -19,18 +19,21 @@
...
+// clang-format on
+#include "builders/protobuf/transaction_responses/proto_transaction_status_builder.hpp"
 #include "common/byteutils.hpp"
 #include "common/is_any.hpp"
 #include "common/timeout.hpp"
 #include "common/types.hpp"
 #include "cryptography/default_hash_provider.hpp"
-#include "endpoint.pb.h"
+#include "interfaces/iroha_internal/transaction_sequence.hpp"
```
kamilsa: this is already included in transaction_sequence_builder

shared_model/builders/protobuf/transaction_sequence_builder.hpp
```diff
@@ -32,10 +32,11 @@ namespace shared_model {
       template <class T>
       iroha::expected::Result<interface::TransactionSequence, std::string>
       build(T &transport) {
+        auto txs = transport.transactions();
```
kamilsa: const auto& should work as well, but avoids redundant copy

shared_model/validators/default_validator.hpp
```diff
@@ -70,6 +71,11 @@ namespace shared_model {
         SignableModelValidator<DefaultBlockValidator,
                                const interface::Block &,
                                FieldValidator>;
+
+    using DefaultBatchValidator =
```
kamilsa: It does not validate batch. It actually can be used to validate any collection of transactions which may contain several batches. I would name it smth like `DefaultUnsignedTxCollectionValidator`. I don't know how to make the name shorter

test/module/shared_model/builders/protobuf/transport_builder_test.cpp
```diff
@@ -203,10 +204,7 @@ class TransportBuilderTest : public ::testing::Test {
    * @param successCase function invoking if value exists
    * @param failCase function invoking when error returned
    */
-  template <typename SV,
-            typename T,
-            typename SuccessCase,
-            typename FailCase>
+  template <typename SV, typename T, typename SuccessCase, typename FailCase>
```
igor-egorov: Since you have touched that line, could you introduce meaningful names for template parameters SV and T?

Akvinikym: OK, I'll try)

PR #1572, opened 2018-07-17
Refactor QueryProcessor interface

l4l:
### Description of the Change

Since simple queries are performed in synchronous manner there's a reason in queryNotifier removing and returning the response at queryHandle call. Also some functions was refactored, so they accept reference, instead of shared_ptr.

### Benefits

Code is simpler

### Possible Drawbacks 

None?

### DIFF
--- a/irohad/torii/processor/query_processor.hpp
+++ b/irohad/torii/processor/query_processor.hpp
@@ -34,8 +34,9 @@ namespace iroha {
     class QueryProcessor {
      public:
       /**
-       * Register client query
-       * @param query - client intent
+       * Perform client query
+       * @param qry - client intent
+       * @return resulted response
        */
       virtual std::unique_ptr<shared_model::interface::QueryResponse>
       queryHandle(const shared_model::interface::Query &qry) = 0;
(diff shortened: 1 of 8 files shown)


irohad/torii/processor/query_processor.hpp
```diff
@@ -37,24 +37,16 @@ namespace iroha {
        * Register client query
        * @param query - client intent
        */
-      virtual void queryHandle(
-          std::shared_ptr<shared_model::interface::Query> qry) = 0;
+      virtual std::unique_ptr<shared_model::interface::QueryResponse>
+      queryHandle(const shared_model::interface::Query &qry) = 0;
       /**
        * Register client blocks query
        * @param query - client intent
```
Akvinikym: Not consistent with function argument's name. This problem arises in many places, please, rename variables in comments at least in functions you touched, such as this or the one upward, for example

irohad/torii/processor/query_processor_impl.hpp
```diff
@@ -44,8 +44,8 @@ namespace iroha {
        * Register client query
        * @param query - client intent
        */
-      void queryHandle(
-          std::shared_ptr<shared_model::interface::Query> qry) override;
+      std::unique_ptr<shared_model::interface::QueryResponse> queryHandle(
+          const shared_model::interface::Query &qry) override;
 
       /**
```
Akvinikym: Maybe remove comments in this file? They do not add anything new to their ancestors in interface

PR #1583, opened 2018-07-20
Implement test batch builder

muratovv:
### Description of the Change
Add test batch creator which is useful for tests in MST

### Benefits
Add composable functions and tests for it

### How to check
```
cmake -H. -Bbuild && cmake --build build --target proto_batch_test && build/test_bin/proto_batch_test
```

### DIFF
--- a/test/framework/batch_helper.hpp
+++ b/test/framework/batch_helper.hpp
@@ -208,7 +208,6 @@ namespace framework {
        * @return vector with transactions hashes
        */
       template <typename TxBuilderCollection>
-      // forward decl
       auto fetchReducedHashes(
           const std::initializer_list<TxBuilderCollection> &builders) {
         std::vector<shared_model::interface::types::HashType> hashes;
--- a/test/framework/batch_helper.hpp
+++ b/test/framework/batch_helper.hpp
@@ -229,15 +229,15 @@ namespace framework {
 
       /**
        * Function create transactions from builder and passed hashes
-       * @tparam TxBuildersCollection - type of one builder
+       * @tparam TxBuilder - type of one builder
        * @param hashes - vector of hashes of transactions
        * @param builders - initializer list of tx builders
        * @return transactions with correct batch meta
        */
-      template <typename TxBuildersCollection>
+      template <typename TxBuilder>
       auto makeTxBatchCollection(
           const std::vector<shared_model::interface::types::HashType> &hashes,
-          std::initializer_list<TxBuildersCollection> builders) {
+          std::initializer_list<TxBuilder> builders) {
         shared_model::interface::types::SharedTxsCollectionType transactions;
 
         std::transform(
--- a/test/framework/batch_helper.hpp
+++ b/test/framework/batch_helper.hpp
@@ -195,6 +195,98 @@ namespace framework {
       return framework::expected::val(result_batch).value().value;
     }
 
+    /**
+     * Namespace provides useful functions which are related to implementation
+     * but they are internal API
+     */
+    namespace internal {
+
+      /**
+       * Create list of hashes
+       * @tparam TxBuilderCollection - type of builder
+       * @param builders - initializer list which contains all builders
+       * @return vector with transactions hashes
+       */
+      template <typename TxBuilderCollection>
+      // forward decl
+      auto fetchReducedHashes(
+          const std::initializer_list<TxBuilderCollection> &builders) {
+        std::vector<shared_model::interface::types::HashType> hashes;
+        std::for_each(
+            builders.begin(), builders.end(), [&hashes](const auto &builder) {
+              hashes.push_back(builder.build().reducedHash());
+            });
+        return hashes;
+      }
+
+      /**
+       * Variadic to initializer list adapter
+       */
+      template <typename... TxBuilders>
+      auto fetchReducedHashes(const TxBuilders &... builders) {
+        return fetchReducedHashes({builders...});
+      }
+
+      /**
+       * Function create transactions from builder and passed hashes
+       * @tparam TxBuildersCollection - type of one builder
+       * @param hashes - vector of hashes of transactions
+       * @param builders - initializer list of tx builders
+       * @return transactions with correct batch meta
+       */
+      template <typename TxBuildersCollection>
+      auto makeTxBatchCollection(
+          const std::vector<shared_model::interface::types::HashType> &hashes,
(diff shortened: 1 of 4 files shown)


test/framework/batch_helper.hpp
```diff
@@ -195,6 +195,97 @@ namespace framework {
...
+              hashes.push_back(builder.build().reducedHash());
+            });
+        return hashes;
+      }
+
+      /**
+       * Variadic to initializer list adapter
+       */
+      template <typename... TxBuilders>
+      auto fetchReducedHashes(const TxBuilders &... builders) {
```
igor-egorov: Seems that `TxBuilders` and `TxBuilderCollection` represent the same type. For the purposes of clarity and consistency, please use the same name in both places.

test/framework/batch_helper.hpp
```diff
@@ -195,6 +195,97 @@ namespace framework {
       return framework::expected::val(result_batch).value().value;
     }
 
+    /**
+     * Namespace provides useful functions which are related to implementation
+     * but they are internal API
+     */
+    namespace internal {
```
kamilsa: We can split this file to .cpp and .hpp and then put all from internal into unnamed namespace in .cpp file or make all functions from `internal` static to .cpp file. Thus you can hide all methods inside

test/framework/batch_helper.hpp
```diff
@@ -195,6 +195,97 @@ namespace framework {
...
+       * Create list of hashes
+       * @tparam TxBuilderCollection - type of builder
+       * @param builders - initializer list which contains all builders
+       * @return vector with transactions hashes
+       */
+      template <typename TxBuilderCollection>
+      auto fetchReducedHashes(
+          const std::initializer_list<TxBuilderCollection> &builders) {
+        std::vector<shared_model::interface::types::HashType> hashes;
+        std::for_each(
```
kamilsa:
```
std::transfrom(builder.begin(), builders.end(), std::back_inserter(hashes), [](...){/*your lambda*/});
```

test/framework/batch_helper.hpp
```diff
@@ -195,6 +195,97 @@ namespace framework {
...
+       * @param builders - initializer list of tx builders
+       * @return transactions with correct batch meta
+       */
+      template <typename TxBuildersCollection>
+      auto makeTxBatchCollection(
+          const std::vector<shared_model::interface::types::HashType> &hashes,
+          std::initializer_list<TxBuildersCollection> builders) {
+        shared_model::interface::types::SharedTxsCollectionType transactions;
+
+        std::for_each(
```
kamilsa: sab

test/framework/batch_helper.hpp
```diff
@@ -195,6 +195,97 @@ namespace framework {
...
+              transactions.push_back(makePolyTxFromBuilder(builder.batchMeta(
+                  shared_model::interface::types::BatchType::ATOMIC, hashes)));
+            });
+        return transactions;
+      }
+
+      /**
+       * Variadic to initializer list adapter
+       */
+      template <typename... TxBuilders>
```
igor-egorov: If the previous comment is applicable here, change the type name here too.

test/module/shared_model/backend_proto/proto_batch_test.cpp
```diff
@@ -191,3 +192,76 @@ TEST(TransactionBatchTest, BatchWithMissingSignatures) {
...
 }
+
+/**
+ * Create test transaction builder
+ * @param acc_quorum - quorum number for setAccountDetail
+ * @param created_time - time of creation
+ * @param quorum - tx quorum number
+ * @return test tx builder
+ */
+inline auto makeTx(
```
kamilsa: Maybe rename that into `makeTxBuilder` if it returns builder, but not transaction

test/framework/batch_helper.hpp
```diff
@@ -195,6 +195,97 @@ namespace framework {
...
+                                     {std::forward<TxBuilders>(builders)...});
+      }
+    }  // namespace internal
+
+    /**
+     * Create test batch from passed transaction builders
+     * @tparam TxBuilders - variadic types of tx builders
+     * @return shared_ptr for batch
+     */
+    template <typename... TxBuilders>
```
igor-egorov: The same note about the naming. Please change if applicable.

test/module/shared_model/backend_proto/proto_batch_test.cpp
```diff
@@ -4,6 +4,7 @@
  */
 
 #include <gmock/gmock.h>
+#include <builders/protobuf/transaction.hpp>
```
igor-egorov: Use quotes instead of angle brackets for project-local includes.

test/module/shared_model/builders/protobuf/test_transaction_builder.hpp
```diff
@@ -18,22 +18,41 @@
...
+ */
 using TestUnsignedTransactionBuilder =
     shared_model::proto::TemplateTransactionBuilder<
         (1 << shared_model::proto::TemplateTransactionBuilder<>::total) - 1,
         shared_model::validation::AlwaysValidValidator,
-        shared_model::proto::UnsignedWrapper<shared_model::proto::Transaction>>;
+        shared_model::proto::UnsignedWrapper<ProtoTxType>>;
+
+/**
+ * Wrapper for making shared_ptr from the passed builder
```
kamilsa: should be written: .. for making shared_ptr on transaction from the passed builder

test/module/shared_model/backend_proto/proto_batch_test.cpp
kamilsa: Use "" instead <>

test/framework/batch_helper.hpp
muratovv: My point is not to hide those methods but just mark them. Those methods are useful in different cases of creation batches.

test/framework/batch_helper.hpp
```diff
@@ -256,7 +256,8 @@ namespace framework {
        */
       template <typename... TxBuilders>
       auto makeTxBatchCollection(TxBuilders &&... builders) {
-        return makeTxBatchCollection(fetchReducedHashes(builders...),
+        auto hashes = fetchReducedHashes(builders...);
```
kamilsa: auto&&. Otherwise copy is made

PR #1585, opened 2018-07-23
Feature/sql validation in sql executors

victordrobny:
### Description of the Change
isValidate method of CommandValidator is moved to PostgresCommandExecutor

### Benefits
Less db communication


irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -244,18 +248,84 @@ namespace iroha {
...
+              +
+              R"( ELSE 4
+            END AS result
+)";
+      st.exchange(soci::use(account_id, "account_id"));
+      st.exchange(soci::use(role_name, "role_id"));
+      st.exchange(soci::use(creator_account_id_, "creator_id"));
+      std::vector<std::function<std::string()>> message_gen = {
+          [&] {
+            return "is valid command validation failed: no permissions in role "
```
nickaleks: do we need "is valid" prefix for each message?

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -618,17 +763,63 @@ namespace iroha {
...
-      auto message_gen = [&] {
-        return (boost::format(
-                    "failed to update account, account id: '%s', quorum: '%s'")
-                % account_id % quorum)
-            .str();
+      soci::statement st = sql_.prepare << std::string(R"(WITH
+            )")
+              + (is_genesis_ ? "" :
+                             R"(
+          get_signatories AS (
```
nickaleks: I guess we can use views for that in the future

irohad/ametsuchi/impl/mutable_storage_impl.cpp
```diff
@@ -52,6 +52,7 @@ namespace iroha {
             function) {
       auto execute_transaction = [this](auto &transaction) {
         command_executor_->setCreatorAccountId(transaction.creatorAccountId());
+        command_executor_->setIsGenesis(true);
```
nickaleks: Does it mean that all transactions are executed as genesis block?

victordrobny: Yes, since mutable storage doesn't have CommandValidator, this flag is set here every time.

irohad/ametsuchi/impl/postgres_command_executor.cpp
victordrobny: I think we don't

irohad/ametsuchi/impl/mutable_storage_impl.cpp
nickaleks: Please rename to something like doValidation()

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -244,18 +248,84 @@ namespace iroha {
...
+          [&] {
+            return "command validation failed: no permissions in role "
+                + command.roleName();
+          },
+          [&] {
+            return "command validation failed: no roles in account "
+                + creator_account_id_;
+          },
+          [&] {
+            return "command validation failed: account "
```
nickaleks: use boost::format for consistency please

irohad/ametsuchi/impl/postgres_wsv_query.cpp
```diff
@@ -122,16 +122,21 @@ namespace iroha {
       shared_model::interface::RolePermissionSet set;
       soci::indicator ind;
       std::string row;
+      bool has_role = false;
```
nickaleks: I guess you can remove it here

PR #1592, opened 2018-07-24
On demand ordering service fixes

lebdron:
### Description of the Change
Fixes for on demand ordering service:
- remove unused argument in `onCollaborationOutcome`
- add only unique transaction to a proposal
- add outcome type for committed proposal, which does not increase round number
- make test for OS more deterministic by manually incrementing the timestamp

### Benefits
Correct logic

### Possible Drawbacks 
None.


irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -4,9 +4,13 @@
  */
 
 #include "ordering/impl/on_demand_ordering_service_impl.hpp"
-#include <boost/range/algorithm/for_each.hpp>
 
-#include "builders/protobuf/proposal.hpp"
+#include <unordered_set>
+
+#include <boost/range/adaptor/transformed.hpp>
+#include <boost/range/algorithm/for_each.hpp>
```
nickaleks: Please separate boost includes from ours

nickaleks: Please separate boost includes from ours

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -111,12 +118,14 @@ OnDemandOrderingServiceImpl::emitProposal() {
   TransactionType current_tx;
   using ProtoTxType = shared_model::proto::Transaction;
   std::vector<TransactionType> collection;
+  std::unordered_set<std::string> inserted;
```
nickaleks: types::HashType instead of string?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -111,12 +118,14 @@ OnDemandOrderingServiceImpl::emitProposal() {
...
 
   // outer method should guarantee availability of at least one transaction in
   // queue, also, code shouldn't fetch all transactions from queue. The rest
   // will be lost.
   while (current_proposal_.second.try_pop(current_tx)
-         and collection.size() < transaction_limit_) {
+         and collection.size() < transaction_limit_
+         and inserted.insert(current_tx->hash().hex()).second) {
```
nickaleks: Do we clear up inserted set somewhere? It can grow large after a while. Also what about peer restart?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
lebdron: Not sure how does it fit into google codestyle in this case : https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes , because boost headers are not C or C++ system files.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
lebdron: It is local to the method, so it is cleared after each proposal.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
nickaleks: Yeas you are right

irohad/ordering/on_demand_ordering_service.hpp
```diff
@@ -28,11 +26,11 @@ namespace iroha {
        * @param outcome - status of collaboration on proposal
        * @param round - number of round
```
muratovv: Param `round` is redundant.

PR #1593, opened 2018-07-24
Add hash to proposal

lebdron:
### Description of the Change
Since it is required to vote for proposals, and YAC operates with hashes, a corresponding method is added to proposal interface.

### Benefits
Can vote for proposals.

### Possible Drawbacks 
None.

### DIFF
--- a/shared_model/interfaces/iroha_internal/proposal.hpp
+++ b/shared_model/interfaces/iroha_internal/proposal.hpp
@@ -21,6 +21,7 @@
 #include "interfaces/base/model_primitive.hpp"
 #include "interfaces/common_objects/types.hpp"
 #include "interfaces/transaction.hpp"
+#include "utils/lazy_initializer.hpp"
 
 namespace shared_model {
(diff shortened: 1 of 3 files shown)


shared_model/interfaces/iroha_internal/proposal.hpp
```diff
@@ -47,16 +48,28 @@ namespace shared_model {
             and createdTime() == rhs.createdTime();
       }
 
+      virtual const types::BlobType &blob() const = 0;
+
+      const types::HashType &hash() const {
```
l4l: Looks like `class Signable` need to be splitted to Signable & Hashable. What do you think about it?

lebdron: I guess not, because they were merged in the past https://github.com/hyperledger/iroha/pull/1131

l4l: :( and now they're needed

PR #1595, opened 2018-07-25
Add batchMeta to transaction builder

x3medima17:
Signed-off-by: Dumitru <dimasavva17@gmail.com>

### Description of the Change
Adds `batchMeta` setter for transaction builder.

### Benefits
Now clients can set batch information.

### Possible Drawbacks 
None

### DIFF
--- a/shared_model/bindings/model_transaction_builder.hpp
+++ b/shared_model/bindings/model_transaction_builder.hpp
@@ -63,8 +63,8 @@ namespace shared_model {
 
       /**
        * Sets batch meta
-       * @param type one of ATOMIC or ORDERED
-       * @param hashes vector of hashes of transactions in this batch
+       * @param type - one of ATOMIC or ORDERED
+       * @param hashes - vector of hashes of transactions in this batch
        * @return builder with batchMeta set
        */
       ModelTransactionBuilder batchMeta(
--- a/shared_model/bindings/model_transaction_builder.hpp
+++ b/shared_model/bindings/model_transaction_builder.hpp
@@ -68,7 +68,7 @@ namespace shared_model {
        * @return builder with batchMeta set
        */
       ModelTransactionBuilder batchMeta(
-          const interface::types::BatchType &type,
+          interface::types::BatchType type,
           const std::vector<interface::types::HashType> &hashes);
 
       /**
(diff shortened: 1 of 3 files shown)


shared_model/builders/protobuf/builder_templates/transaction_template.hpp
```diff
@@ -125,14 +126,15 @@ namespace shared_model {
         });
       }
 
-      auto createdTime(interface::types::TimestampType created_time) const {
+      auto createdTime(
+          const interface::types::TimestampType &created_time) const {
```
l4l: That fix barely related to the title :\ So does the following

x3medima17: Clang format ¯\_(ツ)_/¯

x3medima17: My inner perfectionist could not leave it like that

shared_model/bindings/model_transaction_builder.hpp
```diff
@@ -61,6 +61,16 @@ namespace shared_model {
        */
       ModelTransactionBuilder quorum(interface::types::QuorumType quorum);
 
+      /**
+       * Sets batch meta
+       * @param type one of ATOMIC or ORDERED
+       * @param hashes vector of hashes of transactions in this batch
```
Akvinikym: You should either use dash "hashes - vector of hashes..." or reformulate the sentence, so that argument's name would be a correct part of it. The same for ```type``` argument slightly upwards

shared_model/builders/protobuf/builder_templates/transaction_template.hpp
```diff
@@ -125,14 +126,15 @@ namespace shared_model {
...
         return transform<CreatedTime>([&](auto &tx) {
           tx.mutable_payload()->mutable_reduced_payload()->set_created_time(
               created_time);
         });
       }
 
-      auto quorum(interface::types::QuorumType quorum) const {
+      auto quorum(const interface::types::QuorumType &quorum) const {
```
Akvinikym: ```QuorumType``` is an ```unsigned int```, and it should no be passed by reference, if you are not going to change it. The same thing happens also in 278 line

PR #1604, opened 2018-07-26
Single Pointer Cache

Akvinikym:
### Description of the Change

This PR implements a thread-safe data structure for storing single pointer of some generic data type.

### Benefits

In future PRs the structure will be heavily used.

### Possible Drawbacks 

None.


irohad/ametsuchi/impl/consensus_cache_block.hpp
```diff
@@ -0,0 +1,38 @@
...
+#define IROHA_CONSENSUS_CACHE_BLOCK_H
+
+#include "ametsuchi/pointer_cache.hpp"
+
+#include "interfaces/iroha_internal/block_variant.hpp"
+
+/**
+ * Thread-safe implementation of consensus cache for storing block variants
+ */
+class ConsensusCacheBlock
```
nickaleks: Please put inside some namespace

nickaleks: Is ConsensusCacheBlock a cache or a block? It is not very clear from the name

irohad/ametsuchi/impl/consensus_cache_block.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+ConsensusCacheBlock::ConsensusCacheBlock() = default;
+
+ConsensusCacheBlock::ConsensusCacheBlock(PointerCache::DataPointer block)
+    : stored_block_(std::move(block)) {}
+
+void ConsensusCacheBlock::insert(PointerCache::DataPointer data) {
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  stored_block_ = data;
```
nickaleks: Use move here?

irohad/ametsuchi/impl/consensus_cache_block.cpp
```diff
@@ -0,0 +1,32 @@
...
+}
+
+PointerCache<shared_model::interface::BlockVariant>::DataPointer
+ConsensusCacheBlock::get() const {
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  return stored_block_ ? stored_block_ : nullptr;
```
nickaleks: I believe just returning stored_block_ will have the same effect

irohad/ametsuchi/pointer_cache.hpp
```diff
@@ -0,0 +1,40 @@
...
+#define IROHA_CONSENSUS_CACHE_H
+
+#include <memory>
+
+/**
+ * Stores pointer to some data
+ */
+template <typename DataType,
+          template <typename> class PointerType = std::shared_ptr>
+class PointerCache {
```
nickaleks: Is there a need for a base class? Why would we want polymorphism here? Does base class require all of its descendants to be thread safe? If yes maybe use NVI

irohad/ametsuchi/pointer_cache.hpp
```diff
@@ -0,0 +1,40 @@
...
+#ifndef IROHA_CONSENSUS_CACHE_H
+#define IROHA_CONSENSUS_CACHE_H
+
+#include <memory>
+
+/**
+ * Stores pointer to some data
+ */
+template <typename DataType,
+          template <typename> class PointerType = std::shared_ptr>
```
nickaleks: Do we really want different pointers here?

libs/cache/single_pointer_cache.hpp
```diff
@@ -0,0 +1,72 @@
...
+    template <typename DataType>
+    void SinglePointerCache<DataType>::insert(
+        SinglePointerCache::DataPointer data) {
+      std::lock_guard<std::mutex> lock(mutex_);
+
+      stored_data_ = std::move(data);
+    }
+
+    template <typename DataType>
+    typename SinglePointerCache<DataType>::DataPointer
```
nickaleks: you can use out of class using to avoid typename

```C++
template <typename Cache>
using CacheData = Cache::DataPointer;
```

libs/cache/single_pointer_cache.hpp
```diff
@@ -0,0 +1,72 @@
...
+namespace iroha {
+  namespace cache {
+
+    /**
+     * Thread-safely stores and returns shared pointer to an element of template
+     * type
+     */
+    template <typename DataType>
+    class SinglePointerCache {
+      using DataPointer = std::shared_ptr<DataType>;
```
nickaleks: please add private specifier explicitly.

Also why using is private? Users might need it.

test/module/libs/cache/single_pointer_cache_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+ protected:
+  void SetUp() override {
+    consensus_cache_block = std::make_shared<ConsensusBlockCacheType>();
+  }
+
+  void TearDown() override {
+    consensus_cache_block.reset();
+  }
+
+  std::shared_ptr<ConsensusBlockCacheType> consensus_cache_block;
```
nickaleks: consensus_block_cache?

nickaleks: Also why is it pointer?

test/module/libs/cache/single_pointer_cache_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+};
+
+/**
+ * @given up and running consensus cache for blocks
+ * @when checking it for emptiness, inserting and getting the elements inside,
+ * releasing cache and trying to get the element again
+ * @then the cache works properly, returning empty optional when cache is empty
+ * and elements, when it's not
+ */
+TEST_F(ConsensusCacheBlockTest, SingleThreadedCache) {
```
nickaleks: Please write separate test case for each function: get, insert, release

libs/cache/single_pointer_cache.hpp
Akvinikym: Should I? It's used only once, so I think it can be left as-is

test/module/libs/cache/single_pointer_cache_test.cpp
Akvinikym: It's a pointer, so it could easily be reseted in ```TearDown()```

test/module/libs/cache/single_pointer_cache_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+  // we are going to wait until reader wakes up before releasing the cache,
+  // otherwise there's no sense in the test
+  std::thread reader{[this, &reader_has_started, &cv]() {
+    reader_has_started = true;
+    cv.notify_all();
+    ASSERT_EQ(*int_cache.get(), default_int_value);
+  }};
+  std::unique_lock<std::mutex> lock(mutex);
+  cv.wait(lock, [&reader_has_started]() { return reader_has_started; });
```
kamilsa: () can be removed in lambda

test/module/libs/cache/single_pointer_cache_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/**
+ * @given empty int cache
+ * @when first thread inserts value @and another tries to read it @and the first
+ * removes it @and second read operation completes
+ * @then the system could possibly crash, if it was not thread-safe - it would
+ * try to give removed value to the second thread
+ */
+TEST_F(SinglePointerCacheTest, MultithreadedCache) {
```
kamilsa: This test can potentially cause segfault error when ` ASSERT_EQ(*int_cache.get(), default_int_value);` (line 79) is invoked after `int_cache.release();` line 84

PR #1608, opened 2018-07-26
Fix spdlog segfault

igor-egorov:
### Description of the Change

set_pattern could be called simultaneously from several threads affecting global context.
The similar issue - https://github.com/gabime/spdlog/issues/419

### Benefits

No segfault caused by spdlog configuring. 

### Possible Drawbacks 

None

### Usage Examples or Tests 

```diff
cmake -H. -Bbuild
cd build
make create_role_test
while test_bin/create_role_test ; do :; done
+the loop should not be interrupted by spdlog set_pattern segfault anymore
```

thanks to @lebdron

### DIFF
--- a/libs/logger/logger.cpp
+++ b/libs/logger/logger.cpp
@@ -37,22 +37,23 @@ namespace logger {
     return red("<--- " + string);
   }
 
-  static void setGlobalPattern() {
-    spdlog::set_pattern("[%Y-%m-%d %H:%M:%S.%F] %n %v");
+  static void setGlobalPattern(spdlog::logger &logger) {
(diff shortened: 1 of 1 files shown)


libs/logger/logger.cpp
```diff
@@ -37,25 +37,28 @@ namespace logger {
...
     } else {
-      setGlobalPattern();
+      setGlobalPattern(*logger);
     }
-    return spdlog::stdout_color_mt(tag);
+    return logger;
   }
 
   Logger log(const std::string &tag) {
+    static std::mutex mutex;
```
muratovv: Nice trick with static!

PR #1615, opened 2018-07-29
Fix ToriiService Streaming Full Pipeline Test

igor-egorov:

### Description of the Change

The test (StreamingFullPipelineTest) was failing on slow machines. 
Investigation showed that the root cause was the way thread scheduler works, so the code was generally valid.

Streaming client resubscription is a fully valid use case of streaming service.
The client can resubscribe to the stream if the previous stream was ended unexpectedly for the client.
Unexpectedly means that client requests the hash of a transaction that was definitely sent to Iroha and the stream reaches its end with one of the following statuses: `NOT_RECEIVED`, `STATELESS_VALIDATION_SUCCESS`, `STATEFUL_VALIDATION_SUCCESS`.

Before the fix test client was not able to resubscribe to the stream. Now it can.

### Benefits

Legal test client behavior is implemented.
The test is not failing.

### Possible Drawbacks 

None

### Usage Examples or Tests 

```diff
cmake -H. -Bbuild
cd build
make torii_service_test
while test_bin/torii_service_test --gtest_filter=*ull* ; do :; done
+this loop should not be broken up by ASSERT_EQ(last_status, COMMITTED) error message
```

### DIFF
--- a/test/module/irohad/torii/torii_service_test.cpp
+++ b/test/module/irohad/torii/torii_service_test.cpp
@@ -34,6 +34,7 @@ using namespace iroha::torii;
 using namespace std::chrono_literals;
 constexpr std::chrono::milliseconds initial_timeout = 1s;
 constexpr std::chrono::milliseconds nonfinal_timeout = 2 * 10s;
+constexpr unsigned resubscibe_attempts = 3;
 
 using iroha::Commit;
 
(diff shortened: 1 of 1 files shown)


test/module/irohad/torii/torii_service_test.cpp
```diff
@@ -34,6 +34,7 @@ using namespace iroha::torii;
 using namespace std::chrono_literals;
 constexpr std::chrono::milliseconds initial_timeout = 1s;
 constexpr std::chrono::milliseconds nonfinal_timeout = 2 * 10s;
+constexpr unsigned resubscibe_attempts = 3;
```
luckychess: Typo plz.

PR #1618, opened 2018-07-31
Fix/batch any range

muratovv:
### Description of the Change

Fix segfault behavior on any_range with a temporary collection.

List of changes:
* rework any_range with the template parameter
* add toString method
* fix linkage errors because of missed stateless_validation target

### Benefits
After this PR, I will able to continue with MST-tests.

### Possible Drawbacks 
This PR may be separated for 3.
Missed test for any_range reworking. Unfortunately, I don't know how to create a minimal test for it. But this regression will be tested further in MST tests.

### DIFF
--- a/shared_model/interfaces/iroha_internal/transaction_sequence.cpp
+++ b/shared_model/interfaces/iroha_internal/transaction_sequence.cpp
@@ -10,7 +10,6 @@
 #include "validators/transaction_validator.hpp"
 #include "validators/transactions_collection/batch_order_validator.hpp"
 
-#include "cryptography/model_impl/hash.cpp"
 
 namespace shared_model {
   namespace interface {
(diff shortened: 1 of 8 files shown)


shared_model/interfaces/iroha_internal/transaction_sequence.cpp
```diff
@@ -10,6 +10,8 @@
 #include "validators/transaction_validator.hpp"
 #include "validators/transactions_collection/batch_order_validator.hpp"
 
+#include "cryptography/model_impl/hash.cpp"
```
igor-egorov: 🤔

muratovv: 🤔

muratovv:
CI doesn't pass because of client_test. This test will be fixed in https://github.com/hyperledger/iroha/pull/1619.

PR #1619, opened 2018-07-31
Fix non-determinism in client test

igor-egorov:

### Description of the Change

The client test was failing on slow machines (CI).
The nature of failure was described and fixed in https://github.com/hyperledger/iroha/pull/1615 (please take a look for the details).

### Benefits

Non-failing test.

### Possible Drawbacks 

None

### Usage Examples or Tests 

```diff
cmake -H. -Bbuild
cd build
make torii_service_test
while test_bin/client_test ; do :; done
+this loop should not be broken up with the error from linux build https://jenkins.soramitsu.co.jp/blue/organizations/jenkins/iroha%2Firoha-hyperledger/detail/PR-1597/39/pipeline/42
```

### DIFF
--- a/test/module/iroha-cli/client_test.cpp
+++ b/test/module/iroha-cli/client_test.cpp
@@ -45,6 +45,16 @@ using namespace shared_model::proto;
 using namespace std::chrono_literals;
 constexpr std::chrono::milliseconds initial_timeout = 1s;
 constexpr std::chrono::milliseconds nonfinal_timeout = 2 * 10s;
+
+/**
+Here we imitate the behavior of StatusStram client but on a bit lower level. So
+the do-while cycle imitates client resubscription to the stream. Stream
+"expiration" is a valid designed case (see pr #1615 for the details).
+
+The number of attempts (3) is a magic constant here. The idea behind this number
+is the following: only one resubscription is usually enough to pass the test; if
+three resubscribes were not enough, then most likely there is another bug.
+ */
 constexpr uint32_t status_read_attempts = 3;
 
 class ClientServerTest : public testing::Test {
--- a/test/module/iroha-cli/client_test.cpp
+++ b/test/module/iroha-cli/client_test.cpp
@@ -45,6 +45,7 @@ using namespace shared_model::proto;
 using namespace std::chrono_literals;
 constexpr std::chrono::milliseconds initial_timeout = 1s;
 constexpr std::chrono::milliseconds nonfinal_timeout = 2 * 10s;
+constexpr unsigned status_read_attempts = 3;
 
 class ClientServerTest : public testing::Test {
  public:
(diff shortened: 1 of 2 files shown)


test/module/iroha-cli/client_test.cpp
```diff
@@ -45,6 +45,7 @@ using namespace shared_model::proto;
 using namespace std::chrono_literals;
 constexpr std::chrono::milliseconds initial_timeout = 1s;
 constexpr std::chrono::milliseconds nonfinal_timeout = 2 * 10s;
+constexpr unsigned status_read_attempts = 3;
```
muratovv: Why are 3 attempts enough? Seems like we improve the probability of test passing instead of doing it deterministic.

test/module/iroha-cli/client_test.cpp
```diff
@@ -183,12 +184,21 @@ TEST_F(ClientServerTest, SendTxWhenStatelessInvalid) {
...
-  auto res = iroha_cli::CliClient(ip, port).getTxStatus(
-      shared_model::crypto::toBinaryString(tx_hash));
-  ASSERT_EQ(res.answer.tx_status(),
+  auto getAnswer = [&]() {
+    return iroha_cli::CliClient(ip, port)
+        .getTxStatus(shared_model::crypto::toBinaryString(shm_tx.hash()))
+        .answer;
+  };
+  decltype(getAnswer()) answer;
+  unsigned read_attempt_counter(status_read_attempts);
```
muratovv: better to use `auto` here. But if you want to state concrete type use platform-independent option: uint32_t, for example.

test/module/iroha-cli/client_test.cpp
```diff
@@ -239,10 +249,18 @@ TEST_F(ClientServerTest, SendTxWhenStatefulInvalid) {
...
-  // check it really failed with specific message
-  auto answer =
-      client.getTxStatus(shared_model::crypto::toBinaryString(tx.hash()))
-          .answer;
+  auto getAnswer = [&]() {
+    return client.getTxStatus(shared_model::crypto::toBinaryString(tx.hash()))
+        .answer;
+  };
+  decltype(getAnswer()) answer;
+  unsigned read_attempt_counter(status_read_attempts);
```
muratovv: same about type.

test/module/iroha-cli/client_test.cpp
lebdron: The deterministic way to do it is to access the status bus and do additional synchronization. But the idea from discussion was to replicate the conditions of the client which tries to use the API.

igor-egorov: Here we imitate the behavior of StatusStram client but on a bit lower level. So the do-while cycle imitates client resubscription to the stream. Stream "expiration" is a valid designed case (see #1615 for the details).

The number of attempts (3) is a magic constant here. The idea behind this number is the following: only one resubscription is usually enough to pass the test; if three resubscribes were not enough, then most likely threre is another bug.

igor-egorov: Discussed offline with @muratovv: decided to leave used approach.

PR #1621, opened 2018-07-31
Synchronizer with Block Variant

Akvinikym:
### Description of the Change

With this PR ability for Iroha to finally process empty blocks is added - now they are not committed to ledger, but still used to download chain, if something is missing.

### Benefits

Empty block pipeline ends.

### Possible Drawbacks 

None.

During the development process I faced a problem: `iroha::visit_in_place(..)` method does not allow lambdas with capture parameters without copy constructors, because in implementation they are passed by value. I think, it can should be fixed, and lambdas could be passed by references, because in future we may clash with this problem again.


test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -88,6 +91,25 @@ class SynchronizerTest : public ::testing::Test {
     return std::make_shared<shared_model::proto::Block>(std::move(block));
   }
 
+  std::shared_ptr<shared_model::interface::EmptyBlock> makeEmptyCommit(
+      size_t time = iroha::time::now()) const {
+    using TestUnsignedEmptyBlockBuilder =
```
kamilsa: Please move that to `test_empty_block_builder.hpp`

test/module/irohad/validation/chain_validation_test.cpp
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <framework/specified_visitor.hpp>
```
kamilsa: Fix brackets

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -46,85 +46,160 @@ namespace iroha {
...
+                ? blocks.back()->hash() == committed_block_variant.prevHash()
+                : blocks.back()->hash() == committed_block_variant.hash();
+
+            if (validator->validateChain(chain, *storage)
+                and chain_ends_with_right_block) {
+              // peer sent valid chain
+              notifier.get_subscriber().on_next(chain);
+
+              for (const auto &block : blocks) {
+                storage->apply(*block, trueStorageApplyPredicate);
```
kamilsa: Put a comment why we don't need check here

irohad/validation/impl/chain_validator_impl.cpp
```diff
@@ -20,35 +20,37 @@
...
-        : supermajority_checker_(supermajority_checker) {
-      log_ = logger::log("ChainValidator");
-    }
+        : supermajority_checker_(supermajority_checker),
+          log_(logger::log("ChainValidator")) {}
 
     bool ChainValidatorImpl::validateBlock(
-        const shared_model::interface::Block &block,
+        const shared_model::interface::BlockVariant &block_variant,
         ametsuchi::MutableStorage &storage) {
```
kamilsa: I think now this method can be const

irohad/validation/impl/chain_validator_impl.cpp
```diff
@@ -20,35 +20,37 @@
...
+                        block_var.signatures(), peers.value());
           };
 
-      // Apply to temporary storage
-      return storage.apply(block, apply_block);
+      // check inside of temporary storage
+      return storage.check(block_variant, check_block);
     }
 
     bool ChainValidatorImpl::validateChain(
```
kamilsa: same, this method can be const

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -46,85 +46,160 @@ namespace iroha {
...
+       * Committed block variant is not applied, because it's either empty or
+       * already exists in downloaded chain
+       * @param committed_block_variant to be processed
+       * @param notifier to notify the pcs about commit
+       * @param mutable_factory to commit results
+       * @param storage to validate and apply chain
+       * @param block_loader to download the blocks
+       * @param validator to validate chain
+       */
+      void processUnapplicableBlock(
```
nickaleks: Please split this function, it is hard to understand. I think the algorithm you outlined in the comments can be used to implement a separate function for each state

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -46,85 +47,117 @@ namespace iroha {
...
-            return;
-          }
-          auto network_chain = blockLoader_->retrieveBlocks(
-              shared_model::crypto::PublicKey(signature.publicKey()));
-          // Check chain last commit
+    void SynchronizerImpl::downloadAndApplyMissingChain(
+        const shared_model::interface::BlockVariant &committed_block_variant,
+        std::unique_ptr<ametsuchi::MutableStorage> storage) const {
+      while (true) {
+        for (const auto &signature : committed_block_variant.signatures()) {
```
nickaleks: peer_signature would be a better name

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -46,85 +47,117 @@ namespace iroha {
...
           std::vector<std::shared_ptr<shared_model::interface::Block>> blocks;
-          network_chain.as_blocking().subscribe(
-              [&blocks](auto block) { blocks.push_back(block); });
-          auto is_chain_end_expected =
-              blocks.back()->hash() == commit_message->hash();
+          block_loader_
+              ->retrieveBlocks(
```
nickaleks: Please consider not downloading whole chain

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -46,85 +47,117 @@ namespace iroha {
...
             return;
           }
         }
       }
     }
 
+    void SynchronizerImpl::process_commit(
+        const shared_model::interface::BlockVariant &committed_block_variant) {
+      log_->info("processing commit");
+      auto storage = createTemporaryStorage();
```
nickaleks: Please consider moving it to else clause

irohad/synchronizer/impl/synchronizer_impl.hpp
```diff
@@ -34,23 +34,53 @@ namespace iroha {
...
       rxcpp::subjects::subject<Commit> notifier_;
       rxcpp::composite_subscription subscription_;
 
       logger::Logger log_;
+
+      /**
+       * Creates a temporary storage out of the provided factory
+       * @return pointer to created storage
+       */
+      std::unique_ptr<ametsuchi::MutableStorage> createTemporaryStorage() const;
```
nickaleks: Is MutableStorage temporary?

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -37,25 +39,55 @@ class MstPipelineTest : public AcceptanceFixture {
   }
 
   /**
-   * Creates the transaction with the user creation commands
-   * @param perms are the permissions of the user
-   * @return built tx and a hash of its payload
+   * Creates a mst user
+   * @param itf, in which the user will be created
+   * @param sigs - signatories of that mst user
```
nickaleks: I believe it is more of a number of signatories

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -46,85 +46,160 @@ namespace iroha {
       subscription_.unsubscribe();
     }
 
+    namespace {
+      /**
+       * Lambda always returning true specially for applying blocks to storage
+       */
+      auto trueStorageApplyPredicate = [](const auto &, auto &, const auto &) {
```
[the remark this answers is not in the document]
Akvinikym: Unfortunately, it cannot, problem with references arises during run-time

irohad/synchronizer/impl/synchronizer_impl.cpp
Akvinikym: It's used during validation in `if`, so it must be created here

irohad/synchronizer/impl/synchronizer_impl.hpp
Akvinikym: Yes

irohad/synchronizer/impl/synchronizer_impl.cpp
Akvinikym: It will not. `retrieveBlocks(..)` automatically starts to download blocks from current top

irohad/synchronizer/impl/synchronizer_impl.hpp
```diff
@@ -34,23 +34,50 @@ namespace iroha {
           std::shared_ptr<ametsuchi::MutableFactory> mutableFactory,
           std::shared_ptr<network::BlockLoader> blockLoader);
 
-      ~SynchronizerImpl();
+      ~SynchronizerImpl() override;
 
-      void process_commit(
-          const shared_model::interface::BlockVariant &commit_message) override;
+      void process_commit(const shared_model::interface::BlockVariant
+                              &committed_block_variant) override;
```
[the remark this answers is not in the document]
Akvinikym: Yeap, `clang` does not see anything wrong

irohad/synchronizer/synchronizer.hpp
```diff
@@ -39,7 +39,10 @@ namespace iroha {
 
       /**
        * Emit committed blocks
-       * Note: from one block received on consensus
+       * Note 1: from one block received on consensus
+       * Note 2: commit may be empty; in that case, it can contain empty
```
nickaleks: Can empty commit contain non-empty pointer? If no then "can contain empty pointer" does not need "can" in it.

irohad/synchronizer/synchronizer.hpp
```diff
@@ -39,7 +39,10 @@ namespace iroha {
...
+       * pointer, if ledger state on this peer is up-to-date, so there is
```
nickaleks: I believe "if ledger..." can be a separate sentence. Also consider replacing "so" with "then".

It is not immediately clear that there is a causation between first and second part of the statement. Maybe write something like:

"if commit is empty, it means that the ledger state is up to date..."

PR #1623, opened 2018-08-01
Proposal gate

lebdron:
### Description of the Change
YAC proposal gate implementation:
- Use hash gate for consensus
- Hash provider with round number as round indentifier and optional<hash> with vote

### Unrelated changes note
Bind operator `operator|` was updated to allow non-copyable objects. "Legacy" code in old model and iroha-cli was not compliant with changes and had to be updated.

### DIFF
--- a/irohad/consensus/yac/impl/yac_proposal_gate_impl.cpp
+++ b/irohad/consensus/yac/impl/yac_proposal_gate_impl.cpp
@@ -1,0 +1,68 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "consensus/yac/impl/yac_proposal_gate_impl.hpp"
+
+#include "common/visitor.hpp"
+#include "consensus/yac/cluster_order.hpp"
+#include "consensus/yac/messages.hpp"
+#include "interfaces/iroha_internal/proposal.hpp"
+
+using namespace iroha;
+using namespace iroha::network;
+using namespace iroha::consensus::yac;
+
+YacProposalGateImpl::YacProposalGateImpl(
+    std::shared_ptr<HashGate> hash_gate,
+    std::shared_ptr<YacPeerOrderer> orderer,
+    std::shared_ptr<YacProposalHashProvider> hash_provider)
+    : hash_gate_(std::move(hash_gate)),
+      orderer_(std::move(orderer)),
+      hash_provider_(std::move(hash_provider)),
+      log_(logger::log("YacProposalGateImpl")) {}
+
+expected::Result<void, std::string> YacProposalGateImpl::vote(
+    ProposalVote vote) {
+  auto hash = hash_provider_->makeHash(vote);
+  log_->info("vote for proposal ({}, {}, {})",
+             vote.proposal ? vote.proposal.value()->hash().toString() : "''",
+             vote.round.first,
+             vote.round.second);
+  auto order = orderer_->getOrdering(hash);
+  if (not order) {
+    return expected::makeError("Orderer doesn't provide peers");
+  }
+  last_voted_proposal_.emplace(hash, std::move(vote));
+  hash_gate_->vote(hash, *order);
+  return {};
+}
+
+rxcpp::observable<ProposalOutcomeType> YacProposalGateImpl::outcomes() {
+  return hash_gate_->onOutcome().map([this](auto outcome) {
+    auto hash = visit_in_place(
+        outcome, [](auto outcome) { return outcome.votes.at(0).hash; });
+    auto proposal_info = hash_provider_->makeProposalInfo(hash);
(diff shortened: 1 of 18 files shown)


test/module/irohad/consensus/yac/yac_proposal_gate_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+ * @then commit emitted
+ */
+TEST_F(YacProposalGateTest, CommitAchieved) {
+  // make hash from proposal vote
+  EXPECT_CALL(*hash_provider, makeHash(_)).WillOnce(Return(yac_hash));
+
+  // generate order of peers
+  EXPECT_CALL(*peer_orderer, getOrdering(_)).WillOnce(Return(order));
+
+  // ya consensus
```
kamilsa: ya`c`

irohad/network/proposal_gate.hpp
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Allows to vote for proposals and receive consensus outcomes
+     */
+    class ProposalGate {
+     public:
+      /**
+       * Vote for proposal
+       * @param vote - structure with optional proposal and round number
+       */
+      virtual void vote(ProposalVote vote) = 0;
```
kamilsa: Can it be const ref?

test/module/irohad/consensus/yac/yac_proposal_hash_provider_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+  YacProposalHashProviderImpl hash_provider;
+  std::unique_ptr<MockProposal> proposal;
+  shared_model::interface::types::BlobType blob;
+  shared_model::interface::types::HashType hash;
+  iroha::ordering::transport::RoundType round{1, 1};
+};
+
+/**
+ * @given non-empty proposal optional and round number
+ * @when proposal, round -> yac hash -> proposal info
```
muratovv: `->` is not so understandable here, maybe rework it with a verb?

irohad/network/proposal_gate.hpp
lebdron: Probably yes, but semantic will be different. Because vote contains a unique pointer, value argument means that vote should be moved inside proposal gate. It is done because vote is stored inside proposal gate for commit comparison.

irohad/consensus/yac/impl/yac_proposal_gate_impl.cpp
```diff
@@ -0,0 +1,67 @@
...
+void YacProposalGateImpl::vote(ProposalVote vote) {
+  auto hash = hash_provider_->makeHash(vote);
+  log_->info("vote for proposal ({}, {}, {})",
+             vote.proposal ? vote.proposal.value()->hash().toString() : "''",
+             vote.round.first,
+             vote.round.second);
+  auto order = orderer_->getOrdering(hash);
+  if (not order) {
+    log_->error("ordering doesn't provide peers => pass round");
+    return;
```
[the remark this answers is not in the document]
lebdron: This behavior is defined by the caller. Method will return `Result`, which wil allow to distinguish bad outcomes.

irohad/consensus/yac/impl/yac_proposal_gate_impl.cpp
```diff
@@ -0,0 +1,67 @@
...
+            throw std::logic_error("Not implemented yet");
+          }
+          std::shared_ptr<shared_model::interface::Proposal> proposal(
+              std::move(current_proposal_->second.proposal).value());
+          return ProposalCommit{proposal, model_hash.round};
+        },
+        [&](RejectMessage reject) -> ProposalOutcomeType {
+          return ProposalReject{model_hash.round};
+        });
+    current_proposal_.reset();
```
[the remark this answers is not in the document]
lebdron: Probably naming is not the best, because it represents "last voted value", which is reset after consensus outcome is received.

irohad/consensus/yac/impl/yac_proposal_gate_impl.cpp
```diff
@@ -23,45 +23,46 @@ YacProposalGateImpl::YacProposalGateImpl(
...
   auto hash = hash_provider_->makeHash(vote);
   log_->info("vote for proposal ({}, {}, {})",
              vote.proposal ? vote.proposal.value()->hash().toString() : "''",
              vote.round.first,
              vote.round.second);
   auto order = orderer_->getOrdering(hash);
   if (not order) {
-    log_->error("ordering doesn't provide peers => pass round");
-    return;
+    return expected::makeError("ordering doesn't provide peers => pass round");
```
muratovv: "=> pass round" is redundant because a strategy is provided by outer scope.

muratovv: "=> pass round" is redundant because a strategy is provided by outer scope.

PR #1626, opened 2018-08-02
Add comments for og

muratovv:
### Description of the Change
* Add explanation comments for OG
* fix concurrent semantic for lock acquisition.

### Benefits
Less time for understanding what happens here.

### DIFF
--- a/irohad/ordering/impl/ordering_gate_impl.cpp
+++ b/irohad/ordering/impl/ordering_gate_impl.cpp
@@ -72,6 +72,7 @@ namespace iroha {
         const iroha::network::PeerCommunicationService &pcs) {
       log_->info("setPcs");
 
+      /// observable which contains heights of the top committed blocks
       auto top_block_height =
           pcs.on_commit()
               .transform([](const Commit &commit) {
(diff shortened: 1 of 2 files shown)


irohad/ordering/impl/ordering_gate_impl.cpp
```diff
@@ -103,8 +109,11 @@ namespace iroha {
         std::shared_ptr<shared_model::interface::Proposal> proposal) {
       log_->info("Received new proposal, height: {}", proposal->height());
       proposal_queue_.push(std::move(proposal));
-      std::lock_guard<std::mutex> lock(proposal_mutex_);
-      net_proposals_.get_subscriber().on_next(0);
+      {
```
lebdron: Scope should not be required here. Consider the following code:
```C++
{
A a;
a.smth();
...
B b(a);
b.smth();
}
```
`B` cannot be created before `A`, so constructor will not be called when scope is opened.

muratovv: Yes, standard guarantee order of initialization by statements:
```
int a;
int b;
```
But not guarantee for the list of initialization
```
int a, b;
```

So, seems your issue is valid.

PR #1632, opened 2018-08-07
with thread

itorK:
boostrap needs compiled with thread option.


lebdron:
Hi @itorK ! Thanks for the pull request. Could you please sign off the commit? It is done with `-s` option for `git commit`.

PR #1633, opened 2018-08-07
added run-postgres-dev and stop-postgres-dev scripts

tyvision:
Signed-off-by: tyvision <antykushin@gmail.com>

### Description of the Change

Since running a docker is not always required for local development, a postgresql instance could be enough for certain cases. First script `scripts/run-postgresql-dev.sh` starts the `postgresql` container. Second script `scripts/stop-postgresql-dev.sh` stops and removes the container and its dependencies such as volumes.

### Benefits

Script `run-pg-dev.sh` prints `<ip_address>`:`<port>` of the launched instance

### Possible Drawbacks 

If docker container with name `postgres`exists, please run `stop-pg-dev.sh` script first

To run
```
cd scripts/
./run-pg-dev.sh
```

To stop:
```
./stop-pg-dev.sh
```


igor-egorov:
I suggest slightly change the names to run-pg-dev.sh and stop-pg-dev.sh.

scripts/run-postgresql-dev.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+export PG_CONTAINER_NAME="postgres"
+PG_IMAGE_NAME="postgres:9.5"
+PG_HOST_DATA="/tmp/pgdata"
+PG_CONTAINER_DATA="/var/lib/postgresql/data"
+PG_USER="iroha"
+PG_PASS="helloworld"
```
igor-egorov: Change PG_PASS to "mysecretpassword".
Since the script is designed for development needs, the password should be changed.

https://github.com/hyperledger/iroha/blob/38dd0a1818e0269834625985955a360d34cb8baf/example/config.sample#L5

https://github.com/hyperledger/iroha/blob/38dd0a1818e0269834625985955a360d34cb8baf/example/config.docker#L5

Tests also use mysecretpassword by default:

https://github.com/hyperledger/iroha/blob/38dd0a1818e0269834625985955a360d34cb8baf/test/framework/config_helper.hpp#L27

scripts/run-pg-dev.sh
```diff
@@ -0,0 +1,23 @@
...
+PG_PASS="mysecretpassword"
+export PG_PORT="5432"
+
+export G_ID=$(id -g)
+export U_ID=$(id -u)
+
+mkdir -p ${PG_HOST_DATA}
+docker run -dt --user "${U_ID}:${G_ID}" \
+        --name ${PG_CONTAINER_NAME} \
+        -p ${PG_PORT} \
```
igor-egorov: Consider replacing with `-p 127.0.0.1:5432:5432 \`, otherwise, there is a chance to bind Postgres to a public network interface (what was happen in my case).

scripts/run-pg-dev.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+export PG_CONTAINER_NAME="postgres"
+PG_IMAGE_NAME="postgres:9.5"
+PG_HOST_DATA="/tmp/pgdata"
```
bakhtin: I think we can simply use default location for a volume (i.e., `-v ${PG_CONTAINER_DATA}`). `/tmp` sometimes is a RAM disk.

scripts/run-pg-dev.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+export PG_CONTAINER_NAME="postgres"
```
bakhtin: Why do you `export` some variables while do not others?

scripts/stop-pg-dev.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+export PG_CONTAINER_NAME="postgres"
+PG_HOST_DATA="/tmp/pgdata"
```
bakhtin: Same here with volumes

scripts/stop-pg-dev.sh
```diff
@@ -0,0 +1,10 @@
...
+export PG_PORT="5432"
+
+docker stop ${PG_CONTAINER_NAME}; docker rm -f ${PG_CONTAINER_NAME}
```
bakhtin: you do not need to use `-f`  flag as `stop` shuts down container gracefully

PR #1636, opened 2018-08-09
Feature/get roles test

Carver182:
### Description of the Change

This test checks whether the user can query list of roles

### Benefits

It will be possible to test if the user have necessary permissions to require list of roles

### Possible Drawbacks 

none


test/integration/acceptance/CMakeLists.txt
```diff
@@ -90,3 +90,7 @@ addtest(set_account_detail_test set_account_detail_test.cpp)
 target_link_libraries(set_account_detail_test
     acceptance_fixture
     )
+addtest(get_roles_test get_roles_test.cpp)
+target_link_libraries(get_roles_test
+    acceptance_fixture
+    )
```
igor-egorov: Add line break to the end of the file. That is the style we use.

igor-egorov:
* It was committed alongside with merges to actualize feature branch state after receiving reviewers' approvals.

test/integration/acceptance/CMakeLists.txt
Carver182: done

test/integration/acceptance/get_roles_test.cpp
```diff
@@ -59,28 +52,26 @@ TEST_F(GetRoles, CanGetRoles) {
  * @then there is should be an exception
  */
 TEST_F(GetRoles, CanNotGetRoles) {
-    auto checkQuery = [](auto &queryResponse) {
-        ASSERT_ANY_THROW(boost::apply_visitor(
-                framework::SpecifiedVisitor<shared_model::interface::RolesResponse>(), queryResponse.get()));
-    };
+  auto checkQuery = [](auto &queryResponse) {
+    ASSERT_ANY_THROW(boost::apply_visitor(
```
Akvinikym: I think, it's better not to check against throwing just any exception, but use `SpecifiedVisitor` to unwrap response to a specific error type. I assume, it would be StatefulValidationError response. You can check the actual return type in debugger

Akvinikym: Also, if you implement it, update the comment accordingly

test/integration/acceptance/get_roles_test.cpp
```diff
@@ -4,53 +4,46 @@
...
 
-class GetRoles : public AcceptanceFixture {
-public:
-//    auto makeUserWithPerms(const interface::RolePermissionSet &perms = {
-//            interface::permissions::Role::kGetRoles}) {
-//        return AcceptanceFixture::makeUserWithPerms(perms);
-//    }
-
-};
+class GetRoles : public AcceptanceFixture {};
```
igor-egorov: There is no need in empty `GetRoles` class. Remove it. Use `AcceptanceFixture` below instead of `GetRoles`.

test/integration/acceptance/get_roles_test.cpp
```diff
@@ -4,53 +4,46 @@
...
  * @given a user with CanGetRoles permission
  * @when execute query with getRoles command
  * @then there is should be no exception
  */
 TEST_F(GetRoles, CanGetRoles) {
-    auto checkQuery = [](auto &queryResponse) {
-        ASSERT_NO_THROW(boost::apply_visitor(
-                framework::SpecifiedVisitor<shared_model::interface::RolesResponse>(), queryResponse.get()));
-    };
+  auto checkQuery = [](auto &queryResponse) {
```
igor-egorov: Rewrite "then" section of the docstring. It should indicate that the list of roles can be received.
Presence or missing of an exception is an implementation detail of the test and should not be mentioned in the docstring.

test/integration/acceptance/get_roles_test.cpp
```diff
@@ -59,28 +52,34 @@ TEST_F(GetRoles, CanGetRoles) {
...
-
-    auto query = TestUnsignedQueryBuilder()
-            .createdTime(iroha::time::now())
-            .creatorAccountId(kUserId)
-            .queryCounter(1)
-            .getRoles()
-            .build()
-            .signAndAddSignature(kUserKeypair)
-            .finish();
+  auto checkQuery = [](auto &queryResponse) {
```
igor-egorov: Rewrite "then" section. The same comment as for previous test case. In this test case, the description of "then" is outdated.

PR #1641, opened 2018-08-12
Improve permissions doc compiler

igor-egorov:

### Description of the Change

Now script generates also a table of contents for permissions list.
We received feedback about such missing table more than once. 

### Benefits

All the existing permissions can be viewed on a single screen.
There is no need anymore to go through the full definitions of permissions while searching a certain one.

### Possible Drawbacks 

None

### Usage Examples or Tests

[DEMO](https://html-huuroiewos.now.sh/maintenance/permissions)  - left side menu navigation is broken threre - thats is ok. 

or compile by yourself
```
cd docs/source
make permissions
make html
firefox _build/html/maintenance/permissions.html
```


neewy:
Great, I like "Grantable" tags

PR #1646, opened 2018-08-14
Fix the rest of thread scheduling issues for torii service test

igor-egorov:

### Description of the Change

The time showed us that the rest of torii_service_test s also require the fix for thread scheduling issues on slow machines like our CI environment. 

The idea behind the fix is emulating client behavior while getting tx statuses - resubscribing.

### Benefits

Passing tests in a various environments.
Test behavior now is a bit closer to the behavior of a real Iroha user.

### Possible Drawbacks 

None. 

### Usage Examples or Tests

```
cmake -H. -Bbuild
cmake --build build --target torii_service_test
cd build
ctest -R torii_service_test --output-on-failure
```

### DIFF
--- a/test/module/irohad/torii/torii_service_test.cpp
+++ b/test/module/irohad/torii/torii_service_test.cpp
@@ -208,7 +208,12 @@ TEST_F(ToriiServiceTest, StatusWhenTxWasNotReceivedBlocking) {
     tx_request.set_tx_hash(
         shared_model::crypto::toBinaryString(tx_hashes.at(i)));
     iroha::protocol::ToriiResponse toriiResponse;
-    client.Status(tx_request, toriiResponse);
+    auto resub_counter(resubscribe_attempts);
+    do {
+      client.Status(tx_request, toriiResponse);
+    } while (toriiResponse.tx_status()
+                 != iroha::protocol::TxStatus::NOT_RECEIVED
+             and --resub_counter);
     ASSERT_EQ(toriiResponse.tx_status(),
               iroha::protocol::TxStatus::NOT_RECEIVED);
   }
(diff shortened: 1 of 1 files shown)


test/module/irohad/torii/torii_service_test.cpp
```diff
@@ -208,7 +208,12 @@ TEST_F(ToriiServiceTest, StatusWhenTxWasNotReceivedBlocking) {
     tx_request.set_tx_hash(
         shared_model::crypto::toBinaryString(tx_hashes.at(i)));
     iroha::protocol::ToriiResponse toriiResponse;
-    client.Status(tx_request, toriiResponse);
+    auto resub_counter(resubscribe_attempts);
+    do {
+      client.Status(tx_request, toriiResponse);
+    } while (toriiResponse.tx_status()
+                 != iroha::protocol::TxStatus::NOT_RECEIVED
```
lebdron: `NOT_RECEIVED` should not require several requests, maybe remove this loop?

test/module/irohad/torii/torii_service_test.cpp
```diff
@@ -383,9 +388,12 @@ TEST_F(ToriiServiceTest, CheckHash) {
...
     tx_request.set_tx_hash(shared_model::crypto::toBinaryString(hash));
     iroha::protocol::ToriiResponse toriiResponse;
-    // when
-    client.Status(tx_request, toriiResponse);
-    // then
+    auto resub_counter(resubscribe_attempts);
+    do {
+      client.Status(tx_request, toriiResponse);
+    } while (toriiResponse.tx_hash()
+                 != shared_model::crypto::toBinaryString(hash)
```
lebdron: Please move `shared_model::crypto::toBinaryString(hash)` to separate variable to avoid duplication in line 398. Same with two test cases below.

PR #1649, opened 2018-08-14
MST Pending Status

Akvinikym:
### Description of the Change

This PR adds a MST_PENDING status for transactions, which were sent for further processing into MST processor.

### Benefits

Now users will be able to understand, if their transaction is being processed by MST.

### Possible Drawbacks 

None.


irohad/torii/processor/impl/transaction_processor_impl.cpp
```diff
@@ -142,6 +142,11 @@ namespace iroha {
       if (boost::size(transaction->signatures()) < transaction->quorum()) {
         log_->info("waiting for quorum signatures");
         mst_processor_->propagateTransaction(transaction);
+        status_bus_->publish(
```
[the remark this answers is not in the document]
Akvinikym: This status publishing is done in such way throughout this file, and I think it will be unnecessary hard to move it into a separate function, because we need to put the status somehow, and it's done through the builder. In case of separate function we can pass the builder, but why?

irohad/torii/processor/impl/transaction_processor_impl.cpp
```diff
@@ -142,6 +142,11 @@ namespace iroha {
...
+            shared_model::builder::DefaultTransactionStatusBuilder()
+                .mstPending()
+                .txHash(transaction->hash())
+                .build());
         return;
       }
```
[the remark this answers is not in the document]
Akvinikym: I think, stateless success is a good variant for signed MST, because, in fact, it passes stateless validation only after all needed signatures appear

irohad/torii/processor/impl/transaction_processor_impl.cpp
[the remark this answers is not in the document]
Akvinikym: So, after a little discuss, I decided that this task actually is about creating something like status factory, which should be done outside this PR

shared_model/backend/protobuf/transaction_responses/proto_tx_response.hpp
```diff
@@ -54,6 +60,23 @@ namespace shared_model {
...
 
+      /**
+       * Compare priorities of two transaction responses
+       * @param other response
+       * @return:
+       *    - -1, if this response's priority is less, than other's
+       *    -  0, if it's equal
+       *    -  1, if it's greater
+       */
+      int comparePriorities(const TransactionResponse &other) const noexcept {
```
l4l: Use `enum class` instead

irohad/torii/impl/command_service.cpp
```diff
@@ -259,16 +261,14 @@ namespace torii {
...
-            cache_->findItem(hash).value_or([&] {
-              log_->debug("tx not received: {}", hash.toString());
-              return shared_model::proto::TransactionStatusBuilder()
-                  .txHash(hash)
-                  .notReceived()
-                  .build()
-                  .getTransport();
-            }())));
+    ResponsePtrType initial_status = cache_->findItem(hash).value_or([&] {
+      log_->debug("tx not received: {}", hash.toString());
```
l4l: tx **is** not ...

shared_model/backend/protobuf/transaction_responses/proto_tx_response.hpp
```diff
@@ -85,6 +108,23 @@ namespace shared_model {
       // stub hash
       const Lazy<crypto::Hash> hash_{
           [this] { return crypto::Hash(this->proto_->tx_hash()); }};
+
+      /**
+       * @return priority this transaction response
+       */
+      int priority() const noexcept {
+        return iroha::visit_in_place(
```
l4l: `return variant_->which()` ?

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -104,15 +107,22 @@ TEST_F(MstPipelineTest, OnePeerSendsTest) {
   auto tx = baseTx()
                 .setAccountDetail(kUserId, "fav_meme", "doge")
                 .quorum(kSignatories + 1);
+  auto checkMstPendingTxStatus =
+      [](const shared_model::proto::TransactionResponse &resp) {
+        ASSERT_NO_THROW(boost::apply_visitor(
+            SpecifiedVisitor<interface::MstPendingResponse>(), resp.get()));
+      };
+//  auto checkMstPassedTxStatus =
```
l4l: Maybe delete commented lines?

shared_model/backend/protobuf/transaction_responses/proto_tx_response.hpp
Akvinikym: No, this won't work, because then we should place classes in variant according to their priorities, and this is not as obvious as this `visit_in_place` function IMHO

shared_model/backend/protobuf/transaction_responses/proto_tx_response.hpp
Akvinikym: What do you mean?

l4l:
```cpp
enum class Result { LE, EQ, GT, };

Result comparePriorities(const T&) ...
```

irohad/torii/processor/impl/transaction_processor_impl.cpp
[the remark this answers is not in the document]
muratovv: Offline discussion:
Move status_bus::publish to the separate method and add a task for refactoring this method in a separate PR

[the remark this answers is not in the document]
muratovv: Offline discussion:
Move status_bus::publish to the separate method and add a task for refactoring this method in a separate PR

shared_model/interfaces/transaction_responses/mst_pending_response.hpp
```diff
@@ -0,0 +1,24 @@
...
+#define IROHA_MST_PENDING_RESPONSE_HPP
+
+namespace shared_model {
+  namespace interface {
+    /**
+     * Transaction was sent to MST processing and not yet signed by all required
+     * signatories
+     */
+    class MstPendingResponse : public AbstractTxResponse<MstPendingResponse> {
+     private:
```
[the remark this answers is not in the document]
muratovv: Well, in this case, add a task for refactoring this with the lowest priority.

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -104,15 +107,22 @@ TEST_F(MstPipelineTest, OnePeerSendsTest) {
   auto tx = baseTx()
                 .setAccountDetail(kUserId, "fav_meme", "doge")
                 .quorum(kSignatories + 1);
+  auto checkMstPendingTxStatus =
+      [](const shared_model::proto::TransactionResponse &resp) {
+        ASSERT_NO_THROW(boost::apply_visitor(
+            SpecifiedVisitor<interface::MstPendingResponse>(), resp.get()));
+      };
+//  auto checkEnoughSignaturesCollectedStatus =
```
kamilsa: Did you forget to delete that?

shared_model/backend/protobuf/transaction_responses/proto_tx_response.hpp
```diff
@@ -85,6 +109,23 @@ namespace shared_model {
       // stub hash
       const Lazy<crypto::Hash> hash_{
           [this] { return crypto::Hash(this->proto_->tx_hash()); }};
+
+      /**
+       * @return priority this transaction response
```
kamilsa: Maybe explain what priority means?

shared_model/backend/protobuf/transaction_responses/proto_tx_response.hpp
Akvinikym: OK, I implemented this method looking at old C functions, strcmp etc, where it's done in this way :) will change it

muratovv: Maybe relay operator<, >, == to transactionResponse from priority directly?

shared_model/backend/protobuf/transaction_responses/proto_tx_response.hpp
```diff
@@ -85,6 +109,24 @@ namespace shared_model {
       // stub hash
       const Lazy<crypto::Hash> hash_{
           [this] { return crypto::Hash(this->proto_->tx_hash()); }};
+
+      /**
+       * @return priority of this transaction response; transaction response can
+       * only be replaced with one with higher priority
+       */
+      int priority() const noexcept {
```
muratovv: Seems, this table should be in the interface, not it proto backend.

shared_model/backend/protobuf/transaction_responses/proto_tx_response.hpp
```diff
@@ -85,6 +109,24 @@ namespace shared_model {
...
+            *variant_,
+            [](const StatelessValidTxResponse &) { return 1; },
+            [](const MstPendingResponse &) { return 2; },
+            [](const EnoughSignaturesCollectedResponse &) { return 3; },
+            [](const StatefulValidTxResponse &) { return 4; },
+            [](const CommittedTxResponse &) { return 5; },
+            [](const StatelessFailedTxResponse &) { return 6; },
+            [](const StatefulFailedTxResponse &) { return 7; },
+            [](const MstExpiredResponse &) { return 8; },
+            [](const NotReceivedTxResponse &) { return 9; });
```
muratovv: Currently, statuses are misleading. For example, why MstExpired is greater than committed?

I suggest reworking priorities with final status priority, which means maximal priority and reflects the impossibility of achieving another status for the transaction. Also, take care of intermediate statuses.

If there will be some troubles, we may arrange an offline meeting.

irohad/torii/impl/command_service.cpp
```diff
@@ -57,11 +57,10 @@ namespace torii {
       auto tx_hash = proto_response->transactionHash();
       auto cached_tx_state = cache_->findItem(tx_hash);
       if (cached_tx_state
-          and proto_response->getTransport().tx_status()
-              <= cached_tx_state->tx_status()) {
+          and proto_response->comparePriorities(**cached_tx_state) <= 0) {
```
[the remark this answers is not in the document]
Akvinikym: It was a value, which compare function gave, so that `<= 0` meant that first value is less or equal, than the second one, C-style. But as it's going to be replaced by enum class, doesn't matter

irohad/torii/processor/impl/transaction_processor_impl.cpp
```diff
@@ -17,23 +17,25 @@ namespace iroha {
 
     using network::PeerCommunicationService;
 
-    static std::string composeErrorMessage(
-        const validation::TransactionError &tx_error) {
-      if (not tx_error.first.tx_passed_initial_validation) {
-        return (boost::format("Stateful validation error: transaction %s "
-                              "did not pass initial verification: "
-                              "checking '%s', error message '%s'")
+    namespace {
```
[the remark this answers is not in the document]
Akvinikym: Alternative is private static function, but as [this question](https://stackoverflow.com/questions/154469/unnamed-anonymous-namespaces-vs-static-functions) tells, it's better to use anonymous namespace to define transaction unit only methods.

test/module/irohad/torii/processor/transaction_processor_test.cpp
```diff
@@ -147,12 +147,10 @@ TEST_F(TransactionProcessorTest, TransactionProcessorOnProposalTest) {
   for (size_t i = 0; i < proposal_size; i++) {
     auto &&tx = addSignaturesFromKeyPairs(baseTestTx(), makeKey());
     txs.push_back(tx);
-    status_map[tx.hash()] =
-        status_builder.notReceived().txHash(tx.hash()).build();
   }
 
   EXPECT_CALL(*status_bus, publish(_))
-      .Times(proposal_size)
+      .Times(proposal_size * 2)
```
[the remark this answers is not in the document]
Akvinikym: Because new status - `kEnoughSignaturesCollected` - was introduced, and now transactions receive it as well

irohad/torii/processor/impl/transaction_processor_impl.cpp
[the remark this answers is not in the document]
muratovv: As I understand, it is related only to header files, isn't it?

test/module/irohad/torii/processor/transaction_processor_test.cpp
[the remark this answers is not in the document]
Akvinikym: It's +1 for every transaction in proposal, so number of times is doubled

irohad/torii/processor/impl/transaction_processor_impl.cpp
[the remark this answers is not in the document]
Akvinikym: As I explored, this makes no sense in header files, because it will be duplicated while compiling in every file this header is included, but for .cpp it works perfectly

PR #1651, opened 2018-08-15
Feature/query benchmark

victordrobny:
### Description of the Change
In order to measure Iroha query execution performance, a benchmark that executes GetAccount query was introduced.

### Benefits
A performance measure is introduced.


test/benchmark/query_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "utils/query_error_response_visitor.hpp"
+
+const std::string kUser = "user";
+const std::string kUserId = kUser + "@test";
+const shared_model::crypto::Keypair kAdminKeypair =
+    shared_model::crypto::DefaultCryptoAlgorithmType::generateKeypair();
+const shared_model::crypto::Keypair kUserKeypair =
+    shared_model::crypto::DefaultCryptoAlgorithmType::generateKeypair();
+
+TestUnsignedTransactionBuilder createUserWithPerms(
```
Alexey-N-Chernyshov: Please, add couple words of description

test/benchmark/CMakeLists.txt
```diff
@@ -22,6 +22,18 @@ target_include_directories(bm_proto_creation PUBLIC
 
 target_link_libraries(bm_proto_creation
     benchmark
+    gtest::gtest
```
lebdron: Please remove gtest and gmock here and below, because they do not look used in the code.

victordrobny: They are used in test validator, which is used in test transaction builder

PR #1659, opened 2018-08-17
Feature/reduced hash wrapper

x3medima17:
### Description of the Change
Makes `reducedHash` visible to libraries.

### Benefits
Now clients can call `reducedHash`.

### Possible Drawbacks 
More wrappers.

### DIFF
--- a/shared_model/backend/protobuf/transaction.hpp
+++ b/shared_model/backend/protobuf/transaction.hpp
@@ -29,6 +29,7 @@
 
 namespace shared_model {
   namespace proto {
+    using HashProvider = shared_model::crypto::Sha3_256;
     class Transaction FINAL : public CopyableProto<interface::Transaction,
                                                    iroha::protocol::Transaction,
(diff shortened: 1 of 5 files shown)


shared_model/backend/protobuf/transaction.hpp
```diff
@@ -29,6 +29,7 @@
 
 namespace shared_model {
   namespace proto {
+    using HashProvider = shared_model::crypto::Sha3_256;
```
nickaleks: Please move HashProvider inside the class so it does not leak to the namespace

shared_model/builders/protobuf/unsigned_proto.hpp
```diff
@@ -102,7 +103,7 @@ namespace shared_model {
       typename std::enable_if<
```
nickaleks: please replace typename std::enable_if<...> with std::enable_if_t

PR #1660, opened 2018-08-20
Send tx sequence method in ITF

kamilsa:
### Description of the Change

This PR adds ability to send transaction sequences to the ITF and check stateless valid/invalid status

### Benefits

We can send sequences of transactions to the ITF

### Possible Drawbacks 

None

### Usage Examples or Tests

The test was added to pipeline test file

### DIFF
--- a/test/framework/integration_framework/integration_test_framework.cpp
+++ b/test/framework/integration_framework/integration_test_framework.cpp
@@ -233,6 +233,8 @@ namespace integration_framework {
 
     boost::barrier bar(2);
 
+    // subscribe on status bus and save all stateless statuses into a vector
+    std::vector<shared_model::proto::TransactionResponse> statuses;
     iroha_instance_->instance_->getStatusBus()
         ->statuses()
         .filter([&transactions](auto s) {
@@ -253,7 +255,12 @@ namespace integration_framework {
           return it != transactions.end();
         })
         .take(transactions.size())
-        .subscribe([](auto &&) {}, [&bar] { bar.wait(); });
+        .subscribe(
+            [&statuses](auto s) {
+              statuses.push_back(*std::static_pointer_cast<
+                                 shared_model::proto::TransactionResponse>(s));
+            },
+            [&bar] { bar.wait(); });
 
     // put all transactions to the TxList and send them to iroha
     iroha::protocol::TxList tx_list;
@@ -268,14 +275,6 @@ namespace integration_framework {
 
     // make sure that the first (stateless) status is come
     bar.wait();
-
-    // collect all statuses and process them
-    std::vector<shared_model::proto::TransactionResponse> statuses;
-    for (const auto tx : transactions) {
-      getTxStatus(tx->hash(), [&statuses](const auto &status) {
-        statuses.push_back(status);
-      });
-    }
 
     validation(statuses);
     return *this;
--- a/test/framework/integration_framework/integration_test_framework.cpp
+++ b/test/framework/integration_framework/integration_test_framework.cpp
@@ -228,48 +228,52 @@ namespace integration_framework {
       const shared_model::interface::TransactionSequence &tx_sequence,
       std::function<void(std::vector<shared_model::proto::TransactionResponse>
                              &)> validation) {
+    log_->info("send transactions");
     auto transactions = tx_sequence.transactions();
 
-    boost::barrier bar1(tx_sequence.transactions().size() + 1);
-    auto bar2 =
-        std::make_shared<boost::barrier>(tx_sequence.transactions().size() + 1);
+    boost::barrier bar(2);
 
     iroha_instance_->instance_->getStatusBus()
         ->statuses()
-        .filter([&](auto s) {
-          return std::find_if(transactions.begin(),
-                              transactions.end(),
-                              [&s](const auto tx) {
-                                return s->transactionHash() == tx->hash();
-                              })
-              == transactions.end();
+        .filter([transactions](auto s) {
+          // filter statuses for transactions from sequence
+          auto it = std::find_if(
+              transactions.begin(), transactions.end(), [&s](const auto tx) {
+                // check if status is either stateless valid or failed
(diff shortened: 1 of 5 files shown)


test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -224,6 +224,68 @@ namespace integration_framework {
     return *this;
   }
 
+  IntegrationTestFramework &IntegrationTestFramework::sendTxSequence(
+      const shared_model::interface::TransactionSequence &tx_sequence,
+      std::function<void(std::vector<shared_model::proto::TransactionResponse>
+                             &)> validation) {
+    log_->info("send transactions");
+    auto transactions = tx_sequence.transactions();
```
lebdron: No need to copy here, maybe use reference?

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -224,6 +224,68 @@ namespace integration_framework {
...
+
+    boost::barrier bar(2);
+
+    iroha_instance_->instance_->getStatusBus()
+        ->statuses()
+        .filter([transactions](auto s) {
```
lebdron: Same with capture by reference.

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -224,6 +224,68 @@ namespace integration_framework {
...
+          // filter statuses for transactions from sequence
+          auto it = std::find_if(
+              transactions.begin(), transactions.end(), [&s](const auto tx) {
```
lebdron: `const auto &tx`

test/integration/pipeline/pipeline_test.cpp
```diff
@@ -98,3 +107,53 @@ TEST(PipelineIntegrationTest, SendTx) {
...
+ * @then receive STATELESS_VALIDATION_SUCCESS status on that transactions,
+ * all transactions are passed to proposal and does not appear in block
+ */
+TEST(PipelineIntegrationTest, SendTxSequence) {
+  size_t tx_size = 5;
+
+  shared_model::interface::types::SharedTxsCollectionType txs;
+
+  for (size_t i = 0; i < tx_size; i++) {
+    auto &&tx = prepareAddAssetQtyTransaction(iroha::time::now() + i);
```
lebdron: `iroha::time::now()` is better to be called once before the loop for simplicity.

test/framework/integration_framework/integration_test_framework.cpp
kamilsa: its a test, but ok

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -224,6 +224,73 @@ namespace integration_framework {
...
+    const auto &transactions = tx_sequence.transactions();
+
+    boost::barrier bar(2);
+
+    iroha_instance_->instance_->getStatusBus()
+        ->statuses()
+        .filter([&transactions](auto s) {
+          // filter statuses for transactions from sequence
+          auto it = std::find_if(
+              transactions.begin(), transactions.end(), [&s](const auto tx) {
```
lebdron: -> `const auto &tx`

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -224,6 +224,73 @@ namespace integration_framework {
...
+                    [](const shared_model::interface::StatelessValidTxResponse
+                           &stateless_valid_response) { return true; },
+                    [](const auto &other_responses) { return false; });
+                return is_stateless_status
+                    and s->transactionHash() == tx->hash();
+              });
+          return it != transactions.end();
+        })
+        .take(transactions.size())
+        .subscribe([](auto &&) {}, [&bar] { bar.wait(); });
```
lebdron: Unfortunately, it seems that `getTxStatus` call will not work in a deterministic way. Consider the following scenario:
1. `ListTorii` is called, thread switch to gRPC, transactions are executed, `wait` on line 256 is called
2. thread switch to test, `wait` on line 270 is called, it continues execution because the barrier requirement of 2 is satisfied
3. thread switch to gRPC, it continues to run up to stateful validation
4. thread switch to test, `getTxStatus` is executed, but the results are stateful valid/invalid already

Therefore, it seems necessary to capture the statuses in `statuses` subscription. ` statuses.push_back(status);` can be moved inside `OnNext` lambda, and the barrier will be placed between `ListTorii` and `validation` calls.

test/framework/integration_framework/integration_test_framework.cpp
kamilsa: it is shared ptr

test/regression/regression_test.cpp
```diff
@@ -23,7 +23,7 @@
 #include "framework/integration_framework/integration_test_framework.hpp"
 #include "framework/specified_visitor.hpp"
 
-constexpr auto kUser = "user@test";
+constexpr auto kAdmin = "user@test";
```
Akvinikym: `kAdmin` with credentials of the user? Should be changed, I think

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -224,6 +224,72 @@ namespace integration_framework {
...
+                    [](const auto &other_responses) { return false; });
+                return is_stateless_status
+                    and s->transactionHash() == tx->hash();
+              });
+          return it != transactions.end();
+        })
+        .take(transactions.size())
+        .subscribe(
+            [&statuses](auto s) {
+              statuses.push_back(*std::static_pointer_cast<
```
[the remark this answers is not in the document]
kamilsa: It won't work. It cannot cast interface::TransactionResponse to proto::TransactionResponse

test/regression/regression_test.cpp
kamilsa: oh my.. It was unintentionally refactored

PR #1661, opened 2018-08-21
Integration tests for GetPendingTransactions()

Akvinikym:
### Description of the Change

This PR does two things:

1. Fixes MST - before it did not send batches, which were inserted or updated with new signatures, to the pending storage, which lead to inability to retrieve them
2. Implements a couple of tests for `getPendingTransactions` query

### Benefits

Tests for a new query.

### Possible Drawbacks 

None.

### Usage of tests

All tests, which are added, are located inside `test/integration/pipeline/multisig_tx_pipeline_test.cpp`. They can be executed by building and running the corresponding target, `multisig_tx_pipeline_test`


irohad/multi_sig_transactions/impl/mst_processor_impl.cpp
```diff
@@ -32,6 +32,13 @@ namespace iroha {
                     });
     }
   }
+  void shareState(
+      ConstRefState state,
+      rxcpp::subjects::subject<std::shared_ptr<MstState>> &subject) {
```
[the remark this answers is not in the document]
Akvinikym: Compile error, if you try to do it :(

irohad/multi_sig_transactions/impl/mst_processor_impl.cpp
```diff
@@ -32,6 +32,13 @@ namespace iroha {
                     });
     }
   }
+  void shareState(
```
muratovv: Add documentation for the function and maybe rename it or previous shareState. Now it isn't obvious that they are doing the not same job.

irohad/multi_sig_transactions/impl/mst_processor_impl.cpp
muratovv: Use alias for DataType instead

irohad/multi_sig_transactions/impl/mst_processor_impl.cpp
```diff
@@ -56,7 +63,12 @@ namespace iroha {
 
   auto FairMstProcessor::propagateBatchImpl(const iroha::DataType &batch)
       -> decltype(propagateBatch(batch)) {
-    shareState(storage_->updateOwnState(batch), batches_subject_);
+    auto state_update = storage_->updateOwnState(batch);
```
muratovv: I don't see a reason to introduce pair with bool response here. Method isEmpty is already implemented in MstState. 
Also, the current interface of updateOwnState is totally misleading - for one situation it returns one value, for another situation - another value.

UpdateOwnState possibly may return pair of two states: state with completed transactions and state with new signatures.

igor-egorov: AFAIK the second parameter reflects the state of a batch - completed or just updated. 
Nevertheless, the current implementation with std::pair is unclear and misleading.
My vote for the simplicity.

irohad/multi_sig_transactions/state/mst_state.hpp
```diff
@@ -166,8 +166,9 @@ namespace iroha {
      * Insert batch in own state and push it in out_state if required
      * @param out_state - state for inserting completed batches
      * @param rhs_tx - batch for insert
+     * @return true if batch is complete after that insertion; false otherwise
      */
-    void insertOne(MstState &out_state, const DataType &rhs_tx);
+    bool insertOne(MstState &out_state, const DataType &rhs_tx);
```
muratovv: Seems, that we can avoid this bool with two passed states

irohad/multi_sig_transactions/storage/mst_storage.hpp
```diff
@@ -39,20 +39,20 @@ namespace iroha {
...
-    MstState apply(
+    StateUpdateResult apply(
         const std::shared_ptr<shared_model::interface::Peer> &target_peer,
         const MstState &new_state);
 
     /**
      * Provide updating state of current peer with new transaction
      * @param tx - new transaction for insertion in state
-     * @return State with completed transaction
+     * @return completed or updated mst state
```
muratovv: s/or/and

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -120,3 +150,91 @@ TEST_F(MstPipelineTest, OnePeerSendsTest) {
...
+      .sendTx(signTx(pending_tx, signatories[0]))
+      .sendQuery(makeGetPendingTxsQuery(kUserId, kUserKeypair),
+                 signatory_check(1))
+      .sendTx(signTx(pending_tx, signatories[1]))
+      .sendQuery(makeGetPendingTxsQuery(kUserId, kUserKeypair),
+                 signatory_check(2));
+}
+
+/**
+ * @given a ledger with pending transactions
```
muratovv: Seems that ledger initially empty or at least doesn't contains pending transactions.

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
```diff
@@ -191,18 +248,18 @@ TEST_F(MstProcessorTest, expiredTransactionUsecase) {
  * that contains TX with another signature
  *
  * @then check that:
- * state updated
+ * state observer is not called
  * AND 1 prepared transaction (although quorum 1)
```
lebdron: Please remove `although quorum 1`.

PR #1673, opened 2018-08-24
Add SendTx and SendQuery system tests

igor-egorov:

### Description of the Change

Add a test that:
* runs separate independent Iroha instance as our users do
* sends a transaction and awaits its committed status
* sends a query and verifies the query response

Additionally, the implementation of KeysManager was refactored.

### Benefits

Coverage is increased. 
No need to include tx-example.py to CI cycle.

### Possible Drawbacks 

None.

### Usage Examples or Tests

```
cmake -H. -Bbuild
cd build
make irohad_test
ctest -R irohad_test --output-on-failure
```

### DIFF
--- a/libs/crypto/keys_manager.hpp
+++ b/libs/crypto/keys_manager.hpp
@@ -26,7 +26,7 @@ namespace shared_model {
   namespace crypto {
     class Keypair;
   }
-}
+}  // namespace shared_model
 
 namespace iroha {
   /**
@@ -37,25 +37,12 @@ namespace iroha {
     virtual ~KeysManager() = default;
 
     /**
-     * Create a new keypair and store it as is on disk
-     * @return false if create account failed
-     */
-    virtual bool createKeys() = 0;
-
-    /**
      * Create keys a new keypair and store it encrypted on disk
-     * @param pass_phrase is a password for the keys
+     * @param pass_phrase is used for private key encryption. Optional, default
+     * value is an empty string.
      * @return false if create account failed
      */
-    virtual bool createKeys(const std::string &pass_phrase) = 0;
-
-    /**
-     * Load plain-text keys associated with the manager, then validate loaded
-     * keypair by signing and verifying signature of test message
-     * @return nullopt if no keypair found locally, or verification failure;
-     *         related keypair otherwise
-     */
-    virtual boost::optional<shared_model::crypto::Keypair> loadKeys() = 0;
+    virtual bool createKeys(const std::string &pass_phrase = "") = 0;
 
     /**
      * Load encrypted keys associated with the manager, then validate loaded
@@ -65,7 +52,7 @@ namespace iroha {
      *         related keypair otherwise
      */
     virtual boost::optional<shared_model::crypto::Keypair> loadKeys(
-        const std::string &pass_phrase) = 0;
+        const std::string &pass_phrase = "") = 0;
   };
 
 }  // namespace iroha
--- a/libs/crypto/keys_manager.hpp
+++ b/libs/crypto/keys_manager.hpp
@@ -1,18 +1,6 @@
 /**
- * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
- * http://soramitsu.co.jp
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *        http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
  */
 
 #ifndef IROHA_KEYS_MANAGER_HPP
@@ -21,12 +9,7 @@
 #include <string>
 
 #include <boost/optional.hpp>
-
-namespace shared_model {
-  namespace crypto {
-    class Keypair;
-  }
-}  // namespace shared_model
+#include "cryptography/keypair.hpp"
 
 namespace iroha {
   /**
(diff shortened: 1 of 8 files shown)


igor-egorov:
To reviewers: I saw compilation errors. Please review despite them. I will be able to fix them after the Sunday. Locally I had no such errors, probably because of cache.

test/system/irohad_test.cpp
```diff
@@ -39,10 +34,22 @@ using namespace boost::filesystem;
 using namespace std::chrono_literals;
 using iroha::operator|;
 
-class IrohadTest : public testing::Test {
+/**
+The do-while cycle imitates client resubscription to the stream. Stream
```
lebdron: cycle -> loop

test/system/irohad_test.cpp
```diff
@@ -135,15 +156,69 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
...
+  tx_request.set_tx_hash(shared_model::crypto::toBinaryString(tx.hash()));
+
+  auto client = torii::CommandSyncClient(kAddress, kPort);
+  client.Torii(tx.getTransport());
+
+  auto resub_counter(resubscribe_attempts);
+  do {
+    std::this_thread::sleep_for(resubscribe_timeout);
+    client.Status(tx_request, torii_response);
+  } while (torii_response.tx_status() != iroha::protocol::TxStatus::COMMITTED
```
lebdron: `iroha::protocol::TxStatus::COMMITTED` can be moved to a separate variable, so that the literal is not duplicated both in line 195 and 198.

libs/crypto/keys_manager.hpp
```diff
@@ -26,7 +26,7 @@ namespace shared_model {
   namespace crypto {
     class Keypair;
   }
```
lebdron: `// namespace crypto`

libs/crypto/keys_manager.hpp
```diff
@@ -36,26 +36,13 @@ namespace iroha {
...
-    virtual bool createKeys(const std::string &pass_phrase) = 0;
-
-    /**
-     * Load plain-text keys associated with the manager, then validate loaded
-     * keypair by signing and verifying signature of test message
-     * @return nullopt if no keypair found locally, or verification failure;
-     *         related keypair otherwise
-     */
-    virtual boost::optional<shared_model::crypto::Keypair> loadKeys() = 0;
+    virtual bool createKeys(const std::string &pass_phrase = "") = 0;
```
lebdron: Not sure if this is a good idea, same below
http://www.codingstandard.com/rule/8-3-3-do-not-use-default-arguments/
>Use of default arguments can make code maintenance and refactoring more difficult. Overloaded forwarding functions can be used instead without having to change existing function calls.

It should be possible to define a function in the interface, such as
```C++
bool createKeys() {
  return createKeys("");
}
```

libs/crypto/keys_manager_impl.hpp
```diff
@@ -28,14 +17,30 @@ namespace iroha {
...
+     * @param path_to_keypair - path to directory that contains priv and pub key
+     * of an account
+     *
+     * Here we use an empty string as a default value of path to file, since
+     * there are usages of KeysManagerImpl with path passed as a part of
+     * account_id.
+     */
+    explicit KeysManagerImpl(
+        const std::string &account_id,
+        const boost::filesystem::path &path_to_keypair = "");
```
lebdron: Same here, a constructor with one parameter can be defined below:
```C++
explicit KeysManagerImpl(
  const std::string &account_id) 
  : KeysManagerImpl(account_id, "") {}
```

libs/crypto/keys_manager_impl.hpp
```diff
@@ -28,14 +17,30 @@ namespace iroha {
...
 
-    bool createKeys() override;
-    bool createKeys(const std::string &pass_phrase) override;
+    bool createKeys(const std::string &pass_phrase = "") override;
 
-    boost::optional<shared_model::crypto::Keypair> loadKeys() override;
+    /**
+     * Loads keypair from disk
```
lebdron: It should not be requred to write the whole doc if the behavior is already defined here:
https://github.com/hyperledger/iroha/blob/63cb6a186dccb3e40f222af3da561998a05a4f5d/libs/crypto/keys_manager.hpp#L47-L54

libs/crypto/keys_manager_impl.cpp
```diff
@@ -73,55 +64,37 @@ namespace iroha {
...
-      return boost::none;
-
-    Keypair keypair = Keypair(PublicKey(Blob::fromHexString(pub_key)),
-                              PrivateKey(Blob::fromHexString(priv_key)));
 
-    return this->validate(keypair) ? boost::make_optional(keypair)
-                                   : boost::none;
+    std::string contents;
+    file >> contents;
+    return boost::make_optional<std::string>(contents);
```
lebdron: `return contents;` should be OK, there should be an implicit ctor for the optional.

test/system/irohad_test.cpp
```diff
@@ -39,10 +34,22 @@ using namespace boost::filesystem;
...
-class IrohadTest : public testing::Test {
+/**
+The do-while cycle imitates client resubscription to the stream. Stream
+"expiration" is a valid designed case (see pr #1615 for the details).
+
+The number of attempts (5) is a magic constant here. The idea behind this number
+is the following: five resubscription with 3 seconds timeout is usually enough
+to pass the test; if not - most likely there is another bug.
+ */
+constexpr uint32_t resubscribe_attempts = 5;
```
lebdron: Maybe store the variables in the fixture?

libs/crypto/keys_manager.hpp
igor-egorov: I will add. Usually such comments are auto-generated by clang-format

libs/crypto/keys_manager.hpp
igor-egorov: Thanks! Did not know about the preferred way with overloading rather than default arguments.

libs/crypto/keys_manager_impl.cpp
```diff
@@ -73,55 +72,45 @@ namespace iroha {
...
-    std::ifstream file(filename);
+  boost::optional<std::string> KeysManagerImpl::loadFile(
+      const boost::filesystem::path &path) const {
+    auto file_path = path.string();
+    std::ifstream file(file_path);
     if (not file) {
-      log_->error("Cannot read '" + filename + "'");
-      return false;
+      log_->error("Cannot read '" + file_path + "'");
+      return {};
```
luckychess: Does it work as ```boost::none```?

test/system/irohad_test.cpp
```diff
@@ -122,9 +128,26 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
...
+   * "expiration" is a valid designed case (see pr #1615 for the details).
+   *
+   * The number of attempts (5) is a magic constant here. The idea behind this
+   * number is the following: five resubscription with 3 seconds timeout is
+   * usually enough to pass the test; if not - most likely there is another bug.
+   */
+  const uint32_t resubscribe_attempts = 5;
+  const std::chrono::seconds resubscribe_timeout = std::chrono::seconds(3);
+
+ protected:
```
luckychess: Why not private?

test/system/irohad_test.cpp
```diff
@@ -122,9 +128,26 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
 
  public:
   boost::filesystem::path irohad_executable;
-  std::chrono::milliseconds timeout = 1s;
+  const std::chrono::milliseconds kTimeout = std::chrono::seconds(1);
+  const std::string kAddress;
+  const uint16_t kPort;
 
- private:
+  boost::optional<child> iroha_process_;
```
luckychess: What the point to make it optional?

libs/crypto/keys_manager_impl.cpp
igor-egorov: Right, `{}` will produce `boost::none`;

test/system/irohad_test.cpp
igor-egorov: You are right. They can be private. I will change this.

test/system/irohad_test.cpp
igor-egorov: The optional reflects the possibility of missing initialized Iroha. The field `iroha_process_` cannot be just the class member because it will require initialization in fixture constructor's initializer list. We cannot do it because the initialization parameters for `iroha_process_` are not known by the moment of the fixture construction.

PR #1675, opened 2018-08-24
On-demand ordering gate and related fixes

lebdron:
### Description of the Change
On-demand ordering gate, which requests proposals from the on-demand ordering service, and uses voting to pass committed proposal to the pipeline

### Benefits
Ordering gate which does not require block reject if peers do not agree on a proposal

### Possible Drawbacks 
TransactionBatch is modified because it was impossible to build ordering gate with validator linking issues


irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -0,0 +1,114 @@
...
+                         current_round_ = nextRejectRound(current_round_);
+                       });
+        this->vote();
+      })),
+      current_round_(initial_round) {}
+
+void OnDemandOrderingGate::propagateTransaction(
+    std::shared_ptr<const shared_model::interface::Transaction> transaction)
+    const {
+  propagateBatch(batch_factory_(std::move(transaction)));
```
muratovv: I suggest not doing support this method because we have been moving to batch API.
So, maybe throw an exception instead?

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -0,0 +1,114 @@
...
+      proposal_gate_(std::move(proposal_gate)),
+      batch_factory_(std::move(batch_factory)),
+      subscription_(events.subscribe([this](auto event) {
+        // exclusive lock
+        std::lock_guard<std::shared_timed_mutex> lock(mutex_);
+
+        visit_in_place(event,
+                       [this](BlockEvent block) {
+                         // block committed, increment block round
+                         current_round_ = {block.height, 1};
```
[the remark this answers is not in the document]
lebdron: It is used only once now, so it should not be required.

shared_model/interfaces/iroha_internal/transaction_batch_factory.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "common/result.hpp"
+#include "interfaces/common_objects/transaction_sequence_common.hpp"
+#include "interfaces/iroha_internal/transaction_batch.hpp"
+#include "validators/transactions_collection/transactions_collection_validator.hpp"
+
+namespace shared_model {
+  namespace interface {
+
+    class TransactionBatchFactory {
```
kamilsa: Please add short description of the class

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -0,0 +1,75 @@
...
+    std::unique_ptr<shared_model::interface::UnsafeProposalFactory> factory,
+    transport::RoundType initial_round)
+    : ordering_service_(std::move(ordering_service)),
+      network_client_(std::move(network_client)),
+      events_subscription_(events.subscribe([this](auto event) {
+        // exclusive lock
+        std::lock_guard<std::shared_timed_mutex> lock(mutex_);
+
+        visit_in_place(event,
+                       [this](BlockEvent block) {
```
kamilsa: why not const ref?

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -0,0 +1,75 @@
...
+                         // block committed, increment block round
+                         current_round_ = {block.height, 1};
+                       },
+                       [this](EmptyEvent empty) {
```
kamilsa: why not const ref?

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -0,0 +1,75 @@
...
+                         // no blocks committed, increment reject round
+                         current_round_ = {current_round_.first,
```
kamilsa: can we use something else instead of pair. I mean round.first and round.second is hard to understand if you don't know that round is pair of (current round and reject round if I am not mistaken)

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -0,0 +1,75 @@
...
+                  current_round_.first, current_round_.second, {});
+            }));
+      })),
+      proposal_factory_(std::move(factory)),
+      current_round_(initial_round) {}
+
+void OnDemandOrderingGate::propagateTransaction(
+    std::shared_ptr<const shared_model::interface::Transaction> transaction)
+    const {
+  throw std::logic_error("Not implemented");
```
muratovv: Maybe write "deprecated" instead of "logic error"?

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+rxcpp::observable<std::shared_ptr<shared_model::interface::Proposal>>
+OnDemandOrderingGate::on_proposal() {
+  return proposal_notifier_.get_observable();
+}
+
+void OnDemandOrderingGate::setPcs(
+    const iroha::network::PeerCommunicationService &pcs) {
+  throw std::logic_error("Not implemented");
```
muratovv: Here is also, possibly to provide the reason of an exception, isn't it?

irohad/ordering/impl/on_demand_ordering_gate.cpp
muratovv: I miss with that. I mean "deprecated" instead of "Not implemented". Sorry about that.

PR #1685, opened 2018-08-28
Fix send tx sequence method

kamilsa:

### Description of the Change

Due to unclear reasons boost::barrier throws segfault in send tx sequence test (once out from 100 runs on average). To avoid that it was replaced with std::condition_variable. 
### Benefits

No segfaults are thrown anymore(test was executed ~2000 times)

### Possible Drawbacks 

Inconsistency with `sendTx` method which still uses `boost::barrier`

### DIFF
--- a/test/framework/integration_framework/integration_test_framework.cpp
+++ b/test/framework/integration_framework/integration_test_framework.cpp
@@ -135,6 +135,7 @@ namespace integration_framework {
         ->getPeerCommunicationService()
         ->on_proposal()
         .subscribe([this](auto proposal) {
+          log_->info("Before push to proposal queue");
           proposal_queue_.push(proposal);
           log_->info("proposal");
           queue_cond.notify_all();
@@ -236,7 +237,9 @@ namespace integration_framework {
     log_->info("send transactions");
     const auto &transactions = tx_sequence.transactions();
 
-    boost::barrier bar(2);
+    std::mutex m;
+    std::condition_variable cv;
+    bool processed = false;
 
     // subscribe on status bus and save all stateless statuses into a vector
     std::vector<shared_model::proto::TransactionResponse> statuses;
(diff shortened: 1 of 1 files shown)


test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -279,7 +286,10 @@ namespace integration_framework {
         tx_list);
 
     // make sure that the first (stateless) status is come
-    bar.wait();
+    {
```
lebdron: Code block is redundant.

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -265,7 +268,11 @@ namespace integration_framework {
               statuses.push_back(*std::static_pointer_cast<
                                  shared_model::proto::TransactionResponse>(s));
             },
-            [&bar] { bar.wait(); });
+            [&cv, &m, &processed] {
+              std::lock_guard<std::mutex> lock(m);
```
lebdron: lock seems redundant here, what is its purpose? `processed` cannot be modified concurrently because on_complete is called once only.

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -279,7 +286,10 @@ namespace integration_framework {
         tx_list);
 
     // make sure that the first (stateless) status is come
-    bar.wait();
+    {
+      std::unique_lock<std::mutex> lk(m);
+      cv.wait(lk, [&] { return processed; });
```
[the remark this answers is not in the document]
lebdron: Why? This is equivalent to `while (!pred()) { wait(lock); }`, so it will unlock after the observable is completed.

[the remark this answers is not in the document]
l4l: http://www.cplusplus.com/reference/condition_variable/condition_variable/notify_all/
> If no threads are waiting, the function does nothing.

cv sucks in c++ :( that's why we've decided to use barrier

test/framework/integration_framework/integration_test_framework.cpp
kamilsa: According to https://en.cppreference.com/w/cpp/thread/condition_variable it is needed

PR #1688, opened 2018-08-28
Move common object builders to tests

lebdron:
### Description of the Change
Since builders are replaced with factories, the deprecated builders are moved to tests for further refactoring and removal.

### Benefits
Less hard-coded dependencies

### Possible Drawbacks 
None


irohad/consensus/yac/impl/yac_crypto_provider_impl.cpp
```diff
@@ -61,19 +63,17 @@ namespace iroha {
...
-                   });
+        factory_->createSignature(pubkey, signature)
+            .match(
+                [&](iroha::expected::Value<
+                    std::unique_ptr<shared_model::interface::Signature>> &sig) {
+                  vote.signature = std::move(sig.value);
+                },
+                [&](iroha::expected::Error<std::string> &reason) {
+                  logger::log("YacCryptoProvider::getVote")
+                      ->error("Cannot build vote signature: {}", reason.error);
```
l4l: That branch will probably lead to UB, since `vote.signature` is undefined

lebdron: Behavior is defined, it will result in segfault on the attempt to access `vote.signature` in next method, because the default value of shared pointer is nullptr. This pull request is not aimed at fixing the behavior, and since this error is logged, it is possible to fix it in another task.

irohad/consensus/yac/impl/yac_crypto_provider_impl.cpp
```diff
@@ -61,19 +63,17 @@ namespace iroha {
...
-                         ->error("Cannot build vote signature: {}",
-                                 *reason.error);
-                   });
+        factory_->createSignature(pubkey, signature)
+            .match(
+                [&](iroha::expected::Value<
+                    std::unique_ptr<shared_model::interface::Signature>> &sig) {
+                  vote.signature = std::move(sig.value);
+                },
+                [&](iroha::expected::Error<std::string> &reason) {
```
nickaleks: I believe you do not need the capture here

irohad/consensus/yac/impl/yac_crypto_provider_impl.hpp
```diff
@@ -1,33 +1,25 @@
...
+
 #include "cryptography/keypair.hpp"
+#include "interfaces/common_objects/common_objects_factory.hpp"
 
 namespace iroha {
   namespace consensus {
     namespace yac {
       class CryptoProviderImpl : public YacCryptoProvider {
        public:
         explicit CryptoProviderImpl(
```
nickaleks: Explicit can be removed

irohad/main/application.cpp
```diff
@@ -100,14 +100,14 @@ void Irohad::dropStorage() {
...
-  auto factory =
+  common_objects_factory_ =
       std::make_shared<shared_model::proto::ProtoCommonObjectsFactory<
           shared_model::validation::FieldValidator>>();
   auto block_converter =
       std::make_shared<shared_model::proto::ProtoBlockJsonConverter>();
   auto storageResult = StorageImpl::create(block_store_dir_,
                                            pg_conn_,
-                                           std::move(factory),
+                                           common_objects_factory_,
```
nickaleks: please keep std::move

test/module/irohad/multi_sig_transactions/transport_test.cpp
```diff
@@ -3,11 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include "multi_sig_transactions/transport/mst_transport_grpc.hpp"
+
 #include <gtest/gtest.h>
```
nickaleks: Please add empty line after this include

test/module/shared_model/factory_mocks.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IROHA_SHARED_MODEL_FACTORY_MOCKS_HPP
+#define IROHA_SHARED_MODEL_FACTORY_MOCKS_HPP
+
+#include <gmock/gmock.h>
+#include "interfaces/common_objects/common_objects_factory.hpp"
+
+struct MockCommonObjectsFactory
```
nickaleks: Do we need this as a part of this pr?

irohad/main/application.cpp
lebdron: This shared pointer is reused in several components, so it cannot be moved.

test/module/irohad/multi_sig_transactions/transport_test.cpp
lebdron: Google codestyle suggests the following order:
class header
empty line
system headers (e.g. vector, iostream)
empty line
other includes

PR #1690, opened 2018-08-29
Eliminate fields naming inconsistency in proto files 

igor-egorov:

### Description of the Change

Some fields in proto files were named inconsistently. That was causing the need of workarounds implementation on frontend JS side. This pr eliminates such inconsistency and applies required fixes to related sources.

### Benefits

Naming consistency inside proto files increased. No need anymore to do workarounds on frontend side. 

### Possible Drawbacks 

Unknown

### Usage Examples or Tests

```
./clean.sh
cmake -H. -Bbuild -DSWIG_JAVA=ON -DSWIG_PYTHON=ON
cd build
make
make python_test
make irohapy
make irohajava
make binary_test
ctest --output-on-failure
ctest -R python --output-on-failure
ctest -R java --output-on-failure
```


l4l:
Consider also `Signature::pubkey` -> `Signature::public_key`

test/module/shared_model/validators/field_validator_test.cpp
```diff
@@ -76,7 +64,7 @@ class FieldValidatorTest : public ValidatorsTest {
 
  public:
   FieldValidatorTest() {
-    for (const auto &field : {"public_key", "main_pubkey", "pubkey"}) {
+    for (const auto &field : {"public_key", "public_key", "pubkey"}) {
```
lebdron: One "public_key" should be enough here, because it populates a collection. Is "pubkey" used anywhere? It can be removed if it is unused.

test/module/shared_model/validators/validators_fixture.hpp
```diff
@@ -50,7 +50,7 @@ class ValidatorsTest : public ::testing::Test {
     for (const auto &id : {"account_id", "src_account_id"}) {
       field_setters[id] = setString(account_id);
     }
-    for (const auto &id : {"public_key", "main_pubkey"}) {
+    for (const auto &id : {"public_key", "public_key"}) {
```
lebdron: Same for one "public key".

PR #1694, opened 2018-08-29
Fix json escaping

l4l:
### Description of the Change

Double quotes need to be escaped (including user code) explicitly. This pr improves this by performing such actions internally. Couple of tests are also added

### Benefits

Cleaner interface

### Possible Drawbacks 

None

### DIFF
--- a/test/integration/acceptance/set_account_detail_test.cpp
+++ b/test/integration/acceptance/set_account_detail_test.cpp
@@ -67,6 +67,44 @@ TEST_F(SetAccountDetail, Self) {
 }
 
 /**
+ * @given a user with can_set_detail perm
+ *        AND a json-object value with single quotes
+ * @when execute tx with SetAccountDetail command aimed to the user
+ * @then there is the tx in proposal
+ */
+TEST_F(SetAccountDetail, JsonObject) {
+  IntegrationTestFramework(1)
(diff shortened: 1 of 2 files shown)


test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -66,6 +66,44 @@ TEST_F(SetAccountDetail, Self) {
...
+ * @when execute tx with SetAccountDetail command aimed to the user
+ * @then there is the tx in proposal
+ */
+TEST_F(SetAccountDetail, JsonObject) {
+  IntegrationTestFramework(1)
+      .setInitialState(kAdminKeypair)
+      .sendTx(makeUserWithPerms())
+      .skipProposal()
+      .skipBlock()
+      .sendTx(complete(baseTx(kUserId, kKey, "{'value' : ['abc', 'dca']}")))
```
igor-egorov: I suggest a bit stronger verification for both test cases:

Set the tested value as const at the beginning of a test case.
SendTx(Set the value as an account detail) - already done.
Query the detail.
ASSERT_EQ(const_detail, query_response_detail);

l4l:
Will be reworked with [new schema](https://gist.github.com/l4l/d9b278001587a3b98b0546e405d933bc)

PR #1695, opened 2018-08-29
Fix container timestamp validation

lebdron:
### Description of the Change
Since timestamps of transactions are required to be validated against block or proposal timestamp, a separate method which has additional timestamp argument is required, because nested field validator is inaccessible.
Timestamp validation is removed from proposal and block because they require timestamp validation, which differs from transactions.

### Benefits
No more "proposals from future".

### Possible Drawbacks 
Even more complex validators

### Usage Examples or Tests 
```
ctest -R container_validator_test
```


neewy:
@lebdron I would consider making a hotfix if this problem persists in master branch

igor-egorov:
`module_transport_builder_test` has failed on both platforms - Linux and Mac.

shared_model/validators/container_validator.hpp
```diff
@@ -58,11 +44,15 @@ namespace shared_model {
                 transactions_collection_validator),
             field_validator_(field_validator) {}
 
-      Answer validate(const Iface &cont, std::string reason_name) const {
+      template <typename Validator>
+      Answer validate(const Iface &cont,
+                      std::string reason_name,
```
Akvinikym: Shouldn't it be passed by const ref?

shared_model/validators/container_validator.hpp
```diff
@@ -71,6 +61,10 @@ namespace shared_model {
         return answer;
       }
 
+      Answer validate(const Iface &cont, std::string reason_name) const {
```
Akvinikym: The same, not const ref?

test/module/shared_model/builders/protobuf/transport_builder_test.cpp
```diff
@@ -111,40 +112,40 @@ class TransportBuilderTest : public ::testing::Test {
   auto getBaseBlockBuilder() {
     return BlockBuilder()
         .transactions(std::vector<Transaction>({createTransaction()}))
-        .height(1)
-        .prevHash(crypto::Hash("asd"));
+        .createdTime(created_time)
+        .height(1);
```
Akvinikym: Was there a reason to change order?

shared_model/validators/transaction_validator.hpp
```diff
@@ -294,6 +286,33 @@ namespace shared_model {
...
+       * @param tx - transaction to validate
+       * @return Answer containing found error if any
+       */
+      Answer validate(const interface::Transaction &tx) const {
+        return validateImpl(tx, [this](auto &reason, auto time) {
+          field_validator_.validateCreatedTime(reason, time);
+        });
+      }
+
+      Answer validate(const interface::Transaction &tx,
```
igor-egorov: Please add docstring.

shared_model/validators/field_validator.cpp
```diff
@@ -275,26 +263,31 @@ namespace shared_model {
 
     void FieldValidator::validateCreatedTime(
         ReasonsGroupType &reason,
-        const interface::types::TimestampType &timestamp) const {
-      iroha::ts64_t now = time_provider_();
-
-      if (now + future_gap_ < timestamp) {
+        interface::types::TimestampType timestamp,
+        interface::types::TimestampType current_timestamp) const {
```
igor-egorov: Please add docstrings for both `validateCreatedTime` methods. They are needed here more than elsewhere since the lists of methods arguments is quite similar and unclear at the same time.

lebdron:
@neewy Master branch does not contain the affected code, so a hotfix is not required.

shared_model/validators/signable_validator.hpp
```diff
@@ -43,6 +40,25 @@ namespace shared_model {
...
 
+     public:
+      explicit SignableModelValidator(
+          FieldValidator &&validator = FieldValidator())
+          : ModelValidator(validator), field_validator_(std::move(validator)) {}
+
+      Answer validate(const Model &model,
+                      interface::types::TimestampType current_timestamp) const {
+        return validateImpl(
+            model, [this, &current_timestamp](const auto &model) {
```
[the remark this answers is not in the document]
lebdron: After an offline discussion, it was agreed that alternative solutions require default arguments, which are not recommended to be used http://www.codingstandard.com/rule/8-3-3-do-not-use-default-arguments/

shared_model/validators/transaction_validator.hpp
```diff
@@ -294,6 +286,33 @@ namespace shared_model {
...
+          : field_validator_(field_validator),
+            command_validator_(command_validator) {}
+
+      /**
+       * Applies validation to given transaction
+       * @param tx - transaction to validate
+       * @return Answer containing found error if any
+       */
+      Answer validate(const interface::Transaction &tx) const {
+        return validateImpl(tx, [this](auto &reason, auto time) {
```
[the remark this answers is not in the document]
lebdron: Same with the default arguments

shared_model/validators/transactions_collection/transactions_collection_validator.cpp
```diff
@@ -70,19 +56,51 @@ namespace shared_model {
       return res;
     }
 
-    template <typename TransactionValidator, typename FieldValidator>
+    template <typename TransactionValidator>
+    Answer TransactionsCollectionValidator<TransactionValidator>::validate(
+        const shared_model::interface::types::TransactionsForwardCollectionType
+            &transactions) const {
+      return validateImpl(transactions, [this](const auto &tx) {
```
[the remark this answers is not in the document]
lebdron: Same with the default arguments

lebdron:
integration_pipeline_test is not deterministic, as it failed on iteration 29 on a local machine. It is not related to the pull request and will be fixed in another task.

PR #1698, opened 2018-08-30
Add AddSignatory acceptance test

l4l:
### Description of the Change

This pr introduce integration tests for AddSignature command 

### Benefits

Move coverage


test/integration/acceptance/add_signatory_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+              kUser2Id, interface::permissions::Grantable::kAddMySignatory)),
+          CHECK_BLOCK(1))
+      .sendTxAwait(complete(baseTx().creatorAccountId(kUser2Id).addSignatory(
+                                kUserId, kUser2Keypair.publicKey()),
+                            kUser2Keypair),
+                   CHECK_BLOCK(1));
+}
+
+/**
+ * C226 Add signatory to account, which haven't granted such permission
```
nickaleks: "did not grant" is more correct from here imo

test/integration/acceptance/add_signatory_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+              checkStatelessInvalid);
+}
+
+/**
+ * @given some user with CanAddMySignatory permission
+ * @when execute tx with AddSignatory with a valid key which isn't associated
+ *       with any user
+ * @then there is no tx in proposal
+ */
+TEST_F(AddSignatory, NonexistedKey) {
```
nickaleks: Maybe use "NonExistent" for correctness and to be similar to NonExistentUser test

test/integration/acceptance/add_signatory_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+  const std::string kRole2 = "roletwo";
+  const std::string kUser2 = "usertwo";
+  const std::string kUser2Id = kUser2 + "@test";
+  const std::string kAmount = "1.0";
+  const crypto::Keypair kUser2Keypair =
+      crypto::DefaultCryptoAlgorithmType::generateKeypair();
+};
+
+#define CHECK_BLOCK(i) \
```
nickaleks: There was no consensus about the usage of macros, maybe discuss it with the team before merging

test/integration/acceptance/add_signatory_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+  [](auto &block) { ASSERT_EQ(block->transactions().size(), i); }
+
+/**
+ * C224 Add existing public key of other user
+ * @given some user with CanAddSignatory permission and a second user
+ * @when execute tx with AddSignatory where the first is a creator and the
+ *       second's key is added as a signatory
+ * @then there is the tx in proposal
+ */
+TEST_F(AddSignatory, Basic) {
```
nickaleks: Maybe run a query to determine if the signatory was added?

test/integration/acceptance/add_signatory_test.cpp
l4l: Passive voice is required, also I would even change it to `isn't granted`

test/integration/acceptance/add_signatory_test.cpp
l4l: It was merged earlier, so it just copy-pasted from some other test
I'd better refactor it as a static function of AcceptanceFixture in future

test/integration/acceptance/add_signatory_test.cpp
l4l: added

PR #1701, opened 2018-08-31
Feature/prepared sql executors

victordrobny:
### Description of the Change
SQL statements in command executor were wrapped into prepared statements

### Benefits
Better performance. AddAssetQuantity benchmark became 2.5-3 times faster.


irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -36,7 +37,7 @@ namespace {
   }
 
   /**
-   * Transforms soci statement to CommandResult,
+   * Executes sql query,
```
nickaleks: If this function executes a query, maybe rename it to indicate that

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -528,227 +311,599 @@ namespace iroha {
...
+    }
+
+    CommandResult PostgresCommandExecutor::operator()(
+        const shared_model::interface::AddPeer &command) {
+      auto &peer = command.peer();
+
+      auto cmd = boost::format("EXECUTE %1% ('%2%', '%3%', '%4%')");
+
+      if (do_validation_) {
+        cmd = (cmd % "addAddPeerWithValidation");
```
nickaleks: addAddPeer?

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -21,6 +22,14 @@
 namespace iroha {
   namespace ametsuchi {
 
+    std::map<std::string, std::shared_ptr<soci::statement>>
+    &getPreparedStatements(
```
nickaleks: Does this function have a definition?

l4l: Looks like an internal function

test/module/iroha-cli/CMakeLists.txt
```diff
@@ -15,13 +15,14 @@
...
-target_link_libraries(client_test
-    client
-    processors
-    server_runner
-    query_execution
-    )
-target_include_directories(client_test PUBLIC
-    ${PROJECT_SOURCE_DIR}/iroha-cli
-    )
+#TODO: muratovv, IR-1589, fix compilation errors and pass tests
```
l4l: why it's not done in that pr?
Same for the rest tests

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -156,40 +162,18 @@ namespace {
...
-      auto &account_id = creator_account_id_;
-      auto &asset_id = command.assetId();
-      auto amount = command.amount().toStringRepr();
-      auto precision = command.amount().precision();
-      boost::format cmd(R"(
+    const std::string PostgresCommandExecutor::addAssetQuantityBase = R"(
+          PREPARE %s (text, text, int, text) AS
           WITH has_account AS (SELECT account_id FROM account
-                               WHERE account_id = :account_id LIMIT 1),
+                               WHERE account_id = $1 LIMIT 1),
```
l4l: Why named aliases replaced with anonymous ids? It's harder to read/undertstand the query and more error-prone

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1494,7 +1268,438 @@ namespace iroha {
...
+           % "addAssetQuantityWithValidation"
+           % (boost::format(R"(has_perm AS (%s),)")
+              % checkAccountRolePermission(
+                    shared_model::interface::permissions::Role::kAddAssetQty,
+                    "$1"))
+                 .str()
+           % "AND (SELECT * from has_perm)"
+           % "WHEN NOT (SELECT * from has_perm) THEN 1")
+              .str();
+      std::string addAssetQuantityWithoutValidation =
```
l4l: seems it can be `const static`

irohad/ametsuchi/impl/storage_impl.hpp
```diff
@@ -109,6 +109,8 @@ namespace iroha {
                       converter,
                   size_t pool_size);
 
+      void prepareStatements();
```
l4l: Why it cannot be static (passing connection pool & its size as args)
If the pervious point is positive, than it can removed from interface and have internal linkage

irohad/ametsuchi/impl/postgres_command_executor.cpp
nickaleks: Because postgres understands only positional arguments. Previously we substituted parameters in our code, now it is done as a prepared statement

l4l: Got it, I though it was soci args, thanks!

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -36,26 +44,25 @@ namespace {
...
    * construction in successful case.
-   * @return CommandResult with combined error message
-   * in case of result contains error
+   * @return CommandResult with command name and error message
    */
-  iroha::ametsuchi::CommandResult makeCommandResultByReturnedValue(
-      soci::statement &st,
+  iroha::ametsuchi::CommandResult executeQuery(
+      soci::session &sql,
+      std::string &&cmd,
```
muratovv: Why not use const ref instead?

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -67,53 +74,59 @@ namespace {
 
   std::string checkAccountRolePermission(
       shared_model::interface::permissions::Role permission,
-      const std::string &account_alias = "role_account_id") {
+      const std::string &account_alias = ":role_account_id") {
```
muratovv: `:` looks misleading. Maybe move it to separate constant and make concat here?

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -67,53 +74,59 @@ namespace {
...
     const auto perm_str =
         shared_model::interface::RolePermissionSet({permission}).toBitstring();
     const auto bits = shared_model::interface::RolePermissionSet::size();
     std::string query = (boost::format(R"(
           SELECT COALESCE(bit_or(rp.permission), '0'::bit(%1%))
```
muratovv: Is it possible to move all SQLs to separate files and load it into the runtime. In this case we will more flexible in terms of business logic right now.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -67,53 +74,59 @@ namespace {
...
+              WHERE ar.account_id = %3%)")
                          % bits % perm_str % account_alias)
                             .str();
     return query;
   }
 
   std::string checkAccountGrantablePermission(
-      shared_model::interface::permissions::Grantable permission) {
+      shared_model::interface::permissions::Grantable permission,
+      std::string creator_id = ":grantable_permittee_account_id",
```
muratovv: Same issue about `:` and further.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -67,53 +74,59 @@ namespace {
...
+      std::string account_id = ":grantable_account_id") {
     const auto perm_str =
         shared_model::interface::GrantablePermissionSet({permission})
             .toBitstring();
     const auto bits = shared_model::interface::GrantablePermissionSet::size();
     std::string query = (boost::format(R"(
```
muratovv: Same issue about fetching SQLs here and further.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -528,228 +318,593 @@ namespace iroha {
...
+        const shared_model::interface::AddAssetQuantity &command) {
+      auto &account_id = creator_account_id_;
+      auto &asset_id = command.assetId();
+      auto amount = command.amount().toStringRepr();
+      int precision = command.amount().precision();
+
+      auto cmd = boost::format("EXECUTE %1% ('%2%', '%3%', %4%, '%5%')");
+
+      if (do_validation_) {
+        cmd = (cmd % "addAssetQuantityWithValidation");
```
muratovv: Seems this code is duplicated many times. I guess better to refactor it with separate function.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -528,228 +318,593 @@ namespace iroha {
...
+            return std::string("Asset with given precision does not exist");
+          },
+          [] { return std::string("Summation overflows uint256"); },
+      };
+      return executeQuery(sql_, cmd.str(), "AddAssetQuantity", message_gen);
+    }
+
+    CommandResult PostgresCommandExecutor::operator()(
+        const shared_model::interface::AddPeer &command) {
+      auto &peer = command.peer();
```
muratovv: Each execute functions contains steps:
1. convert params to string
2. prepare execture string
3. add handlers
4. combine all previous.
Maybe we may create one function which will include all those steps and operator() will pass correct params to this function?

irohad/ametsuchi/impl/postgres_command_executor.cpp
nickaleks: Added TODO

irohad/ametsuchi/impl/postgres_command_executor.cpp
nickaleks: Added TODO

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -28,6 +29,42 @@
...
+    std::string command_name;
+    std::string command_base;
+    std::vector<std::string> permission_checks;
+  };
+
+  std::pair<std::string, std::string> compileStatement(
+      const PreparedStatement &statement) {
+    auto initial = boost::format(statement.command_base);
+
+    auto with_validation =
```
l4l: Would be nice to comment most of the actions in that function, since it very generic, thus it can be easily misused

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -28,6 +29,42 @@
...
 namespace {
+  struct PreparedStatement {
+    std::string command_name;
+    std::string command_base;
+    std::vector<std::string> permission_checks;
+  };
+
+  std::pair<std::string, std::string> compileStatement(
+      const PreparedStatement &statement) {
+    auto initial = boost::format(statement.command_base);
```
l4l: I think this variable value can be inlined twice, so the code is easier to read

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1494,7 +1251,348 @@ namespace iroha {
...
+
+      prepareStatement(sql, grant_permission);
+
+      PreparedStatement remove_signatory{
+          "removeSignatory",
+          removeSignatoryBase,
+          {(boost::format(R"(
+          has_perm AS (%s),
+          get_account AS (
+              SELECT quorum FROM account WHERE account_id = :account_id LIMIT 1
```
lebdron: Since there is no more `soci::use`, please check that this parameter is used.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1494,7 +1251,348 @@ namespace iroha {
...
+           ),
+          get_signatories AS (
+              SELECT public_key FROM account_has_signatory
+              WHERE account_id = :account_id
```
lebdron: Same as above.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -67,57 +108,63 @@ namespace {
...
+                         % creator_id % account_id)
                             .str();
     return query;
   }
 
   std::string checkAccountHasRoleOrGrantablePerm(
       shared_model::interface::permissions::Role role,
-      shared_model::interface::permissions::Grantable grantable) {
+      shared_model::interface::permissions::Grantable grantable,
+      const std::string &creator_id,
```
lebdron: std::string -> shared_model::interface::types::AccountIdType

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -67,57 +108,63 @@ namespace {
...
+              WHERE ar.account_id = %3%)")
                          % bits % perm_str % account_alias)
                             .str();
     return query;
   }
 
   std::string checkAccountGrantablePermission(
-      shared_model::interface::permissions::Grantable permission) {
+      shared_model::interface::permissions::Grantable permission,
+      const std::string &creator_id,
```
lebdron: std::string -> shared_model::interface::types::AccountIdType

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -67,57 +108,63 @@ namespace {
 
   std::string checkAccountRolePermission(
       shared_model::interface::permissions::Role permission,
-      const std::string &account_alias = "role_account_id") {
+      const std::string &account_alias) {
```
lebdron: std::string -> shared_model::interface::types::AccountIdType
also, why "alias" instead of "id"?

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -151,45 +198,31 @@ namespace {
             % permittee)
         .str();
   }
+
+  template <typename Format>
+  void appendCommandName(const std::string &name,
+                         Format &cmd,
+                         bool do_validation) {
+    auto command_name =
+        name + (do_validation ? "WithValidation" : "WithoutValidation");
```
lebdron: It looks like "WithValidation" and "WithoutValidation" should be defined as constants in struct PreparedStatement, so that they can be accessed here.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -528,230 +360,567 @@ namespace iroha {
...
+      };
+      return executeQuery(sql_, cmd.str(), "AppendRole", message_gen);
+    }
+
+    CommandResult PostgresCommandExecutor::operator()(
+        const shared_model::interface::CreateAccount &command) {
+      auto &account_name = command.accountName();
+      auto &domain_id = command.domainId();
+      auto &pubkey = command.pubkey().hex();
+      std::string account_id = account_name + "@" + domain_id;
```
lebdron: std::string -> shared_model::interface::types::AccountIdType

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1494,7 +1251,348 @@ namespace iroha {
...
+                SELECT COALESCE(bit_or(rp.permission), '0'::bit(%2%)) &
+                    (SELECT * FROM role_permissions) =
+                    (SELECT * FROM role_permissions)
+                FROM role_has_permissions AS rp
+                JOIN account_has_roles AS ar on ar.role_id = rp.role_id
+                WHERE ar.account_id = $1
+            ),)")
+            % checkAccountRolePermission(
+                  shared_model::interface::permissions::Role::kAppendRole, "$1")
+            % std::to_string(bits))
```
lebdron: to_string should not be necessary.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1494,7 +1251,348 @@ namespace iroha {
...
+          createRoleBase,
+          {(boost::format(R"(
+          account_has_role_permissions AS (
+                SELECT COALESCE(bit_or(rp.permission), '0'::bit(%s)) &
+                    $3 = $3
+                FROM role_has_permissions AS rp
+                JOIN account_has_roles AS ar on ar.role_id = rp.role_id
+                WHERE ar.account_id = $1),
+          has_perm AS (%s),)")
+            % std::to_string(bits)
```
lebdron: same for to_string, please check.

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1494,7 +1251,348 @@ namespace iroha {
...
+          {(boost::format(R"(
+            has_perm AS (%s),)")
+            % checkAccountRolePermission(
+                  shared_model::interface::permissions::Role::kDetachRole,
+                  "$1"))
+               .str(),
+           R"(AND (SELECT * FROM has_perm))",
+           R"(WHEN NOT (SELECT * FROM has_perm) THEN 1)"}};
+
+      prepareStatement(sql, detach_role);
```
lebdron: Maybe create a vector and do emplace_back instead of declaring variables? A loop can be used to prepare all the statements then.

PR #1703, opened 2018-08-31
BFT OS components

lebdron:
### Description of the Change
Includes the following changes:
- YAC refactoring with states instead of vote, commit, and reject
- Transaction batch factory separated from transaction batch
- BFT OS components: On-demand ordering service, client and server gRPC transport, connection manager, and ordering gate

### Benefits
One step towards CFT and BFT

### Possible Drawbacks 
None


irohad/consensus/yac/impl/yac_crypto_provider_impl.hpp
```diff
@@ -21,11 +21,7 @@ namespace iroha {
             std::shared_ptr<shared_model::interface::CommonObjectsFactory>
                 factory);
 
-        bool verify(CommitMessage msg) override;
-
-        bool verify(RejectMessage msg) override;
-
-        bool verify(VoteMessage msg) override;
+        bool verify(std::vector<VoteMessage> msg) override;
```
kamilsa: why not const ref?

irohad/consensus/yac/transport/impl/network_impl.hpp
```diff
@@ -18,6 +18,7 @@
 #ifndef IROHA_NETWORK_IMPL_HPP
 #define IROHA_NETWORK_IMPL_HPP
 
+#include <consensus/yac/messages.hpp>
```
kamilsa: use `""`

PR #1707, opened 2018-09-05
Fix failure after empty block

kamilsa:
### Description of the Change

This PR fixes a bug of failure of transactions commitment after the situation where simulator returns empty block. Also this is the first step of removing empty block entity from the iroha

Now simulator always produces Block instead of BlockVariant

### Benefits

Now we can execute transactions after empty block

### Possible Drawbacks 

None

### Usage Examples or Tests

New test checking successful commitment was added to pipeline test

### DIFF
--- a/irohad/main/application.cpp
+++ b/irohad/main/application.cpp
@@ -184,6 +184,12 @@ void Irohad::initOrderingGate() {
  */
 void Irohad::initSimulator() {
   auto block_factory = std::make_unique<shared_model::proto::ProtoBlockFactory>(
+      //  Block factory in simulator uses UnsignedBlockValidator because it is
+      //  not required to check signatures of block here, as they will be
+      //  checked when supermajority of peers will sign the block. It is also
+      //  not required to validate signatures of transactions here because they
+      //  are validated in the ordering gate, where they are received from the
+      //  ordering service.
       std::make_unique<
           shared_model::validation::DefaultUnsignedBlockValidator>());
   simulator = std::make_shared<Simulator>(ordering_gate,
--- a/irohad/main/application.cpp
+++ b/irohad/main/application.cpp
@@ -184,7 +184,8 @@ void Irohad::initOrderingGate() {
  */
 void Irohad::initSimulator() {
   auto block_factory = std::make_unique<shared_model::proto::ProtoBlockFactory>(
-      std::make_unique<shared_model::validation::BlockVariantValidator>());
+      std::make_unique<
+          shared_model::validation::DefaultUnsignedBlockValidator>());
   simulator = std::make_shared<Simulator>(ordering_gate,
                                           stateful_validator,
                                           storage,
--- a/irohad/main/impl/block_loader_init.cpp
+++ b/irohad/main/impl/block_loader_init.cpp
@@ -26,26 +26,26 @@ using namespace iroha::network;
 auto BlockLoaderInit::createService(
     std::shared_ptr<BlockQueryFactory> block_query_factory,
     std::shared_ptr<consensus::ConsensusResultCache> consensus_result_cache) {
-  return std::make_shared<BlockLoaderService>(block_query_factory,
-                                              std::move(consensus_result_cache));
+  return std::make_shared<BlockLoaderService>(
+      block_query_factory, std::move(consensus_result_cache));
 }
 
 auto BlockLoaderInit::createLoader(
     std::shared_ptr<PeerQueryFactory> peer_query_factory,
     std::shared_ptr<BlockQueryFactory> block_query_factory) {
   shared_model::proto::ProtoBlockFactory factory(
-      std::make_unique<shared_model::validation::BlockVariantValidator>());
-  return std::make_shared<BlockLoaderImpl>(peer_query_factory,
-                                           block_query_factory,
-                                           std::move(factory));
+      std::make_unique<
+          shared_model::validation::DefaultSignedBlockValidator>());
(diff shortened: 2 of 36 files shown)


irohad/main/application.cpp
```diff
@@ -184,7 +184,8 @@ void Irohad::initOrderingGate() {
  */
 void Irohad::initSimulator() {
   auto block_factory = std::make_unique<shared_model::proto::ProtoBlockFactory>(
-      std::make_unique<shared_model::validation::BlockVariantValidator>());
+      std::make_unique<
+          shared_model::validation::DefaultUnsignedBlockValidator>());
```
lebdron: Please make a comment here that "unsigned" also applies for the transaction signatures, besides the block signatures. It is not required to validate them here because they are validated in the ordering gate, where they are received from the ordering service.

irohad/main/impl/block_loader_init.cpp
```diff
@@ -26,26 +26,26 @@ using namespace iroha::network;
...
 auto BlockLoaderInit::createLoader(
     std::shared_ptr<PeerQueryFactory> peer_query_factory,
     std::shared_ptr<BlockQueryFactory> block_query_factory) {
   shared_model::proto::ProtoBlockFactory factory(
-      std::make_unique<shared_model::validation::BlockVariantValidator>());
-  return std::make_shared<BlockLoaderImpl>(peer_query_factory,
-                                           block_query_factory,
-                                           std::move(factory));
+      std::make_unique<
+          shared_model::validation::DefaultSignedBlockValidator>());
```
lebdron: https://github.com/hyperledger/iroha/blob/8883ac9000730229cd49a67c18bd993f2293fa08/shared_model/validators/default_validator.hpp#L108-L120

While it seems OK not to check the transaction signatures, because it will be checked later that the supermajority of peers have signed the block, and therefore transaction signatures should not be forged, a comment can be useful here.

test/integration/pipeline/pipeline_test.cpp
```diff
@@ -193,3 +200,33 @@ TEST(PipelineIntegrationTest, SendTxSequenceAwait) {
...
       .done();
 }
+
+/**
+ * Check that after no transactions were committed we are able to send and apply
+ * new transactions
+ *
+ * @given createFirstDomain and createSecondDomain transactions
+ * @when first domain is created second time
+ * @then block with no transactions is created @and after that
```
lebdron: Other tests use slightly different style, with "AND" on a new line. Maybe use it here too?

test/integration/pipeline/pipeline_test.cpp
```diff
@@ -24,24 +24,69 @@
 #include "framework/batch_helper.hpp"
 #include "framework/integration_framework/integration_test_framework.hpp"
 #include "framework/specified_visitor.hpp"
+#include "integration/acceptance/acceptance_fixture.hpp"
 #include "utils/query_error_response_visitor.hpp"
 
-constexpr auto kAdmin = "admin@test";
-const shared_model::crypto::Keypair kAdminKeypair =
-    shared_model::crypto::DefaultCryptoAlgorithmType::generateKeypair();
+using ::testing::IsEmpty;
```
lebdron: Please remove, because it is unused.

test/module/irohad/network/block_loader_test.cpp
```diff
@@ -250,24 +253,20 @@ TEST_F(BlockLoaderTest, ValidWhenMultipleBlocks) {
...
-      getBaseBlockBuilder().build().signAndAddSignature(key).finish()));
+  auto block = std::make_shared<shared_model::proto::Block>(
+      getBaseBlockBuilder().build().signAndAddSignature(key).finish());
+  auto requested = wrapBlock(block);
   block_cache->insert(requested);
 
   EXPECT_CALL(*peer_query, getLedgerPeers())
       .WillOnce(Return(std::vector<wPeer>{peer}));
-  EXPECT_CALL(*validator, validate(*requested)).WillOnce(Return(Answer{}));
+  EXPECT_CALL(*validator, validate(RefAndPointerEq(block)))
```
lebdron: Is it possible to write `validate(*block)` instead, so that `RefAndPointerEq` is not required?

kamilsa: Then it tries to copy abstract block

shared_model/validators/block_variant_validator.hpp
```diff
@@ -20,7 +20,8 @@ namespace shared_model {
         : public shared_model::validation::AbstractValidator<
               shared_model::interface::BlockVariant> {
      public:
-      Answer validate(const shared_model::interface::BlockVariant &m) override {
+      Answer validate(
```
nickaleks: Do we still need block variant validator?

kamilsa: It will be removed in next PR as soon as EmptyBlock will be removed completely

irohad/simulator/impl/simulator.cpp
```diff
@@ -110,15 +110,16 @@ namespace iroha {
         log_->error("Unable to query top block height");
         return;
       }
-      auto block = block_factory_->unsafeCreateBlock(height,
-                                                     last_block->hash(),
-                                                     proposal.createdTime(),
-                                                     proposal.transactions());
-      crypto_signer_->sign(block);
+      std::shared_ptr<shared_model::interface::Block> block =
```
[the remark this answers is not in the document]
kamilsa: yes, does not compile otherwise

PR #1709, opened 2018-09-06
Add test for running Iroha with --overwrite-ledger flag

vadimreutskiy:
Signed-off-by: Vadim Reutskiy <reutskiy@soramitsu.co.jp>

### Description of the Change
Added two tests:
1. Test, which starts Iroha, adds transaction, restarts with --overwrite-ledger flags and checks that there is only one block in blockstorage and Iroha still can add transactions.
2. Test, which starts Iroha, adds transaction, restarts Iroha checks that amount of blocks in blockstorage is the same as before restart and Iroha still can add transactions.

### Benefits
Now system tests can check correctness of restarting Iroha both with keeping blocks in blockstorage and with wiping all data and rewriting ledger from the beginning.

### Possible Drawbacks 
None

Target: irohad_test

### DIFF
--- a/test/system/irohad_test.cpp
+++ b/test/system/irohad_test.cpp
@@ -63,9 +63,34 @@ class IrohadTest : public AcceptanceFixture {
   }
 
   void launchIroha() {
-    iroha_process_.emplace(irohad_executable.string() + setDefaultParams());
+    launchIroha(setDefaultParams());
+  }
+
+  void launchIroha(std::string parameters) {
+    iroha_process_.emplace(irohad_executable.string() + parameters);
     std::this_thread::sleep_for(kTimeout);
     ASSERT_TRUE(iroha_process_->running());
+  }
+
+  void launchIroha(const boost::optional<std::string> &config_path,
+                   const boost::optional<std::string> &genesis_block,
+                   const boost::optional<std::string> &keypair_path,
+                   const boost::optional<std::string> &additional_params) {
+    launchIroha(
+        params(config_path, genesis_block, keypair_path, additional_params));
+  }
+
+  int getBlockCount() {
+    int block_count = 0;
+
+    for (directory_iterator itr(blockstore_path_); itr != directory_iterator();
+         ++itr) {
+      if (is_regular_file(itr->path())) {
+        ++block_count;
+      }
+    }
+
+    return block_count;
   }
 
   void TearDown() override {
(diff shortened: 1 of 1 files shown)


test/system/irohad_test.cpp
```diff
@@ -80,16 +105,49 @@ class IrohadTest : public AcceptanceFixture {
...
   }
 
   std::string setDefaultParams() {
-    return params(config_copy_, path_genesis_.string(), path_keypair_.string());
+    return params(
+        config_copy_, path_genesis_.string(), path_keypair_.string(), {});
+  }
+
+  iroha::protocol::ToriiResponse sendDefaultTx(
+      boost::optional<shared_model::crypto::Keypair> key_pair) {
```
Akvinikym: You don't check, if this optional contains value inside the function. Consider either passing raw value or adding the check.
Also, I think, in both cases argument should be passed by const ref

test/system/irohad_test.cpp
```diff
@@ -80,16 +105,49 @@ class IrohadTest : public AcceptanceFixture {
...
+
+    auto tx = complete(baseTx(kAdminId).setAccountQuorum(kAdminId, 1),
+                       key_pair.get());
+    tx_request.set_tx_hash(shared_model::crypto::toBinaryString(tx.hash()));
+
+    auto client = torii::CommandSyncClient(kAddress, kPort);
+    client.Torii(tx.getTransport());
+
+    auto resub_counter(resubscribe_attempts);
+    const auto committed_status = iroha::protocol::TxStatus::COMMITTED;
```
Akvinikym: As I know, nearly always it's better to use `constexpr`, if possible

test/system/irohad_test.cpp
```diff
@@ -80,16 +105,49 @@ class IrohadTest : public AcceptanceFixture {
...
+      std::this_thread::sleep_for(resubscribe_timeout);
+      client.Status(tx_request, torii_response);
+    } while (torii_response.tx_status() != committed_status
+             and --resub_counter);
+
+    return torii_response;
+  }
+
+  void sendDefaultTxAndCheck(
+      boost::optional<shared_model::crypto::Keypair> key_pair) {
```
Akvinikym: The same

test/system/irohad_test.cpp
```diff
@@ -224,3 +267,63 @@ TEST_F(IrohadTest, SendQuery) {
...
+              std::string("--overwrite-ledger"));
+
+  ASSERT_EQ(getBlockCount(), 1);
+
+  SCOPED_TRACE("From restart with --overwrite-ledger flag test");
+  sendDefaultTxAndCheck(key_pair);
+}
+
+/**
+ * Test verifies that after common restart of Iroha it will keep height
```
Akvinikym: Not clear, what does the word "common" mean. Maybe, "usual" will suit better?
Also, please add some details about height, which should be kept - again, for now it's not crystal clear

test/system/irohad_test.cpp
```diff
@@ -80,16 +105,49 @@ class IrohadTest : public AcceptanceFixture {
...
     return res;
   }
 
   std::string setDefaultParams() {
-    return params(config_copy_, path_genesis_.string(), path_keypair_.string());
+    return params(
+        config_copy_, path_genesis_.string(), path_keypair_.string(), {});
+  }
+
+  iroha::protocol::ToriiResponse sendDefaultTx(
```
igor-egorov: Please add the docstring for the method.
Among the rest, it should denote that the method tries to retrieve tx status several times until `COMMITTED` will be received.

test/system/irohad_test.cpp
igor-egorov: I also suggest passing keypair as `const shared_model::crypto::Keypair &key_pair` since we use `boost::optional` when the filed may be missing.

test/system/irohad_test.cpp
igor-egorov: My suggestion is to omit technical details about blockchain height. 
In my view, it would be better to describe the case from the viewpoint of the ability to accept and serve new transactions.

given an Iroha with some transactions committed ontop of the genesis block
when the Iroha is restarted without --overwrite-ledger flag
then the state is successfully restored, the Iroha accepts and able to commit new transactions

neewy:
why this is not merged yet??

PR #1712, opened 2018-09-07
Fix/json escaping

l4l:
Reworked #1694


l4l:
Decided to delay this issue, since `DetailEntry` need to be separated before and it's not that trivial.

l4l:
Outdated, eventually will be reworked

PR #1715, opened 2018-09-10
Fix/bash params

l4l:
### Description of the Change

Relates to f5705f2

* fixes bash params
* use uniformly env vars for setting shell


shared_model/clean.sh
```diff
@@ -1,3 +1,3 @@
-#!/bin/bash
+#!/usr/bin/env bash
 rm -rf external
```
Akvinikym: Maybe add empty line between 1 and 2?

scripts/build-local-image.sh
```diff
@@ -1,4 +1,5 @@
-#!/bin/sh -x
+#!/usr/bin/env bash
+set -x
```
Akvinikym: Extra empty line?

PR #1716, opened 2018-09-11
Move out transport logic from CommandService 

l4l:
### Description of the Change

Before, `CommandService` has messed up logic, both for transport and main logic. This pr introduces `CommandService`, `CommandServiceImpl` and `CommandServiceTransport`. First one is an internal interface and the second its implementation. The last one serve for transport handling (timeouts, stateless checks, endpoint)

### Benefits

Unit test can be easily written

### Possible Drawbacks 

None?

### Alternate Designs

It seems that CommandService API may be not that comfy, feel free to propose improvals

### DIFF
--- a/irohad/torii/impl/command_service_transport_grpc.cpp
+++ b/irohad/torii/impl/command_service_transport_grpc.cpp
@@ -57,15 +57,12 @@ namespace torii {
       }
 
       std::string folded_hashes =
-          std::accumulate(tx_hashes.begin(),
+          std::accumulate(std::next(tx_hashes.begin()),
                           tx_hashes.end(),
-                          std::string(),
+                          tx_hashes[0].hex(),
                           [](auto &&acc, const auto &h) -> std::string {
-                            return acc + h.hex() + ", ";
+                            return acc + ", " + h.hex();
                           });
-
-      // remove leading ", "
-      folded_hashes.resize(folded_hashes.size() - 2);
 
       return (boost::format(
                   "Stateless invalid tx in transaction sequence, error: %s\n"
@@ -124,11 +121,11 @@ namespace torii {
       iroha::protocol::ToriiResponse *response) {
     auto status = command_service_->getStatus(
         shared_model::crypto::Hash(request->tx_hash()));
-    response->set_tx_hash(
-        shared_model::crypto::toBinaryString(status->transactionHash()));
-    response->set_error_message(std::move(status->errorMessage()));
-    response->set_tx_status(
-        static_cast<iroha::protocol::TxStatus>(status->get().which()));
+    *response =
+        std::static_pointer_cast<shared_model::proto::TransactionResponse>(
+            command_service_->getStatus(
+                shared_model::crypto::Hash(request->tx_hash())))
+            ->getTransport();
     return grpc::Status::OK;
   }
 
--- a/irohad/torii/impl/command_service_transport_grpc.hpp
+++ b/irohad/torii/impl/command_service_transport_grpc.hpp
@@ -40,7 +40,7 @@ namespace torii {
      * @param context - call context (see grpc docs for details)
      * @param request - transaction received
      * @param response - no actual response (grpc stub for empty answer)
-     * @return - grpc::Status
+     * @return status
      */
     grpc::Status Torii(grpc::ServerContext *context,
                        const iroha::protocol::Transaction *request,
@@ -51,7 +51,7 @@ namespace torii {
      * @param context - call context (see grpc docs for details)
      * @param request - list of transactions received
      * @param response - no actual response (grpc stub for empty answer)
-     * @return - grpc::Status
+     * @return status
      */
     grpc::Status ListTorii(grpc::ServerContext *context,
                            const iroha::protocol::TxList *request,
@@ -64,7 +64,7 @@ namespace torii {
      * uniquely
      * @param response - ToriiResponse which contains a current state of
      * requested transaction
-     * @return - grpc::Status
+     * @return status
      */
     grpc::Status Status(grpc::ServerContext *context,
                         const iroha::protocol::TxStatusRequest *request,
@@ -77,7 +77,7 @@ namespace torii {
      * uniquely
      * @param response_writer - grpc::ServerWriter which can repeatedly send
      * transaction statuses back to the client
-     * @return - grpc::Status
+     * @return status
      */
     grpc::Status StatusStream(grpc::ServerContext *context,
                               const iroha::protocol::TxStatusRequest *request,
--- a/test/module/irohad/torii/torii_service_test.cpp
+++ b/test/module/irohad/torii/torii_service_test.cpp
@@ -27,6 +27,7 @@ constexpr size_t TimesToriiBlocking = 5;
 using ::testing::_;
 using ::testing::A;
 using ::testing::AtLeast;
+using ::testing::HasSubstr;
 using ::testing::Return;
 
 using namespace iroha::network;
@@ -609,9 +610,8 @@ TEST_F(ToriiServiceTest, FailedListOfTxs) {
         ASSERT_EQ(toriiResponse.tx_status(),
                   iroha::protocol::TxStatus::STATELESS_VALIDATION_FAILED);
         auto msg = toriiResponse.error_message();
-        ASSERT_EQ(
-            msg.find("Stateless invalid tx in transaction sequence, error: "),
-            0);
+        ASSERT_THAT(toriiResponse.error_message(),
+                    HasSubstr("bad timestamp: sent from future"));
         ASSERT_NE(msg.find(hash.hex()), std::string::npos);
       });
 }


irohad/torii/command_service.hpp
```diff
@@ -6,105 +6,42 @@
...
-     * @param request - list of transactions received
-     * @param response - no actual response (grpc stub for empty answer)
-     * @return - grpc::Status
-     */
-    virtual grpc::Status ListTorii(grpc::ServerContext *context,
-                                   const iroha::protocol::TxList *request,
-                                   google::protobuf::Empty *response) override;
+    virtual void handleTransactionList(
+        const shared_model::interface::TransactionSequence &tx_list) = 0;
+    virtual void handleTransactionListError(
```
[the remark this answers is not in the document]
Akvinikym: Also, there are no docs about it

irohad/torii/command_service.hpp
```diff
@@ -6,105 +6,42 @@
...
-    void Status(const iroha::protocol::TxStatusRequest &request,
-                iroha::protocol::ToriiResponse &response);
-
-    /**
-     * Status call via grpc
-     * @param context - call context
-     * @param request - TxStatusRequest object which identifies transaction
-     * uniquely
-     * @param response - ToriiResponse which contains a current state of
+     * @param response - TransactionResponse which contains a current state of
```
Akvinikym: I assume, it's `@return`, not `@param`

irohad/torii/command_service.hpp
```diff
@@ -6,105 +6,42 @@
...
      * requested transaction
-     * @return - grpc::Status
      */
-    virtual grpc::Status Status(
-        grpc::ServerContext *context,
-        const iroha::protocol::TxStatusRequest *request,
-        iroha::protocol::ToriiResponse *response) override;
+    virtual std::shared_ptr<shared_model::interface::TransactionResponse>
+    getStatus(const shared_model::crypto::Hash &request) = 0;
```
Akvinikym: Can't it be `const` method?

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+namespace torii {
+
+  CommandServiceImpl::CommandServiceImpl(
+      std::shared_ptr<iroha::torii::TransactionProcessor> tx_processor,
+      std::shared_ptr<iroha::ametsuchi::Storage> storage,
+      std::shared_ptr<iroha::torii::StatusBus> status_bus)
+      : tx_processor_(tx_processor),
+        storage_(storage),
+        status_bus_(status_bus),
```
Akvinikym: Maybe, you should move `tx_processor`, `storage` and `status_bus` here?

irohad/torii/impl/command_service_impl.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include "backend/protobuf/transaction_responses/proto_tx_response.hpp"
+#include "cache/cache.hpp"
+#include "cryptography/hash.hpp"
+#include "logger/logger.hpp"
+#include "torii/processor/transaction_processor.hpp"
+#include "torii/status_bus.hpp"
+
+namespace torii {
+  /**
+   * Actual implementation of sync CommandServiceImpl.
```
Akvinikym: Shouldn't that be "of sync CommandService"?

irohad/torii/impl/command_service_transport_grpc.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "endpoint.pb.h"
+#include "logger/logger.hpp"
+
+namespace torii {
+  class CommandServiceTransportGrpc
+      : public iroha::protocol::CommandService::Service {
+   public:
+    /**
+     * Creates a new instance of CommandServiceTransportGrpc
+     * @param tx_processor - processor of received transactions
```
Akvinikym: Not used anymore

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -0,0 +1,226 @@
...
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "torii/impl/command_service_impl.hpp"
+
+#include <thread>
+
+#include "ametsuchi/block_query.hpp"
+#include "builders/protobuf/transaction_responses/proto_transaction_status_builder.hpp"
```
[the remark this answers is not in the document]
l4l: For now response cannot be built otherwise.

irohad/torii/command_service.hpp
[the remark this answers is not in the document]
l4l: Agreed, then there's two options:
- StatusBus need to be shared to CommandServiceTransportGrpc as well
- Make CommandService::pushStatus public

I like the first one, what do you think

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -171,12 +173,11 @@ namespace integration_framework {
...
       std::function<void(const shared_model::proto::TransactionResponse &)>
           validation) {
-    iroha::protocol::TxStatusRequest request;
-    request.set_tx_hash(shared_model::crypto::toBinaryString(hash));
-    iroha::protocol::ToriiResponse response;
-    iroha_instance_->getIrohaInstance()->getCommandService()->Status(request,
-                                                                     response);
-    validation(shared_model::proto::TransactionResponse(std::move(response)));
+    auto response =
+        iroha_instance_->getIrohaInstance()->getCommandService()->getStatus(
```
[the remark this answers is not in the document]
l4l: We probably need to add a todo for covering that test case

irohad/torii/command_service.hpp
l4l: Nope, there's logging at least and probably some other mutable components present

irohad/torii/impl/command_service_impl.cpp
l4l: not a big deal

irohad/torii/command_service.hpp
Akvinikym: But logging does not change state of the object, so `const`-ness is possible, if only that is the issue

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+      std::string folded_hashes =
+          std::accumulate(tx_hashes.begin(),
+                          tx_hashes.end(),
+                          std::string(),
+                          [](auto &&acc, const auto &h) -> std::string {
+                            return acc + h.hex() + ", ";
+                          });
+
+      // remove leading ", "
```
lebdron: Please use the following approach to avoid any removals:
```C++
std::string s = std::accumulate(std::next(v.begin()), v.end(),
                                std::to_string(v[0]), // start with first element
                                [](std::string a, int b) {
                                    return a + '-' + std::to_string(b);
                                });
```

test/module/irohad/torii/torii_service_test.cpp
```diff
@@ -612,11 +608,12 @@ TEST_F(ToriiServiceTest, FailedListOfTxs) {
...
                  and --resub_counter);
 
-        auto error_beginning = toriiResponse.error_message().substr(
-            0, toriiResponse.error_message().find_first_of('.'));
-
         ASSERT_EQ(toriiResponse.tx_status(),
                   iroha::protocol::TxStatus::STATELESS_VALIDATION_FAILED);
-        ASSERT_EQ(error_beginning, error_msg_beginning);
+        auto msg = toriiResponse.error_message();
+        ASSERT_EQ(
```
lebdron: Please use the following instead, it looks more obvious:
https://github.com/hyperledger/iroha/blob/ec9c10acf4273cd49d857a77fa84cc4001ea8b68/test/module/shared_model/validators/container_validator_test.cpp#L104-L105

irohad/torii/impl/command_service_transport_grpc.hpp
```diff
@@ -0,0 +1,97 @@
...
+        std::chrono::milliseconds nonfinal_timeout,
+        std::shared_ptr<shared_model::interface::TxStatusFactory>
+            status_factory);
+
+    /**
+     * Torii call via grpc
+     * @param context - call context (see grpc docs for details)
+     * @param request - transaction received
+     * @param response - no actual response (grpc stub for empty answer)
+     * @return - grpc::Status
```
lebdron: This line does not provide any useful information to the reader, same for methods below. Consider removing, or adding information about possible values and their cases.

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -0,0 +1,206 @@
...
+              });
+        });
+    return grpc::Status::OK;
+  }
+
+  grpc::Status CommandServiceTransportGrpc::Status(
+      grpc::ServerContext *context,
+      const iroha::protocol::TxStatusRequest *request,
+      iroha::protocol::ToriiResponse *response) {
+    auto status = command_service_->getStatus(
```
lebdron: Is it possible to use `getTransport`? Like `*response = command_service_->getStatus(shared_model::crypto::Hash(request->tx_hash()))->getTransport();`

PR #1718, opened 2018-09-13
Remove propagateTransaction method

lebdron:
### Description of the Change
Remove propagateTransaction from PCS and OrderingGate interface since it is replaced with propagateBatch. 
Remove OrderingGateServiceTest, since tests are disabled, and SinglePeerOrderingService will be deprecated soon.

### Benefits
Less unused code to support

### Possible Drawbacks 
None


irohad/ordering/impl/ordering_gate_transport_grpc.hpp
```diff
@@ -48,10 +48,6 @@ namespace iroha {
                               const protocol::Proposal *request,
                               ::google::protobuf::Empty *response) override;
 
-      void propagateTransaction(
```
Akvinikym: Do we need `interface::Transaction` forward declaration in this file now?

PR #1452, opened 2018-06-11
Fix bugs related to SetAccountDetail

l4l:
### Description of the Change

Makes a required fixes so that following tests accepted:
- SetAccountDetail on itself can be set w/o any perm

### Benefits

Less bugs, more coverage

### Possible Drawbacks 

None

PR #1723, opened 2018-09-15
Add more acceptance tests for SetAccountDetail

l4l:
### Description of the Change

Add more acceptance tests for SetAccountDetail, since some was already added in #1452 

### Benefits

More coverage

### Possible Drawbacks 

None

`test_bin/set_account_detail_test`

### DIFF
--- a/test/integration/acceptance/set_account_detail_test.cpp
+++ b/test/integration/acceptance/set_account_detail_test.cpp
@@ -49,6 +49,7 @@ class SetAccountDetail : public AcceptanceFixture {
 };
 
 /**
+ * C274
  * @given a user without can_set_detail permission
  * @when execute tx with SetAccountDetail command aimed to the user
  * @then there is the tx in proposal
@@ -65,6 +66,24 @@ TEST_F(SetAccountDetail, Self) {
 }
 
 /**
+ * C273
+ * @given a user with required permission
+ * @when execute tx with SetAccountDetail command with inexistent user
+ * @then there is no tx in proposal
+ */
+TEST_F(SetAccountDetail, NonExistentUser) {
+  IntegrationTestFramework(1)
+      .setInitialState(kAdminKeypair)
+      .sendTx(makeUserWithPerms({interface::permissions::Role::kSetDetail}))
+      .skipProposal()
+      .skipBlock()
+      .sendTxAwait(
+          complete(baseTx("inexistent@" + kDomain, kKey, kValue)),
+          [](auto &block) { ASSERT_EQ(block->transactions().size(), 0); });
+}
+
+/**
+ * C280
  * @given a pair of users and first one without permissions
  * @when the first one tries to use SetAccountDetail on the second
  * @then there is an empty verified proposal
@@ -113,6 +132,7 @@ TEST_F(SetAccountDetail, WithPerm) {
 }
 
 /**
+ * C275
  * @given a pair of users
  *        @and second has been granted can_set_my_detail from the first
  * @when the first one tries to use SetAccountDetail on the second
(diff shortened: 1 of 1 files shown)


test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -65,6 +66,24 @@ TEST_F(SetAccountDetail, Self) {
...
+ * @then there is no tx in proposal
+ */
+TEST_F(SetAccountDetail, NonExistentUser) {
+  IntegrationTestFramework(1)
+      .setInitialState(kAdminKeypair)
+      .sendTx(makeUserWithPerms({interface::permissions::Role::kSetDetail}))
+      .skipProposal()
+      .skipBlock()
+      .sendTxAwait(
+          complete(baseTx("inexistent@" + kDomain, kKey, kValue)),
```
igor-egorov: Please move all non-standard(default) values as const variables to the beginning of the test case. 
This is our best practice we are trying to follow.

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
...
+ * @then there is no tx in proposal
+ */
+TEST_F(SetAccountDetail, HugeKey) {
+  IntegrationTestFramework(1)
+      .setInitialState(kAdminKeypair)
+      .sendTx(makeUserWithPerms())
+      .skipProposal()
+      .skipBlock()
+      .sendTxAwait(
+          complete(baseTx(kUserId, std::string(64, 'a'), kValue)),
```
igor-egorov: Please move `std::string(64, 'a')` as a const var to the beginning of the test case. 
The reason is the same as in the first review comment.

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
...
+ * @when execute tx with SetAccountDetail command with empty key
+ * @then there is no tx in proposal
+ */
+TEST_F(SetAccountDetail, EmptyKey) {
+  IntegrationTestFramework(1)
+      .setInitialState(kAdminKeypair)
+      .sendTx(makeUserWithPerms())
+      .skipProposal()
+      .skipBlock()
+      .sendTx(complete(baseTx(kUserId, "", kValue)), checkStatelessInvalid);
```
igor-egorov: Please move exceptional value as a const var to the beginning of the test case.
The reason is the same as in the first review comment.

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
...
+ * @when execute tx with SetAccountDetail command with empty value
+ * @then there is no tx in proposal
+ */
+TEST_F(SetAccountDetail, EmptyValue) {
+  IntegrationTestFramework(1)
+      .setInitialState(kAdminKeypair)
+      .sendTx(makeUserWithPerms())
+      .skipProposal()
+      .skipBlock()
+      .sendTxAwait(complete(baseTx(kUserId, kKey, "")), [](auto &block) {
```
igor-egorov: Please move exceptional value as a const var to the beginning of the test case.
The reason is the same as in the first review comment.

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
...
+ * @then there is no tx in proposal
+ */
+TEST_F(SetAccountDetail, HugeKeyValue) {
+  IntegrationTestFramework(1)
+      .setInitialState(kAdminKeypair)
+      .sendTx(makeUserWithPerms())
+      .skipProposal()
+      .skipBlock()
+      .sendTx(complete(baseTx(
+                  kUserId, std::string(10000, 'a'), std::string(10000, 'b'))),
```
igor-egorov: Please move exceptional values as a const vars to the beginning of the test case.
The reason is the same as in the first review comment.

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
         ASSERT_EQ(block->transactions().size(), 1);
       });
 }
+
+/**
+ * C276
+ * @given a user with required permission
+ * @when execute tx with SetAccountDetail command with max key
+ * @then there is no tx in proposal
```
igor-egorov: there **IS** a tx in proposal
Moreover, there is a tx in a block

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
...
+ */
+TEST_F(SetAccountDetail, HugeKey) {
```
igor-egorov: You have two test cases: `HugeKey` and `HugeKeyValue`.
The meaning of "huge" is never defined. Does it mean huge but still valid or not?

Surprisingly, here "huge" is a big valid value, but in `HugeKeyValue` test case it is not.

Please rename both test cases with more clear names.

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
...
+      .sendTxAwait(
+          complete(baseTx(kUserId, std::string(64, 'a'), kValue)),
+          [](auto &block) { ASSERT_EQ(block->transactions().size(), 1); });
+}
+
+/**
+ * C277
+ * @given a user with required permission
+ * @when execute tx with SetAccountDetail command with empty key
+ * @then there is no tx in proposal
```
igor-egorov: Actually, in this test case, we are not checking the state of the proposal.
Please correct the description.

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
...
+      .skipProposal()
+      .skipBlock()
+      .sendTx(complete(baseTx(kUserId, "", kValue)), checkStatelessInvalid);
+}
+
+/**
+ * C278
+ * @given a user with required permission
+ * @when execute tx with SetAccountDetail command with empty value
+ * @then there is no tx in proposal
```
igor-egorov: There **IS** a tx in proposal
Moreover, there is a tx in a block

test/integration/acceptance/set_account_detail_test.cpp
```diff
@@ -147,3 +167,69 @@ TEST_F(SetAccountDetail, WithGrantablePerm) {
...
+      .sendTxAwait(complete(baseTx(kUserId, kKey, "")), [](auto &block) {
+        ASSERT_EQ(block->transactions().size(), 1);
+      });
+}
+
+/**
+ * C279
+ * @given a user with required permission
+ * @when execute tx with SetAccountDetail command with huge both key and value
+ * @then there is no tx in proposal
```
igor-egorov: Actually, in this test case, we are not checking the state of the proposal.
Please correct the description.

test/integration/acceptance/set_account_detail_test.cpp
Akvinikym: I don't like that we say "no tx in proposal" while actually checking the block. We not have the `checkVerifiedProposal(..)`, so maybe use it or write "no tx in block"?

PR #1725, opened 2018-09-17
Restrict storage lifetime in simulator and synchronizer

lebdron:
### Description of the Change
 - Restrict storage lifetime to callback in simulator and synchronizer. This prevents storage deadlocks, since resources are released before calling `on_next`, which in turn calls subscribers callbacks.
 - Refactor `operator|` bind calls and `match` calls to remove null pointers using `boost::get`.

### Benefits
No more deadlocks when storage connection is closed

### Possible Drawbacks 
`boost::get` calls are hard to read, a task will be created to refactor `Result` class.


lebdron:
@kamilsa unfortunately it fails in dev

PR #1728, opened 2018-09-20
Remove unnecessary template instantiations in YacGateImpl

lebdron:
### Description of the Change
Templight report for YacGateImpl showed that boost::variant templates for command type were instantiated. However, command type is unused in YacGate, so they should not be instantiated in compile time. Changes in this pull request remove these instantiations, and reduce the compile time from 400s to 60s according to Templight reports.

Main changes:
- Forward declarations instead of includes for interface command:
https://github.com/hyperledger/iroha/blob/bcfb800ac296892624b9fb87f8b7ffcaf744199b/shared_model/interfaces/commands/command.hpp#L15-L30
- Extern templates:
https://github.com/hyperledger/iroha/blob/bcfb800ac296892624b9fb87f8b7ffcaf744199b/shared_model/interfaces/commands/command.hpp#L79-L97
- Separate transaction sequence factory from transaction sequence to remove protobuf and validator dependency from shared model interfaces
- Remove proto backend usage in stateful validator

Note:
Why are exactly the following instantiations needed?
https://github.com/hyperledger/iroha/blob/bcfb800ac296892624b9fb87f8b7ffcaf744199b/shared_model/interfaces/commands/impl/command.cpp#L27-L34
It is because they were the only templates reported by the compiler as undefined references.

### Benefits
- Shorter compile time
- More clear dependency separation

### Possible Drawbacks 
None


shared_model/backend/protobuf/proto_proposal_factory.hpp
```diff
@@ -20,17 +21,17 @@ namespace shared_model {
       FactoryResult<std::unique_ptr<interface::Proposal>> createProposal(
           interface::types::HeightType height,
           interface::types::TimestampType created_time,
-          const interface::types::TransactionsCollectionType &transactions)
-          override {
+          boost::any_range<interface::Transaction, boost::forward_traversal_tag>
```
nickaleks: Why change the type?

shared_model/interfaces/iroha_internal/transaction_batch_helpers.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include "cryptography/hash.hpp"
+
+namespace shared_model {
+  namespace interface {
+
+    /**
+     * Provides a method that calculates reduced batch hash
+     */
+    class TransactionBatchHelpers {
```
nickaleks: Why make a class with only static functions?

lebdron: This is for compatibility reasons, since it was previously in TransactionBatch class. It can probably be made a free function in a separate task.

shared_model/backend/protobuf/proto_proposal_factory.hpp
lebdron: It is otherwise not possible to apply `boost::adaptors::transformed`:
https://github.com/hyperledger/iroha/blob/bcfb800ac296892624b9fb87f8b7ffcaf744199b/irohad/validation/impl/stateful_validator_impl.cpp#L238-L241
This will be refactored with `using` to avoid a convoluted type in interface.

PR #1731, opened 2018-09-24
BFT YacGate

l4l:
### Description of the Change

YacGate that supports rounds and use more precise output on its outcome

### Benefits

One more step toward bft consensus

### Usage Examples or Tests

For now only `yac` target can be built


irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -48,104 +39,102 @@ namespace iroha {
...
             consensus_result_cache_(std::move(consensus_result_cache)),
             log_(logger::log("YacGate")) {
-        block_creator_->on_block().subscribe(
-            [this](auto block) { this->vote(block); });
+        block_creator_->on_block().subscribe([this](auto block) {
+          // TODO(@l4l) 24/09/18 IR-1717
+          // update BlockCreator iface according to YacGate
+          this->vote(
+              std::shared_ptr<shared_model::interface::Proposal>(
+                  (shared_model::proto::ProtoProposalFactory<
```
lebdron: Proto should not be used even as a temporary solution. Consider making a dummy interface implementation in the same file.

irohad/consensus/yac/yac_hash_provider.hpp
```diff
@@ -52,6 +42,9 @@ namespace iroha {
          */
         std::string block_hash;
 
+        /// Consensus round
+        ordering::transport::Round round;
```
lebdron: No need to add this field in this pull request if storage and transport logic is unchanged.

irohad/consensus/yac/yac_hash_provider.hpp
```diff
@@ -78,7 +71,8 @@ namespace iroha {
          * @return hashed value of block
          */
         virtual YacHash makeHash(
-            const shared_model::interface::Block &block) const = 0;
+            const shared_model::interface::Block &block,
+            const shared_model::interface::Proposal &proposal) const = 0;
```
lebdron: Please add round parameter here.

irohad/network/consensus_gate.hpp
```diff
@@ -1,52 +1,71 @@
...
+    struct VoteOther {
+      consensus::yac::YacHash hash;
+    };
+
+    /// Reject on proposal
+    struct ProposalReject {};
+
+    /// Reject on block
+    struct BlockReject {
+      consensus::yac::YacHash hash;
```
lebdron: YacHash is an implementation of consensus, it should not be present in this file

irohad/network/consensus_gate.hpp
```diff
@@ -1,52 +1,71 @@
...
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
  */
 
 #ifndef IROHA_CONSENSUS_GATE_HPP
 #define IROHA_CONSENSUS_GATE_HPP
 
 #include <rxcpp/rx.hpp>
 
+#include "consensus/yac/yac_hash_provider.hpp"
```
lebdron: Please remove this include from interface, because it is a part of implementation.

irohad/consensus/yac/impl/yac_hash_provider_impl.cpp
```diff
@@ -1,33 +1,22 @@
...
       YacHash YacHashProviderImpl::makeHash(
-          const shared_model::interface::Block &block) const {
+          const shared_model::interface::Block &block,
+          const shared_model::interface::Proposal &proposal) const {
         YacHash result;
-        auto hex_hash = block.hash().hex();
-        result.proposal_hash = hex_hash;
-        result.block_hash = hex_hash;
+        result.proposal_hash = proposal.hash().hex();
+        result.block_hash = block.hash().hex();
```
lebdron: No need to change this method in this pull request if other parts of Yac are not changed.

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -48,104 +39,102 @@ namespace iroha {
...
+          current_block_->addSignature(sig->signedData(), sig->publicKey());
         }
       }
+
+      rxcpp::observable<YacGateImpl::GateObject> YacGateImpl::handleCommit(
+          const CommitMessage &msg) {
+        return rxcpp::observable<>::create<GateObject>([&](auto subscriber) {
+          const auto hash = getHash(msg.votes);
+          if (not hash) {
+            log_->info("Invalid commit message, hashes are different");
```
lebdron: Commit message cannot be invalid, since it is guaranteed by CommitMessage interface. Consider removing this conditional.

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,104 +35,91 @@ namespace iroha {
...
-                        },
-                        // if load has failed, no peers provided the block
-                        [this, subscriber](std::exception_ptr) {
-                          log_->error("Cannot load committed block");
-                          subscriber.on_completed();
-                        });
-              });
+          return visit_in_place(
+              message,
+              [this](const CommitMessage &msg) { return handleCommit(msg); },
```
lebdron: Please create a nested static visitor class instead.

irohad/network/consensus_gate.hpp
```diff
@@ -1,52 +1,70 @@
...
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
  */
 
 #ifndef IROHA_CONSENSUS_GATE_HPP
 #define IROHA_CONSENSUS_GATE_HPP
 
 #include <rxcpp/rx.hpp>
 
+#include "ordering/on_demand_os_transport.hpp"
```
lebdron: Use this instead
https://github.com/hyperledger/iroha/blob/114e25d9b642e0b3e2c797641e17d95f6ea637ae/irohad/consensus/round.hpp#L28

irohad/network/consensus_gate.hpp
```diff
@@ -1,52 +1,70 @@
...
        * Providing data for consensus for voting
        * @param block is the block for which current node is voting
        */
       virtual void vote(
-          std::shared_ptr<shared_model::interface::Block> block) = 0;
+          std::shared_ptr<shared_model::interface::Proposal> proposal,
+          std::shared_ptr<shared_model::interface::Block> block,
+          Round round) = 0;
+
+      using GateObject = boost::variant<PairValid,
```
lebdron: This will require creating a `.cpp` file for the interface, and another header.
1. Include `variant_fwd.hpp` in this header file
2. Create a separate header with `variant.hpp` include and `extern template for the variant
3. Create a `.cpp` file with variant instantiations

irohad/network/consensus_gate.hpp
```diff
@@ -1,52 +1,70 @@
...
-  }
+    class Proposal;
+  }  // namespace interface
 }  // namespace shared_model
 
 namespace iroha {
   namespace network {
+    /// Current pair is valid
+    struct PairValid {
+      std::shared_ptr<shared_model::interface::Block> block_;
```
lebdron: No need for trailing underscore, since it is a public member.

irohad/network/consensus_gate.hpp
```diff
@@ -1,52 +1,70 @@
...
+    };
+
+    /// Network votes for another pair and round
+    struct VoteOther {
+      std::shared_ptr<shared_model::interface::Block> block_;
```
lebdron: Is current block required for synchronizer to load the committed block? Maybe use something else instead?

irohad/network/consensus_gate.hpp
```diff
@@ -1,52 +1,70 @@
...
+    };
+
+    /// Reject on proposal
+    struct ProposalReject {};
+
+    /// Reject on block
+    struct BlockReject {
+      std::shared_ptr<shared_model::interface::Block> block_;
```
lebdron: Which block is stored in reject message?

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,104 +35,91 @@ namespace iroha {
...
       }
 
-      void YacGateImpl::vote(std::shared_ptr<shared_model::interface::Block> block) {
-        auto hash = hash_provider_->makeHash(*block);
-        log_->info("vote for block ({}, {})",
-                   hash.vote_hashes.proposal_hash,
-                   block->hash().toString());
-        auto order = orderer_->getOrdering(hash);
+      void YacGateImpl::vote(
+          std::shared_ptr<shared_model::interface::Proposal> proposal,
```
lebdron: Optionals should be used for both proposal and block, because there could be no proposal from the current ordering service, and therefore no block.

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,104 +35,91 @@ namespace iroha {
...
           auto sig = vote.hash.block_signature;
-          current_block_.second->addSignature(sig->signedData(),
-                                              sig->publicKey());
+          current_block_->addSignature(sig->signedData(), sig->publicKey());
         }
       }
+
+      rxcpp::observable<YacGateImpl::GateObject> YacGateImpl::handleCommit(
+          const CommitMessage &msg) {
+        return rxcpp::observable<>::create<GateObject>([&](auto subscriber) {
```
lebdron: It is better not to manually create an observable, but use existing method `just` to emit a single message, or `empty` if nothing is emitted.

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,104 +35,91 @@ namespace iroha {
...
+          const auto hash = getHash(msg.votes).value();
+          if (hash.vote_hashes.proposal_hash.size() == 0) {
```
lebdron: Maybe use `.empty()` instead?

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,104 +35,91 @@ namespace iroha {
...
+            log_->info("Voted for another block, waiting for sync");
+            subscriber.on_next(network::VoteOther{current_block_});
+          }
+          subscriber.on_completed();
+        });
+      }
+
+      rxcpp::observable<YacGateImpl::GateObject> YacGateImpl::handleReject(
+          const RejectMessage &msg) {
+        return rxcpp::observable<>::create<GateObject>([&](auto subscriber) {
```
lebdron: Same for using `rxcpp::observable<>::just`

irohad/consensus/yac/yac_hash_provider.hpp
```diff
@@ -24,6 +12,7 @@
 #include "consensus/round.hpp"
 #include "consensus/yac/storage/yac_common.hpp"
 #include "interfaces/common_objects/types.hpp"
+#include "ordering/on_demand_os_transport.hpp"
```
lebdron: This include should not be required in this file.

irohad/network/consensus_gate.hpp
l4l: I'm not really aware of it. I just guessed that Block will be enough.

irohad/consensus/yac/impl/yac_gate_impl.cpp
l4l: Optional sounds redundant, maybe just use (& notice in docs) `nullptr` as no proposa/block case?

irohad/consensus/yac/impl/yac_gate_impl.cpp
l4l: Why?

irohad/consensus/yac/impl/yac_gate_impl.cpp
l4l: What's the profit?

lebdron: No redundant `visit_in_place` with two proxy methods.

irohad/consensus/yac/impl/yac_gate_impl.cpp
lebdron: No, since we guarantee that pointers are not nullable in server code.

irohad/consensus/yac/impl/yac_gate_impl.cpp
lebdron: To simplify the code and eliminate any possibilty to forget `on_completed` call.

irohad/consensus/yac/impl/yac_gate_impl.cpp
l4l: ...will be replaced with redundant visit class with two methods?
I don't notice any profit again, could you describe it?

l4l: ...will be replaced with redundant visit class with two methods?
I don't notice any profit again, could you describe it?

irohad/network/consensus_gate.hpp
l4l: The one is voted for and that is probably incorrect, meh, thank you for noticing
What should be placed there?

irohad/consensus/yac/impl/yac_gate_impl.hpp
```diff
@@ -78,8 +70,8 @@ namespace iroha {
 
         logger::Logger log_;
 
-        std::pair<YacHash, std::shared_ptr<shared_model::interface::Block>>
-            current_block_;
+        std::shared_ptr<shared_model::interface::Block> current_block_;
```
lebdron: -> `optional<shared_ptr>`, same reason as `vote` arguments

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,103 +35,101 @@ namespace iroha {
...
-        auto order = orderer_->getOrdering(hash);
+      void YacGateImpl::vote(
+          boost::optional<std::shared_ptr<shared_model::interface::Proposal>>
+              proposal,
+          boost::optional<std::shared_ptr<shared_model::interface::Block>>
+              block,
+          Round round) {
+        bool is_none = not proposal or not block;
+        if (is_none) {
+          current_block_ = nullptr;
```
lebdron: -> `nullopt`

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,103 +35,101 @@ namespace iroha {
...
+          current_hash_ = {};
```
lebdron: Round should be set here, please leave a comment.

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,103 +35,101 @@ namespace iroha {
...
+                     current_block_->height(),
+                     current_block_->hash().hex());
+          return rxcpp::observable<>::just<GateObject>(
+              network::PairValid{current_block_});
+        } else {
+          log_->info("Voted for another block, waiting for sync");
+          return rxcpp::observable<>::just<GateObject>(
+              network::VoteOther{current_block_});
+        }
+        return rxcpp::observable<>::empty<GateObject>();
```
lebdron: This line is unreachable since conditionals cover all the cases, please remove.

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -35,103 +35,101 @@ namespace iroha {
...
           return;
         }
-        current_block_ = std::make_pair(hash, block);
-        hash_gate_->vote(hash, *order);
+
+        hash_gate_->vote(current_hash_, *order);
 
         // insert the block we voted for to the consensus cache
-        consensus_result_cache_->insert(block);
+        if (is_none) {
```
lebdron: Code in this conditional might throw an exception in case there was no block:
https://github.com/hyperledger/iroha/blob/a52ca3e89997bcb45e04d3032dc8258da436673d/irohad/consensus/yac/impl/yac_gate_impl.cpp#L53
Check if this case is covered by the tests and change the conditional.

irohad/network/consensus_gate.hpp
l4l: todo added

PR #1736, opened 2018-09-25
Docker image isolated dependencies

bakhtin:

### Description of the Change
- Swap out develop/Dockerfile with dependencies/Dockerfile
- Append Dockerfile with `CMAKE_PREFIX_PATH` to point CMake look for dependencies on additional paths
- Add README files 

### Benefits
Resolves ambiguous situations during the build when some dependencies may break it in unpredictable ways. Or, vice versa, may affect the build in the way the code actually compiles when it should not. We isolate each dependency in a separate folder so they do not intersect with ones installed in the system.

### Possible Drawbacks 
None (leave the comment if you come up with any)


docker/develop/Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN set -e; \
         libpcre3-dev autoconf bison  \
         # other
         wget curl file gdb gdbserver ccache \
-        gcovr cppcheck doxygen rsync graphviz graphviz-dev unzip zip; \
+        gcovr cppcheck doxygen graphviz graphviz-dev unzip zip; \
```
l4l: why remove rsync?

docker/README.md
```diff
@@ -1,3 +1,7 @@
-# Docker
+# Description
+This directory contains Docker files for building different flavors of Docker images:
+- *develop* contains Dockerfile with all the dependencies required for compiling Iroha source code. We use it in development process, thus all depencencies are kept up-to-date
+- *release* contains Dockerifle with all the dependencies required for running Iroha binaries compiled with *Release* flag. Each time commit is pushed into `develop` or `master` branch, automated routine is executed to build Docker image containing Iroha binaries. It is then pushed to [Dockerhub](https://hub.docker.com/r/hyperledger/iroha/)
```
Akvinikym: Doesn't everything, which goes after words "Each time commit is pushed..." relate to both *develop* and *release*? Or only *release* is updated, even if we push into 'develop'?
If does relate to both, please move this part to the beginning.

bakhtin:
We have issues having all the dependencies in separate directories. Mitigation needed

PR #1737, opened 2018-09-25
Add MST test for possible replays

igor-egorov:

### Description of the Change

Add a test that checks that all the transactions (even fully signed, where quorum == amount of signatures) are going through MST and there is no chance for the transaction with the same hash be left as pending.

MstPipelineTest fixture was refactored a bit as a side effect.

### Benefits

Increased tests coverage. The bug revealed (IR-1329)

### Possible Drawbacks 

None

### Usage Examples or Tests

multisig_tx_pipeline_test


test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -105,14 +82,40 @@ class MstPipelineTest : public AcceptanceFixture {
...
+    return complete(baseQry(creator).getPendingTransactions(), key);
+  }
+
+  /**
+   * Query validation lambda - check that empty transactions response returned
+   * @param response - query response
+   */
+  static void no_txs_check(const shared_model::proto::QueryResponse &response) {
+    ASSERT_NO_THROW({
+      const auto &pending_tx_resp = boost::apply_visitor(
```
Akvinikym: It's better to use `boost::get<shared_model::interface::TransactionsResponse>(response.get())` to get the same result, instead of `SpecifiedVisitor`

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -105,14 +82,40 @@ class MstPipelineTest : public AcceptanceFixture {
...
+          framework::SpecifiedVisitor<
+              shared_model::interface::TransactionsResponse>(),
+          response.get());
+      ASSERT_TRUE(pending_tx_resp.transactions().empty());
+    });
+  }
+
+  /**
+   * Checks the number of signatures of the first pending transaction
```
Akvinikym: Method does not check, but returns a lambda checking

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -105,14 +82,40 @@ class MstPipelineTest : public AcceptanceFixture {
...
+   * @param expected_signatures_number
+   * @return query validation lambda
+   */
+  auto signatory_check(size_t expected_signatures_number) {
+    return [expected_signatures_number](auto &response) {
+      ASSERT_NO_THROW({
+        const auto &pending_tx_resp = boost::apply_visitor(
```
Akvinikym: Same about visitor

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -105,14 +82,40 @@ class MstPipelineTest : public AcceptanceFixture {
...
+    });
+  }
+
+  /**
+   * Checks the number of signatures of the first pending transaction
+   * @param expected_signatures_number
+   * @return query validation lambda
+   */
+  auto signatory_check(size_t expected_signatures_number) {
+    return [expected_signatures_number](auto &response) {
```
Akvinikym: As I see, there are no reasons not to make the `response` a const reference

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -238,19 +229,37 @@ TEST_F(MstPipelineTest, GetPendingTxsNoSignedTxs) {
...
-      .sendTx(signTx(pending_tx, signatories[1]))
-      .sendTx(signTx(pending_tx, kUserKeypair))
+  mst_itf.sendTx(complete(pending_tx, signatories[0]))
+      .sendTx(complete(pending_tx, signatories[1]))
+      .sendTx(complete(pending_tx, kUserKeypair))
+      .sendQuery(makeGetPendingTxsQuery(kUserId, kUserKeypair), no_txs_check);
+}
+
+/**
+ * Disabled because fully signed transaction goes no through MST and pending
```
Akvinikym: "goes no" -> "doesn't go"

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -238,19 +229,37 @@ TEST_F(MstPipelineTest, GetPendingTxsNoSignedTxs) {
...
+ * transaction remains in query response IR-1329
+ * @given a ledger with mst user (quorum=3) created
+ * @when the user sends a transactions with only one signature, then sends the
```
Akvinikym: "sends a transactions" -> "sends a transaction"

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -105,14 +82,40 @@ class MstPipelineTest : public AcceptanceFixture {
...
-        .queryCounter(1)
-        .getPendingTransactions()
-        .build()
-        .signAndAddSignature(key)
-        .finish();
+    return complete(baseQry(creator).getPendingTransactions(), key);
+  }
+
+  /**
+   * Query validation lambda - check that empty transactions response returned
```
Akvinikym: It's not a lambda, but a full method. Maybe just "Check that empty..." as a first line would be better?

igor-egorov: It's meaning to be a query validation lambda. Nothing to fix.

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -105,14 +78,39 @@ class MstPipelineTest : public AcceptanceFixture {
...
+   * @param response - query response
+   */
+  static void no_txs_check(const proto::QueryResponse &response) {
```
lebdron: Please rename to camel case according to codestyle. Same for other methods

test/integration/pipeline/multisig_tx_pipeline_test.cpp
```diff
@@ -105,14 +78,39 @@ class MstPipelineTest : public AcceptanceFixture {
...
+    ASSERT_NO_THROW({
+      const auto &pending_txs_resp =
+          boost::get<interface::TransactionsResponse>(response.get());
```
lebdron: `interface::TransactionsResponse` -> `const interface::TransactionsResponse &` 
Same in other methods
You can make a local shortcut which will accept `T` and do `boost::get<const T &>`

igor-egorov:
Some extra time is required to solve the testing methodology problem with multisig_tx_pipeline_test.

PR #1581, opened 2018-07-20
Fix rare-case amount overflows

l4l:
### Description of the Change

- Makes overflowed amounts stateless invalid
- Fix increaseValuePrecision overflow 
- Refactor amount_utils a bit
- Add related tests

### Benefits

Lesser bugs, bigger coverage

### Possible Drawbacks 

Pretty much poor-readable code

### Alternate Designs

Part of the fixes (with whole amout_utils, actually) can be simplified with just checked cpp_int. Though I don't think, whether enabling exceptions is a good idea (and it seems that they are disabled).

PR #1742, opened 2018-09-26
Add proper amount validation

l4l:
### Description of the Change

Add proper stateless validation splitted from #1581

### Benefits

Amount stateless validation satisfy doc reference


shared_model/interfaces/common_objects/impl/amount.cpp
```diff
@@ -16,23 +17,23 @@ namespace shared_model {
         : amount_(std::move(amount)),
           precision_(0),
           multiprecision_repr_([this] {
-            static const std::regex r("([0-9]+)(\\.([0-9]+))?");
+            static const std::regex r(
+                "[1-9][0-9]*(\\.([0-9]+))?|0\\.([0-9]*[1-9])");
```
Akvinikym: The same regex pattern is repeated twice - not good, of something will change. Can't you reuse it from `FieldValidator::amount_pattern_`? Or maybe declare the way both classes would see and use it?

l4l: They're not the same actually. Take in account groups

neewy:
This pull request introduces unwanted changes to API

PR #1753, opened 2018-10-02
Separate range types; add missing MST link library

lebdron:
### Description of the Change
Separate `range_types.hpp` from `types.hpp` to avoid redundant `boost` include where it is not required.
Move non-template implementations from headers to cpp in MST.
Replace includes with forward declarations in `mst_types.hpp`.
Add missing `boost` link for `mst_state`

### Benefits
More strict dependency management

### Possible Drawbacks 
None

### DIFF
--- a/irohad/multi_sig_transactions/hash.hpp
+++ b/irohad/multi_sig_transactions/hash.hpp
@@ -6,11 +6,6 @@
 #ifndef IROHA_HASH_HPP
 #define IROHA_HASH_HPP
 
-#include <functional>
-#include <string>
-
-#include "cryptography/public_key.hpp"
-#include "interfaces/common_objects/peer.hpp"
 #include "multi_sig_transactions/mst_types.hpp"
 
 namespace iroha {
@@ -18,15 +13,9 @@ namespace iroha {
     /**
      * Hash calculation factory for batch
      */
-    template <typename BatchType>
     class PointerBatchHasher {
      public:
-      size_t operator()(const BatchType &batch) const {
-        return string_hasher(batch->reducedHash().hex());
-      }
-
-     private:
-      std::hash<std::string> string_hasher;
+      size_t operator()(const DataType &batch) const;
     };
 
     /**
@@ -35,12 +24,7 @@ namespace iroha {
     class PeerHasher {
      public:
       std::size_t operator()(
-          const std::shared_ptr<shared_model::interface::Peer> &obj) const {
-        return hasher(obj->address() + obj->pubkey().hex());
-      }
-
-     private:
-      std::hash<std::string> hasher;
+          const std::shared_ptr<shared_model::interface::Peer> &obj) const;
     };
   }  // namespace model
 }  // namespace iroha
--- a/irohad/multi_sig_transactions/impl/hash.cpp
+++ b/irohad/multi_sig_transactions/impl/hash.cpp
@@ -1,0 +1,28 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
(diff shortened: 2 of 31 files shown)


test/module/irohad/multi_sig_transactions/gossip_propagation_strategy_test.cpp
```diff
@@ -1,35 +1,23 @@
...
 #include <memory>
 #include <numeric>
-#include <rxcpp/rx.hpp>
 #include <string>
 #include <thread>
 #include <vector>
 
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <boost/range/irange.hpp>
```
l4l: shouldn't it be higher?

irohad/multi_sig_transactions/impl/hash.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <functional>
+#include <string>
+
+#include "cryptography/public_key.hpp"
+#include "interfaces/common_objects/peer.hpp"
+#include "interfaces/iroha_internal/transaction_batch.hpp"
+
+namespace iroha {
+  namespace model {
```
l4l: strange namespace

irohad/multi_sig_transactions/hash.hpp
```diff
@@ -35,12 +24,7 @@ namespace iroha {
     class PeerHasher {
      public:
       std::size_t operator()(
-          const std::shared_ptr<shared_model::interface::Peer> &obj) const {
-        return hasher(obj->address() + obj->pubkey().hex());
-      }
-
-     private:
-      std::hash<std::string> hasher;
+          const std::shared_ptr<shared_model::interface::Peer> &obj) const;
```
l4l: Add <memory> & iface::Peer fdecl please

shared_model/interfaces/common_objects/range_types.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+    class Signature;
+    class Transaction;
+    class AccountAsset;
+
+    namespace types {
+
+      /// Type of signature range, which returns when signatures are invoked
+      using SignatureRangeType = boost::any_range<interface::Signature,
+                                                  boost::forward_traversal_tag,
```
nickaleks: Why is it forward traversal?

test/module/irohad/multi_sig_transactions/gossip_propagation_strategy_test.cpp
lebdron: No, since these are not system libraries.

irohad/multi_sig_transactions/impl/hash.cpp
lebdron: Well, that is what was in the header file.

shared_model/interfaces/common_objects/range_types.hpp
lebdron: https://en.cppreference.com/w/cpp/container/unordered_set

> iterator | Constant ForwardIterator

I think it is because `unordered_set` has forward iterator, and this range is used in signable.

irohad/multi_sig_transactions/hash.hpp
lebdron: https://github.com/hyperledger/iroha/blob/99b8a32a65fb69575f01b9a5950042ecde290754/irohad/multi_sig_transactions/mst_types.hpp#L17
Since `mst_types` is included in hash, maybe there is no need to add another forward declaration?

lebdron:
Since pr-head CI was successful, it seems that test which failed in branch is not deterministic. This has to be investigated in a separate task.

PR #1754, opened 2018-10-03
Remove wsv query dependency from chain validator

nickaleks:
### Description of the Change

This PR removes wsv_query dependency from chain validator

### Benefits

Wsv Query is less used at the moment

### Possible Drawbacks 

Mutable Storage interface had to be changed


irohad/ametsuchi/mutable_storage.hpp
```diff
@@ -53,8 +54,12 @@ namespace iroha {
        * false otherwise
        * @return result of predicate
        */
-      virtual bool check(const shared_model::interface::Block &block,
-                         MutableStoragePredicateType function) = 0;
+      virtual bool check(
+          const shared_model::interface::Block &block,
+          std::function<bool(const shared_model::interface::Block &,
```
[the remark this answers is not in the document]
Akvinikym: For example, the `MutableStoragePredicateType` can be declared as follows:
`template <typename QueryType>`
`using MutableStoragePredicateType =`
           `std::function<bool(const shared_model::interface::Block &,`
                                 `QueryType &,`
                                 `const shared_model::interface::types::HashType &)>;`
And thus be used in both functions

irohad/ametsuchi/mutable_storage.hpp
```diff
@@ -71,7 +74,7 @@ namespace iroha {
        * @return True if block was successfully applied, false otherwise.
        */
       virtual bool apply(const shared_model::interface::Block &block,
-                         MutableStoragePredicateType function) = 0;
+                         MutableStoragePredicate<WsvQuery> function) = 0;
```
lebdron: It should be possible to remove the predicate from this method
https://github.com/hyperledger/iroha/blob/edf8ee73af2dcbeefa1471dde16de5b08a4fd362/irohad/synchronizer/impl/synchronizer_impl.cpp#L29-L31

irohad/ametsuchi/mutable_storage.hpp
```diff
@@ -70,8 +72,7 @@ namespace iroha {
        * otherwise.
        * @return True if block was successfully applied, false otherwise.
        */
```
Akvinikym: Please, update docs, there's no `@param function` anymore

PR #1755, opened 2018-10-03
fix irohad_test

Solonets:
### Description of the Change

irohad was given 1 second to initialize, if irohad failed to initialized in 1 second, test framework started to send some transactions to not oppened port. Now test wait's for irohad to send "iroha initilized"

### Benefits

test passed

### Possible Drawbacks 

--

### DIFF
--- a/test/system/irohad_test.cpp
+++ b/test/system/irohad_test.cpp
@@ -71,11 +71,10 @@ class IrohadTest : public AcceptanceFixture {
     auto channel = grpc::CreateChannel(kAddress + ":" + std::to_string(kPort),
                                        grpc::InsecureChannelCredentials());
     auto state = channel->GetState(true);
-    auto start = std::chrono::system_clock::now();
+    auto deadline = std::chrono::system_clock::now() + kTimeout;
     while (state != grpc_connectivity_state::GRPC_CHANNEL_READY
-           and start + kTimeout > std::chrono::system_clock::now()) {
-      channel->WaitForStateChange(state,
-                                  std::chrono::system_clock::now() + kTimeout);
+           and deadline > std::chrono::system_clock::now()) {
+      channel->WaitForStateChange(state, deadline);
       state = channel->GetState(true);
     }
     ASSERT_EQ(state, grpc_connectivity_state::GRPC_CHANNEL_READY);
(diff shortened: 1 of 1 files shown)


test/system/irohad_test.cpp
```diff
@@ -67,11 +67,20 @@ class IrohadTest : public AcceptanceFixture {
   }
 
   void launchIroha(const std::string &parameters) {
-    iroha_process_.emplace(irohad_executable.string() + parameters);
-    std::this_thread::sleep_for(kTimeout);
+    boost::process::ipstream output;
+    std::thread wait_init([&output] {
+        std::string line;
+        while (std::getline(output, line)) {
+          if (line.find("iroha initialized") != std::string::npos) {
```
kamilsa: There should be some another way to fix that. For now it is okay I think. But please leave todo that will require to remove this check as soon as healthcare is introduced

l4l:
Imo it's better solve with a spinlock, e.g:

```
auto timeout = now() + 20s; // should be really big
while (not iroha_process_->running() && now() < timeout);
ASSERT_TRUE(iroha_process_->running());
```

test/system/irohad_test.cpp
Solonets: @kamilsa what can be a healthcare in this case?

Solonets:
@l4l you have to insure that iroha have been initialized and started to listen the port, not that process have been started, for this we need to track irohad output

l4l:
Indeed. I guess this task should be solved with the healthcheck issue in future

test/system/irohad_test.cpp
```diff
@@ -67,20 +67,18 @@ class IrohadTest : public AcceptanceFixture {
...
-          if (line.find("iroha initialized") != std::string::npos) {
-            return;
-          }
-        }
-
-    });
-    iroha_process_.emplace(irohad_executable.string() + parameters, boost::process::std_out > output_);
-    reader_.join();
+    iroha_process_.emplace(irohad_executable.string() + parameters);
+    auto channel = grpc::CreateChannel(kAddress + ":" + std::to_string(kPort),
```
l4l: Should be internal port, not torii. Otherwise the first one may have an issue that is not covered

l4l: Since the transaction are sent in some tests, this way is fine (and probably even better), resolved

test/system/irohad_test.cpp
```diff
@@ -208,6 +206,7 @@ DROP TABLE IF EXISTS index_by_id_height_asset;
 
  public:
   boost::filesystem::path irohad_executable;
+  const std::chrono::milliseconds kTimeout = std::chrono::seconds(5);
```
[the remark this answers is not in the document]
kamilsa: make bigger timestamp

PR #1761, opened 2018-10-04
Add platform independent Python lib and Batch example

igor-egorov:

### Description of the Change

Introduced SWIG-free Iroha lib for Python.
Batches usage example included.

### Benefits

A little step towards user-friendly Iroha.

### Possible Drawbacks 

None ?

### Usage Examples or Tests

Read irohalib.md and do the preparation steps
Run ```python3 batch-example.py```


example/python/irohalib.py
```diff
@@ -0,0 +1,310 @@
...
+        core_payload.created_time = created_time
+        core_payload.creator_account_id = creator_account
+        core_payload.commands.extend(commands)
+        return tx
+
+    @staticmethod
+    def command(name, **kwargs):
+        """
+        Creates a protobuf command to be inserted into a transaction
+        :param name: CamelCased name of command
```
lebdron: Make these consistent in command and query please, CamelCased or CamelCase

example/python/irohalib.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+
+    def __init__(self, address=None):
+        self.address = address if address else '127.0.0.1:50051'
+
+    def send_tx(self, transaction):
+        """
+        Send a transaction to Iroha
+        :param transaction: protobuf Transaction
+        :return: None
```
lebdron: It might be useful to return gRPC status to check whether the transaction was successfully sent. Same for other methods

example/python/irohalib.py
```diff
@@ -0,0 +1,310 @@
...
+        """
+        channel = grpc.insecure_channel(self.address)
```
lebdron: It looks like channel and stub can be made class fields

example/python/batch-example.py
```diff
@@ -0,0 +1,180 @@
...
+        )],
+        creator_account='alice@test',
+        quorum=2
+    )
+    bob_tx = iroha.transaction(
+        [iroha.command(
+            'TransferAsset', src_account_id='bob@test', dest_account_id='alice@test', asset_id='dogecoin#test',
+            amount='2'
+        )],
+        creator_account='bob@test'
```
lebdron: Maybe leave a comment to clarify that alice does not know about bob's quorum.

example/python/batch-example.py
```diff
@@ -0,0 +1,183 @@
...
+    b'f101537e319568c765b2cc89698325604991dca57b9716b58016b253506caba3',
+    b'f101537e319568c765b2cc89698325604991dca57b9716b58016b253506caba4'
+]
+bob_public_keys = [ic.derive_public_key(x) for x in bob_private_keys]
+
+
+def send_transaction_and_print_status(transaction):
+    global net
+    hex_hash = binascii.hexlify(ic.hash(transaction))
+    print('+' * 40)
```
l4l: Doesn't really help much without reading code itself. Consider using `print("Enter <function_name>")`

example/python/batch-example.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+    bob_tx = iroha.transaction(
+        [iroha.command(
+            'TransferAsset', src_account_id='bob@test', dest_account_id='alice@test', asset_id='dogecoin#test',
+            amount='2'
+        )],
+        creator_account='bob@test'
+        # we intentionally omit here bob's quorum, since alice is the originator of the exchange and in general case
+        # alice does not know bob's quorum.
+        # bob knowing onw quorum in case of accept should sign the tx using all the number of missing keys at once
```
l4l: `onw` - looks like a typo

example/python/batch-example.py
```diff
@@ -0,0 +1,183 @@
...
+            ic.sign_query(
+                iroha.query('GetPendingTransactions', creator_account='bob@test'),
+                bob_private_keys[0]
+            )
+        )
+    )
+    print(' ~~~')
+
+
+def bob_declines_exchange_request():
```
l4l: This function looks exactly like `bob_accepts_exchange_request` except one line. Consider adding function with a parameter and move logic there (so `bob_accepts_exchange_request` & `bob_declines_exchange_request` are just be the wrappings around the new one)

example/python/irohalib.md
```diff
@@ -0,0 +1,30 @@
+# Python Iroha Lib Description
+
+irohalib.py is a possible implemetation of a platform-independent way of interaction with Iroha network via Python 3.
```
l4l: possible -> example?

example/python/batch-example.py
```diff
@@ -0,0 +1,183 @@
...
+
+import binascii
+
+iroha = Iroha('admin@test')
+net = IrohaGrpc()
+
+admin_private_key = open('../admin@test.priv').read()
+
+alice_private_keys = [
+    b'f101537e319568c765b2cc89698325604991dca57b9716b58016b253506caba1',
```
l4l: Where's that key comes from?

example/python/irohalib.py
```diff
@@ -0,0 +1,316 @@
...
+    @staticmethod
+    def _camel_case_to_snake_case(camel_case_string):
+        """Transforms"""
+        tmp = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', camel_case_string)
+        return re.sub('([a-z0-9])([A-Z])', r'\1_\2', tmp).lower()
+
+    @staticmethod
+    def now():
+        """Current timestamp in milliseconds"""
+        return int(round(time.time() * 1000))
```
l4l: Not a big deal, but datetime should be considered for msecs

example/python/irohalib.py
```diff
@@ -0,0 +1,316 @@
...
+    @staticmethod
+    def command(name, **kwargs):
+        """
+        Creates a protobuf command to be inserted into a transaction
+        :param name: CamelCased name of command
+        :param kwargs: command arguments as they defined in schema
+        :return: a proto command
+
+        Usage example:
+        cmd = Iroha.command('CreateDomain', domain_id='test', default_role='user')
```
l4l: :heart:

example/python/irohalib.py
```diff
@@ -0,0 +1,316 @@
...
+class Iroha(object):
+    """
+    Collection of factory methods for transactions and queries creation
+    """
+
+    def __init__(self, creator_account=None):
+        self.creator_account = creator_account
+
+    @staticmethod
+    def _camel_case_to_snake_case(camel_case_string):
```
l4l: private methods usually have double underscore prefix in python

example/python/batch-example.py
lebdron: Not sure if this is the point of example code. Having two distinct functions makes it easier to understand the use cases for starters

example/python/batch-example.py
igor-egorov: Thanks! Added a tracer for methods' calls. Now each call will be accomplished with "Entering/Leaving X". 
Thus, it will be clear what each output means.

example/python/batch-example.py
igor-egorov: They have a lot in common due to the common nature of those two actions.
For clarity, I am not going to change this part of the example.

example/python/irohalib.md
igor-egorov: Possible is ok here and perfectly reflects the initial intention.

Thanks for commenting anyway - I have found a type in "implementation" word!

PR #1772, opened 2018-10-11
Refactor block index to construct a single query for each block

nickaleks:
### Description of the Change

This PR refactors postgers_block_index to construct a single query instead of multiple calls to the database.

### Benefits

Supposedly improved performance.

### Possible Drawbacks 

None

### DIFF
--- a/irohad/ametsuchi/impl/postgres_block_index.cpp
+++ b/irohad/ametsuchi/impl/postgres_block_index.cpp
@@ -6,24 +6,32 @@
 #include "ametsuchi/impl/postgres_block_index.hpp"
 
 #include <numeric>
-#include <unordered_set>
 
 #include <boost/format.hpp>
-#include <boost/range/adaptor/filtered.hpp>
 #include <boost/range/adaptor/indexed.hpp>
-#include <boost/range/algorithm/for_each.hpp>
-#include <boost/range/numeric.hpp>
 #include "common/types.hpp"
 #include "common/visitor.hpp"
-#include "cryptography/ed25519_sha3_impl/internal/sha3_hash.hpp"
 #include "interfaces/commands/command.hpp"
 #include "interfaces/commands/transfer_asset.hpp"
 #include "interfaces/common_objects/types.hpp"
 #include "interfaces/iroha_internal/block.hpp"
 
 namespace {
+  // Return transfer asset if command contains it
+  boost::optional<const shared_model::interface::TransferAsset &>
+  getTransferAsset(const shared_model::interface::Command &cmd) noexcept {
+    using ReturnType =
+        boost::optional<const shared_model::interface::TransferAsset &>;
+    return iroha::visit_in_place(
+        cmd.get(),
+        [](const shared_model::interface::TransferAsset &c) {
+          return ReturnType(c);
+        },
+        [&](const auto &command) -> ReturnType { return boost::none; });
+  }
+
   // tx hash -> block where hash is stored
-  std::string makeHashIndexQuery(
+  std::string makeHashIndex(
       const shared_model::interface::types::HashType &hash,
       const std::string &height) {
     boost::format base(
@@ -34,7 +42,7 @@ namespace {
 
   // make index account_id:height -> list of tx indexes
   // (where tx is placed in the block)
-  std::string makeCreatorHeightIndexQuery(
+  std::string makeCreatorHeightIndex(
       const shared_model::interface::types::AccountIdType creator,
       const std::string &height,
       const std::string &tx_index) {
@@ -45,19 +53,19 @@ namespace {
   }
 
   // Make index account_id -> list of blocks where his txs exist
-  std::string makeAccountHeightIndexQuery(const std::string &account_id,
-                                          const std::string &height) {
+  std::string makeAccountHeightIndex(const std::string &account_id,
+                                     const std::string &height) {
     boost::format base(
         "INSERT INTO height_by_account_set(account_id, "
         "height) VALUES "
         "('%s', '%s');");
-
     return (base % account_id % height).str();
   }
 
   // Collect all assets belonging to creator, sender, and receiver
   // to make account_id:height:asset_id -> list of tx indexes
-  std::string makeAccountAssetIndexQuery(
+  // for transfer asset in command
+  std::string makeAccountAssetIndex(
       const shared_model::interface::types::AccountIdType &account_id,
       const std::string &height,
       const std::string &index,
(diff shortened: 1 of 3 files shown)


irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,127 @@
...
+  boost::optional<const shared_model::interface::TransferAsset &>
+  getTransferAsset(const shared_model::interface::Command &cmd) noexcept {
+    using ReturnType =
+        boost::optional<const shared_model::interface::TransferAsset &>;
+    return iroha::visit_in_place(
+        cmd.get(),
+        [](const shared_model::interface::TransferAsset &c) {
+          return ReturnType(c);
+        },
+        [&](const auto &command) -> ReturnType { return boost::none; });
```
Akvinikym: Not necessary to specify argument name `command`, I assume

Akvinikym: Also, is reference lambda capture needed here?

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,127 @@
...
-    bool execute(soci::statement &st) {
-      st.define_and_bind();
-      try {
-        st.execute(true);
-        return true;
-      } catch (const std::exception &e) {
-        return false;
-      }
-    }
+  // tx hash -> block where hash is stored
```
Akvinikym: Very strange comment, did you want to add docs for a function?

Akvinikym: As I see, all methods in this namespace have the same comments, maybe change them to usual docs format?

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,127 @@
...
+}  // namespace
+
+namespace iroha {
+  namespace ametsuchi {
+    PostgresBlockIndex::PostgresBlockIndex(soci::session &sql)
+        : sql_(sql), log_(logger::log("PostgresBlockIndex")) {}
 
     void PostgresBlockIndex::index(
         const shared_model::interface::Block &block) {
       const auto &height = std::to_string(block.height());
```
Akvinikym: Why stringifying the height, if it can be passed directly as `HeightType` and then inserted into boost format as `uint` value?

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,127 @@
...
-            // (where tx is placed in the block)
-            st = (sql_.prepare
-                      << "INSERT INTO index_by_creator_height(creator_id, "
-                         "height, index) "
-                         "VALUES (:id, :height, :index)",
-                  soci::use(creator_id),
-                  soci::use(height),
-                  soci::use(index));
-            execute(st);
+            const auto index = std::to_string(tx.index());
```
Akvinikym: And index as well?

irohad/ametsuchi/impl/postgres_block_index.cpp
nickaleks: They are private methods of the file. I do not see the point of adding doc comments for since they are not supposed to be used outside. I will add more descriptive comments though

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,128 @@
...
 
-namespace iroha {
-  namespace ametsuchi {
+namespace {
+  // Return transfer asset if command contains it
+  boost::optional<const shared_model::interface::TransferAsset &>
+  getTransferAsset(const shared_model::interface::Command &cmd) noexcept {
+    using ReturnType =
+        boost::optional<const shared_model::interface::TransferAsset &>;
+    return iroha::visit_in_place(
```
lebdron: Using `boost::get` with pointer may result in simpler code here, and visitor include will not be required.

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,128 @@
...
 
-            this->indexAccountAssets(
+            query += makeAccountHeightIndex(creator_id, height);
+            query += makeAccountAssetIndex(
                 creator_id, height, index, tx.value().commands());
+            query += makeHashIndex(tx.value().hash(), height);
+            query += makeCreatorHeightIndex(creator_id, height, index);
+            return query;
           });
+      sql_ << index_query;
```
lebdron: Should be wrapped with try-catch.

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,128 @@
...
-#include "cryptography/ed25519_sha3_impl/internal/sha3_hash.hpp"
 #include "interfaces/commands/command.hpp"
 #include "interfaces/commands/transfer_asset.hpp"
+#include "interfaces/common_objects/types.hpp"
 #include "interfaces/iroha_internal/block.hpp"
 
-namespace iroha {
-  namespace ametsuchi {
+namespace {
+  // Return transfer asset if command contains it
```
lebdron: I would rather use `///` since it is a docstring. Same for other methods

irohad/ametsuchi/impl/postgres_block_index.hpp
```diff
@@ -31,32 +31,21 @@ namespace iroha {
...
-       * Collect all assets belonging to creator, sender, and receiver
-       * to make account_id:height:asset_id -> list of tx indexes (where
-       * tx with certain asset is placed in the block)
-       * @param account_id of transaction creator
-       * @param height of block
-       * @param index of transaction in the block
-       * @param commands in the transaction
+       * Create several indices for block. Namely:
+       * transaction hash -> block, where this transaction is stored
+       * transaction creator -> block where his transaction is located
```
lebdron: It may be useful to mention "account_id:height -> list of tx indexes" index here as well.

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,128 @@
...
+    PostgresBlockIndex::PostgresBlockIndex(soci::session &sql)
+        : sql_(sql), log_(logger::log("PostgresBlockIndex")) {}
 
     void PostgresBlockIndex::index(
         const shared_model::interface::Block &block) {
-      const auto &height = std::to_string(block.height());
-      boost::for_each(
-          block.transactions() | boost::adaptors::indexed(0),
-          [&](const auto &tx) {
+      const auto &height = block.height();
```
lebdron: `auto` is sufficient for simple types like height.

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,128 @@
...
+      auto indexed_txs = block.transactions() | boost::adaptors::indexed(0);
+      std::string index_query = std::accumulate(
+          indexed_txs.begin(),
+          indexed_txs.end(),
+          std::string{},
+          [&height](auto query, const auto &tx) {
```
lebdron: `height` can be captured by value.

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -5,125 +5,128 @@
...
 
-#include <unordered_set>
+#include <numeric>
 
-#include <boost/range/adaptor/filtered.hpp>
+#include <boost/format.hpp>
 #include <boost/range/adaptor/indexed.hpp>
-#include <boost/range/algorithm/for_each.hpp>
-#include <boost/range/numeric.hpp>
 #include "common/types.hpp"
```
lebdron: Please remove if it is unused.

irohad/ametsuchi/impl/postgres_block_index.cpp
nickaleks: It will just replace visit_in_place with an if statement. I doubt there will be much improvement

irohad/ametsuchi/impl/postgres_block_index.cpp
nickaleks: types.hpp is used for various types in function parameters

lebdron: `interfaces/common_objects/types.hpp` is used indeed, but I am not sure about `common/types.hpp`

irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -126,7 +126,11 @@ namespace iroha {
             query += makeCreatorHeightIndex(creator_id, height, index);
             return query;
           });
-      sql_ << index_query;
+      try {
+        sql_ << index_query;
+      } catch (std::exception &e) {
```
lebdron: There was a suggestion to use `const &` due to multithreaded environment.

PR #1774, opened 2018-10-11
Copy postgres healthcheck file to docker container

neewy:
### Description of the Change

We now copy `wait-for-it.sh` file, which checks that Postgres is running, to a docker container.

### Benefits

Iroha won't start without Postgres

### Possible Drawbacks 

None

### DIFF
--- a/docker/release/Dockerfile
+++ b/docker/release/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get install -y /tmp/iroha.deb; \
 
 WORKDIR /opt/iroha_data
 
-COPY entrypoint.sh /entrypoint.sh
-COPY wait-for-it.sh /wait-for-it.sh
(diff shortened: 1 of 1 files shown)


docker/release/Dockerfile
```diff
@@ -11,5 +11,6 @@ RUN apt-get install -y /tmp/iroha.deb; \
 WORKDIR /opt/iroha_data
 
 COPY entrypoint.sh /entrypoint.sh
+COPY wait-for-it.sh /wait-for-it.sh
```
l4l: Use `COPY entrypoint.sh wait-for-it.sh /` instead, so there's no additional layer

Also checkout [this link](https://stackoverflow.com/questions/30256386/how-to-copy-multiple-files-in-one-layer-using-a-dockerfile), I might miss smth

PR #1783, opened 2018-10-14
MST: fixed mergeSignaturesInBatch

MBoldyrev:
### Description of the Change

Most likely a bug fix in iroha::mergeSignaturesInBatch. In my experiment, boost::combine iterator of different-sized containers passed far over both containers' end. This behavior could be used to crash or even exploit irohad.

### Benefits

Bugfix

### Possible Drawbacks 

Additional time to compare sizes.


irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -107,6 +107,9 @@ namespace iroha {
    * @return return if at least one new signature was inserted
    */
   bool mergeSignaturesInBatch(DataType &target, const DataType &donor) {
+    if (target->transactions().size() != donor->transactions().size()) {
+      return false;
+    }
     auto inserted_new_signatures = false;
     for (auto zip :
          boost::combine(target->transactions(), donor->transactions())) {
```
muratovv: This pull request doesn't contain a test for the fix. Pls, add it.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -107,6 +107,9 @@ namespace iroha {
    * @return return if at least one new signature was inserted
    */
   bool mergeSignaturesInBatch(DataType &target, const DataType &donor) {
+    if (target->transactions().size() != donor->transactions().size()) {
```
muratovv: Seems, the fix breaks functionality of MST. I guess it should be reworked with another approach, maybe remove `combine` or something else.

PR #1787, opened 2018-10-17
Prepare Consensus for Two-Phase Commit

nickaleks:
### Description of the Change
This pull request adds information about whether or not the commit is the same as we voted for.
This allows implementation of two phase commit

### Benefits
Can implement two phase commit

### Possible Drawbacks 
None


l4l:
Looks quite similar to #1731

irohad/network/consensus_gate.hpp
```diff
@@ -28,6 +28,21 @@ namespace shared_model {
 
 namespace iroha {
   namespace network {
+
+    enum class CommitType {
+      kVoted,  // peer voted for the same block
+      kOther   // peer has not voted for this block
```
Akvinikym: Don't like this names very much :( it is not clear without comments, what does `CommitType::kOther` means. Consider renaming, please

Akvinikym: For example, the enum class class itself can be renamed, such as `PeerVotedFor::{kThisBlock, kOtherBlock}`. This kind of radical solution, by, I think, more readable

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -125,12 +125,12 @@ TEST_F(YacGateTest, YacGateSubscriptionTest) {
 
   // verify that yac gate emit expected block
   auto gate_wrapper = make_test_subscriber<CallExact>(gate->on_commit(), 1);
-  gate_wrapper.subscribe([this](auto block) {
-    ASSERT_EQ(block, expected_block);
+  gate_wrapper.subscribe([this](auto commit) {
+    ASSERT_EQ(commit.block, expected_block);
```
Akvinikym: Also, I think, you can check the added `CommitType` here and in all changes places as well

nickaleks:
@l4l Yes, but I need these changes asap as to work further. #1731 Is big and will take some time get merged.

l4l:
<details><summary>offtop but true</summary>

</details>

test/module/irohad/consensus/yac/yac_gate_test.cpp
nickaleks: I think that will ruin the unit testing aspect since we only want to test one thing in each test.

irohad/network/consensus_gate.hpp
nickaleks: Good idea

PR #1788, opened 2018-10-17
Feature/itf mst interface

MBoldyrev:
### Description of the Change

This patch adds `sendMstState` function to ITF, making it possible to send arbitrary MstState messages to the Iroha instance.

### Benefits

Extends testing facilities of ITF.

### Possible Drawbacks 

Yet to be found.


irohad/main/application.hpp
```diff
@@ -82,7 +86,10 @@ class Irohad {
          std::chrono::milliseconds proposal_delay,
          std::chrono::milliseconds vote_delay,
          const shared_model::crypto::Keypair &keypair,
-         bool is_mst_supported);
+         bool is_mst_supported,
+         boost::optional<std::chrono::milliseconds> mst_gossip_emitting_period =
+             boost::none,
+         boost::optional<uint32_t> mst_gossip_amount_per_once = boost::none);
```
l4l: Maybe replace last three params with: `boost::optional<std::pair<milliseconds, uint32_t>>`?

MBoldyrev:
The change including adding of mock peers grows big, and I will submit it in another PR.

irohad/main/application.hpp
MBoldyrev: done with a dedicated struct

irohad/multi_sig_transactions/gossip_propagation_strategy_prams.hpp
```diff
@@ -0,0 +1,35 @@
...
+   */
+  struct GossipPropagationStrategyParams {
+    /// period of emitting data in ms
+    std::chrono::milliseconds period{kDefaultPeriod};
+
+    /// amount of data (peers) emitted per once
+    uint32_t amount{kDefaultAmount};
+  };
+
+  using OptGossipPropagationStrategyParams = boost::optional<GossipPropagationStrategyParams>;
```
l4l: Looks like a useless alias:
- hides the real type
- doesn't make it shorter :(

irohad/multi_sig_transactions/gossip_propagation_strategy_prams.hpp
```diff
@@ -0,0 +1,35 @@
...
+#ifndef IROHA_GOSSIP_PROPAGATION_STRATEGY_PARAMS_HPP
+#define IROHA_GOSSIP_PROPAGATION_STRATEGY_PARAMS_HPP
+
+#include <chrono>
+
+#include <boost/optional.hpp>
+
+// TODO: IR-1317 @l4l (02/05/18) magics should be replaced with options via
+// cli parameters
+static const std::chrono::milliseconds kDefaultPeriod =
```
l4l: It seems [there's a `constexpr` ctor](https://en.cppreference.com/w/cpp/chrono/duration/duration) for duration, so you may use it instead

irohad/main/application.cpp
```diff
@@ -59,7 +59,8 @@ Irohad::Irohad(const std::string &block_store_dir,
       max_proposal_size_(max_proposal_size),
       proposal_delay_(proposal_delay),
       vote_delay_(vote_delay),
-      is_mst_supported_(is_mst_supported),
+      is_mst_supported_(opt_mst_gossip_params),
```
l4l: Is that field is still required?

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
```diff
@@ -22,15 +22,15 @@ namespace iroha {
...
         non_visited({}),
         emit_worker(emit_worker),
-        emitent(rxcpp::observable<>::interval(steady_clock::now(), period)
-                    .map([this, amount](int) {
+        emitent(rxcpp::observable<>::interval(steady_clock::now(),
+                                              params.period)
+                    .map([this, params](int) {
                       PropagationData vec;
-                      auto range = boost::irange(0u, amount);
+                      auto range = boost::irange(0u, params.amount);
```
l4l: amount quite misleading name (at least for non-internal type), maybe you have a better idea for that?

test/framework/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ target_link_libraries(integration_framework
     query_client
     )
 
-target_include_directories(integration_framework PUBLIC ${PROJECT_SOURCE_DIR}/test)
+target_include_directories(integration_framework PUBLIC ${PROJECT_SOURCE_DIR}/test PRIVATE "${gmock_INCLUDE_DIR};${gtest_INCLUDE_DIR}")
```
l4l: That fix quite strange, ITF shouldn't depend on gtest

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -23,15 +25,41 @@
...
+    shared_model::interface::Transaction,
+    shared_model::proto::Transaction>;
+using AbstractTransactionValidator =
+    shared_model::validation::AbstractValidator<
+        shared_model::interface::Transaction>;
+using AlwaysValidTransactionValidator =
+    shared_model::validation::AlwaysValidModelValidator<
+        shared_model::interface::Transaction>;
+
+static std::string kLocalHost = "127.0.0.1";
```
l4l: Might be a good idea to wrap these consts & function in anonymous namespace

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -23,15 +25,41 @@
...
+static constexpr size_t kDefaultToriiPort = 11501;
+static constexpr size_t kDefaultInternalPort = 50541;
+
+template <size_t default_port>
+static size_t getNextPort() {
+  static size_t increment = 0;
+  return default_port + (++increment);
```
l4l: Not a real situation, but I think it might be nice to additionally check for overflow (since ports are usually lesser than ~65k) and throw exception if any

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -51,14 +79,30 @@ namespace integration_framework {
...
+        query_client_("127.0.0.1", torii_port_),
         proposal_waiting(proposal_waiting),
         block_waiting(block_waiting),
         tx_response_waiting(tx_response_waiting),
         maximum_proposal_size_(maximum_proposal_size),
+        common_objects_factory_(
+            std::make_shared<AlwaysValidProtoCommonObjectsFactory>()),
+        transaction_factory_(std::make_shared<ProtoTransactionFactory>(
+            std::unique_ptr<AbstractTransactionValidator>(
+                new AlwaysValidTransactionValidator()))),
```
l4l: Won't it work?:
```
transaction_factory_(std::make_shared<...>(std::make_unique<AlwaysValidTransactionValidator>()))
```

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -337,6 +434,13 @@ namespace integration_framework {
     return *this;
   }
 
+  IntegrationTestFramework &IntegrationTestFramework::sendMstState(
+      const shared_model::crypto::PublicKey &src_key,
```
l4l: Seems unused

test/framework/integration_framework/iroha_instance.cpp
```diff
@@ -52,17 +56,29 @@ namespace integration_framework {
     instance_->storage->insertBlock(block);
   }
 
+  void IrohaInstance::setMstGossipParams(
+      std::chrono::milliseconds mst_gossip_emitting_period,
+      uint32_t mst_gossip_amount_per_once) {
+    BOOST_ASSERT_MSG(
+        !instance_,
+        "Gossip propafation params must be set before Irohad is started!");
```
l4l: propafation -> propagation

test/framework/integration_framework/iroha_instance.cpp
```diff
@@ -52,17 +56,29 @@ namespace integration_framework {
     instance_->storage->insertBlock(block);
   }
 
+  void IrohaInstance::setMstGossipParams(
```
l4l: Quite strange interface. Why not just pass such param via ctor?

irohad/main/application.cpp
MBoldyrev: Yes, it is used two times. I do not want to substitute it with optional gossip params variable, because an explicit bool is much more expressive.

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
MBoldyrev: I do not think it is misleading. It is the amount of items being output per round.

test/framework/integration_framework/integration_test_framework.cpp
MBoldyrev: thanks, I missed it

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
l4l: It is. But it isn't implied from the `params.amount`. The one may only deduce "some amount parameter" from that

test/module/shared_model/validators/always_valid_validators.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#ifndef IROHA_ALWAYS_VALID_VALIDATORS_HPP_
+#define IROHA_ALWAYS_VALID_VALIDATORS_HPP_
+
+#include "validators/abstract_validator.hpp"
+
+namespace shared_model {
+  namespace validation {
+
+    struct AlwaysValidFieldValidator final {
```
muratovv: Add a reason to do such kind of structures and add use cases for that.

test/module/shared_model/validators/always_valid_validators.hpp
```diff
@@ -0,0 +1,76 @@
...
+      template <typename... Args>
+      void validateBatchMeta(Args...) const {}
+      template <typename... Args>
+      void validateHeight(Args...) const {}
+      template <typename... Args>
+      void validateHash(Args...) const {}
+    };
+
+    template <typename Model>
+    struct AlwaysValidModelValidator final : public AbstractValidator<Model> {
```
muratovv: Same as before

irohad/main/application.cpp
MBoldyrev: If we add a separate constructor for every optional feature - we will end up with 2^n constructors =). Also it would involve copy-pasting common parts of constructor code. So I do not see the benefits of such approach. Please clarify it if I miss something.

MBoldyrev: > I guess maybe better to add another ctor of irohad, one will not receive both is_mst_supported and params and another will receive only params without optional wrapper.

If we add a separate constructor for every optional feature - we will end up with 2^n constructors =). Also it would involve copy-pasting common parts of constructor code. So I do not see the benefits of such approach. Please clarify it if I miss something.

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -23,15 +26,44 @@
...
+namespace {
+  std::string kLocalHost = "127.0.0.1";
+  constexpr size_t kDefaultToriiPort = 11501;
+  constexpr size_t kDefaultInternalPort = 50541;
+  constexpr size_t kMaxPort = 65535;
+
+  template <size_t default_port>
+  size_t getNextPort() {
+    static size_t increment = 0;
+    assert(increment <= kMaxPort && "The maximum port value reached");
```
l4l: There's no asserts in release :(

test/framework/integration_framework/iroha_instance.cpp
```diff
@@ -52,17 +56,29 @@ namespace integration_framework {
...
+      std::chrono::milliseconds mst_gossip_emitting_period,
+      uint32_t mst_gossip_amount_per_once) {
+    BOOST_ASSERT_MSG(
+        !instance_,
+        "Gossip propagation params must be set before Irohad is started!");
+    opt_mst_gossip_params_ = iroha::GossipPropagationStrategyParams{
+        .period = mst_gossip_emitting_period,
+        .amount = mst_gossip_amount_per_once};
```
l4l: That's from C90, not C++ :(
[It seems](https://en.cppreference.com/w/cpp/language/aggregate_initialization) it going to be in 20th standard only

test/module/shared_model/validators/validators.hpp
```diff
@@ -6,6 +6,7 @@
 #ifndef IROHA_VALIDATOR_MOCKS_HPP
 #define IROHA_VALIDATOR_MOCKS_HPP
 
+#include "module/shared_model/validators/always_valid_validators.hpp"
```
l4l: Looks redundant

test/module/irohad/multi_sig_transactions/gossip_propagation_strategy_test.cpp
```diff
@@ -79,8 +79,11 @@ PropagationData subscribeAndEmit(boost::optional<PropagationData> data,
   EXPECT_CALL(*pbfactory, createPeerQuery())
       .WillRepeatedly(testing::Return(boost::make_optional(
           std::shared_ptr<iroha::ametsuchi::PeerQuery>(query))));
+  iroha::GossipPropagationStrategyParams gossip_params;
+  gossip_params.period = period;
```
l4l: Since fields have different type it's ok to use just initializer list, eg:
`iroha::GossipPropagationStrategyParams gossip_params{period, amount};`

Same for other places

MBoldyrev: I do not like it in such form, when you do not specify the names of members explicitly. If somebody changes the structure and forgets to check all such initializations, he might introduce an error.

MBoldyrev: But the C90 variant of explicit member initialization is wonderful, and I wait for it in C++20 (as you pointed out above).

l4l: > If somebody changes the structure and forgets to check all such initializations, he might introduce an error.

The types are different, so that probably won't be happened

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -131,6 +188,21 @@ namespace integration_framework {
       const shared_model::crypto::Keypair &keypair) {
     log_->info("init state");
     // peer initialization
+    common_objects_factory_->createPeer(kLocalHost, keypair.publicKey())
```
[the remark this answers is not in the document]
MBoldyrev: Yes, but nevertheless I think ITF should be able to create and send any kind of junk data, and that it does not need any validation on its side. Whatever test data we want to use, ITF should not stop us by validation rules.

test/module/irohad/multi_sig_transactions/gossip_propagation_strategy_test.cpp
MBoldyrev: And what about using adding same types, or having implicit conversions? I would rather avoid that risks.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -38,6 +41,11 @@ namespace shared_model {
     class Block;
   }
 }  // namespace shared_model
+namespace iroha {
+  namespace network {
+    class MstTransportGrpc;
```
lebdron: It does not seem to be used in this file, is it required here?

test/framework/integration_framework/iroha_instance.cpp
```diff
@@ -52,17 +56,30 @@ namespace integration_framework {
     instance_->storage->insertBlock(block);
   }
 
+  void IrohaInstance::setMstGossipParams(
+      std::chrono::milliseconds mst_gossip_emitting_period,
+      uint32_t mst_gossip_amount_per_once) {
+    BOOST_ASSERT_MSG(
+        !instance_,
```
lebdron:
```suggestion
        not instance_,
```
since this notation is mostly used instead in the project.

irohad/main/application.cpp
MBoldyrev: > It is more error-prone to track two variables though, because of the opt_mst_gossip_params => is_mst_supported_ relation.
It is a part of internal logic stated explicitly in the constructor: we enable the MST support if we got parameters for its initialization. After that we use each of the two variables only in its own sense, and I think it is less error prone than if we would check the parameters presence each time we need to know whether we support MST, as that would be  real mess.

irohad/main/application.cpp
```diff
@@ -322,10 +324,7 @@ void Irohad::initMstProcessor() {
     // TODO: IR-1317 @l4l (02/05/18) magics should be replaced with options via
```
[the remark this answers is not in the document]
MBoldyrev: Thanks! I already placed this todo where the magics are initialized (now they are called defaults), but forgot to remove it from here.

irohad/main/application.hpp
```diff
@@ -18,6 +18,7 @@
 #include "main/impl/consensus_init.hpp"
 #include "main/impl/ordering_init.hpp"
 #include "main/server_runner.hpp"
+#include "multi_sig_transactions/gossip_propagation_strategy_prams.hpp"
```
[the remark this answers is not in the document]
MBoldyrev: sure, thanks

test/framework/integration_framework/integration_test_framework.hpp
MBoldyrev: Thanks, that is odd.

irohad/multi_sig_transactions/impl/gossip_propagation_strategy.cpp
MBoldyrev: changed the names (see 9611ee1)

PR #1789, opened 2018-10-17
Switch from java-8-oracle to java-9-openjdk

bakhtin:

### Description of the Change
Switch Java version in develop Dockerfile from Oracle Java 8 to OpenJDK 9. Previous version is no more available by the URL (returns 404)

### Benefits
develop Dockerfile build in now passing successfully

### Possible Drawbacks 
A small chance that bindings generated with Oracle Java can be incompatible with OpenJDK

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -411,7 +411,7 @@ pipeline {
           }
           agent { label 'docker-build-agent' }
           environment {
-            JAVA_HOME = "/usr/lib/jvm/java-8-oracle"
+            JAVA_HOME = "/usr/lib/jvm/java-9-openjdk"
           }
           steps {
             script {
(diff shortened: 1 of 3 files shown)


bakhtin:
Do you use bindings generated with Java from that image? Or use this image directly?

bakhtin:
As long as Java update would not break anything for SORA team, I'll keep this version bump. Having the most recent version of the software is always a good practice.

neewy:
@bakhtin did you check that bindings, generated with this version, actually work?

bakhtin:
I checked it can be compiled. Although, not use them to interact with Iroha instance

bakhtin:
`ppa:webupd8team/java` is now live (it was down for some time) again. And Java 9 has reached EOL. As Java 8 is LTS we'll stick to this version for now. I'm closing this PR

PR #1791, opened 2018-10-18
Fix bug with 100% peer load

Akvinikym:
Signed-off-by: Akvinikym <anarant12@gmail.com>

### Description of the Change

Bug with 100%+ Iroha node overload was appearing because of block streaming service, which was based into RxCpp blocking observable. In version, used in current Iroha, it had a serious issue - spinlock - which was addressed here and fixed August 6 2018. So, this pull request updated a version of RxCpp up to the current one.

### Benefits

No more useless node overload.

### Possible Drawbacks 

With the current RxCpp version it seems that it is not sufficient to `#include <rxcpp/rx-observable.hpp>`, but a whole `#include <rxcpp/rx.hpp>`, which can theoretically lead to increased compilation time.

### Usage Examples or Tests

The use case can be tested manually through Python's scripts `blocks-query.py`, which is to launched first, and then `tx-example.py`. Because of subscription to blocks in the first script, the 100% load appeared before fixing the bug. Now, it isn't.

### DIFF
--- a/cmake/Modules/Findrxcpp.cmake
+++ b/cmake/Modules/Findrxcpp.cmake
@@ -9,7 +9,7 @@ find_package_handle_standard_args(rxcpp DEFAULT_MSG
 
 
 set(URL https://github.com/Reactive-Extensions/rxcpp.git)
-set(VERSION 1b2e0589f19cb34d8cd58803677701dcf2161876)
+set(VERSION a7d5856385f126e874db6010d9dbfd37290c61de)
 set_target_description(rxcpp "Library for reactive programming" ${URL} ${VERSION})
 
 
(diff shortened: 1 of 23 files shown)


cmake/Modules/Findrxcpp.cmake
```diff
@@ -9,7 +9,7 @@ find_package_handle_standard_args(rxcpp DEFAULT_MSG
 
 
 set(URL https://github.com/Reactive-Extensions/rxcpp.git)
-set(VERSION 1b2e0589f19cb34d8cd58803677701dcf2161876)
+set(VERSION a7d5856385f126e874db6010d9dbfd37290c61de)
```
l4l: Please add either a version comment, or why this particular commit is used (if it has no tag)

l4l: Also would be nice to add version (if any) to other places, e.g dockerfiles

l4l: Comments, yeah

PR #1792, opened 2018-10-18
Fix/empty block appliance

Akvinikym:
### Description of the Change

Final part of fixing synchronization - after this PR uncynchronized peer will be able to download and apply empty blocks as well. Fixing it involved allowing ProtoBlockFactory to create empty block.

### Benefits

Bug is fixed.

### Possible Drawbacks 

None.


shared_model/validators/transactions_collection/transactions_collection_validator.cpp
```diff
@@ -19,22 +19,28 @@
...
+    Answer TransactionsCollectionValidator<TransactionValidator,
+                                           CollectionCanBeEmpty>::
+        validateImpl(const interface::types::TransactionsForwardCollectionType
+                         &transactions,
+                     Validator &&validator) const {
       Answer res;
       ReasonsGroupType reason;
       reason.first = "Transaction list";
 
       if (boost::empty(transactions)) {
```
lebdron: `if (boost::empty(transactions) and not CollectionCanBeEmpty)` looks simpler than a nested `if`

PR #1793, opened 2018-10-18
Fix race condition on lazy fields in ProtoTransactionResponse

igor-egorov:

### Description of the Change

We encountered one more race condition caused by LazyInitializer. Removed.

RC was occurred when publishStatus was called for batch of transactions.

### Benefits

Improved stability.

### Possible Drawbacks 

Need to remove all LazyInitializers from the whole project.

### Usage Examples or Tests

```
test_bin/pipeline_test --gtest_filter=PipelineIntegrationTest.SendTxSequenceAwait --gtest_repeat=500
test_bin/shared_proto_tx_response_test
```


test/module/shared_model/backend_proto/shared_proto_tx_response_test.cpp
```diff
@@ -46,3 +48,32 @@ TEST(ProtoTxResponse, TxResponseLoad) {
               shared_model::crypto::Hash(hash));
   });
 }
+
+/**
+ * @given TransactionResponse that previously had lazy fields
```
l4l: "uninitialized lazy fields"

igor-egorov:
to @l4l,
I don't think that it is possible to construct the same object at the same time from two threads
so now, it is kind of safe

test/module/shared_model/backend_proto/shared_proto_tx_response_test.cpp
igor-egorov: The point is not on their initialisation state. The point is on presence of lazy fields.

test/module/shared_model/backend_proto/shared_proto_tx_response_test.cpp
```diff
@@ -46,3 +48,32 @@ TEST(ProtoTxResponse, TxResponseLoad) {
...
+ * @when those lazy fields are simultaneously accessed
+ * @then there is no race condition and segfaults
+ */
+TEST(TxResponse, SafeToReadFromMultipleThreads) {
+  iroha::protocol::ToriiResponse response;
+  const std::string hash = "123";
```
l4l: Seems that var used only once so it might be inlined

test/module/shared_model/backend_proto/shared_proto_tx_response_test.cpp
```diff
@@ -46,3 +48,32 @@ TEST(ProtoTxResponse, TxResponseLoad) {
...
+  response.set_tx_hash(hash);
+  response.set_tx_status(iroha::protocol::TxStatus::COMMITTED);
+  auto model_response = shared_model::proto::TransactionResponse(response);
+
+  const auto num_threads = 20;
```
l4l: Can be declared later

test/module/shared_model/backend_proto/shared_proto_tx_response_test.cpp
```diff
@@ -46,3 +48,32 @@ TEST(ProtoTxResponse, TxResponseLoad) {
...
+
+  auto multiple_access = [&model_response] {
+    // old good way to cause race condition on lazy fields
+    ASSERT_TRUE(model_response == model_response);
```
l4l: Is it legal to use `ASSERT_EQ` here?

test/module/shared_model/backend_proto/shared_proto_tx_response_test.cpp
igor-egorov: You are right. It could be inlined. I would like leave it as is, since it does not violate any of our guides.

test/module/shared_model/backend_proto/shared_proto_tx_response_test.cpp
igor-egorov: It looks like we can substitute ASSERT_TRUE with ASSERT_EQ, but ASSERT_TRUE was intentionally used here to be independent of the implementation of ASSERT_EQ. Usage of  ASSERT_TRUE can make us sure that operator== is called.

l4l: What else can be called here?

igor-egorov: Evaluation order can be mangled by gtest stuff. We need to cause fields initialization as much simultaneously as it is possible

PR #1795, opened 2018-10-19
Feature/mst prop enable

l4l:
### Description of the Change

This pr changes behavior of mst enablement flag. Before it is used to completely disable MST component. And for now it's remain proper handling mst txes but without sharing its state among peers.

Also mst_network_test is added to ensure that nothing was sent to the network

### Benefits

MST features can be locally used even when it disabled

### Possible Drawbacks 

Code duplication in test :(

### Usage Examples or Tests

`make mst_network_test && test_bin/mst_network_test`

### Alternate Designs

Instead of making copy of additional Irohad deriviative we might consider using builder instead 89860fb


irohad/multi_sig_transactions/mst_propagation_strategy_stub.hpp
```diff
@@ -0,0 +1,18 @@
...
+#ifndef IROHA_PROPAGATION_STRATEGY_STUB_HPP
+#define IROHA_PROPAGATION_STRATEGY_STUB_HPP
+
+#include "multi_sig_transactions/mst_propagation_strategy.hpp"
+
+namespace iroha {
+  class PropagationStrategyStub : public PropagationStrategy {
+    rxcpp::observable<PropagationStrategy::PropagationData> emitter() override;
+  };
+
```
Akvinikym: Extra line

irohad/multi_sig_transactions/transport/mst_transport_stub.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include "network/impl/async_grpc_client.hpp"
+
+namespace iroha {
+  namespace network {
+    class MstTransportStub : public MstTransportGrpc {
+     public:
+      using MstTransportGrpc::MstTransportGrpc;
+
+      void sendState(const shared_model::interface::Peer &to,
+                     ConstRefState providing_state) override {}
```
Akvinikym: Can't arguments' names be avoided, as we don't use them?

test/framework/integration_framework/test_mstnet_irohad.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include "multi_sig_transactions/storage/mst_storage_impl.hpp"
+
+namespace integration_framework {
+
+  struct MockTransportGrpc : public iroha::network::MstTransportGrpc {
+    using MstTransportGrpc::MstTransportGrpc;
+    MOCK_METHOD2(sendState,
+                 void(const shared_model::interface::Peer &,
+                      iroha::ConstRefState));
+  };
```
Akvinikym: Maybe it's better to put it into a separate file?

test/framework/integration_framework/test_mstnet_irohad.hpp
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Class for integration testing of Irohad.
```
Akvinikym: But isn't it about MST integration testing of Irohad?

test/framework/integration_framework/test_mstnet_irohad.hpp
```diff
@@ -0,0 +1,114 @@
...
+      } else {
+        mst_propagation = std::make_shared<iroha::PropagationStrategyStub>();
+      }
+
+      auto mst_time = std::make_shared<iroha::MstTimeProviderImpl>();
+      auto fair_mst_processor = std::make_shared<iroha::FairMstProcessor>(
+          mst_transport, mst_storage, mst_propagation, mst_time);
+      mst_processor = fair_mst_processor;
+      mst_transport->subscribe(fair_mst_processor);
+      log_->info("[Init] => MST processor");
```
Akvinikym: Everything after `mst_transport = mock;` is the same as in real Irohad, I assume. Maybe it's possible to put this body into a static method, which will be shared among both real and this Irohades?

test/integration/network/mst_network_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+  shared_model::proto::Block block;
+  const shared_model::interface::types::AssetIdType kAssetId =
+      IntegrationTestFramework::kAssetName + "#"
+      + IntegrationTestFramework::kDefaultDomain;
+  std::unique_ptr<TestMstnetIrohad> instance;
+};
+
+TEST_F(MstNetwork, NoPropagationOnDisabled) {
+  init([](auto mock) {
+    // shouldn't be called
```
Akvinikym: Why not then putting `EXPECT_CALL(*mock, sendState(_, _)).Times(0);`, as you did in the next test?

irohad/multi_sig_transactions/transport/mst_transport_stub.hpp
```diff
@@ -0,0 +1,32 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IROHA_MST_TRANSPORT_STUB_HPP
+#define IROHA_MST_TRANSPORT_STUB_HPP
+
+#include "mst.grpc.pb.h"
+#include "network/mst_transport.hpp"
+
+#include <google/protobuf/empty.pb.h>
```
muratovv: Pb includes are not allowed here.

irohad/multi_sig_transactions/transport/mst_transport_stub.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include "interfaces/common_objects/common_objects_factory.hpp"
+#include "interfaces/iroha_internal/abstract_transport_factory.hpp"
+#include "interfaces/iroha_internal/transaction_batch_factory.hpp"
+#include "interfaces/iroha_internal/transaction_batch_parser.hpp"
+#include "logger/logger.hpp"
+#include "network/impl/async_grpc_client.hpp"
+
+namespace iroha {
+  namespace network {
+    class MstTransportStub : public MstTransportGrpc {
```
muratovv: Seems that enough to inherit from `iroha::network::MstTransport`. There is no reason to do additional service which will not work.

test/framework/integration_framework/test_mstnet_irohad.hpp
```diff
@@ -0,0 +1,114 @@
...
+    TestMstnetIrohad(
+        const std::string &block_store_dir,
+        const std::string &pg_conn,
+        size_t torii_port,
+        size_t internal_port,
+        size_t max_proposal_size,
+        std::chrono::milliseconds proposal_delay,
+        std::chrono::milliseconds vote_delay,
+        const shared_model::crypto::Keypair &keypair,
+        bool is_mst_supported,
```
muratovv: Why is not enough to pass is_mst_supported=true?

PR #1799, opened 2018-10-21
Remove proto dependency in Posgtres Executors

l4l:
### Description of the Change

Removed protobuf dependency from PostgresQueryExecutor and PostgresCommandExecutor.

### Benefits

Weaker coupling of ametsuchi on proto

### Possible Drawbacks 

None

### DIFF
--- a/irohad/ametsuchi/impl/postgres_command_executor.cpp
+++ b/irohad/ametsuchi/impl/postgres_command_executor.cpp
@@ -8,7 +8,6 @@
 #include <soci/postgresql/soci-postgresql.h>
 #include <boost/format.hpp>
 #include "ametsuchi/impl/soci_utils.hpp"
-#include "backend/protobuf/permissions.hpp"
 #include "cryptography/public_key.hpp"
 #include "interfaces/commands/add_asset_quantity.hpp"
 #include "interfaces/commands/add_peer.hpp"
@@ -171,22 +170,25 @@ namespace {
 
   std::string missRolePerm(
       shared_model::interface::types::AccountIdType account,
-      shared_model::interface::permissions::Role perm) {
+      shared_model::interface::permissions::Role perm,
+      std::shared_ptr<shared_model::interface::PermissionToString>
+          perm_converter) {
     return (boost::format("command validation failed: account %s"
                           " does not have permission %s (role)")
-            % account % shared_model::proto::permissions::toString(perm))
+            % account % perm_converter->toString(perm))
         .str();
   }
 
   std::string missGrantablePerm(
       shared_model::interface::types::AccountIdType account,
       shared_model::interface::types::AccountIdType permittee,
-      shared_model::interface::permissions::Grantable perm) {
+      shared_model::interface::permissions::Grantable perm,
+      std::shared_ptr<shared_model::interface::PermissionToString>
+          perm_converter) {
     return (boost::format(
                 "command validation failed: account %s"
                 " does not have permission %s (grantable) for account %s")
-            % account % shared_model::proto::permissions::toString(perm)
-            % permittee)
+            % account % perm_converter->toString(perm) % permittee)
         .str();
   }
 
@@ -194,13 +196,14 @@ namespace {
       shared_model::interface::types::AccountIdType account,
       shared_model::interface::types::AccountIdType permittee,
       shared_model::interface::permissions::Role role_perm,
-      shared_model::interface::permissions::Grantable grantable_perm) {
+      shared_model::interface::permissions::Grantable grantable_perm,
+      std::shared_ptr<shared_model::interface::PermissionToString>
+          perm_converter) {
     return (boost::format("command validation failed: account %s"
                           " does not have permission %s (role)"
                           " and permission %s (grantable) for account %s")
-            % account % shared_model::proto::permissions::toString(role_perm)
-            % shared_model::proto::permissions::toString(grantable_perm)
-            % permittee)
+            % account % perm_converter->toString(role_perm)
+            % perm_converter->toString(grantable_perm) % permittee)
         .str();
   }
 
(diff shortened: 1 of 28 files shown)


irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -171,36 +170,40 @@ namespace {
 
   std::string missRolePerm(
```
Akvinikym: I would move these three functions from anonymous namespace to `PostgresCommandExecutor` so that `perm_converter` will not be passed in every function call. This looks a little ugly from my point of view :(

irohad/main/application.cpp
```diff
@@ -109,12 +110,15 @@ void Irohad::initStorage() {
...
+  perm_converter_ =
+      std::make_shared<shared_model::proto::ProtoPermissionToString>();
   auto block_converter =
       std::make_shared<shared_model::proto::ProtoBlockJsonConverter>();
   auto storageResult = StorageImpl::create(block_store_dir_,
                                            pg_conn_,
                                            common_objects_factory_,
-                                           std::move(block_converter));
+                                           std::move(block_converter),
+                                           perm_converter_);
```
Akvinikym: For now, it isn't needed anywhere in `application` except this initialization. Maybe make it not a class variable, but a local one, and move?

shared_model/backend/protobuf/impl/proto_permission_to_string.cpp
```diff
@@ -0,0 +1,50 @@
...
+      std::vector<std::string> v;
+      for (size_t i = 0; i < set.size(); ++i) {
+        auto perm = static_cast<interface::permissions::Grantable>(i);
+        if (set.test(perm)) {
+          v.push_back(toString(perm));
+        }
+      }
+      return v;
+    }
+  }  // namespace proto
```
Akvinikym: New line should be added

shared_model/interfaces/permission_to_string.hpp
```diff
@@ -0,0 +1,48 @@
...
+#define PERMISSION_TO_STRING_HPP
+
+#include <string>
+#include <vector>
+
+#include "interfaces/permissions.hpp"
+
+namespace shared_model {
+  namespace interface {
+    class PermissionToString {
```
Akvinikym: I don't like name `PermissionToString` for a class :( I think, it should some noun, for example, `PermissionStringifier` or `PermissionConverter`, though the second variant is less precise

l4l: `PermissionStringifier` sounds even uglier :( That class is just a method collection w/o any actual data. So imo the name is fine

nickaleks:
Also, if possible, remove shared_model_proto_backend from ametsuchi dependencies in CMake

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1009,7 +977,7 @@ namespace iroha {
                     % permittee_account_id
                     % creator_account_id_
                     // TODO(@l4l) 26/06/18 need to be simplified at IR-1479
```
Akvinikym: Isn't IR-1479 about this PR? If so, ToDos should be removed

irohad/ametsuchi/impl/postgres_query_executor.cpp
```diff
@@ -130,13 +129,16 @@ namespace {
...
-  auto notEnoughPermissionsResponse(Roles... roles) {
-    return [roles...] {
+  auto notEnoughPermissionsResponse(
+      std::shared_ptr<shared_model::interface::PermissionToString>
+          perm_converter,
+      Roles... roles) {
+    return [perm_converter, roles...] {
       std::string error = "user must have at least one of the permissions: ";
       for (auto role : {roles...}) {
         // TODO [IR-1758] Akvinikym 12.10.18: get rid of this protobuf
```
Akvinikym: And this ToDo and task as well

irohad/main/application.cpp
```diff
@@ -109,12 +110,15 @@ void Irohad::initStorage() {
   common_objects_factory_ =
       std::make_shared<shared_model::proto::ProtoCommonObjectsFactory<
           shared_model::validation::FieldValidator>>();
+  auto perm_converter_ =
```
Akvinikym: Underscore should be removed in variable's name

irohad/main/application.hpp
```diff
@@ -25,6 +25,7 @@
 #include "interfaces/common_objects/common_objects_factory.hpp"
 #include "interfaces/iroha_internal/query_response_factory.hpp"
 #include "interfaces/iroha_internal/transaction_batch_factory.hpp"
+#include "interfaces/permission_to_string.hpp"
```
Akvinikym: Don't need it

irohad/torii/processor/query_processor_impl.hpp
```diff
@@ -52,6 +52,8 @@ namespace iroha {
       std::shared_ptr<iroha::PendingTransactionStorage> pending_transactions_;
       std::shared_ptr<shared_model::interface::QueryResponseFactory>
           response_factory_;
+      std::shared_ptr<shared_model::interface::PermissionToString>
```
Akvinikym: A strange one, where is it initialized and used?

irohad/ametsuchi/query_executor_factory.hpp
```diff
@@ -10,6 +10,7 @@
 
 #include "ametsuchi/query_executor.hpp"
 #include "interfaces/iroha_internal/query_response_factory.hpp"
+#include "interfaces/permission_to_string.hpp"
```
Akvinikym: Do you need it here?

irohad/torii/processor/query_processor_impl.hpp
l4l: https://github.com/hyperledger/iroha/blob/a0da9d9f5ba1952ee37a1ebcdb19fe41a3317d7d/irohad/ametsuchi/impl/storage_impl.cpp#L58
https://github.com/hyperledger/iroha/blob/a0da9d9f5ba1952ee37a1ebcdb19fe41a3317d7d/irohad/ametsuchi/impl/storage_impl.cpp#L74-L76
https://github.com/hyperledger/iroha/blob/a0da9d9f5ba1952ee37a1ebcdb19fe41a3317d7d/irohad/ametsuchi/impl/storage_impl.cpp#L102
https://github.com/hyperledger/iroha/blob/a0da9d9f5ba1952ee37a1ebcdb19fe41a3317d7d/irohad/ametsuchi/impl/storage_impl.cpp#L156

PR #1803, opened 2018-10-24
Use proposal factory for OnDemandOrderingService

nickaleks:
### Description of the Change
Instead of manually constructing protobuf object, add a factory for on demand ordering service.

### Benefits
No dependency on protobuf

### Possible Drawbacks 
none


test/module/irohad/ordering/on_demand_os_test.cpp
```diff
@@ -179,3 +204,24 @@ TEST_F(OnDemandOsTest, EraseReject) {
         {reject_round.block_round, i + 1 - proposal_limit}));
   }
 }
+
+/**
+ * @given initialized on-demand OS @and some transactions are sent to is
```
l4l: to is -> to **it**

PR #1809, opened 2018-10-25
pretty printing YAC messages

MBoldyrev:
### Description of the Change

toString() method for iroha::consensus::yac::CommitMessage, iroha::consensus::yac::RejectMessage and their members.

### Benefits

pretty logs

### Possible Drawbacks 

more hearer includes

### DIFF
--- a/shared_model/utils/string_builder.hpp
+++ b/shared_model/utils/string_builder.hpp
@@ -63,6 +63,25 @@ namespace shared_model {
       }
 
       /**
+       * Appends a new named collection to string
+       * @tparam Collection - type of collection
+       * @tparam Transform - type of transformation function
+       * @param name - field name to append
+       * @param c - collection to append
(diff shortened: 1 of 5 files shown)


shared_model/utils/string_builder.hpp
```diff
@@ -62,6 +62,25 @@ namespace shared_model {
...
+      /**
+       * Appends a new named collection to string
+       * @tparam Collection - type of collection
+       * @tparam Transform - type of transformation function
+       * @param name - field name to append
+       * @param c - collection to append
+       * @param t - transformation function
+       */
+      template <typename Collection, typename Transform>
+      PrettyStringBuilder &appendAll(
```
Akvinikym: Can't these overloads of `appendAll` be merged into one function by giving the `name` argument default value of empty string and checking it in the body?

PR #1812, opened 2018-10-26
ITF: consensus gate testing interface

MBoldyrev:
### Description of the Change

Added an interface to send YAC states to tested iroha instance and to listen for YacGate on_commit messages.

### Benefits

The ability to test consensus.

### Possible Drawbacks 

To be found.

### DIFF
--- a/test/framework/integration_framework/integration_test_framework.cpp
+++ b/test/framework/integration_framework/integration_test_framework.cpp
@@ -326,6 +326,11 @@ namespace integration_framework {
         ->onExpiredBatches();
   }
 
+  rxcpp::observable<iroha::network::Commit>
+  IntegrationTestFramework::getYacOnCommitObservable() {
+    return iroha_instance_->getIrohaInstance()->getConsensusGate()->on_commit();
+  }
+
   IntegrationTestFramework &
   IntegrationTestFramework::subscribeForAllMstNotifications(
       std::shared_ptr<iroha::network::MstTransportNotification> notification) {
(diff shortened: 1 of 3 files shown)


test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -324,6 +326,11 @@ namespace integration_framework {
         ->onExpiredBatches();
   }
 
+  rxcpp::observable<iroha::network::Commit>
+  IntegrationTestFramework::getYacOnCommitObservable() {
+    return iroha_instance_->getIrohaInstance()->getConsensusGate()->on_commit();
```
l4l: Maybe it's better use push to the queue instead of sharing observable at the interface?

MBoldyrev: I think observables are more flexible as they allow (for example) ignoring, filtering and tee'ing objects, and much more.

test/framework/integration_framework/fake_peer/yac_network_notifier.cpp
```diff
@@ -11,12 +11,6 @@
 
 namespace integration_framework {
 
-  void YacNetworkNotifier::subscribe(
-      const std::shared_ptr<iroha::consensus::yac::NetworkImpl>
-          &yac_transport) {
-    yac_transport->subscribe(shared_from_this());
```
igor-egorov: Since the method `shared_from_this()` is no more used. Can we remove `std::enable_shared_from_this` inheritance from `iroha/test/framework/integration_framework/fake_peer/yac_network_notifier.hpp` ?

PR #1814, opened 2018-10-27
refactor query response factory to accept some sm objects by its para…

victordrobny:
### Description of the Change

Refactor query response factory to accept some shared model objects by its parameters. Also refactor Postgres query executor to use new methods in order to simplify its calls.

### DIFF
--- a/shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
+++ b/shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
@@ -52,22 +52,20 @@ namespace {
 
 std::unique_ptr<shared_model::interface::QueryResponse>
 shared_model::proto::ProtoQueryResponseFactory::createAccountAssetResponse(
-    const std::vector<interface::types::AccountIdType> account_ids,
-    const std::vector<interface::types::AssetIdType> asset_ids,
-    const std::vector<shared_model::interface::Amount> balances,
+    std::vector<std::tuple<interface::types::AccountIdType,
+                           interface::types::AssetIdType,
+                           shared_model::interface::Amount>> assets,
     const crypto::Hash &query_hash) const {
   return createQueryResponse(
-      [account_ids = std::move(account_ids),
-       asset_ids = std::move(asset_ids),
-       balances = std::move(balances)](
+      [assets = std::move(assets)](
           iroha::protocol::QueryResponse &protocol_query_response) {
         iroha::protocol::AccountAssetResponse *protocol_specific_response =
             protocol_query_response.mutable_account_assets_response();
-        for (size_t i = 0; i < account_ids.size(); i++) {
+        for (size_t i = 0; i < assets.size(); i++) {
           auto *asset = protocol_specific_response->add_account_assets();
-          asset->set_account_id(std::move(account_ids.at(i)));
-          asset->set_asset_id(std::move(asset_ids.at(i)));
-          asset->set_balance(balances.at(i).toStringRepr());
+          asset->set_account_id(std::move(std::get<0>(assets.at(i))));
+          asset->set_asset_id(std::move(std::get<1>(assets.at(i))));
+          asset->set_balance(std::get<2>(assets.at(i)).toStringRepr());
         }
       },
       query_hash);
--- a/shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
+++ b/shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
@@ -69,6 +69,29 @@ shared_model::proto::ProtoQueryResponseFactory::createAccountAssetResponse(
 }
 
 std::unique_ptr<shared_model::interface::QueryResponse>
+shared_model::proto::ProtoQueryResponseFactory::createAccountAssetResponse(
+    const std::vector<interface::types::AccountIdType> &account_ids,
+    const std::vector<interface::types::AssetIdType> &asset_ids,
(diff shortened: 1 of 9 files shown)


shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
```diff
@@ -68,6 +68,29 @@ shared_model::proto::ProtoQueryResponseFactory::createAccountAssetResponse(
       query_hash);
 }
 
+std::unique_ptr<shared_model::interface::QueryResponse>
+shared_model::proto::ProtoQueryResponseFactory::createAccountAssetResponse(
+    const std::vector<interface::types::AccountIdType> &account_ids,
```
Akvinikym: Better accept these arguments in touched functions by value, otherwise moving them here will make no sense

shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
```diff
@@ -68,6 +68,29 @@ shared_model::proto::ProtoQueryResponseFactory::createAccountAssetResponse(
...
+  return createQueryResponse(
+      [account_ids = std::move(account_ids),
+       asset_ids = std::move(asset_ids),
+       balances = std::move(balances)](
+          iroha::protocol::QueryResponse &protocol_query_response) {
+        iroha::protocol::AccountAssetResponse *protocol_specific_response =
+            protocol_query_response.mutable_account_assets_response();
+        for (decltype(account_ids.size()) i = 0; i < account_ids.size(); i++) {
+          auto *asset = protocol_specific_response->add_account_assets();
+          asset->set_account_id(account_ids.at(i));
```
Akvinikym: I am not sure, but won't it be useful, if you `std::move(account_ids.at(i))` and all other values?

shared_model/interfaces/iroha_internal/query_response_factory.hpp
```diff
@@ -44,17 +45,30 @@ namespace shared_model {
...
 
+      /**
+       * Create response for account asset query
+       * @param account_ids of assets to be inserted into the response
+       * @param asset_ids of assets to be inserted into the response
+       * @param balances of assets to be inserted into the response
+       * @param query_hash - hash of the query, for which response is created
+       * @return account asset response
+       */
+      virtual std::unique_ptr<QueryResponse> createAccountAssetResponse(
```
Akvinikym: Why haven't you removed the previous versions of functions? Those, which accept objects

test/module/shared_model/backend_proto/proto_query_response_factory_test.cpp
```diff
@@ -68,8 +68,13 @@ TEST_F(ProtoQueryResponseFactoryTest, CreateAccountAssetResponse) {
   const std::string kAccountId = "doge@meme";
   const std::string kAssetId = "dogecoin#iroha";
 
+  std::vector<std::unique_ptr<shared_model::interface::QueryResponse>>
+      query_responses;
   std::vector<std::unique_ptr<shared_model::interface::AccountAsset>> assets,
       assets_test_copy;
+  std::vector<std::string> account_ids;
```
Akvinikym: `std::vector<AccountIdType>` etc?

shared_model/interfaces/iroha_internal/query_response_factory.hpp
victordrobny: Because they are stil used in some tests

shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
```diff
@@ -52,17 +52,22 @@ namespace {
 
 std::unique_ptr<shared_model::interface::QueryResponse>
 shared_model::proto::ProtoQueryResponseFactory::createAccountAssetResponse(
-    std::vector<std::unique_ptr<shared_model::interface::AccountAsset>> assets,
+    const std::vector<interface::types::AccountIdType> account_ids,
+    const std::vector<interface::types::AssetIdType> asset_ids,
+    const std::vector<shared_model::interface::Amount> balances,
```
Akvinikym: I don't think these parameters should be const. Purpose of accepting them by-value was allow moving them without copying inside the functions, but now they are getting copied to lambda because of const modifier. So, I think, const should be removed from here and other functions, which were touched

kamilsa: Please rework with vector of tuples. Otherwise it is not guaranteed by interface that all vectors have the same size

shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
```diff
@@ -52,17 +52,20 @@ namespace {
...
           iroha::protocol::QueryResponse &protocol_query_response) {
         iroha::protocol::AccountAssetResponse *protocol_specific_response =
             protocol_query_response.mutable_account_assets_response();
-        for (const auto &asset : assets) {
-          *protocol_specific_response->add_account_assets() =
-              static_cast<shared_model::proto::AccountAsset *>(asset.get())
-                  ->getTransport();
+        for (size_t i = 0; i < assets.size(); i++) {
+          auto *asset = protocol_specific_response->add_account_assets();
+          asset->set_account_id(std::move(std::get<0>(assets.at(i))));
```
Akvinikym: Won't it be better from performance point of view, if you `auto asset = std::move(assets.at(i));`? I think, three `at(i)`-s are an overhead

PR #1815, opened 2018-10-28
Store rejected transactions' hashes in blockchain

MBoldyrev:
### Description of the Change

The hashes of transactions rejected after proposal verification by simulator are passed to block creator along with the verified proposal. Then these hashes are added to a newly created field in the block.

### Benefits

The rejected transactions' hashes stored in the blockchain let us guarantee a client that certain transaction was rejected and will never be executed.

### Possible Drawbacks 

A malicious peer can bloat the blockchain with tons of fake invalid transactions' hashes.


test/module/irohad/simulator/simulator_test.cpp
```diff
@@ -298,8 +308,8 @@ TEST_F(SimulatorTest, RightNumberOfFailedTxs) {
 
   EXPECT_CALL(*query, getTopBlockHeight()).WillOnce(Return(2));
 
-  EXPECT_CALL(*validator, validate(_, _))
-      .WillOnce(Return(std::make_pair(verified_proposal, tx_errors)));
+  EXPECT_CALL(*validator, validateProxy(_, _))
+      .WillOnce(Return(verified_proposal_and_errors));
```
kamilsa: can't we use invoke here and return unique ptr instead of using proxy?

shared_model/schema/block.proto
```diff
@@ -12,10 +12,16 @@ message Block {
...
+    uint32 tx_number = 2;  ///< The number of accepted transactions inside.
+                           ///< Maximum 16384 or 2^14.
+    uint64 height = 3;          ///< The current block number in a ledger.
+    bytes prev_block_hash = 5;  ///< Previous block hash.
     uint64 created_time = 6;
+
+    /// Hashes of the transactions that did not pass stateful validation.
+    /// Needed here to be able to guarantee the client that this transaction
+    /// was not and will never be executed.
+    repeated bytes rejected_transactions_hashes = 7;
```
kamilsa: can we make this field optional? check out batch meta in transaction.proto. That way genesis block will still work, otherwise please check that genesis block in example folder if it launches with iroha version in your branch

MBoldyrev: I think this field already has the ability of optional value to contain nothing (as an empty sequence). The functions regarding genesis block seem to be unchanged, as the new block should deserialize properly without the new field (please see test `[irohad_test](https://github.com/hyperledger/iroha/blob/feature/rejected-txs-in-block/test/system/irohad_test.cpp)`).

test/module/irohad/simulator/simulator_test.cpp
MBoldyrev: Thanks, it worked.

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -225,6 +232,8 @@ namespace integration_framework {
      * @param validation - callback that receives object of type \relates
      * std::shared_ptr<shared_model::interface::Proposal> by reference
      * @return this
+     * TODO mboldyrev 27.10.2018: make validation function accept
```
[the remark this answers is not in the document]
MBoldyrev: done

test/module/irohad/simulator/simulator_test.cpp
```diff
@@ -261,35 +266,45 @@ TEST_F(SimulatorTest, FailWhenSameAsProposalHeight) {
...
-                    shared_model::crypto::DefaultCryptoAlgorithmType::
-                        generateKeypair())
-                .finish();
-
-  std::vector<shared_model::proto::Transaction> txs = {tx, tx, tx};
+  const int kNumTransactions = 3;
+  std::vector<shared_model::proto::Transaction> txs;
+  for (int i = 0; i < kNumTransactions; ++i) {
+    txs.emplace_back(shared_model::proto::TransactionBuilder()
+                         .createdTime(iroha::time::now() + i)
```
luckychess: I guess we have a special wrapper for this, do we?

test/module/irohad/simulator/simulator_test.cpp
```diff
@@ -313,14 +330,16 @@ TEST_F(SimulatorTest, RightNumberOfFailedTxs) {
...
-    ASSERT_TRUE(verified_proposal_->second.size() == tx_errors.size());
-  });
+  proposal_wrapper.subscribe(
+      [&](auto verified_proposal_) {
+        // assure that txs in verified proposal do not include failed ones
+        ASSERT_EQ(verified_proposal_->verified_proposal->height(),
+                  verified_proposal_height);
+        ASSERT_EQ(verified_proposal_->verified_proposal->transactions(),
+                  verified_proposal_transactions);
+        ASSERT_TRUE(verified_proposal_->rejected_transactions.size()
```
luckychess: May be compare hashes as well?

irohad/model/converters/impl/json_block_factory.cpp
```diff
@@ -57,6 +57,16 @@ namespace iroha {
         }
         document.AddMember("transactions", commands, allocator);
 
+        Value rejected_transactions;
```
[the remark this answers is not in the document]
MBoldyrev: Thanks. If you meant `test/module/shared_model/builders/protobuf/builder_templates/block_template.hpp`, then I added it.

test/module/irohad/simulator/simulator_test.cpp
MBoldyrev: Although the test is called `RightNumberOfFailedTxs`, I think you are right and that is a reasonable change. Now the test also checks the hashes, although the checked result is created in this same test.

test/module/irohad/simulator/simulator_test.cpp
MBoldyrev: I think it is convenient to increment the value regardless of the units of time it uses. If you know a clearer way of doing this, please suggest it.

luckychess: I finally found this - we have a ```getUniqueTime()``` function for exactly this task. Unfortunately it has 2 implementation and both of them are out of this test scope. It should be refactored but not here I think.

irohad/model/converters/impl/json_block_factory.cpp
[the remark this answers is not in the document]
MBoldyrev: done

irohad/simulator/block_creator.hpp
```diff
@@ -40,7 +44,8 @@ namespace iroha {
        * @param proposal - object for validation
        */
       virtual void process_verified_proposal(
-          const shared_model::interface::Proposal &) = 0;
+          const std::shared_ptr<iroha::validation::VerifiedProposalAndErrors>
```
kamilsa: iroha namespace can be omitted

shared_model/interfaces/common_objects/range_types.hpp
```diff
@@ -30,6 +31,11 @@ namespace shared_model {
           boost::any_range<AccountAsset,
                            boost::random_access_traversal_tag,
                            const AccountAsset &>;
+      /// Type of hash collection
+      using HashCollectionType =
+          boost::any_range<HashType,
```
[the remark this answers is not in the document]
kamilsa: I think it is ok for now, as we have it in other places. We will add task to replace it in future

test/module/shared_model/backend_proto/proto_block_factory_test.cpp
```diff
@@ -40,7 +40,11 @@ TEST_F(ProtoBlockFactoryTest, UnsafeBlockCreation) {
   std::vector<shared_model::proto::Transaction> txs;
   txs.emplace_back(iroha::protocol::Transaction{});
 
-  auto block = factory->unsafeCreateBlock(height, prev_hash, created_time, txs);
+  std::vector<shared_model::crypto::Hash> rejected_txs{
+      shared_model::crypto::Hash::fromHexString("rubble_devaluation")};
```
kamilsa: :) you could actually just create empty vector, but it's fine if you want to keep that

PR #1820, opened 2018-10-30
check torii and internal port when start irohad

victordrobny:

### Description of the Change
Check torii and internal port during start of iroha.

### Benefits
If torii or internal port is busy iroha won't start and won't crash in the future.

### Possible Drawbacks 
Since grpc is not able to check this and assign port even if it is already open.


irohad/main/port_check.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+bool checkTcpPortInUse(std::string address, size_t port) {
+    struct sockaddr_in client;
+    int sock;
+
+    client.sin_family = AF_INET;
+    client.sin_port = htons(port);
+    client.sin_addr.s_addr = inet_addr(address.c_str());
+
+    sock = (int) socket(AF_INET, SOCK_STREAM, 0);
```
l4l: `sys/socket.h` is required for that

irohad/main/port_check.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    int result = connect(sock, (struct sockaddr *) &client,sizeof(client));
```
l4l: Not really the best way to check the port. Port might be allocated, but unused, so connect will be failed. Imo it's better check it with `bind` + SO_REUSEADDR

irohad/main/port_check.hpp
l4l: `sock` must be checked

irohad/main/port_check.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    if (result == 0) {
+        close(sock);
```
l4l: sock must be closed anyway, so:

```cpp
close(sock);
return result != 0;
```

irohad/main/port_check.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IROHA_PORT_CHECK_HPP
+#define IROHA_PORT_CHECK_HPP
+
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
+bool checkTcpPortInUse(std::string address, size_t port) {
```
l4l: Must be `inline` (and even better `static inline`) or impl moved to cpp

irohad/main/irohad.cpp
```diff
@@ -211,6 +212,19 @@ int main(int argc, char *argv[]) {
   std::signal(SIGTERM, handler);
   std::signal(SIGQUIT, handler);
 
+  // Check that torii and internal ports are not in use
+  if (checkTcpPortInUse(kListenIp, config[mbr::ToriiPort].GetUint())) {
```
l4l: Port is usually has represented by the `short` type, check the `checkTcpPortInUse` signature as well

irohad/main/irohad.cpp
```diff
@@ -211,6 +212,19 @@ int main(int argc, char *argv[]) {
...
+    log->error("Torii port {} is in use",
+               config[mbr::ToriiPort].GetUint());
```
l4l: Port might be extracted once, there's no need to call json methods every time

PR #1827, opened 2018-11-01
Libs refactorings

l4l:
### Description of the Change

None logic should be changed, if you found one please report

- Split common/types.hpp to numerous files and related fixes
- Remove redundant files
- Small header fixes at the old model
- Cleanup libs targets

### Benefits

Tech debt decreasing


libs/common/CMakeLists.txt
```diff
@@ -1,18 +1,21 @@
...
-target_include_directories(common INTERFACE
-    ${PROJECT_SOURCE_DIR}/libs/common
-    )
+  boost
+  )
+
+add_library(libs_files
+  files.cpp
+  )
+target_link_libraries(libs_files INTERFACE
```
muratovv: Why `libs_files` marked as `INTERFACE` if there is `files.cpp`?

libs/common/CMakeLists.txt
```diff
@@ -1,18 +1,21 @@
...
+  logger
+  common
+  )
 
+add_library(libs_timeout INTERFACE)
+target_link_libraries(libs_timeout INTERFACE
+  common
+  rxcpp
+  )
```
muratovv: Also, pls, add library for datetime with a separation to cpp and header

l4l: There're two functions there and one of it is templated, the second one is simple enough for being inlined. Moreover return type is deduced here. 
Thus there's no need for such action

irohad/model/block.hpp
```diff
@@ -18,12 +18,13 @@
 #ifndef IROHA_BLOCK_HPP
 #define IROHA_BLOCK_HPP
 
-#include <common/types.hpp>
-#include <model/proposal.hpp>
-#include <model/signature.hpp>
-#include <model/transaction.hpp>
 #include <vector>
 
+#include "crypto/hash_types.hpp"
```
muratovv: I am sad about supporting legacy code. This is of course not an issue for your pr.

muratovv:
I missed this part in the review. I think better to add comments for the interface targets in `libs/common/CMakeLists.txt`. I expect something like this:

```
add_library(libs_timeout INTERFACE
	# timeout.hpp
	)
```
This approach will improve understating what dependency you require and helpful in further refactorings.

Thank you.

PR #1830, opened 2018-11-02
Two phase commit

nickaleks:
### Description of the Change

This PR introduces 2 phase commit which speeds up application of blocks in several cases.

### Benefits
Two phase commit is faster than applying the state again.

### Possible Drawbacks 
Might be some subtle bugs.


Warchant:
2PC can be blocked in some cases. 
See https://en.wikipedia.org/wiki/Two-phase_commit_protocol#Disadvantages

nickaleks:
@Warchant yes, we handle it when we receive different commit.

nickaleks:
@bakhtin 👍

irohad/ametsuchi/impl/temporary_wsv_impl.cpp
```diff
@@ -123,6 +123,7 @@ namespace iroha {
     }
 
     TemporaryWsvImpl::~TemporaryWsvImpl() {
+      log_->info("rolling back");
```
lebdron:
```suggestion
      log_->debug("rolling back");
```

test/module/irohad/ametsuchi/ametsuchi_test.cpp
```diff
@@ -85,7 +86,7 @@ void validateAccountAsset(W &&wsv,
   ASSERT_TRUE(account_asset);
   ASSERT_EQ((*account_asset)->accountId(), account);
   ASSERT_EQ((*account_asset)->assetId(), asset);
-  ASSERT_EQ((*account_asset)->balance(), amount);
+  ASSERT_EQ((*account_asset)->balance().toString(), amount.toString());
```
lebdron:
```suggestion
  ASSERT_EQ((*account_asset)->balance(), amount);
```

test/module/irohad/ametsuchi/ametsuchi_test.cpp
```diff
@@ -809,3 +809,219 @@ TEST_F(AmetsuchiTest, TestRestoreWSV) {
...
+}
+
+/**
+ * @given Storage with prepared state
+ * @when another block is applied
+ * @then commitPrepared fails @and prepared state is not applied
+ */
+TEST_F(PreparedBlockTest, CommitPreparedFailsAfterCommit) {
+  // tx which we prepare
+  auto tx = shared_model::proto::TransactionBuilder()
```
lebdron: It looks like `tx`, `other_tx` can be moved to fixture and reused in test cases, and some shortcut for `block` can be created - a method which only takes a vector of transactions, for example.

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -310,3 +311,125 @@ TEST_F(SynchronizerTest, RetrieveBlockTwoFailures) {
 
   ASSERT_TRUE(wrapper.validate());
 }
+
+/**
+ * @given synchronizer receives a commit with the block peer voted for
+ * @when synchronizer processes the commit
```
lebdron:
```suggestion
 * @when synchronizer processes the commit @and commit prepared is successful
```

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -310,3 +311,125 @@ TEST_F(SynchronizerTest, RetrieveBlockTwoFailures) {
...
+ * @then commitPrepared will be called @and if successful, commit is not called
```
lebdron:
```suggestion
 * @then commitPrepared is called @and commit is not called
```

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -310,3 +311,125 @@ TEST_F(SynchronizerTest, RetrieveBlockTwoFailures) {
 
   ASSERT_TRUE(wrapper.validate());
 }
+
+/**
+ * @given synchronizer receives a commit with the block peer voted for
```
lebdron: "given...receives" does not sound quite right, maybe rephrase and move "receives" part to "when"? Same in other test cases.

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -230,6 +252,11 @@ namespace iroha {
         log_->warn("Tried to free connections without active connection");
         return;
       }
+      // rollback possible prepared transaction
```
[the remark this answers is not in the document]
nickaleks: Destructor seems like a good place since we do not want any prepared transactions left hanging in case of shutdown/reboot of the peer.

test/module/irohad/synchronizer/synchronizer_test.cpp
nickaleks: Since we test that commitPrepared is called in this test, it is better to put it in a then clause, since it is the effect we are testing

irohad/synchronizer/synchronizer.hpp
```diff
@@ -24,8 +24,11 @@
 #include "synchronizer/synchronizer_common.hpp"
 
 namespace iroha {
-  namespace synchronizer {
+  namespace network {
+    struct Commit;
```
muratovv: Seems bit confusing that `struct` was used.
I know that `struct` and `class` are totally the same in this case. But other code typically uses `class` for the forward decl.

.jenkinsci/debug-build.groovy
```diff
@@ -45,74 +45,71 @@ def doDebugBuild(coverageEnabled=false) {
...
-      + " -e IROHA_POSTGRES_HOST=${env.IROHA_POSTGRES_HOST}"
-      + " -e IROHA_POSTGRES_PORT=${env.IROHA_POSTGRES_PORT}"
-      + " -e IROHA_POSTGRES_USER=${env.IROHA_POSTGRES_USER}"
-      + " -e IROHA_POSTGRES_PASSWORD=${env.IROHA_POSTGRES_PASSWORD}"
-      + " --network=${env.IROHA_NETWORK}"
-      + " -v /var/jenkins/ccache:${CCACHE_DIR}"
-      + " -v /tmp/${GIT_COMMIT}-${BUILD_NUMBER}:/tmp/${GIT_COMMIT}") {
+  sh "docker run -td -e POSTGRES_USER=${env.IROHA_POSTGRES_USER} \
+  -e POSTGRES_PASSWORD=${env.IROHA_POSTGRES_PASSWORD} --name ${env.IROHA_POSTGRES_HOST} \
+  --network=${env.IROHA_NETWORK} postgres:9.5 -c 'max_prepared_transactions=100'"
```
muratovv: Pls, add an explanation of `max_prepared_transactions` and reason to put it as `100`.

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -26,7 +27,17 @@ namespace {
...
       iroha::ametsuchi::PostgresCommandExecutor::prepareStatements(session);
     }
-  };
+  }
+
+  /**
+   * Verify whether postgres supports prepared transactions
+   */
+  bool preparedTransactionsAvailable(soci::session &sql) {
+    int prepared_txs_count = 0;
```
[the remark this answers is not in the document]
nickaleks: No, its either 0 or bigger

irohad/synchronizer/synchronizer.hpp
nickaleks: Since commit is a struct, we have to use struct in forward declaration.

PR #1831, opened 2018-11-02
Trunk/bft yac

l4l:
### Description of the Change

Introduces BFT Yac

### Benefits

Final (?) step toward complete BFT consensus

### Possible Drawbacks 

None?


lebdron:
Changed base to `trunk/bft-os-integration` since block loading in `VoteOther` case is not handled yet.

PR #1833, opened 2018-11-05
Remove client_test

l4l:
### Description of the Change

Finally it comes.
It seems that client_test doesn't really check behavior of CliClient as intended, but instead duplicate the integration tests that already present in unit test folder (😢). Since it's painful for the support we discussed with @muratovv that it should be removed from the project.

### Benefits

Lesser stuff to support

### Possible Drawbacks

No tests for CliClient, cli is even less maintained


l4l:
The problem there, that CliClient class is hard to test without any refactorings, as example it cannot connect to some api, instead of real port. Since it will be removed in future (and I hope soon), it probably shouldn't be supported, like an old model

lebdron:
@l4l @luckychess @muratovv can we make a decision on this test?

luckychess:
Well, finally we decided to remove this test and put TODO about its refactoring without running whole command and query services but only with a code to make endpoints work.

PR #1834, opened 2018-11-06
Fix consensus VoteOther case

lebdron:
### Description of the Change
The information in VoteOther struct was not enough to handle block loading, so it is replaced with 
- public keys of peers who made the latest commit
- hash of committed block

### Benefits
Can synchronize in case different block was committed

### Possible Drawbacks 
None?


irohad/synchronizer/impl/synchronizer_impl.hpp
```diff
@@ -35,15 +36,19 @@ namespace iroha {
        * apply the missing blocks
        */
       SynchronizationEvent downloadMissingBlocks(
-          std::shared_ptr<shared_model::interface::Block> commit_message,
+          const shared_model::interface::types::PublicKeyCollectionType
```
nickaleks: Maybe pass VoteOther to avoid specifying all parameters in functions?

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -68,18 +69,21 @@ class YacGateTest : public ::testing::Test {
...
+    peer_orderer = peer_orderer_ptr.get();
     hash_provider = make_shared<MockYacHashProvider>();
     block_creator = make_shared<MockBlockCreator>();
     block_cache = make_shared<ConsensusResultCache>();
-  }
 
-  void init() {
-    gate = std::make_shared<YacGateImpl>(std::move(hash_gate),
-                                         std::move(peer_orderer),
+    EXPECT_CALL(*block_creator, on_block())
```
nickaleks: Maybe use ON_CALL so that we do not set any unnecessary invariants for tests?

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -179,15 +181,74 @@ TEST_F(YacGateTest, YacGateSubscribtionTestFailCase) {
...
+      .WillOnce(Return(actual_hash));
+
+  // verify that block we voted for is in the cache
+  auto cache_block = block_cache->get();
+  ASSERT_EQ(cache_block, expected_block);
+
+  // verify that yac gate emit expected block
+  auto gate_wrapper = make_test_subscriber<CallExact>(gate->onOutcome(), 1);
+  gate_wrapper.subscribe([actual_hash, actual_pubkey](auto outcome) {
+    auto concete_outcome = boost::get<iroha::consensus::VoteOther>(outcome);
```
nickaleks: typo in variable name?

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -68,18 +69,21 @@ class YacGateTest : public ::testing::Test {
     message.hash = expected_hash;
     message.signature = signature;
     commit_message = CommitMessage({message});
-    expected_commit = rxcpp::observable<>::just(Answer(commit_message));
+    expected_commit = commit_message;
 
-    hash_gate = make_unique<MockHashGate>();
-    peer_orderer = make_unique<MockYacPeerOrderer>();
+    auto hash_gate_ptr = make_unique<MockHashGate>();
```
l4l: `using namespace std`
😱 (not a pr issue)

PR #1837, opened 2018-11-08
Stateful Response Error Codes

Akvinikym:
### Description of the Change

This PR trunk introduces error codes for stateful response. Numeric values, described in the following [POW](https://soramitsu.atlassian.net/wiki/spaces/IS/pages/439189508/POW+N+075+Stateful+response), are returned the to the client instead of long and insecure error messages.

### Benefits

More security and easier to parse by the client.

### Possible Drawbacks 

Less information about the error, which happened during validation.


irohad/torii/processor/impl/transaction_processor_impl.cpp
```diff
@@ -136,51 +138,49 @@ namespace iroha {
...
         case TxStatusType::kMstPending: {
-          builder = builder.mstPending();
-          break;
+          status_bus_->publish(status_factory_->makeMstPending(hash, tx_error));
+          return;
         };
         case TxStatusType::kEnoughSignaturesCollected: {
-          builder = builder.enoughSignaturesCollected();
-          break;
+          status_bus_->publish(status_factory_->makeEnoughSignaturesCollected(hash, tx_error));
```
nickaleks: Formatting?

PR #1839, opened 2018-11-08
Add fuzzing for BlockLoader endpoints

luckychess:
### Description of the Change

Fuzzing for block loader endpoints (**retrieveBlock** and **retrieveBlocks**).

### Benefits
Another endpoint is covered by fuzzing.

### Possible Drawbacks 
Hard to support until fuzzing in CI is not ready.

### Usage Examples or Tests 
```
cmake -DCMAKE_C_COMPILER=/usr/local/opt/llvm/bin/clang -DCMAKE_CXX_COMPILER=/usr/local/opt/llvm/bin/clang++  -DFUZZING=ON ..
make retrieve_block_fuzz
make retrieve_blocks_fuzz
```

### DIFF
--- a/test/fuzzing/retrieve_blocks_fuzz.cpp
+++ b/test/fuzzing/retrieve_blocks_fuzz.cpp
@@ -5,6 +5,17 @@
 
 #include "block_loader_fixture.hpp"
 
+namespace fuzzing {
+  class MockServerWriter
+      : public grpc::ServerWriterInterface<iroha::protocol::Block> {
+    MOCK_METHOD1(Write, void(iroha::protocol::Block));
+    MOCK_METHOD2(Write,
+                 bool(const iroha::protocol::Block &, grpc::WriteOptions));
+    MOCK_METHOD0(SendInitialMetadata, void());
+    MOCK_METHOD1(NextMessageSize, bool(uint32_t *));
+  };
+}  // namespace fuzzing
+
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, std::size_t size) {
   static fuzzing::BlockLoaderFixture fixture;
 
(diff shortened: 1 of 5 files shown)


test/fuzzing/retrieve_blocks_fuzz.cpp
```diff
@@ -0,0 +1,23 @@
...
+  static fuzzing::BlockLoaderFixture fixture;
+
+  if (size < 1) {
+    return 0;
+  }
+
+  iroha::network::proto::BlocksRequest request;
+  if (protobuf_mutator::libfuzzer::LoadProtoInput(true, data, size, &request)) {
+    grpc::ServerContext context;
+    MockServerWriter writer;
```
l4l: ServerWriter cannot be mocked :( I've tried with ServerWriterInterface, but it cannot be casted safely (e.g only via `reinterpret_cast`. Here's an example of such mock (path _test/module/vendor/grpc/grpc_mocks.hpp_):

```cpp
class MockServerWriter
    : public grpc::ServerWriterInterface<iroha::protocol::ToriiResponse> {
  MOCK_METHOD1(Write, void(iroha::protocol::ToriiResponse));
  MOCK_METHOD2(Write,
               bool(const iroha::protocol::ToriiResponse &,
                    grpc::WriteOptions));
  MOCK_METHOD0(SendInitialMetadata, void());
  MOCK_METHOD1(NextMessageSize, bool(uint32_t *));
};
```

But it's not really a good idea to use it. Need to think about that issue

test/fuzzing/retrieve_blocks_fuzz.cpp
```diff
@@ -5,6 +5,17 @@
 
 #include "block_loader_fixture.hpp"
 
+namespace fuzzing {
+  class MockServerWriter
```
l4l: plz move it to `test/module/vendor/grpc_mocks.hpp`, I will also need that mock

test/fuzzing/retrieve_blocks_fuzz.cpp
```diff
@@ -15,7 +26,12 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, std::size_t size) {
   iroha::network::proto::BlocksRequest request;
   if (protobuf_mutator::libfuzzer::LoadProtoInput(true, data, size, &request)) {
     grpc::ServerContext context;
-    fixture.block_loader_service_->retrieveBlocks(&context, &request, nullptr);
+    fuzzing::MockServerWriter serverWriter;
```
l4l: Should be static, I guess
Moreover some calls probably should be mocked (particularly the ones that return boolean)

luckychess: I don't think it should be static - we pass a non-const raw point to it and have no idea what will happen with it inside.
I'm not sure what call do you mean - the only one which returns boolean is ```Write``` with 2 args but we are not interested in it.

PR #1845, opened 2018-11-12
MST fuzzing

luckychess:
### Description of the Change
Fuzzing for MST endpoint (**SendState**).

### Benefits
One more endpoint covered.

### Possible Drawbacks 
CI is needed.
Also it looks like a false positive warning from Address Sanitizer (at least I couldn't find a problem - see https://gist.github.com/luckychess/b65bd40e958887273e86a0a535cc896d).

### Usage Examples or Tests 
```
cmake -DCMAKE_C_COMPILER=/usr/local/opt/llvm/bin/clang -DCMAKE_CXX_COMPILER=/usr/local/opt/llvm/bin/clang++  -DFUZZING=ON ..
make mst_fuzz
```

### DIFF
--- a/test/fuzzing/mst_fuzz.cpp
+++ b/test/fuzzing/mst_fuzz.cpp
@@ -26,6 +26,7 @@ namespace fuzzing {
           iroha::network::AsyncGrpcClient<google::protobuf::Empty>>();
       std::unique_ptr<shared_model::validation::AbstractValidator<
           shared_model::interface::Transaction>>
+          // TODO luckychess 20.11.2018 Reuse validator from application.cpp IR-1886
           tx_validator =
               std::make_unique<shared_model::validation::
                                    DefaultOptionalSignedTransactionValidator>();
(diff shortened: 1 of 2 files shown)


test/fuzzing/mst_fuzz.cpp
```diff
@@ -0,0 +1,66 @@
...
+    std::shared_ptr<MstTransportGrpc> mst_transport_grpc_;
+
+    MstFixture() {
+      auto async_call_ = std::make_shared<
+          iroha::network::AsyncGrpcClient<google::protobuf::Empty>>();
+      std::unique_ptr<shared_model::validation::AbstractValidator<
+          shared_model::interface::Transaction>>
+          tx_validator =
+              std::make_unique<shared_model::validation::
+                                   DefaultOptionalSignedTransactionValidator>();
```
lebdron: The problem with this approach is that the possible changes in `application.cpp` will not be reflected in this file. Please think of a solution where components in `application.cpp` are referenced here.

luckychess: After discussion with @igor-egorov we decided to leave TODO and fix this issue later since it affect other tasks too.

PR #1846, opened 2018-11-13
Feature/rejected txs index

kamilsa:
### Description of the Change

We should be able to check whether transaction has been processed by ledger, even if it was rejected. This PR adds this functionality by adding hasRejectedTxWithHash query in block query. In future it can be used by committed transactions cache.
### Benefits

Now we have all means to implement fair implementation of committed transactions cache

### Possible Drawbacks 

None

### Usage Examples or Tests

Corresponding tests on block query were provided

### DIFF
--- a/irohad/ametsuchi/impl/postgres_block_index.cpp
+++ b/irohad/ametsuchi/impl/postgres_block_index.cpp
@@ -39,6 +39,15 @@ namespace {
     return (base % hash.hex() % height).str();
   }
 
+  std::string makeRejectedTxHashIndex(
+      const shared_model::interface::types::HashType &rejected_tx_hash,
+      shared_model::interface::types::HeightType height) {
+    boost::format base(
+        "INSERT INTO height_by_rejected_hash(hash, height) VALUES ('%s', "
+        "'%s');");
+    return (base % rejected_tx_hash.hex() % height).str();
+  }
+
   // make index account_id:height -> list of tx indexes
   // (where tx is placed in the block)
   std::string makeCreatorHeightIndex(
@@ -110,7 +119,8 @@ namespace iroha {
         const shared_model::interface::Block &block) {
       auto height = block.height();
       auto indexed_txs = block.transactions() | boost::adaptors::indexed(0);
-      std::string index_query = std::accumulate(
+      auto rejected_txs_hashes = block.rejected_transactions_hashes();
+      std::string tx_index_query = std::accumulate(
           indexed_txs.begin(),
           indexed_txs.end(),
           std::string{},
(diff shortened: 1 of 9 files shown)


irohad/ametsuchi/impl/postgres_block_index.cpp
```diff
@@ -39,6 +39,15 @@ namespace {
     return (base % hash.hex() % height).str();
   }
 
+  std::string makeRejectedTxHashIndex(
```
igor-egorov: I suggest a bit different naming - `storeRejectedTxHash`

Otherwise `make index` sounds like `CREATE INDEX index_name ON table_name (column1, column2, ...);` will be executed under the hood :)

irohad/ametsuchi/impl/postgres_block_query.cpp
```diff
@@ -242,11 +242,27 @@ namespace iroha {
         };
     }
 
-    bool PostgresBlockQuery::hasTxWithHash(
+    bool PostgresBlockQuery::hasCommittedTxWithHash(
         const shared_model::crypto::Hash &hash) {
       return getBlockId(hash) != boost::none;
     }
 
+    bool PostgresBlockQuery::hasRejectedTxWithHash(
```
igor-egorov: Why should we separate hashes storage on two methods? 
Why could we not store both kinds of hashes in the same table with different markers in a separate column?

something like:
```
+-----+----------------+--------------------+-----+
| ... | hash           | committed/rejected | ... |
+-----+----------------+--------------------+-----+
|     | 2ac7b67ca6b76c | committed          |     |
+-----+----------------+--------------------+-----+
|     | 89ac7b867a875a | rejected           |     |
+-----+----------------+--------------------+-----+
```

That approach will let us avoid looking up in two independent sources for the single transaction in the worst case. 
Also, we will have only one method for a transaction look up.

irohad/ametsuchi/impl/postgres_block_index.cpp
kamilsa: I agree with you. My intention was to keep the naming consisted with other methods here. I can add the task to provide better naming for that function

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -508,6 +508,12 @@ CREATE TABLE IF NOT EXISTS height_by_hash (
     hash varchar,
     height text
 );
+
+CREATE TABLE IF NOT EXISTS height_by_rejected_hash (
```
igor-egorov: Add drop table please (somewhere)

PR #1848, opened 2018-11-13
PIMPL for Transaction

Akvinikym:
### Description of the Change

PIMPL idiom for transaction class is introduced in this PR.

### Benefits

Reduced compilation time.

### Possible Drawbacks 

Some of the tests were rewritten, taking into account lack of copy ctor, which was added later, so they now introduce vector wrapping. Corresponding task is already created.

### DIFF
--- a/shared_model/backend/protobuf/impl/transaction.cpp
+++ b/shared_model/backend/protobuf/impl/transaction.cpp
@@ -6,7 +6,6 @@
 #include "backend/protobuf/transaction.hpp"
 
 #include <boost/range/adaptor/transformed.hpp>
-
 #include "backend/protobuf/batch_meta.hpp"
 #include "backend/protobuf/commands/proto_command.hpp"
 #include "backend/protobuf/common_objects/signature.hpp"
@@ -31,14 +30,14 @@ namespace shared_model {
 
       interface::types::BlobType blob_{[this] { return makeBlob(*proto_); }()};
 
-      interface::types::BlobType blob_type_payload_{
+      interface::types::BlobType payload_blob_{
           [this] { return makeBlob(payload_); }()};
 
-      interface::types::BlobType blob_type_reduced_payload_{
+      interface::types::BlobType reduced_payload_blob_{
           [this] { return makeBlob(reduced_payload_); }()};
 
-      interface::types::HashType reduced_hash_ =
-          shared_model::crypto::Sha3_256::makeHash(blob_type_reduced_payload_);
+      interface::types::HashType reduced_hash_{
+          shared_model::crypto::Sha3_256::makeHash(reduced_payload_blob_)};
 
       std::vector<proto::Command> commands_{[this] {
         return std::vector<proto::Command>(reduced_payload_.commands().begin(),
@@ -75,10 +74,8 @@ namespace shared_model {
 
     // TODO [IR-1866] Akvinikym 13.11.18: remove the copy ctor and fix fallen
     // tests
-    Transaction::Transaction(const Transaction &transaction) {
-      this->impl_ =
-          std::make_unique<Transaction::Impl>(*transaction.impl_->proto_);
-    }
+    Transaction::Transaction(const Transaction &transaction)
+        : Transaction(*transaction.impl_->proto_) {}
 
     Transaction::Transaction(Transaction &&transaction) noexcept = default;
 
@@ -98,11 +95,11 @@ namespace shared_model {
     }
 
     const interface::types::BlobType &Transaction::payload() const {
-      return impl_->blob_type_payload_;
+      return impl_->payload_blob_;
     }
 
     const interface::types::BlobType &Transaction::reducedPayload() const {
-      return impl_->blob_type_reduced_payload_;
+      return impl_->reduced_payload_blob_;
     }
 
     interface::types::SignatureRangeType Transaction::signatures() const {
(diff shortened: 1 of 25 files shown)


test/module/shared_model/builders/protobuf/transport_builder_test.cpp
```diff
@@ -107,10 +107,9 @@ class TransportBuilderTest : public ::testing::Test {
   //-------------------------------------Block-------------------------------------
   template <typename BlockBuilder>
   auto getBaseBlockBuilder() {
-    return BlockBuilder()
-        .transactions(std::vector<Transaction>({createTransaction()}))
-        .height(1)
-        .createdTime(created_time);
+    std::vector<shared_model::proto::Transaction> txs;
```
MBoldyrev: Here and in other places you could use:
```suggestion
    std::vector<shared_model::proto::Transaction> txs{createTransaction()};
```

test/module/irohad/validation/stateful_validator_test.cpp
```diff
@@ -147,19 +147,30 @@ class Validator : public testing::Test {
...
-                .createAsset("doge", "coin", 1)
-                .build();
-  auto proposal =
-      TestProposalBuilder()
-          .createdTime(iroha::time::now())
-          .height(3)
-          .transactions(
-              std::vector<shared_model::proto::Transaction>{tx, tx, tx})
-          .build();
+  std::vector<shared_model::proto::Transaction> txs;
```
MBoldyrev: You could void the boilerplate like this
```suggestion
  std::vector<shared_model::proto::Transaction> txs(
      3,
      TestTransactionBuilder()
          .creatorAccountId("doge@master")
          .createdTime(iroha::time::now())
          .quorum(1)
          .createAsset("doge", "coin", 1)
          .build());
```

shared_model/backend/protobuf/impl/transaction.cpp
```diff
@@ -0,0 +1,166 @@
...
+                       })
+          != impl_->signatures_.end()) {
+        return false;
+      }
+
+      auto sig = impl_->proto_->add_signatures();
+      sig->set_signature(crypto::toBinaryString(signed_blob));
+      sig->set_public_key(crypto::toBinaryString(public_key));
+
+      impl_->signatures_ = {[this] {
```
lebdron: `{}` around lambda call look redundant.

test/module/irohad/ametsuchi/ametsuchi_test.cpp
```diff
@@ -655,8 +651,8 @@ TEST_F(AmetsuchiTest, FindTxByHashTest) {
 
   // TODO: 31.10.2017 luckychess move tx3hash case into a separate test after
   // ametsuchi_test redesign
-  auto tx1hash = txn1.hash();
-  auto tx2hash = txn2.hash();
+  auto tx1hash = txs[0].hash();
```
lebdron: It is probably more safe to use `.at()` since it will throw if someone will modify test without noticing.

test/module/irohad/ametsuchi/postgres_query_executor_test.cpp
```diff
@@ -1106,63 +1106,58 @@ namespace iroha {
...
+                          .height(2)
+                          .prevHash(block1.hash())
+                          .build();
 
         apply(storage, block2);
 
-        hash1 = tx1.hash();
-        hash2 = tx2.hash();
-        hash3 = tx3.hash();
+        hash1 = txs1[0].hash();
```
lebdron: Same for `.at()`.

test/module/irohad/validation/stateful_validator_test.cpp
```diff
@@ -178,30 +189,40 @@ TEST_F(Validator, AllTxsValid) {
...
-          .createdTime(iroha::time::now())
-          .height(3)
-          .transactions(std::vector<shared_model::proto::Transaction>{
-              valid_tx, invalid_tx, valid_tx})
-          .build();
 
-  EXPECT_CALL(*temp_wsv_mock, apply(Eq(ByRef(invalid_tx))))
-      .WillOnce(Return(iroha::expected::Error<CommandError>({})));
-  EXPECT_CALL(*temp_wsv_mock, apply(Eq(ByRef(valid_tx))))
+  EXPECT_CALL(*temp_wsv_mock, apply(Eq(ByRef(txs[0]))))
```
lebdron: Same for `.at()`.

test/module/irohad/validation/stateful_validator_test.cpp
lebdron: This will invoke copy ctor:
```C++
vector(size_type count,
       const T& value,
       const Allocator& alloc = Allocator());
```

test/module/shared_model/builders/protobuf/transport_builder_test.cpp
lebdron: Not sure if it will work without copy constructor.

Akvinikym: Andrey is right, this is exactly the thing, which was refactored, because it won't work without a copy constructor. Though it exists now, later it will be removed, so I started the refactoring a little ahead

test/module/irohad/validation/stateful_validator_test.cpp
MBoldyrev: Is there a reason we should avoid copying? The original test (before this refactoring) did copy the tx: https://github.com/hyperledger/iroha/blob/ec3e0ca52e248c1ed6fbb55b5f4db87a97b960bb/test/module/irohad/validation/stateful_validator_test.cpp#L160-L161

MBoldyrev: I see, if you want to avoid using copy ctor, you may use a lambda that generates that tx, an call it several times where needed.

PR #1854, opened 2018-11-14
ITF: PortGuard

MBoldyrev:
### Description of the Change

Meet integration_framework::PortGuard - a class that keeps track of ports you use in ITF tests! Just add an instance in your class where you allocate ports, and ask it for some in a given range. When your class dies (sorry), it will in its final moments mark the ports your instance requested as free.

### Benefits

It should make port reusing easier.

### Possible Drawbacks 

It uses locks for synchronization.

### DIFF
--- a/test/framework/integration_framework/integration_test_framework.hpp
+++ b/test/framework/integration_framework/integration_test_framework.hpp
@@ -62,6 +62,7 @@ namespace integration_framework {
   using std::chrono::milliseconds;
 
   class FakePeer;
+  class PortGuard;
 
   class IntegrationTestFramework {
    private:
@@ -373,6 +374,7 @@ namespace integration_framework {
     std::map<std::string, tbb::concurrent_queue<TxResponseType>>
         responses_queues_;
 
+    std::unique_ptr<PortGuard> port_guard_;
     size_t torii_port_;
     size_t internal_port_;
     std::shared_ptr<IrohaInstance> iroha_instance_;
--- a/test/framework/integration_framework/port_guard.cpp
+++ b/test/framework/integration_framework/port_guard.cpp
@@ -1,0 +1,55 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "framework/integration_framework/port_guard.hpp"
+
+#include <boost/assert.hpp>
+
+namespace integration_framework {
+
+  constexpr PortGuard::PortType PortGuard::kMaxPort;
+  PortGuard::UsedPorts PortGuard::all_used_ports_ = {};
+  std::mutex PortGuard::all_used_ports_mutex_ = {};
+
+  PortGuard::~PortGuard() {
+    std::lock_guard<std::mutex> lock(all_used_ports_mutex_);
+    BOOST_ASSERT_MSG(
+        ((all_used_ports_ | instance_used_ports_) ^ all_used_ports_).none(),
+        "Some ports used by this PortGuard instance are not set in ports "
+        "used by all instances!");
+    all_used_ports_ ^= instance_used_ports_;
+  }
+
+  boost::optional<PortGuard::PortType> PortGuard::tryGetPort(
+      const PortType &min_value, const PortType &max_value) {
+    std::lock_guard<std::mutex> lock(all_used_ports_mutex_);
+    PortType tested_port = min_value;
+    while (all_used_ports_.test(tested_port)) {
+      if (tested_port == max_value) {
+        return boost::none;
+      }
+      ++tested_port;
+    }
+    BOOST_ASSERT_MSG(!all_used_ports_.test(tested_port),
+                     "PortGuard chose an occupied port!");
+    BOOST_ASSERT_MSG(tested_port >= min_value && tested_port <= max_value,
+                     "PortGuard chose a port outside boundaries!");
+    instance_used_ports_.set(tested_port);
+    all_used_ports_.set(tested_port);
+    return tested_port;
+  }
+
+  PortGuard::PortType PortGuard::getPort(const PortType &min_value,
+                                         const PortType &max_value) {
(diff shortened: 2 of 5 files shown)


test/framework/integration_framework/port_guard.cpp
```diff
@@ -0,0 +1,62 @@
...
+                     "PortGuard chose a port outside boundaries!");
+    instance_used_ports_.set(tested_port);
+    all_used_ports_.set(tested_port);
+    return tested_port;
+  }
+
+  PortGuard::PortType PortGuard::getPort(const PortType &min_value,
+                                         const PortType &max_value) {
+    boost::optional<PortType> opt_port = tryGetPort(min_value, max_value);
+    BOOST_VERIFY_MSG(
```
l4l: Why *_VERIFY_* used here and *_ASSERT_* at the function above?

test/module/test/framework/integration_framework/port_guard_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <boost/noncopyable.hpp>
+#include "framework/integration_framework/port_guard.hpp"
+
+using PortGuard = integration_framework::PortGuard;
+
+TEST(PortGuardTest, AllPortsGetUsedAndNoOverlaps) {
+  static constexpr size_t kNumClients = 10; // first-run clients
+  static constexpr size_t kNumClientsKilled = 3; // then these die
+  static constexpr size_t kNumClientsResurrected = 5; // then these are added
+  static constexpr std::pair<size_t, size_t> kRange = {123, 456};
```
l4l: Would be nice to have exact amount of available ports as the required number, since it's a better corner case

test/framework/integration_framework/port_guard.cpp
```diff
@@ -0,0 +1,62 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "framework/integration_framework/port_guard.hpp"
+
+#include <boost/assert.hpp>
+
+namespace integration_framework {
+
+  constexpr PortGuard::PortType PortGuard::kMaxPort;
```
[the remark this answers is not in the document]
MBoldyrev: It needs to be somewhere. If you mean there is a more appropriate place, please suggest it.

test/framework/integration_framework/port_guard.cpp
MBoldyrev: Because this error condition can happen in runtime and it is not mitigated in any way except that macro. And the assertions above are just for testing and clarification purposes and the conditions checked there cannot fail because they are mitigated by the code logic (until it is not tampered with, that is the purpose of asserts).

test/module/test/framework/integration_framework/port_guard_test.cpp
MBoldyrev: Just tried with the whole port range - it was running for 1 min 16 sec. Too long, I think.

l4l: Maybe reduce the range then?

test/module/test/framework/integration_framework/port_guard_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+          << "Port " << port << " was not given out to any client!";
+    }
+  };
+
+  // first run
+  run_new_clients(kNumClients);
+  check();
+
+  // kill some clients...
+  clients.resize(kNumClients - kNumClientsKilled);
```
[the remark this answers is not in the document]
nickaleks: If it is to check that ports are freed, maybe write it as a separate test

test/module/test/framework/integration_framework/port_guard_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+      while (true) {
+        auto port = client.port_guard.tryGetPort(kRange.first, kRange.second);
+        if (!port) {
+          break;
+        }
+        client.used_ports.set(*port);
+      }
+    };
+
+    clients.reserve(clients.size() + num_clients);
```
nickaleks: since you have not put anything into clients, clients.size() will be 0

test/module/test/framework/integration_framework/port_guard_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  // first run
+  run_new_clients(kNumClients);
+  check();
+
+  // kill some clients...
+  clients.resize(kNumClients - kNumClientsKilled);
+  // ... and run some new clients
+  run_new_clients(kNumClientsResurrected);
+  check();
```
nickaleks: Using check() twice will be confusing since it is unclear which one causes test failure

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -373,6 +374,7 @@ namespace integration_framework {
     std::map<std::string, tbb::concurrent_queue<TxResponseType>>
         responses_queues_;
 
+    std::unique_ptr<PortGuard> port_guard_;
```
nickaleks: Why keep it in a pointer?

test/module/test/framework/integration_framework/port_guard_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include <gtest/gtest.h>
+#include <boost/thread/barrier.hpp>
+#include <boost/noncopyable.hpp>
+#include "framework/integration_framework/port_guard.hpp"
+
+using PortGuard = integration_framework::PortGuard;
+
+TEST(PortGuardTest, AllPortsGetUsedAndNoOverlaps) {
+  static constexpr size_t kNumClients = 10; // first-run clients
```
nickaleks: static constexpr out of class context does not make much sense.

test/framework/integration_framework/integration_test_framework.hpp
MBoldyrev: It is to hide the implementation.

PR #1856, opened 2018-11-14
Fix SetAccountQuorum validation rule

igor-egorov:

### Description of the Change

SetAccountQuorum validation rule had inverse logic.
It was not possible to set the quorum less than a number of linked signatories.
The true validation rule is - it is not allowed to set the quorum more than a number of linked signatories.

### Benefits

Fixed validation behavior.
Increased code coverage.

### Possible Drawbacks 

none ?

### Usage Examples or Tests

```
cmake -H. -Bbuild
cd build
make set_account_quorum_test
test_bin/set_account_quorum_test
```

### DIFF
--- a/test/integration/acceptance/set_account_quorum_test.cpp
+++ b/test/integration/acceptance/set_account_quorum_test.cpp
@@ -1,0 +1,87 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <gtest/gtest.h>
+#include "framework/integration_framework/integration_test_framework.hpp"
+#include "integration/acceptance/acceptance_fixture.hpp"
+
+using namespace integration_framework;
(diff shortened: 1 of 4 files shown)


test/integration/acceptance/set_account_quorum_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include <gtest/gtest.h>
+#include "framework/integration_framework/integration_test_framework.hpp"
+#include "integration/acceptance/acceptance_fixture.hpp"
+
+using namespace integration_framework;
+using namespace shared_model;
+using namespace common_constants;
+
+#define check(i) [](auto &block) { ASSERT_EQ(block->transactions().size(), i); }
```
nickaleks: can we move that macro to acceptance fixture, since it is used in many tests?

igor-egorov: Let's do it as a separate pull request (IR-1869).

PR #1858, opened 2018-11-15
Makefile for python libs example

MBoldyrev:
### Description of the Change

Replaced `prepare.sh` with a Makefile.

### Benefits

Makefile is a standard way to generate something. It is also easier to extend and is smarter.

### Possible Drawbacks 

The user needs the `make` program, although it is hard to imagine one not having it.

`make`

We could also use `CMake`, but while the project is not so big it seems an overkill.

### DIFF
--- a/example/python/Makefile
+++ b/example/python/Makefile
@@ -1,8 +1,12 @@
 #!/usr/bin/env make
 
-# generate proto files in current dir
+# sources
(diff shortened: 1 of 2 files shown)


igor-egorov:
Please wait for my review too (need to check that all dependencies are generated correctly)

nickaleks:
LGTM

PR #1859, opened 2018-11-15
fake peer ordering usage example

MBoldyrev:
### Description of the Change

This is an example of using Ordering Gate/Service interface introduced in #1849 

### Benefits

Usage example

### Possible Drawbacks 

Pretty stupid code, should be removed after some real ITF tests on Ordering appear.

### Usage Examples or Tests 

`fake_peer_example_test`


MBoldyrev:
The related branch can be safely deleted, and I have not enough permissions to do it myself.

PR #1863, opened 2018-11-16
Fix MST SendState segfault

igor-egorov:

### Description of the Change

It was possible to cause a segfault in Iroha by sharing MST state that contains a transaction with non-typed command.

Now such states are handled correctly - they are dropped during proto validation.

The test is created.

Now proto validators are used for transactions as it was desired initially.

### Benefits

Improved stability and code coverage.

### Possible Drawbacks 

Once a day, non-proto transaction validator can be accidentally used.

### Usage Examples or Tests

```
make transport_test
test_bin/transport_test
```


shared_model/validators/default_proto_validator.hpp
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#ifndef IROHA_DEFAULT_PROTO_VALIDATOR_HPP
```
lebdron: I would add a space between comment and ifndef.

test/module/irohad/multi_sig_transactions/transport_test.cpp
```diff
@@ -87,17 +88,61 @@ TEST(TransportTest, SendAndReceive) {
...
+ * @then there should be no SEGFAULT
+ */
+TEST(TransportTest, TypelessCommand) {
+  auto async_call_ = std::make_shared<
+      iroha::network::AsyncGrpcClient<google::protobuf::Empty>>();
+  std::unique_ptr<shared_model::validation::AbstractValidator<
+      shared_model::interface::Transaction>>
+      tx_validator = std::make_unique<
+          shared_model::validation::
+              DefaultOptionalSignedProtoTransactionValidator>();
```
lebdron: The trouble here is the same as it was in fuzzing pull request - what if the non-mocked validator gets changed in application?

MBoldyrev:
I think this kind of tests should be done on ITF, as it now has the MST handle merged in dev:
https://github.com/hyperledger/iroha/blob/4e7dbf69d42417bd3c533216ec6e097c97f550ce/test/framework/integration_framework/integration_test_framework.hpp#L252-L261
Or you could use a more convenient handle of Fake Peer that is not yet merged:
https://github.com/hyperledger/iroha/blob/312dc36c78a6e371191a7c5689a8c309eb93d6ab/test/framework/integration_framework/fake_peer/fake_peer.hpp#L153-L154

test/module/irohad/multi_sig_transactions/transport_test.cpp
```diff
@@ -101,3 +101,47 @@ TEST(TransportTest, SendAndReceive) {
...
+                                                      std::move(batch_factory),
+                                                      my_key.publicKey());
+
+  auto notifications = std::make_shared<iroha::MockMstTransportNotification>();
+  transport->subscribe(notifications);
+
+  iroha::network::transport::MstState state;
+
+  auto transaction = state.add_transactions();
+  transaction->mutable_payload()->mutable_reduced_payload()->add_commands();
```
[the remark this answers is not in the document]
igor-egorov: It does make a command without content - exactly what we are trying to test. It is ok.

test/module/irohad/multi_sig_transactions/transport_test.cpp
```diff
@@ -101,3 +101,47 @@ TEST(TransportTest, SendAndReceive) {
...
+      std::make_shared<shared_model::interface::TransactionBatchFactoryImpl>();
+
+  auto my_key = makeKey();
+  auto transport = std::make_shared<MstTransportGrpc>(async_call_,
+                                                      std::move(tx_factory),
+                                                      std::move(parser),
+                                                      std::move(batch_factory),
+                                                      my_key.publicKey());
+
+  auto notifications = std::make_shared<iroha::MockMstTransportNotification>();
```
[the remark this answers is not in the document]
igor-egorov: By design, we have to provide some consumer of events to MST transport.

test/module/irohad/multi_sig_transactions/transport_test.cpp
igor-egorov: I did as @MBoldyrev recommended - reimplemented the test via ITF, thus we always will have the same implementation as in `application` target.

test/module/irohad/multi_sig_transactions/mst_net_input_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+  itf.setInitialState(kAdminKeypair);
+  auto internal_port = itf.internalPort();
+
+  iroha::network::transport::MstState state;
+
+  auto transaction = state.add_transactions();
+  transaction->mutable_payload()->mutable_reduced_payload()->add_commands();
+
+  grpc::ClientContext context;
+  google::protobuf::Empty response;
```
MBoldyrev: It would be nice to move these to the place they are used

igor-egorov:
Need to merge despite failing CI, because the pr contains a part of fix.

PR #1866, opened 2018-11-19
Persistent cache implementation

luckychess:
### Description of the Change
Implementation of persistent cache for transactions.

### Benefits
It's impossible to deal with replays without this cache (or something similar).

### Possible Drawbacks 
None.

### Usage Examples or Tests 
```
make tx_presence_cache_test
```

### DIFF
--- a/irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
+++ b/irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
@@ -35,21 +35,20 @@ namespace iroha {
     TxCacheStatusType TxPresenceCacheImpl::checkInStorage(
         const shared_model::crypto::Hash &hash) const {
       TxCacheStatusType cache_status_check;
-      visit_in_place(
-          storage_->getBlockQuery()->checkTxPresence(hash),
-          [&](const tx_cache_status_responses::Committed &status) {
-            memory_cache_.addItem(hash, status);
-            cache_status_check = tx_cache_status_responses::Committed(hash);
-          },
-          [&](const tx_cache_status_responses::Rejected &status) {
-            memory_cache_.addItem(hash, status);
-            cache_status_check = tx_cache_status_responses::Rejected(hash);
-          },
-          [&](const tx_cache_status_responses::Missing &) {
-            // don't put this hash into cache since "Missing" can become
-            // "Committed" or "Rejected" later
-            cache_status_check = tx_cache_status_responses::Missing(hash);
-          });
+      visit_in_place(storage_->getBlockQuery()->checkTxPresence(hash),
+                     [&](const tx_cache_status_responses::Committed &status) {
+                       memory_cache_.addItem(hash, status);
+                       cache_status_check = status;
+                     },
+                     [&](const tx_cache_status_responses::Rejected &status) {
+                       memory_cache_.addItem(hash, status);
+                       cache_status_check = status;
+                     },
+                     [&](const tx_cache_status_responses::Missing &status) {
+                       // don't put this hash into cache since "Missing" can
+                       // become "Committed" or "Rejected" later
+                       cache_status_check = status;
+                     });
 
       return cache_status_check;
     }
--- a/irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
+++ b/irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
@@ -25,16 +25,16 @@ namespace iroha {
           storage_->getBlockQuery()->checkTxPresence(hash),
           [&](const tx_cache_status_responses::Committed &status) {
             memory_cache_.addItem(hash, status);
-            cache_status_check = tx_cache_status_responses::Committed();
+            cache_status_check = tx_cache_status_responses::Committed(hash);
           },
           [&](const tx_cache_status_responses::Rejected &status) {
             memory_cache_.addItem(hash, status);
-            cache_status_check = tx_cache_status_responses::Rejected();
+            cache_status_check = tx_cache_status_responses::Rejected(hash);
           },
           [&](const tx_cache_status_responses::Missing &) {
             // don't put this hash into cache since "Missing" can become
             // "Committed" or "Rejected" later
-            cache_status_check = tx_cache_status_responses::Missing();
+            cache_status_check = tx_cache_status_responses::Missing(hash);
           });
 
       return cache_status_check;
--- a/irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
+++ b/irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
@@ -35,7 +35,8 @@ namespace iroha {
     TxCacheStatusType TxPresenceCacheImpl::checkInStorage(
         const shared_model::crypto::Hash &hash) const {
       TxCacheStatusType cache_status_check;
-      visit_in_place(storage_->getBlockQuery()->checkTxPresence(hash),
+      auto check_status = storage_->getBlockQuery()->checkTxPresence(hash);
+      visit_in_place(check_status,
                      [&](const tx_cache_status_responses::Committed &status) {
                        memory_cache_.addItem(hash, status);
                        cache_status_check = status;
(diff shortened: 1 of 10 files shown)


irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include "ametsuchi/impl/tx_presence_cache_impl.hpp"
+#include "common/visitor.hpp"
+#include "interfaces/iroha_internal/transaction_batch.hpp"
+#include "interfaces/transaction.hpp"
+
+namespace iroha {
+  namespace ametsuchi {
+    TxPresenceCacheImpl::TxPresenceCacheImpl(std::shared_ptr<Storage> storage)
+        : storage_(storage) {}
```
Akvinikym: `std::move(storage)`?

irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
```diff
@@ -0,0 +1,52 @@
...
+            // "Committed" or "Rejected" later
+            cache_status_check = tx_cache_status_responses::Missing(hash);
+          });
+
+      return cache_status_check;
+    }
+
+    TxPresenceCache::BatchStatusCollectionType TxPresenceCacheImpl::check(
+        const shared_model::interface::TransactionBatch &batch) const {
+      TxPresenceCache::BatchStatusCollectionType batch_statuses{};
```
Akvinikym: If I remember correctly, braced initialization is not necessary here

irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
```diff
@@ -0,0 +1,52 @@
...
+        : storage_(storage) {}
+
+    TxCacheStatusType TxPresenceCacheImpl::check(
+        const shared_model::crypto::Hash &hash) const {
+      auto res = memory_cache_.findItem(hash);
+      if (res != boost::none) {
+        return res.get();
+      }
+
+      TxCacheStatusType cache_status_check;
```
Akvinikym: Maybe separate this part into a distinct function? This is not about checking the cache, but making a query to storage, so, I think, something like `checkInStorage(Hash)` will be better for clearness

irohad/ametsuchi/impl/tx_presence_cache_impl.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include "ametsuchi/tx_presence_cache.hpp"
+#include "cache/cache.hpp"
+
+namespace iroha {
+  namespace ametsuchi {
+
+    class TxPresenceCacheImpl : public TxPresenceCache {
+     public:
+      explicit TxPresenceCacheImpl(std::shared_ptr<Storage> storage);
+      TxCacheStatusType check(
```
Akvinikym: Empty line between ctor and `check` should be added

irohad/ametsuchi/impl/tx_presence_cache_impl.hpp
```diff
@@ -0,0 +1,36 @@
...
+          const shared_model::crypto::Hash &hash) const override;
+
+      BatchStatusCollectionType check(
+          const shared_model::interface::TransactionBatch &batch)
+          const override;
+
+     private:
+      std::shared_ptr<Storage> storage_;
+      mutable cache::Cache<shared_model::crypto::Hash,
```
Akvinikym: And, by the way, if you separate the `check` function into two parts, as I proposed earlier, `mutable` will no longer be needed

test/module/irohad/ametsuchi/tx_presence_cache_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+ */
+TEST_F(TxPresenceCacheTest, MissingHashTest) {
+  shared_model::crypto::Hash hash("1");
+  EXPECT_CALL(*mock_block_query, checkTxPresence(hash))
+      .WillOnce(Return(tx_cache_status_responses::Missing()));
+  TxPresenceCacheImpl cache(mock_storage);
+  auto check_result =
+      boost::get<tx_cache_status_responses::Missing>(cache.check(hash));
+  ASSERT_TRUE((std::is_same<tx_cache_status_responses::Missing,
+                            decltype(check_result)>::value));
```
Akvinikym: `boost::get()` throws in case it could not retrieve a value, so the `ASSERT_TRUE` is redundant here. The `auto check_result` initialization along with the following `ASSERT_EQ` should be just wrapped into a `ASSERT_NO_THROW`

Akvinikym: This is the same in all other tests

test/module/irohad/ametsuchi/tx_presence_cache_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+  std::shared_ptr<MockStorage> mock_storage;
+  std::shared_ptr<MockBlockQuery> mock_block_query;
+};
+
+/**
+ * @given hash which is not in storage
+ * @when cache asked for hash status
+ * @then cache returns Missing status
+ */
+TEST_F(TxPresenceCacheTest, MissingHashTest) {
```
kamilsa: This and two following tests are the same. You can make them type-parameterized (https://github.com/abseil/googletest/blob/master/googletest/docs/advanced.md#type-parameterized-tests). You can check the example in permissions_test.cpp

test/module/irohad/ametsuchi/tx_presence_cache_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+                        shared_model::interface::TransactionBatch>>(
+                        std::make_unique<
+                            shared_model::interface::TransactionBatchImpl>(
+                            txs));
+                  }));
+
+  batch_factory->createTransactionBatch(txs).match(
+      [&](iroha::expected::Value<
+          std::unique_ptr<shared_model::interface::TransactionBatch>> &batch) {
+        auto batch_status = cache.check(*batch.value);
```
kamilsa: I think this variable should be renamed to batch_statuses

irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
luckychess: Please, no. Performance optimisation for one-time action doesn't worth all debug time losses if storage creator decides to use this storage after move.

nickaleks: But you pass a copy here, so it's impossible to use it after move since the creator will have a different instance

luckychess: I mean a next scenario:
```
std::shared_ptr<Storage> storage = std::make_shared<...>(...);
auto cache = TxPresenceCacheImpl(storage);
....
storage-> ...   <=== trouble here
```

nickaleks: not really since you pass it by value. the original storage will remain intact

Akvinikym: Nickita is right, the pointer, which cam to this ctor, is already a copy of the original one

irohad/ametsuchi/impl/tx_presence_cache_impl.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+      TxCacheStatusType check(
+          const shared_model::crypto::Hash &hash) const override;
+
+      BatchStatusCollectionType check(
+          const shared_model::interface::TransactionBatch &batch)
+          const override;
+
+     private:
+      TxCacheStatusType checkInStorage(
```
Akvinikym: Please, add a comment

irohad/ametsuchi/impl/tx_presence_cache_impl.cpp
luckychess: True, agree with you guys.

PR #1868, opened 2018-11-20
Replay check to ordering gate

nickaleks:
Signed-off-by: Nikita Alekseev <n.alekseev2104@gmail.com>

### Description of the Change
This PR adds a replay check to ordering gate via transaction cache

### Benefits
No replays in ordering gate

### Possible Drawbacks 
None

### DIFF
--- a/irohad/ordering/impl/on_demand_ordering_gate.cpp
+++ b/irohad/ordering/impl/on_demand_ordering_gate.cpp
@@ -95,7 +95,8 @@ OnDemandOrderingGate::removeReplays(
           return true;
         },
         [](const auto &status) {
-          // TODO: log replayed transactions when log is added
+          // TODO nickaleks 21.11.18: IR-1887 log replayed transactions
+          // when log is added
           return false;
         });
   };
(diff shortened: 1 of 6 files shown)


irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -68,3 +69,39 @@ void OnDemandOrderingGate::setPcs(
...
+    const shared_model::interface::Proposal &proposal) const {
+  auto not_processed_txs = [this](const auto &tx) {
+    auto tx_result = tx_cache_->check(tx.hash());
+    return iroha::visit_in_place(
+        tx_result,
+        [](const ametsuchi::tx_cache_status_responses::Missing &) {
+          return true;
+        },
+        [](const auto &status) {
+          // TODO: log replayed transactions when log is added
```
kamilsa: Add IR number

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -68,3 +69,40 @@ void OnDemandOrderingGate::setPcs(
...
+  if (not *proposal) {
+    return nullptr;
+  }
+  return removeReplays(**proposal);
+}
+
+std::unique_ptr<shared_model::interface::Proposal>
+OnDemandOrderingGate::removeReplays(
+    const shared_model::interface::Proposal &proposal) const {
+  auto not_processed_txs = [this](const auto &tx) {
```
MBoldyrev: I suggest renaming this function to something like `tx_is_not_processed`, because a) it currently looks very similar to the variable below, `unprocessed_txs`, which is the result of applying this filter, and b) it handles single tx, so no "txs"

PR #1872, opened 2018-11-22
Fix startup flags logic (--genesis_block & --overwrite_ledger)

igor-egorov:

### Description of the Change

This is a better implementation of #1776.

The problem is that CI needs to be able to launch `irohad` that will restore the existing state with the same set of flags as that was passed at the first launch.

The new way of launch parameters handling is implemented. All the possible cases is covered.

The new truth table is below:
image

For a better understanding of changes made in code please take a look at two execution flowcharts (old and new one).
#### Old execution flow
old_flow
#### New execution flow
new_flow

### Benefits

Nothing is changed in already exposed API.
No unexpected behavior is added.
All the combinations of three parameters (presence of blockstore, presence of new genesis_block, presence of overwrite_ledger flag) are covered.

### Possible Drawbacks 

None?

### Usage Examples or Tests

```
cmake -H. -Bbuild
cd build
make irohad_test
test_bin/irohad_test
```

### DIFF
--- a/irohad/main/irohad.cpp
+++ b/irohad/main/irohad.cpp
@@ -150,6 +150,24 @@ int main(int argc, char *argv[]) {
     return EXIT_FAILURE;
   }
 
+  /*
+   * The logic implemented below is reflected in the following truth table.
+   *
+  +------------+--------------+------------------+---------------+---------+
+  | Blockstore | New genesis  | Overwrite ledger | Genesis block | Message |
+  | presence   | block is set | flag is set      | that is used  |         |
+  +------------+--------------+------------------+---------------+---------+
+  | 0          | 1            | 0                | new           |         |
+  | 0          | 1            | 1                | new           | warning |
+  | 1          | 1            | 0                | old           | warning |
+  | 1          | 1            | 1                | new           |         |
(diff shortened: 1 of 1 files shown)


irohad/main/irohad.cpp
```diff
@@ -150,57 +150,72 @@ int main(int argc, char *argv[]) {
...
+  /// if there are any blocks in blockstore, then true
+  bool blockstore = irohad.storage->getBlockQuery()->getTopBlockHeight() != 0;
+
+  /// genesis block file is specified as launch parameter
+  bool genesis = not FLAGS_genesis_block.empty();
+
+  /// overwrite ledger flag was set as launch parameter
+  bool overwrite = FLAGS_overwrite_ledger;
+
+  if (genesis) {  // genesis block file is specified
```
kamilsa: Please put truth table from description before this if else. I believe this should help for those who will read this code

bakhtin:
Can we throw a warning when blockstore is present and overwrite ledger flag is set?

igor-egorov:
> Can we throw a warning when blockstore is present and overwrite ledger flag is set?

That is already implemented

PR #1873, opened 2018-11-22
OnDemand OS: propagate outdated batches

MBoldyrev:
### Description of the Change

This change makes an OnDemand Ordering Service (ODOS) send the batches it got for some rounds that are closed for incoming batches to the currently open rounds on corresponding peers.

### Benefits

Some proposals that otherwise would be lost now get used.

### Possible Drawbacks 

To be researched.


MBoldyrev:
The issue of this PR is supposed to be solved after the base branch is merged to dev.

PR #1877, opened 2018-11-24
Synchronize OnDemand OS ConnectionManager propagation data

MBoldyrev:
### Description of the Change

This change makes OdOs Connection Manager aware of the round for which its destination peers are valid. Now it will wait for the peers data to update if the batch it gets has a newer round, or set the destination of the batch to an actual round if the requested destination is outdated.

### Benefits

Aimed at mitigation of (one of causes of) [the lost proposal bug|https://soramitsu.atlassian.net/browse/IR-1888].

### Possible Drawbacks 

Bugs!

### DIFF
--- a/irohad/ordering/impl/on_demand_connection_manager.cpp
+++ b/irohad/ordering/impl/on_demand_connection_manager.cpp
@@ -5,36 +5,72 @@
 
 #include "ordering/impl/on_demand_connection_manager.hpp"
 
+#include <chrono>
+
 #include <boost/range/combine.hpp>
 #include "interfaces/iroha_internal/proposal.hpp"
 #include "ordering/impl/on_demand_common.hpp"
+
+/// The time to wait for the up-to-date propagation data arrives through
+/// propagation_params_observable (see constructor) when we got a request for a
+/// round that is newer than our propagation data
+static constexpr std::chrono::milliseconds kPropagationDataWaitTimeout(10);
 
 using namespace iroha::ordering;
 
 OnDemandConnectionManager::OnDemandConnectionManager(
     std::shared_ptr<transport::OdOsNotificationFactory> factory,
-    rxcpp::observable<CurrentPeers> peers)
+    rxcpp::observable<PropagationParams> propagation_params_observable)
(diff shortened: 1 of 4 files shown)


irohad/ordering/impl/on_demand_connection_manager.cpp
```diff
@@ -5,37 +5,73 @@
...
+  initializeConnections(inital_propagation_params);
 }
 
-void OnDemandConnectionManager::onBatches(consensus::Round round,
+void OnDemandConnectionManager::onBatches(const consensus::Round round,
                                           CollectionType batches) {
   std::shared_lock<std::shared_timed_mutex> lock(mutex_);
 
+  // check if we have appropriate propagation data:
+  if (current_round_ != round) {
```
l4l: might be moved to separate function and reworked with early return, e.g:
```cpp
void func(...) {
  if (current_round_ == round) {
    return
  }
  ...
}
```

irohad/ordering/impl/on_demand_connection_manager.cpp
```diff
@@ -5,37 +5,52 @@
...
+  initializeConnections(inital_propagation_params);
 }
 
-void OnDemandConnectionManager::onBatches(consensus::Round round,
+void OnDemandConnectionManager::onBatches(const consensus::Round round,
                                           CollectionType batches) {
   std::shared_lock<std::shared_timed_mutex> lock(mutex_);
 
+  // check if we have an appropriate propagation data:
+  if (!verifyPropagtionData(round, lock)) {
```
l4l: It's hard to follow the function logic. I'd suggest passing waiting callback instead, e.g:

```cpp
verifyPropagationData(round, propagation_data_update_cv_.wait_for(
          lock, kPropagationDataWaitTimeout, [this, &request_round] {
            return not(this->current_round_ < request_round);
          }));
```

l4l: or maybe even remove function and inline it's body here

MBoldyrev:
The issue of this PR is supposed to be solved after the base branch is merged to dev.

PR #1879, opened 2018-11-26
Fix Protocol and Interface Validators

Akvinikym:
### Description of the Change

Detailed description of the bug can be found [here](https://soramitsu.atlassian.net/browse/IR-1902).

### Benefits

Validation is now performed in the right order, no more segfaults here.

### Possible Drawbacks 

ProtoTransportFactory now must accept two validators instead of one.

`proto_query_validator_test` was added to reflect changes in query validation process.


test/module/irohad/torii/torii_service_test.cpp
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <validators/protobuf/proto_transaction_validator.hpp>
```
lebdron:
```suggestion
#include "validators/protobuf/proto_transaction_validator.hpp"
```

test/module/shared_model/validators/protobuf/proto_query_validator_test.cpp
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "validators/protobuf/proto_query_validator.hpp"
```
lebdron:
```C++
#include "validators/protobuf/proto_query_validator.hpp"

#include "module/shared_model/validators/validators_fixture.hpp"
#include "queries.pb.h"
```
since `proto_query_validator` is the header of the class being tested here

test/module/shared_model/validators/protobuf/proto_query_validator_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+class ProtoQueryValidatorTest : public ValidatorsTest {
+ public:
+  shared_model::validation::ProtoQueryValidator validator;
+};
+
+/**
+ * @given Protobuf query object with unset query
+ * @when validate is called
+ * @then there is a error returned
```
lebdron:
```suggestion
 * @then there is an error returned
```

test/module/shared_model/validators/protobuf/proto_query_validator_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+ */
+TEST_F(ProtoQueryValidatorTest, UnsetQuery) {
+  iroha::protocol::Query qry;
+  qry.mutable_payload()->mutable_meta()->set_created_time(created_time);
+  qry.mutable_payload()->mutable_meta()->set_creator_account_id(account_id);
+  qry.mutable_payload()->mutable_meta()->set_query_counter(counter);
+
+  auto answer = validator.validate(qry);
+  ASSERT_TRUE(answer.hasErrors());
```
lebdron: Since the test checks that an undefined query does not pass validation, it would be great to check that the validation returns error for that particular reason. Maybe add `MATCHES("undefined")` assertion, since there are no error codes at the moment?

PR #1884, opened 2018-11-27
Feature/update qry response factory

kamilsa:
### Description of the Change

Add transactions page responses to the query response factory

### Benefits

Now it is possible to create transaction page responses using factory

### Possible Drawbacks 

None
### Usage Examples or Tests

Test, checking creation of query pagination response using factory has been added


shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
```diff
@@ -196,6 +196,32 @@ shared_model::proto::ProtoQueryResponseFactory::createTransactionsResponse(
       query_hash);
 }
 
+std::unique_ptr<shared_model::interface::QueryResponse>
+shared_model::proto::ProtoQueryResponseFactory::createTransactionsPageResponse(
```
Akvinikym: Maybe, the response should be called `transactionsPagedResponse`? More correct from language point of view

shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
```diff
@@ -196,6 +196,32 @@ shared_model::proto::ProtoQueryResponseFactory::createTransactionsResponse(
...
+    std::vector<std::unique_ptr<shared_model::interface::Transaction>>
+        transactions,
+    const crypto::Hash &next_tx_hash,
+    interface::types::TransactionsNumberType all_transactions_size,
+    const crypto::Hash &query_hash) const {
+  return createQueryResponse(
+      [transactions = std::move(transactions),
+       &next_tx_hash,
+       &all_transactions_size](
```
Akvinikym: If `all_transactions_size` is a "base" class (I don't know, is it something like `uint_32`?), then there's no need to capture it by reference, by value it should be

shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
```diff
@@ -196,6 +196,32 @@ shared_model::proto::ProtoQueryResponseFactory::createTransactionsResponse(
       query_hash);
 }
 
+std::unique_ptr<shared_model::interface::QueryResponse>
+shared_model::proto::ProtoQueryResponseFactory::createTransactionsPageResponse(
+    std::vector<std::unique_ptr<shared_model::interface::Transaction>>
+        transactions,
+    const crypto::Hash &next_tx_hash,
+    interface::types::TransactionsNumberType all_transactions_size,
```
Akvinikym: Don't like name `all_transactions_size`. What exactly does this field in proto mean? Number of transactions, which are to be retrieved via the query? Maybe, it's better then to use `requested_transactions_amount` or something close.

kamilsa: All transactions size is the total number of transactions in all pages. Any better idea for the naming?

shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
```diff
@@ -196,6 +196,32 @@ shared_model::proto::ProtoQueryResponseFactory::createTransactionsResponse(
...
+    const crypto::Hash &query_hash) const {
+  return createQueryResponse(
+      [transactions = std::move(transactions),
+       &next_tx_hash,
+       &all_transactions_size](
+          iroha::protocol::QueryResponse &protocol_query_response) {
+        iroha::protocol::TransactionsPageResponse *protocol_specific_response =
```
nickaleks:
```suggestion
       auto *protocol_specific_response =
```

test/module/shared_model/backend_proto/proto_query_response_factory_test.cpp
```diff
@@ -282,6 +283,47 @@ TEST_F(ProtoQueryResponseFactoryTest, CreateTransactionsResponse) {
...
+
+  ASSERT_TRUE(query_response);
+  ASSERT_EQ(query_response->queryHash(), kQueryHash);
+  ASSERT_NO_THROW({
+    const auto &response =
+        boost::get<const shared_model::interface::TransactionsPageResponse &>(
+            query_response->get());
+
+    for (auto i = 0; i < kTransactionsNumber; ++i) {
+      ASSERT_EQ(response.transactions()[i].creatorAccountId(),
```
nickaleks: Please replace all 3 asserts to expects, since it will check all conditions if one of them fails, where as ASSERT will terminate the test

test/module/shared_model/backend_proto/proto_query_response_factory_test.cpp
```diff
@@ -282,6 +283,47 @@ TEST_F(ProtoQueryResponseFactoryTest, CreateTransactionsResponse) {
...
+      std::move(transactions), kNextTxHash, kTransactionsNumber, kQueryHash);
+
+  ASSERT_TRUE(query_response);
+  ASSERT_EQ(query_response->queryHash(), kQueryHash);
+  ASSERT_NO_THROW({
+    const auto &response =
+        boost::get<const shared_model::interface::TransactionsPageResponse &>(
+            query_response->get());
+
+    for (auto i = 0; i < kTransactionsNumber; ++i) {
```
nickaleks: This test will pass if the response contains extra transactions since they will not be checked.
I suggest first verify that the size of transactions() is the same as kTransactionNumber

shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
kamilsa: Depends how to read that:) TransactionsPage Response. i.e. Response with TransactionsPage. I think it is good, isn't it?

shared_model/backend/protobuf/impl/proto_query_response_factory.cpp
kamilsa: I think both options how to capture it is fine. It is only a matter of the code style

PR #1887, opened 2018-11-27
Schema update for queries with lists of transactions

MBoldyrev:
### Description of the Change

[IR-20](https://jira.hyperledger.org/projects/IR/issues/IR-20)
This change adds the pagination metadata to some queries that request transaction lists: GetAccountTransactions and GetAccountAssetTransactions. It does not concern GetPendingTransactions as it is expected to be rewritten soon.

### Benefits

A small step towards great feature.

### Possible Drawbacks 

bugs.

### Usage Examples or Tests

When a client requests some sort of transaction list, the request must include pagination metadata. That one _must_ have the page size set  to the maximum amount of transactions the client wants to get at once, and _may_ have the first transaction hash field set to the hash of the transaction, from which the returned list is supposed to start.

### DIFF
--- a/shared_model/backend/protobuf/queries/impl/proto_get_account_asset_transactions.cpp
+++ b/shared_model/backend/protobuf/queries/impl/proto_get_account_asset_transactions.cpp
@@ -5,6 +5,8 @@
 
 #include "backend/protobuf/queries/proto_get_account_asset_transactions.hpp"
 
+#include "backend/protobuf/queries/proto_tx_pagination_meta.hpp"
+
 namespace shared_model {
   namespace proto {
 
@@ -37,6 +39,12 @@ namespace shared_model {
     const interface::types::AssetIdType &GetAccountAssetTransactions::assetId()
         const {
       return account_asset_transactions_.asset_id();
+    }
+
+    std::unique_ptr<interface::TxPaginationMeta>
+    GetAccountAssetTransactions::paginationMeta() const {
+      return std::make_unique<TxPaginationMeta>(
+          account_asset_transactions_.pagination_meta());
     }
 
   }  // namespace proto
--- a/shared_model/backend/protobuf/queries/impl/proto_get_account_asset_transactions.cpp
+++ b/shared_model/backend/protobuf/queries/impl/proto_get_account_asset_transactions.cpp
@@ -14,7 +14,8 @@ namespace shared_model {
     GetAccountAssetTransactions::GetAccountAssetTransactions(QueryType &&query)
         : CopyableProto(std::forward<QueryType>(query)),
           account_asset_transactions_{
-              proto_->payload().get_account_asset_transactions()} {}
+              proto_->payload().get_account_asset_transactions()},
+          pagination_meta_{account_asset_transactions_.pagination_meta()} {}
 
     template GetAccountAssetTransactions::GetAccountAssetTransactions(
         GetAccountAssetTransactions::TransportType &);
@@ -41,10 +42,9 @@ namespace shared_model {
       return account_asset_transactions_.asset_id();
     }
 
-    std::unique_ptr<interface::TxPaginationMeta>
+    const interface::TxPaginationMeta &
     GetAccountAssetTransactions::paginationMeta() const {
-      return std::make_unique<TxPaginationMeta>(
-          account_asset_transactions_.pagination_meta());
+      return pagination_meta_;
     }
 
   }  // namespace proto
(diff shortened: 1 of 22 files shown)


shared_model/interfaces/common_objects/types.hpp
```diff
@@ -75,7 +75,7 @@ namespace shared_model {
       using AccountDetailKeyType = std::string;
       /// Type of account detail value
       using AccountDetailValueType = std::string;
-      /// Type of a number of transactions in block
+      /// Type of a number of transactions in block and query responce page
```
Akvinikym: query responSe page

shared_model/backend/protobuf/queries/impl/proto_get_account_asset_transactions.cpp
```diff
@@ -39,5 +41,11 @@ namespace shared_model {
       return account_asset_transactions_.asset_id();
     }
 
+    std::unique_ptr<interface::TxPaginationMeta>
```
Akvinikym: Why do we want to return a unique_ptr? Is the meta going to be ever changed? If not, why not returning a const ref, as it is done in other methods of the class?

shared_model/backend/protobuf/queries/impl/proto_tx_pagination_meta.cpp
```diff
@@ -0,0 +1,37 @@
...
+namespace types = shared_model::interface::types;
+
+using namespace shared_model::proto;
+
+TxPaginationMeta::TxPaginationMeta(const TransportType &query)
+    : CopyableProto(query) {}
+
+TxPaginationMeta::TxPaginationMeta(TransportType &&query)
+    : CopyableProto(
+          static_cast<std::remove_reference<TransportType>::type &&>(query)) {}
```
Akvinikym: Why such complex structure and not a simple `move`?

shared_model/interfaces/queries/tx_pagination_meta.hpp
```diff
@@ -0,0 +1,33 @@
...
+      virtual types::TransactionsNumberType pageSize() const = 0;
+
+      /// Get the first requested transaction hash, if provided.
+      virtual boost::optional<types::HashType> firstTxHash() const = 0;
+
+      std::string toString() const override;
+
+      bool operator==(const ModelType &rhs) const override;
+    };
+  }  // namespace interface
```
Akvinikym: Empty line should be inserted

shared_model/backend/protobuf/queries/impl/proto_get_account_asset_transactions.cpp
MBoldyrev: The other methods return primitive objects, while this requires construction. Returning a constref would require us to keep a copy of `proto::TxPaginationMeta` inside all proto query classes. I think a unique pointer is a more clear way to provide `proto::TxPaginationMeta` when it is needed.

Akvinikym: Please, consider this issue again before merging

shared_model/backend/protobuf/queries/proto_tx_pagination_meta.hpp
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IROHA_SHARED_PROTO_MODEL_QUERY_TX_PAGINATION_META_HPP
+#define IROHA_SHARED_PROTO_MODEL_QUERY_TX_PAGINATION_META_HPP
```
Akvinikym: Include guards in other files in this folder are called `IROHA_SHARED_MODEL_PROTO_...`. Please, rename this one according to the common scheme

shared_model/interfaces/queries/tx_pagination_meta.hpp
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IROHA_SHARED_INTERFACE_MODEL_QUERY_TX_PAGINATION_META_HPP
+#define IROHA_SHARED_INTERFACE_MODEL_QUERY_TX_PAGINATION_META_HPP
```
Akvinikym: The same

shared_model/backend/protobuf/queries/impl/proto_get_account_asset_transactions.cpp
kamilsa: Here we would need to create an object every time pagination meta object is invoked, whereas it is only access methods. Please change to cont ref as it is done for example for amount in transfer asset. It is okay to construct  pagination meta

PR #1888, opened 2018-11-27
Refactor simulator with optional

lebdron:
### Description of the Change
Correctly handle `optional<Proposal>` in simulator, and return `boost::none` instead of block if there was no proposal.

### Benefits
Another step to handle `none` from ordering service.

### Possible Drawbacks 
None?

### DIFF
--- a/irohad/torii/processor/CMakeLists.txt
+++ b/irohad/torii/processor/CMakeLists.txt
@@ -1,3 +1,6 @@
+# Copyright Soramitsu Co., Ltd. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
 add_library(processors
     impl/transaction_processor_impl.cpp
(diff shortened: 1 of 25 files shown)


irohad/torii/processor/CMakeLists.txt
```diff
@@ -11,4 +11,5 @@ target_link_libraries(processors PUBLIC
     shared_model_proto_builders
     status_bus
     common
+    verified_proposal_creator_common
```
igor-egorov: Since you have been touching the file, please add license header.

lebdron: https://github.com/hyperledger/iroha/blob/fe542f16b69c0962943ed9a808dc2a31554308be/irohad/torii/processor/CMakeLists.txt#L1-L2

PR #1891, opened 2018-11-28
Transaction persistent cache

lebdron:
### Description of the Change
Prevent replay attacks by querying the storage for processed transactions in the following components: MST transport, ordering gate, ordering service.

### Benefits
Already processed transactions are rejected when they are received.

### Possible Drawbacks 
Persistent storage lookup overhead.


MBoldyrev:
As I wrote: https://github.com/hyperledger/iroha/pull/1865#issuecomment-442777786, it seems that replay policies should be applied outside "transport" classes. Not only it would granulate responsibilities and suit the OOP style, but also remove the need for tricks where we need a transport with no replay checks. All the logic besides internal types conversion and transfer should be kept outside transport classes, I think.
However this is just a suggestion and it does not directly affect implemented functions, so I suggest creating a separate task for refactoring which would not block this feature.

PR #1893, opened 2018-11-29
Feature/oneof block

kamilsa:
### Description of the Change
Before it was:

```
message Block {
  // block_v1 fields
}
```

From now on it will be
```
message Block_v1 {
// block_v1 fields
}

message Block {
  oneof block_version {
    Block_v1 block_v1 = 1;
  }
}
```

Such changes would allow us to introduce Block_v2 and further versions, by simply adding them to oneof in Block message. Internally we will be able to recognize and manage different version of the blocks.

### Benefits

Iroha will be able to process different versions of the blocks.

### Possible Drawbacks 

None

### DIFF
--- a/iroha-cli/main.cpp
+++ b/iroha-cli/main.cpp
@@ -17,13 +17,14 @@
 
 #include <gflags/gflags.h>
 #include <rapidjson/istreamwrapper.h>
-#include <rapidjson/rapidjson.h>
 #include <boost/filesystem.hpp>
 #include <iostream>
 
+#include "backend/protobuf/proto_block_json_converter.hpp"
 #include "backend/protobuf/queries/proto_query.hpp"
 #include "backend/protobuf/transaction.hpp"
 #include "client.hpp"
+#include "common/result.hpp"
 #include "converters/protobuf/json_proto_converter.hpp"
 #include "crypto/keys_manager_impl.hpp"
 #include "grpc_response_handler.hpp"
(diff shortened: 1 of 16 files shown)


Warchant:
Another drawback is legacy for new networks :(

shared_model/schema/block.proto
```diff
@@ -27,3 +27,9 @@ message Block {
   Payload payload = 1;
   repeated Signature signatures = 2;
 }
+
+message Block {
+  oneof block_version {
```
muratovv: Maybe stay just `version` here?

iroha-cli/main.cpp
```diff
@@ -17,13 +17,14 @@
 
 #include <gflags/gflags.h>
 #include <rapidjson/istreamwrapper.h>
-#include <rapidjson/rapidjson.h>
```
MBoldyrev: I suggest not removing this include, even if it works without it, because we directly create and use `rapidjson::Document` in this file.

iroha-cli/main.cpp
```diff
@@ -122,9 +123,19 @@ int main(int argc, char *argv[]) {
...
+
+    shared_model::proto::ProtoBlockJsonConverter().serialize(bl).match(
+        [&logger,
+         &output_file](const iroha::expected::Value<std::string> &json) {
+          output_file << json.value;
+          logger->info("File saved to genesis.block");
+        },
+        [&logger](const auto &error) {
+          // should not get here
+          logger->error("error while parsing serializing genesis block");
```
MBoldyrev:
```suggestion
          logger->error("error while parsing serialized genesis block: " + error.error);
```

PR #1901, opened 2018-12-03
Query Stateful Validation Error Codes

Akvinikym:
### Description of the Change

Commands have set of stateful validation error codes, but queries still had just error messages. This PR introduces error codes, described in the following [POW](https://soramitsu.atlassian.net/wiki/spaces/IS/pages/439189508/POW+N+075+Stateful+response).

### Benefits

Unified way to pass error message to the client.

### Possible Drawbacks 

None.

### DIFF
--- a/shared_model/interfaces/query_responses/error_query_response.hpp
+++ b/shared_model/interfaces/query_responses/error_query_response.hpp
@@ -1,18 +1,6 @@
 /**
- * Copyright Soramitsu Co., Ltd. 2017 All Rights Reserved.
- * http://soramitsu.co.jp
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *        http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
  */
 
 #ifndef IROHA_SHARED_MODEL_QUERY_ERROR_RESPONSE_HPP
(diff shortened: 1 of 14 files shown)


test/module/irohad/ametsuchi/postgres_query_executor_test.cpp
```diff
@@ -339,11 +376,8 @@ namespace iroha {
                        .getAccount("some@domain")
                        .build();
       auto result = executeQuery(query);
-      ASSERT_TRUE(boost::apply_visitor(
-          shared_model::interface::QueryErrorResponseChecker<
-              shared_model::interface::NoAccountErrorResponse>(),
-          result->get()))
-          << result->toString();
+      checkStatefulError<shared_model::interface::NoAccountErrorResponse>(
+          std::move(result), 0);
```
kamilsa: Can we at add constants for the test? At least local variables with good naming?

Akvinikym: OK, constants for test only can be added, but they are to be substituted by the solution for named constants of both Command and Query Executors

shared_model/interfaces/query_responses/error_query_response.hpp
```diff
@@ -72,6 +60,18 @@ namespace shared_model {
        */
       virtual const ErrorMessageType &errorMessage() const = 0;
 
+      /// Error code type
+      using ErrorCodeType = uint32_t;
```
igor-egorov: Should not this be defined in `shared_model/interfaces/common_objects/types.hpp`?

test/module/irohad/ametsuchi/postgres_query_executor_test.cpp
```diff
@@ -255,13 +304,10 @@ namespace iroha {
...
       auto result = executeQuery(query);
-      ASSERT_NO_THROW({
-        const auto &cast_resp = boost::apply_visitor(
-            framework::SpecifiedVisitor<
-                shared_model::interface::AccountResponse>(),
-            result->get());
-        ASSERT_EQ(cast_resp.account().accountId(), account->accountId());
-      }) << result->toString();
+      checkSuccessfulResult<shared_model::interface::AccountResponse>(
+          std::move(result), [this](auto &&cast_resp) {
```
igor-egorov: I believe `const auto &cast_resp` should work well here. I don't see a reason to use `&&` here.

PR #1913, opened 2018-12-05
Remove ReferenceHolder from some objects

Akvinikym:
### Description of the Change

In QueryResponse, TransactionResponse and Query classes ReferenceHolder is not needed for keeping proto part - storing it by value is enough.

### Benefits

Unnecessary usage of ReferenceHolder is removed.

### Possible Drawbacks 

None.


shared_model/backend/protobuf/impl/transaction.cpp
```diff
@@ -40,8 +40,9 @@ namespace shared_model {
           shared_model::crypto::Sha3_256::makeHash(reduced_payload_blob_)};
 
       std::vector<proto::Command> commands_{[this] {
-        return std::vector<proto::Command>(reduced_payload_.commands().begin(),
-                                           reduced_payload_.commands().end());
+        return std::vector<proto::Command>(
```
nickaleks: Please remove this lambda

shared_model/backend/protobuf/commands/impl/proto_command.cpp
```diff
@@ -87,7 +81,7 @@ namespace shared_model {
     }
 
     Command *Command::clone() const {
-      return new Command(*impl_->proto_);
+      std::terminate();
```
muratovv: Add logging information here

PR #1914, opened 2018-12-05
Enable system test

lebdron:
### Description of the Change
Enable irohad system test
- Pass boost::none instead of empty proposal
- Use delay to reduce the number of reject rounds

### Benefits
Working OS

### Possible Drawbacks 
Yet to discover

### DIFF
--- a/irohad/main/impl/on_demand_ordering_init.cpp
+++ b/irohad/main/impl/on_demand_ordering_init.cpp
@@ -172,22 +172,32 @@ namespace iroha {
             proposal_factory,
         std::shared_ptr<ametsuchi::TxPresenceCache> tx_cache,
         consensus::Round initial_round) {
-      auto delay = [reject_counter = static_cast<uint64_t>(0),
-                    local_counter =
-                        static_cast<uint64_t>(0)](const auto &commit) mutable {
+      // TODO andrei 06.12.18 IR-75 Make counter and generator parametrizable
+      const uint64_t kCounter = 0, kMaxLocalCounter = 2;
+      auto time_generator = [](auto reject_counter) {
+        return std::chrono::seconds(reject_counter);
+      };
+      // reject_counter and local_counter are local mutable variables of lambda
+      auto delay = [reject_counter = kCounter,
+                    local_counter = kCounter,
+                    &time_generator,
+                    kMaxLocalCounter](const auto &commit) mutable {
         using iroha::synchronizer::SynchronizationOutcomeType;
         if (commit.sync_outcome == SynchronizationOutcomeType::kReject
             or commit.sync_outcome == SynchronizationOutcomeType::kNothing) {
-          if (local_counter++ == 2) {
+          // Increment reject_counter each local_counter calls of function
+          ++local_counter;
+          if (local_counter == kMaxLocalCounter) {
             local_counter = 0;
-            if (reject_counter < std::numeric_limits<uint64_t>::max()) {
+            if (reject_counter
+                < std::numeric_limits<decltype(reject_counter)>::max()) {
               reject_counter++;
             }
           }
         } else {
           reject_counter = 0;
         }
-        return std::chrono::seconds(reject_counter);
+        return time_generator(reject_counter);
       };
 
       auto map = [](auto commit) {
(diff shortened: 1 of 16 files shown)


irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -81,6 +81,7 @@ void OnDemandOrderingServiceImpl::onBatches(consensus::Round round,
                        return request_reject_round == reject_round
                            or (request_reject_round >= 2 and reject_round >= 2);
                      });
+        // add log
```
kamilsa: please add log

libs/common/delay.hpp
```diff
@@ -0,0 +1,176 @@
...
+ */
+
+#ifndef IROHA_DELAY_HPP
+#define IROHA_DELAY_HPP
+
+#include <rxcpp/operators/rx-delay.hpp>
+
+namespace iroha {
+
+  template <class T, class Selector, class Coordination>
```
kamilsa: please add explanation of the meaning for that file

irohad/main/impl/on_demand_ordering_init.cpp
```diff
@@ -171,38 +172,62 @@ namespace iroha {
             proposal_factory,
         std::shared_ptr<ametsuchi::TxPresenceCache> tx_cache,
         consensus::Round initial_round) {
+      auto delay = [reject_counter = static_cast<uint64_t>(0),
```
muratovv: What is the point to use a static cast? I've expected to use literals instead: `0u`.

irohad/main/impl/on_demand_ordering_init.cpp
```diff
@@ -171,38 +172,62 @@ namespace iroha {
...
+                    local_counter =
+                        static_cast<uint64_t>(0)](const auto &commit) mutable {
+        using iroha::synchronizer::SynchronizationOutcomeType;
+        if (commit.sync_outcome == SynchronizationOutcomeType::kReject
+            or commit.sync_outcome == SynchronizationOutcomeType::kNothing) {
+          if (local_counter++ == 2) {
```
muratovv: I guess it is better to move out increment operator from `if` statement.

Also, constant `2` is not so meaningful better use names constant here or describe why two is enough.

irohad/main/impl/on_demand_ordering_init.cpp
muratovv: Nice trick with storing variables in closure but it seems that there is a need to add some comments about this fact.

irohad/main/impl/on_demand_ordering_init.cpp
```diff
@@ -171,38 +172,62 @@ namespace iroha {
...
+            local_counter = 0;
+            if (reject_counter < std::numeric_limits<uint64_t>::max()) {
+              reject_counter++;
+            }
+          }
+        } else {
+          reject_counter = 0;
+        }
+        return std::chrono::seconds(reject_counter);
```
muratovv: What is the point to increase delay each reject round? I think better to increase rejects only after f malicious rounds because first f rounds could be conducted by malicious os.

irohad/main/impl/on_demand_ordering_init.cpp
```diff
@@ -171,38 +172,62 @@ namespace iroha {
...
-                [](const auto &)
-                    -> ordering::OnDemandOrderingGate::BlockRoundEventType {
-                  return ordering::OnDemandOrderingGate::EmptyEvent{};
-                });
-          }),
+          notifier.get_observable()
+              .lift<iroha::synchronizer::SynchronizationEvent>(
+                  iroha::makeDelay<iroha::synchronizer::SynchronizationEvent>(
+                      delay, rxcpp::identity_current_thread()))
+              .map(map),
```
muratovv: In which thread map will be invoked?

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -39,7 +39,7 @@ namespace iroha {
               proposal_factory,
           std::shared_ptr<ametsuchi::TxPresenceCache> tx_cache,
           size_t number_of_proposals = 3,
-          const consensus::Round &initial_round = {2, 1});
+          const consensus::Round &initial_round = {2, 0});
```
muratovv: Named constant police

irohad/main/impl/on_demand_ordering_init.cpp
lebdron: `on_next` caller thread of `notifier`.

PR #1917, opened 2018-12-06
Get Account Asset transactions pagination

nickaleks:
### Description of the Change

This PR introduces pagination logic to GetAccountAssetTransactions query.

### Benefits
Pagination for query

### Possible Drawbacks 

Since logic of getAccountTransactions and GetAccountAssetTransactions is almost the same, their logic was moved to a common function. This function might be confusing.

### DIFF
--- a/irohad/ametsuchi/impl/postgres_query_executor.cpp
+++ b/irohad/ametsuchi/impl/postgres_query_executor.cpp
@@ -651,13 +651,11 @@ namespace iroha {
 
       auto cmd = (boost::format(R"(WITH has_perms AS (%s),
       t AS (
-          SELECT DISTINCT has.height, index
-          FROM height_by_account_set AS has
-          JOIN index_by_id_height_asset AS ich ON has.height = ich.height
-          AND has.account_id = ich.id
+          SELECT DISTINCT height, index
+          FROM position_by_account_asset
           WHERE account_id = :account_id
           AND asset_id = :asset_id
-          ORDER BY has.height, index ASC
+          ORDER BY height, index ASC
       )
       SELECT height, index, perm FROM t
       RIGHT OUTER JOIN has_perms ON TRUE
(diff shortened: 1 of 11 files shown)


irohad/ametsuchi/impl/postgres_query_executor.cpp
```diff
@@ -342,6 +342,119 @@ namespace iroha {
...
+            position_by_hash.index >= first_hash.index)
+      ),
+      total_size AS (
+        SELECT COUNT(*) FROM my_txs
+      ),
+      t AS (
+        SELECT my_txs.height, my_txs.index
+        FROM my_txs
+        JOIN previous_txes ON my_txs.height = previous_txes.height
+        AND my_txs.index = previous_txes.index
```
MBoldyrev: And do we need that additional temporary table at all? I mean current `previous_txes`. Seems like we can write
```sql
        JOIN first_hash ON my_txs.height > first_hash.height
        OR (my_txs.height = first_hash.height AND
            my_txs.index >= first_hash.index)
```
and remove the `previous_txes` completely

irohad/ametsuchi/impl/postgres_query_executor.cpp
```diff
@@ -342,6 +342,119 @@ namespace iroha {
...
+        LIMIT :page_size
+      )
+      SELECT height, index, count, perm FROM t
+      RIGHT OUTER JOIN has_perms ON TRUE
```
MBoldyrev: Quite strange join. Maybe place a todo to rewrite it?

nickaleks: We want to add permissions to the result of the query, that's why we join

MBoldyrev: It is strange that we use `RIGHT OUTER JOIN .. ON TRUE`, while it seems we could use just `CROSS JOIN`

PR #1921, opened 2018-12-06
Print user-defined objects to log directly

Akvinikym:
### Description of the Change

Previously, user-defined objects were printed into the log via calling their `toString()` method, which is not very efficient. This PR introduces an ability to log such objects directly, and their `toString()` will be called only if level of logging meets the one turned on.

### Benefits

No more additional logging overhead when unnecessary.

### Possible Drawbacks 

None.

### DIFF
--- a/libs/logger/logger.hpp
+++ b/libs/logger/logger.hpp
@@ -1,27 +1,25 @@
 /**
- * Copyright Soramitsu Co., Ltd. 2018 All Rights Reserved.
- * http://soramitsu.co.jp
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


libs/logger/logger.hpp
```diff
@@ -1,28 +1,26 @@
...
 
-#include <spdlog/spdlog.h>
 #include <memory>
 #include <numeric>  // for std::accumulate
 #include <string>
 
+#include <spdlog/fmt/ostr.h>
+#include <spdlog/spdlog.h>
+
+/// Allows to log objects, which has toString() method without calling it, e.g.
```
MBoldyrev:
```suggestion
/// Allows to log objects, which have toString() method, without calling it, e.g.
```

PR #1924, opened 2018-12-06
Feature/domain asset permission

grimadas:
### Description of the Change

This PR adds domain permissions for commands: AddAssetQty and SubtractAssetQty.

### Benefits

Following changes allow restricting permissions for issuing asset for specific domains. For example, one can create assets only in their domain if has this permission. 

### Possible Drawbacks 

None


neewy:
I would change the target into trunk, where we can update documentation as well @grimadas

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -68,6 +69,8 @@ namespace {
     return {with_validation.str(), without_validation.str()};
   }
 
+  auto hasCommandPermission() {}
```
Akvinikym: What it is this?

grimadas: Thanks, removed

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -264,6 +265,31 @@ namespace {
...
+      shared_model::interface::permissions::Role domain_role,
+      const shared_model::interface::types::AccountIdType &creator_id,
+      const shared_model::interface::types::AccountIdType
+          &id_with_target_domain) {
+    std::string query = (boost::format(R"(WITH
+          has_global_role_perm AS (%1%),
+          has_domain_role_perm AS (%2%)
+          SELECT CASE
+                           WHEN (SELECT * FROM has_global_role_perm) THEN true
+                           WHEN ((split_part(%3%, '@', 2) = split_part(%4%, '@', 2))
```
kamilsa: I think not this WHEN never used. You can change the type of id_with_target_domain to AssetIdType

test/module/irohad/ametsuchi/postgres_executor_test.cpp
```diff
@@ -103,6 +103,23 @@ namespace iroha {
                         true)));
       }
 
+      void addOnePerm(
```
kamilsa: Leave a comment please

PR #1925, opened 2018-12-07
Fix uncommitted cache

kamilsa:

### Description of the Change

Previously uncommitted transactions cache was removing by reduced hash of transaction batch. However block event emitted to ordering gate returns hashes of transactions. So the fix removes batches from the cache by transaction's hash. So, if any batch in transaction cache contains transaction with given transaction hash, then entire batch with that transaction is removed from the cache

### Benefits

No more duplication warnings

### Possible Drawbacks 

None

### Usage Examples or Tests

on_demand_cache_test with remove function was updated

### DIFF
--- a/test/module/shared_model/interface_mocks.hpp
+++ b/test/module/shared_model/interface_mocks.hpp
@@ -7,6 +7,7 @@
 #define IROHA_SHARED_MODEL_INTERFACE_MOCKS_HPP
 
 #include <gmock/gmock.h>
+#include "cryptography/crypto_provider/crypto_defaults.hpp"
 #include "cryptography/public_key.hpp"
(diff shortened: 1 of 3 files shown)


test/module/shared_model/interface_mocks.hpp
```diff
@@ -110,6 +127,27 @@ auto createMockBatchWithHash(
   return res;
 }
 
+/**
+ * Creates mock batch with provided transactions
+ * @param txs -- const ref to hash to be returned by the batch
```
lebdron: 🙈  txs is not a hash.
I suggest to rewrite docstring with one line such as "Creates mock batch with provided transactions and reduced hash".

PR #1929, opened 2018-12-07
Add round check to consensus gate, fix ordering gate round

lebdron:
### Description of the Change
- Ordering gate was generating incorrect reject rounds based on initial stored value. Now it uses the value from observable
- Yac gate was accepting votes and consensus results for old rounds. Now it ignores them

### Benefits
Predictable pipeline behavior

### Possible Drawbacks 
None?

YacGateTest.OlderVote, YacGateTest.OlderCommit, YacGateTest.OlderReject

### DIFF
--- a/irohad/consensus/yac/impl/yac_gate_impl.cpp
+++ b/irohad/consensus/yac/impl/yac_gate_impl.cpp
@@ -33,12 +33,24 @@ namespace iroha {
             hash_provider_(std::move(hash_provider)),
             block_creator_(std::move(block_creator)),
             consensus_result_cache_(std::move(consensus_result_cache)),
-            log_(logger::log("YacGate")) {
+            log_(logger::log("YacGate")),
+            current_hash_() {
         block_creator_->onBlock().subscribe(
             [this](const auto &event) { this->vote(event); });
       }
 
       void YacGateImpl::vote(const simulator::BlockCreatorEvent &event) {
+        if (not(current_hash_.vote_round < event.round)) {
+          log_->info(
+              "Current round [{}, {}] is greater or equal than vote round [{}, "
+              "{}], skipped",
+              current_hash_.vote_round.block_round,
+              current_hash_.vote_round.reject_round,
+              event.round.block_round,
+              event.round.reject_round);
+          return;
+        }
+
         current_hash_ = hash_provider_->makeHash(event);
 
         if (not event.round_data) {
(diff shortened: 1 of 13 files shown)


irohad/ordering/impl/on_demand_common.cpp
```diff
@@ -21,5 +21,13 @@ namespace iroha {
     const consensus::RejectRoundType kNextCommitRoundConsumer =
         kFirstRejectRound;
 
+    consensus::Round nextCommitRound(consensus::Round round) {
```
muratovv: Maybe rework it with const ref?

irohad/ordering/impl/on_demand_ordering_gate.hpp
```diff
@@ -43,7 +43,9 @@ namespace iroha {
       /**
        * Represents no storage modification. Reject round increment
        */
-      struct EmptyEvent {};
+      struct EmptyEvent {
+        consensus::Round round;
```
muratovv: Add meaning of field here.

test/module/irohad/consensus/yac/yac_gate_test.cpp
```diff
@@ -253,3 +247,104 @@ TEST_F(YacGateTest, DifferentCommit) {
 
   ASSERT_TRUE(gate_wrapper.validate());
 }
+
+/**
+ * @given yac gate with current round initialized
+ * @when vote for older round is called
+ * @then vote is ignored
+ */
+TEST_F(YacGateTest, OlderVote) {
```
muratovv: Seems that this and further tests have the same structure. I think better to move initialization logic to a new fixture based on `YacGateTest`.

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -33,12 +33,24 @@ namespace iroha {
...
             consensus_result_cache_(std::move(consensus_result_cache)),
-            log_(logger::log("YacGate")) {
+            log_(logger::log("YacGate")),
+            current_hash_() {
         block_creator_->onBlock().subscribe(
             [this](const auto &event) { this->vote(event); });
       }
 
       void YacGateImpl::vote(const simulator::BlockCreatorEvent &event) {
+        if (not(current_hash_.vote_round < event.round)) {
```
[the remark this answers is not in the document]
lebdron: Found a solution with `boost::less_than_comparable`

PR #1933, opened 2018-12-11
Commands Mocks Factory

Akvinikym:
### Description of the Change

This PR introduces mocks for commands and a factory, which creates those mocks, so that it will be possible to use them in places, where the actual commands are not needed, such as both ExecutorTest-s.

### Benefits

Less dependencies in future.

### Possible Drawbacks 

None.

### DIFF
--- a/test/module/shared_model/mock_objects_factories/mock_command_factory.cpp
+++ b/test/module/shared_model/mock_objects_factories/mock_command_factory.cpp
@@ -6,6 +6,7 @@
 #include "module/shared_model/mock_objects_factories/mock_command_factory.hpp"
 
 using ::testing::Return;
+using ::testing::ReturnRef;
 using ::testing::ReturnRefOfCopy;
 
 namespace shared_model {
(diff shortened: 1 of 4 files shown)


test/module/shared_model/mock_objects_factories/mock_command_factory.cpp
```diff
@@ -0,0 +1,277 @@
...
+        const types::AccountIdType &account_id,
+        permissions::Grantable permission) const {
+      return createFactoryResult<MockGrantPermission>(
+          [&account_id,
+           permission](FactoryResult<MockGrantPermission> specific_cmd_mock) {
+            EXPECT_CALL(*specific_cmd_mock, accountId())
+                .WillRepeatedly(ReturnRefOfCopy(account_id));
+            EXPECT_CALL(*specific_cmd_mock, permissionName())
+                .WillRepeatedly(Return(permission));
+            return specific_cmd_mock;
```
MBoldyrev: Maybe we should provide a meaningful return for the `toString()` method too?

PR #1934, opened 2018-12-11
Remove Reference Holder from Command and Rework CommandExecutorTest

Akvinikym:
### Description of the Change

Removing the reference holder from Command class breaks Command and Query Executors Tests, as it is no longer possible to copy Command objects. To make this work, a Command mock factory was created in previous PR, and this one reworks CommandExecutorTest with that factory. QueryExecutorTest will be covered in next PR.

### Benefits

No more real Command where it is not needed.

### Possible Drawbacks 

None.

### DIFF
--- a/shared_model/backend/protobuf/commands/impl/proto_command.cpp
+++ b/shared_model/backend/protobuf/commands/impl/proto_command.cpp
@@ -21,7 +21,7 @@
 #include "backend/protobuf/commands/proto_set_quorum.hpp"
 #include "backend/protobuf/commands/proto_subtract_asset_quantity.hpp"
 #include "backend/protobuf/commands/proto_transfer_asset.hpp"
-#include "utils/reference_holder.hpp"
+#include "logger/logger.hpp"
 #include "utils/variant_deserializer.hpp"
 
 namespace {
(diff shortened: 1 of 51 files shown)


shared_model/backend/protobuf/commands/impl/proto_command.cpp
```diff
@@ -86,7 +83,12 @@ namespace shared_model {
     }
 
     Command *Command::clone() const {
```
nickaleks: Please add TODO to rework hierarchy, so that clone() can be removed.
Also dependency on logger in the model is not good, so we should remove clone ASAP

test/module/irohad/ametsuchi/postgres_executor_test.cpp
```diff
@@ -93,14 +80,13 @@ namespace iroha {
...
-        ASSERT_TRUE(
-            val(execute(buildCommand(TestTransactionBuilder().createRole(
-                            role_id, permissions)),
-                        true)));
-        ASSERT_TRUE(
-            val(execute(buildCommand(TestTransactionBuilder().appendRole(
-                            account_id, role_id)),
-                        true)));
+
+        ASSERT_TRUE(val(execute(
```
nickaleks: This construct repeats a lot, please move it to separate function,
so it is possible to write:
```suggestion
        ASSERT_TRUE(execute(mock_command_factory->contructWhatever(...)));
```

test/module/irohad/ametsuchi/postgres_executor_test.cpp
```diff
@@ -139,45 +147,35 @@ namespace iroha {
...
-          "57896044618658097711785492504343953926634992332820282019728792003956"
-          "5648"
-          "19966.0";  // 2**255
-      const std::string asset_amount_one_zero = "1.0";
+      const shared_model::interface::Amount uint256_halfmax =
+          shared_model::interface::Amount{
+              "5789604461865809771178549250434395392663499233282028201972879200"
+              "3956"
+              "564819966.0"};  // 2**255
+      const shared_model::interface::Amount asset_amount_one_zero =
```
nickaleks:
```suggestion
      const shared_model::interface::Amount asset_amount_one_zero{"1.0"};
```

PR #1935, opened 2018-12-11
Trunk/block oneof

kamilsa:
### Description of the Change

This PR introduces block versioning. It consists of already approved PRs
### Benefits

Allows to manage new block versions in future

### Possible Drawbacks 

None


irohad/main/impl/raw_block_loader.cpp
```diff
@@ -35,7 +35,7 @@ namespace iroha {
         const std::string &data) {
       return jsonToProto<iroha::protocol::Block>(data) | [](auto &&block) {
         return boost::optional<std::shared_ptr<Block>>(
-            std::make_shared<shared_model::proto::Block>(std::move(block)));
+            std::make_shared<shared_model::proto::Block>(std::move(block.block_v1())));
```
igor-egorov: Should we prepare the method somehow for parsing future block versions?

Probably we need to place some stub with logging for other and unsupported versions.

irohad/model/converters/impl/pb_block_factory.cpp
```diff
@@ -15,8 +15,10 @@
  * limitations under the License.
  */
 
-#include <boost/assert.hpp>
 #include "model/converters/pb_block_factory.hpp"
+
+#include <boost/assert.hpp>
+
```
igor-egorov: Minor
The blank line is not needed. Only stdlib headers should be separated like that.

irohad/model/sha3_hash.cpp
```diff
@@ -33,7 +33,7 @@ namespace iroha {
 
   hash256_t hash(const model::Block &block) {
     auto &&pb_dat = block_factory.serialize(block);
-    return hash(pb_dat);
+    return hash(pb_dat.block_v1());
```
igor-egorov: Probably we have to expect here appearing of some other block versions rather than v1.
Should we print something if some other version of a block was passed here?

igor-egorov: If some fix is possible and needed here, please do the same fix for all other applicable places.

PR #1939, opened 2018-12-11
Map is substituted by a vector

Akvinikym:
[IR-110](https://jira.hyperledger.org/browse/IR-110)

### Description of the Change

Unordered set of transactions during creation of block errors is not appropriate, as the order can vary, for example, on different peers, and then transactions will be mapped to wrong errors. Thus, the unordered collection was substituted by a vector.

### Benefits

Bug is fixed.

### Possible Drawbacks 

None.


irohad/validation/stateful_validator_common.hpp
```diff
@@ -44,18 +44,19 @@ namespace iroha {
...
-    //      IR-1849               which will include the rejected tx hashes
+    /// Hash of transaction with error, which appeared during validation of this
+    /// transaction
+    struct TransactionError {
+      shared_model::crypto::Hash tx_hash;
+      CommandError error;
+    };
+
+    /// Collection of transactions errors
+    using TransactionsErrors = std::vector<TransactionError>;
```
luckychess: Why vector and not just a regular map?

PR #1903, opened 2018-12-03
GetAccountTransactions Pagination

nickaleks:
### Description of the Change

This PR introduces pagination logic to get account transactions query.
You can now specify starting hash and page size of the response.

### Benefits

Pagination

### Possible Drawbacks 

Slower queries due to added complexity

### Alternate Designs *[optional]*

In several places, additional indexes can be introduced in the future to optimize query

PR #1840, opened 2018-11-08
Consensus fuzzing

luckychess:
### Description of the Change
Fuzzing for consensus (**SendState** endpoint).

### Benefits
One more fuzzing-covered endpoint.

### Possible Drawbacks 
Need CI integration.

### Usage Examples or Tests 
```
cmake -DCMAKE_C_COMPILER=/usr/local/opt/llvm/bin/clang -DCMAKE_CXX_COMPILER=/usr/local/opt/llvm/bin/clang++  -DFUZZING=ON ..
make consensus_fuzz
```

PR #1941, opened 2018-12-12
Hyperledger Iroha v1.0 Release Candidate 1

kamilsa:
We are happy to introduce the first release candidate for Hyperledger Iroha!
There were several major features implemented for the first production version of Iroha:

# Multi-signature Transaction (MST) is ready for use!

Now you can request several signatures to verify the transaction. And transactions can receive 2 more statuses: MST_PENDING for transactions waiting for more signatures and ENOUGH_SIGNATURES_COLLECTED for transactions that have already received the number of signatures requested ( #1649).

We have made MST possible with the following changes: 
- adding Storage for Pending Transactions: #1598 
- fixed sending batches to the Storage and retrieving them:  #1661 
- fixed retrieval of pending MST transactions: #1661 
- MST, reworked with batches: #1642 and client tests for MST: #1652  and other improvements.

Here is the example of MST use case: Alice and Bob are performing exchange between bitcoin and dogecoin using MST to perform exchange only when both participants agreed on the transfer #1761

**Important note:** MST is functional. However, there is a known issue: when user sends a completed MST transaction it's version without all signatures is not removed from MST storage, so when one asks for his pending transactions he will still see it. As a temporary workaround, you can remove all signatures after getting pending transactions and then sign them with your signature and send to Iroha.

# New fully BFT On-Demand Ordering Service is here!
Impressive preparations were made for the assembly of fully BFT (Byzantine-fault tolerant) Ordering service:

- BFT OS algorithm and interfaces: #1670
- on-demand ordering gate: #1675
- the transport: #1635 and on-demand connection manager: #1645
- the synchroniser: #1647
- BFT OS components: #1703
- Consensus Round as Key in YacStorages: #1729
- BFT YacGate: #1731
- BFT YaC: #1831 

And, finally, we merged the new BFT Ordering Service: #1923

# API changes:
There were some serious API changes, so please make sure that your code is compatible with:
- Iroha versions in the block: #1893
- GetAccountTransactions query pagination: #1903
- Renamed client services: #1881
- Storing rejected transactions' hashes in the block: #1815 

# Replaced builders with factories
Now the code runs faster with changed design of its components. Here are our new factories within the code:

- factory for a block: #1622
- factory for transaction status creation: #1717
- factory for query responses and tests: #1724
- abstract factory for creating batches from transactions: #1743
- transport factory: #1750

# Error Responses
Real error reasons for real people using Iroha - introducing Query Error Responses so you could know why the query failed: #1770
Stateful Error Codes in Postgres Command Executor, Endpoint and to SFV: #1813, #1817, #1821
Stateful Response Error Codes: #1837

# Replay prevention:
In different Iroha components: #1857, #1868, #1865, #1868
Feature of storing rejected transactions' hashes in blockchain: #1815
Rejected txs index: #1846

# Preparations for backward compatibility:
Added Iroha version into the block: #1893. Then we updated proto block validator for the new structure of the block: #1905.

# And some of the smaller though very important ones…
## Optimised command and query SQL
A complex work was done to optimise the work of SQL for both commands and queries: #1605
## Consensus Cache in Block Loader and YAC Gate
We constantly move towards faster Iroha. 
Now there is no need for a node to wait time to ask another node for synchronisation thanks to the consensus cache. So let us say good-bye to load_delay: #1627
## SWIG-free Iroha library for Python
Platform independent. Lighter. Better solution if you want to use Python with Iroha: #1761
## SubtractAssetQuantity command
Was not previously implemented in Iroha-cli but now works correctly with AddAssetQuantity: #1713
## Domain asset permissions
Now it is possible to restrict ability to add or subtract assets within domain: #1924 
## Benchmarks to measure performance
We are introducing a benchmark that executes GetAccount query and a benchmark that executes AddAssetQuantity command to have more information on the performance: #1651  and #1648
## Empty Roles
The purpose of append role command is to promote an account to some created role in the system, where a role is a set of permissions account has to perform an action (command or query). It was impossible to add a role without permissions to account. Now it is possible: #1654
## reducedHash
reducedHash wrapper is now visible to libraries and can be called by clients: #1659
## TransactionBatchParser
Splits a list of transactions by possible adjacent batches: #1750
## Genesis Parser
It will generate a nice md file with tables of accounts, permissions and other parameters of the genesis block: #1775
## Query pagination: 
Now you can get transaction query results using pagination: #1903, #1927 
Introduce queries pagination by @nickaleks · Pull Request #1927 ·...

## Improved tests
Here are some of the most influential results:

- the ability to send transaction sequences to the ITF and check stateless valid/invalid status: #1660 ITF feature of checking all the specified statuses: #1720
- Fuzzing targets for ordering service endpoints, consensus, MST and for BlockLoader endpoints: #1829, #1839, #1845, #1840

## Fixes
Please also make sure to check out some important fixes and see if your Iroha is not affected by them:

- freeConnections method, which disposes all currently used connections was fixed: #1727
- send tx sequence method: #1685
- Consensus VoteOther case: #1834
- ITF bugs related to BFT OS: #1826
- tx-example.py and blocks-query.py: #1844
- verbosit flag is now working: #1942 

## DevOps news

- CMake version Minimum CMake version requirement is now 3.11.2 to use cool new features of it in Iroha: #1667
- Iroha on Kubernetes cluster 
- Now Iroha can be run on Kubernetes: #1697

Please provide us with your feedback so we could have a great stable version for Hyperledger Iroha 1.0
Contact us via mailing list: iroha@lists.hyperledger.org
Or in chat: https://chat.hyperledger.org/channel/iroha (for general discussions), https://chat.hyperledger.org/channel/iroha-contributors (for contributors).


neewy:
I would put API changes as the second item in the list.

kamilsa:
> I would put API changes as the second item in the list.

I've put it a bit more higher in the notes

PR #1943, opened 2018-12-12
Correct .proto Files Search

Akvinikym:
### Description of the Change

Sometimes, the build of Iroha could be interrupted, because paths, specified in cmake functions were incorrect, and thus it could not find the .proto files. This PR fixes it.

### Benefits

More frequent successful builds.

### Possible Drawbacks 

None.

### DIFF
--- a/cmake/functions.cmake
+++ b/cmake/functions.cmake
@@ -73,9 +73,9 @@ function(compile_proto_to_cpp PROTO)
   add_custom_command(
       OUTPUT ${SCHEMA_OUT_DIR}/${GEN_PB_HEADER} ${SCHEMA_OUT_DIR}/${GEN_PB}
       COMMAND ${GEN_COMMAND}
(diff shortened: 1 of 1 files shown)


cmake/functions.cmake
```diff
@@ -73,9 +73,9 @@ function(compile_proto_to_cpp PROTO)
   add_custom_command(
       OUTPUT ${SCHEMA_OUT_DIR}/${GEN_PB_HEADER} ${SCHEMA_OUT_DIR}/${GEN_PB}
       COMMAND ${GEN_COMMAND}
-      ARGS -I${GEN_ARGS} -I${CMAKE_CURRENT_SOURCE_DIR} ${ARGN} --cpp_out=${SCHEMA_OUT_DIR} ${PROTO}
+      ARGS -I${GEN_ARGS} -I. ${ARGN} --cpp_out=${SCHEMA_OUT_DIR} ${PROTO}
```
l4l: I'm not sure it's cross-platform solution

PR #1949, opened 2018-12-13
Commands Mocks Factory

Akvinikym:
### Description of the Change

This trunk includes PR-s, which are about removing ReferenceHolder from the Command, introducing the Command mocks factory and reworking the related tests.

### Benefits

Less dependencies, no unneccessary ReferenceHolder.

### Possible Drawbacks 

Maybe a little more complicated way to write Executors tests.


test/module/irohad/ametsuchi/postgres_executor_test.cpp
```diff
@@ -64,43 +64,51 @@ namespace iroha {
...
-        executor->doValidation(not do_validation);
-        executor->setCreatorAccountId(creator);
-        return boost::apply_visitor(*executor, command->get());
+      /**
+       * Check that passed result contains value and not an error (without this
+       * method the one below is not able to perform this check
+       * @param result to be checked
+       */
+      void checkCommandResult(const CommandResult &result) {
+        ASSERT_TRUE(val(result));
```
nickaleks: Now it's impossible to understand which assert has failed

nickaleks: Maybe move ASSERT_TRUE to top level, and execute to return val(result)

Akvinikym: We cannot return val(result), because it is not going to work for error cases, where we cannot unwrap result to get value

PR #1959, opened 2018-12-17
Fix MST pending storage

Akvinikym:
### Description of the Change

Previously, if batch was pending, and the same batch with **all** signatures comes, it went straight to the PCS. Because of that, the pending one was not removed and in theory could be replayed. Now, the system checks, if the arrived batch does already exist in MST storage, even if it has all signatures.

### Benefits

No more replays possible, and pending storage gets cleaned.

### Possible Drawbacks 

Overhead for unordered_set's `find`.

### DIFF
--- a/test/module/irohad/multi_sig_transactions/state_test.cpp
+++ b/test/module/irohad/multi_sig_transactions/state_test.cpp
@@ -53,23 +53,31 @@ TEST(StateTest, UpdateExistingState) {
 }
 
 /**
- * @given empty state @and two batches
- * @when inserting first batch and not the second
- * @then "contains" method shows presence of the first batch @and absence of the
- * second one
+ * @given empty state @and a batch
+ * @when inserting the batch
+ * @then "contains" method shows presence of the batch
  */
-TEST(StateTest, ContainsMethodWorkCorrectly) {
+TEST(StateTest, ContainsMethodFindsInsertedBatch) {
   auto state = MstState::empty();
 
   auto first_signature = makeSignature("1", "pub_key_1");
-  auto first_batch = makeTestBatch(txBuilder(1, iroha::time::now()));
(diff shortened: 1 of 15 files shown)


test/module/irohad/multi_sig_transactions/state_test.cpp
```diff
@@ -52,12 +52,32 @@ TEST(StateTest, UpdateExistingState) {
...
+  auto state = MstState::empty();
+
+  auto first_signature = makeSignature("1", "pub_key_1");
+  auto first_batch = makeTestBatch(txBuilder(1, iroha::time::now()));
+  auto tx = addSignatures(first_batch, 0, first_signature);
+  state += tx;
+
+  auto second_batch = makeTestBatch(txBuilder(1, iroha::time::now()));
+
+  ASSERT_TRUE(state.contains(first_batch));
```
nickaleks: Replace asserts with expects please

test/module/irohad/multi_sig_transactions/storage_test.cpp
```diff
@@ -98,3 +98,17 @@ TEST_F(StorageTest, StorageWhenCreate) {
...
+ * @given storage with three batches @and one another batch
+ * @when checking, if one of the three batches @and another one belongs to the
+ * storage
+ * @then one of the batches belongs to the state, while the last one does not
+ */
+TEST_F(StorageTest, StorageFindsBatch) {
+  auto batch_in_state = makeTestBatch(txBuilder(1, creation_time));
+  auto distinct_batch = makeTestBatch(txBuilder(4, creation_time));
+
+  ASSERT_TRUE(storage->batchInStorage(batch_in_state));
```
nickaleks: sab

nickaleks:
Also your test doesn't pass

PR #1966, opened 2018-12-18
Make CreateAccount cmd validation stronger

igor-egorov:

### Description of the Change

Validation of CreateAccount command will respect a set of permissions of transaction creator.
It would not be possible to create an account in a domain which default role contain some permissions that are missing in transaction creator's permissions set.

In short: transaction creator should have all the permissions of default role of a domain where new account will be created.

### Benefits

Permissions escalation is not possible via creating accounts in more priveleged domains.

### Possible Drawbacks

None ?

### Usage Examples or Tests

```create_account_test```

### DIFF
--- a/irohad/ametsuchi/impl/postgres_command_executor.cpp
+++ b/irohad/ametsuchi/impl/postgres_command_executor.cpp
@@ -1286,13 +1286,30 @@ namespace iroha {
           {"createAccount",
            createAccountBase,
            {(boost::format(R"(
-            has_perm AS (%s),)")
+           domain_role_permissions_bits AS (
+                 SELECT COALESCE(bit_or(rhp.permission), '0'::bit(%1%)) AS bits
+                 FROM role_has_permissions AS rhp
+                 WHERE rhp.role_id = (SELECT * FROM get_domain_default_role)),
+           account_permissions AS (
+                 SELECT COALESCE(bit_or(rp.permission), '0'::bit(%1%)) AS perm
+                  FROM role_has_permissions AS rp
+                 JOIN account_has_roles AS ar on ar.role_id = rp.role_id
+                 WHERE ar.account_id = $1
+           ),
+           creator_has_enough_permissions AS (
+                SELECT ap.perm & dpb.bits = dpb.bits
+                FROM account_permissions AS ap, domain_role_permissions_bits AS dpb
+           ),
+           has_perm AS (%2%),
+          )") % bits
              % checkAccountRolePermission(
                    shared_model::interface::permissions::Role::kCreateAccount,
                    "$1"))
                 .str(),
-            R"(AND (SELECT * FROM has_perm))",
-            R"(WHEN NOT (SELECT * FROM has_perm) THEN 2)"}});
+            R"(AND (SELECT * FROM has_perm)
+               AND (SELECT * FROM creator_has_enough_permissions))",
+            R"(WHEN NOT (SELECT * FROM has_perm) THEN 2
+               WHEN NOT (SELECT * FROM creator_has_enough_permissions) THEN 2)"}});
 
       statements.push_back(
           {"createAsset",
--- a/test/integration/acceptance/create_account_test.cpp
+++ b/test/integration/acceptance/create_account_test.cpp
@@ -4,12 +4,16 @@
  */
 
 #include <gtest/gtest.h>
+#include <boost/algorithm/cxx11/any_of.hpp>
 #include "framework/integration_framework/integration_test_framework.hpp"
 #include "integration/acceptance/acceptance_fixture.hpp"
 
 using namespace integration_framework;
 using namespace shared_model;
 using namespace common_constants;
+
+#define check(i) \
+  [](const auto &resp) { ASSERT_EQ(resp->transactions().size(), i); }
 
 class CreateAccount : public AcceptanceFixture {
  public:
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: test/integration/acceptance/create_account_test.cpp [hunk 1]
```diff
@@ -4,13 +4,17 @@
...
 #include <gtest/gtest.h>
+#include <boost/algorithm/cxx11/any_of.hpp>
 #include "framework/integration_framework/integration_test_framework.hpp"
 #include "integration/acceptance/acceptance_fixture.hpp"
 
 using namespace integration_framework;
 using namespace shared_model;
 using namespace common_constants;
 
+#define check(i) \
```
Akvinikym: Please, add a comment, what exactly does this macros check

test/integration/acceptance/create_account_test.cpp
```diff
@@ -156,3 +160,35 @@ TEST_F(CreateAccount, EmptyName) {
...
+      .setInitialState(kAdminKeypair)
+      .sendTxAwait(second_domain_tx, check(1))
+      .sendTxAwait(makeUserWithPerms(), check(1))
+      .sendTx(create_elevated_user)
+      .skipProposal()
+      .checkVerifiedProposal(check(0))
+      .checkBlock([&rejected_hash](const auto &block) {
+        const auto hashes = block->rejected_transactions_hashes();
+        bool target_tx_is_rejected = boost::algorithm::any_of_equal(
+            boost::begin(hashes), boost::end(hashes), rejected_hash);
```
Akvinikym: Don't you expect only one transaction to be here? If so, maybe it's better to `ASSERT`, that hashes contains only one element and then retrieve it, `ASSERT`-ing that it is equal to `rejected_hash`

test/integration/acceptance/create_account_test.cpp
```diff
@@ -156,3 +160,35 @@ TEST_F(CreateAccount, EmptyName) {
...
+
+  IntegrationTestFramework(1)
+      .setInitialState(kAdminKeypair)
+      .sendTxAwait(second_domain_tx, check(1))
+      .sendTxAwait(makeUserWithPerms(), check(1))
+      .sendTx(create_elevated_user)
+      .skipProposal()
+      .checkVerifiedProposal(check(0))
+      .checkBlock([&rejected_hash](const auto &block) {
+        const auto hashes = block->rejected_transactions_hashes();
```
Akvinikym:
```suggestion
        const auto &hashes = block->rejected_transactions_hashes();
```

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1286,13 +1286,30 @@ namespace iroha {
...
            {(boost::format(R"(
-            has_perm AS (%s),)")
+           domain_role_permissions_bits AS (
+                 SELECT COALESCE(bit_or(rhp.permission), '0'::bit(%1%)) AS bits
+                 FROM role_has_permissions AS rhp
+                 WHERE rhp.role_id = (SELECT * FROM get_domain_default_role)),
+           account_permissions AS (
+                 SELECT COALESCE(bit_or(rp.permission), '0'::bit(%1%)) AS perm
+                  FROM role_has_permissions AS rp
+                 JOIN account_has_roles AS ar on ar.role_id = rp.role_id
```
Akvinikym:
```suggestion
                 JOIN account_has_roles AS ar ON ar.role_id = rp.role_id
```

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1286,13 +1286,30 @@ namespace iroha {
...
            createAccountBase,
            {(boost::format(R"(
-            has_perm AS (%s),)")
+           domain_role_permissions_bits AS (
+                 SELECT COALESCE(bit_or(rhp.permission), '0'::bit(%1%)) AS bits
+                 FROM role_has_permissions AS rhp
+                 WHERE rhp.role_id = (SELECT * FROM get_domain_default_role)),
+           account_permissions AS (
+                 SELECT COALESCE(bit_or(rp.permission), '0'::bit(%1%)) AS perm
+                  FROM role_has_permissions AS rp
```
Akvinikym: Extra space?

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1286,13 +1286,30 @@ namespace iroha {
           {"createAccount",
            createAccountBase,
            {(boost::format(R"(
-            has_perm AS (%s),)")
+           domain_role_permissions_bits AS (
+                 SELECT COALESCE(bit_or(rhp.permission), '0'::bit(%1%)) AS bits
+                 FROM role_has_permissions AS rhp
+                 WHERE rhp.role_id = (SELECT * FROM get_domain_default_role)),
```
Akvinikym: Can't this be done with `JOIN` as below? They are more efficient, as far as I know

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1286,13 +1286,30 @@ namespace iroha {
           {"createAccount",
            createAccountBase,
            {(boost::format(R"(
-            has_perm AS (%s),)")
+           domain_role_permissions_bits AS (
+                 SELECT COALESCE(bit_or(rhp.permission), '0'::bit(%1%)) AS bits
+                 FROM role_has_permissions AS rhp
```
Akvinikym: Please, give a little more meaningful names to `rhp` and `rp`

irohad/ametsuchi/impl/postgres_command_executor.cpp
```diff
@@ -1286,13 +1286,30 @@ namespace iroha {
...
+                  FROM role_has_permissions AS rp
+                 JOIN account_has_roles AS ar on ar.role_id = rp.role_id
+                 WHERE ar.account_id = $1
+           ),
+           creator_has_enough_permissions AS (
+                SELECT ap.perm & dpb.bits = dpb.bits
+                FROM account_permissions AS ap, domain_role_permissions_bits AS dpb
+           ),
+           has_perm AS (%2%),
+          )") % bits
```
Akvinikym: Not formatted, I think

test/integration/acceptance/create_account_test.cpp
kamilsa: The code in this lambda can be simplified using gmock's means: 

```suggestion
         ASSERT_THAT(block->rejected_transactions_hashes(), Contains(rejected_hash));
```

irohad/ametsuchi/impl/postgres_command_executor.cpp
igor-egorov: Discussed. Will use rhp in all places

irohad/ametsuchi/impl/postgres_command_executor.cpp
igor-egorov: we are not trying here to perform ```SELECT ... FROM a, b WHERE a.id=b.id```. That is why where statement could not be substituted by join.

irohad/ametsuchi/impl/postgres_command_executor.cpp
igor-egorov: Well... it is

PR #1967, opened 2018-12-18
Refactor round logging; replace remaining toString calls

lebdron:
### Description of the Change
- Log `Round` objects by calling `toString`
- Remove remaining explicit `toString` calls for logging in `irohad`
- Fix round log in on demand ordering init

### Benefits
Simpler code

### Possible Drawbacks 
None?

### DIFF
--- a/irohad/main/impl/on_demand_ordering_init.cpp
+++ b/irohad/main/impl/on_demand_ordering_init.cpp
@@ -132,9 +132,11 @@ namespace iroha {
           // with number of peers
           auto &peer =
               current_peers_[permutation[reject_round % permutation.size()]];
-          log_->debug("For round {}, using OS on peer: {}",
-                      consensus::Round{current_round.block_round, reject_round},
-                      peer->toString());
+          log_->debug(
+              "For {}, using OS on peer: {}",
+              consensus::Round{current_round.block_round + block_round_advance,
+                               reject_round},
+              *peer);
           return peer;
         };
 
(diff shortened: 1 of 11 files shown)


irohad/main/impl/on_demand_ordering_init.cpp
```diff
@@ -132,9 +132,11 @@ namespace iroha {
           // with number of peers
           auto &peer =
               current_peers_[permutation[reject_round % permutation.size()]];
-          log_->debug("For round {}, using OS on peer: {}",
-                      consensus::Round{current_round.block_round, reject_round},
-                      peer->toString());
+          log_->debug(
+              "For {}, using OS on peer: {}",
+              consensus::Round{current_round.block_round + block_round_advance,
```
Akvinikym: I thought, isn't it wasteful to create a struct just to place it into the log? Maybe do a manual output here?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -97,9 +90,8 @@ OnDemandOrderingServiceImpl::onRequestProposal(consensus::Round round) {
   // read lock
   std::shared_lock<std::shared_timed_mutex> guard(lock_);
   auto proposal = proposal_map_.find(round);
-  log_->debug("onRequestProposal, round[{}, {}], {}returning a proposal.",
-              round.block_round,
-              round.reject_round,
+  log_->debug("onRequestProposal, {}, {}returning a proposal.",
```
Akvinikym:
```suggestion
  log_->debug("onRequestProposal, {}, {} returning a proposal.",
```

lebdron: https://github.com/hyperledger/iroha/pull/1871#discussion_r238148963

MBoldyrev: :laughing: maybe we need to place a comment warning that the whitespace is not missing there! To see why, please see the substitution for these braces.

irohad/main/impl/on_demand_ordering_init.cpp
lebdron: I think we can use this approach for simplicity now, and optimize it later if it turns out to be a bottleneck.

PR #1968, opened 2018-12-18
Fix memory leaks on irohad shutdown

luckychess:
### Description of the Change
Pull request fixes couple of memory leaks on irohad shutdown by calling unsubscribe() in destructors of several classes.

### Benefits
Clean output of leak sanitizer, Irohad destructor is called now.

### Possible Drawbacks 
None.


l4l:
How did you test, that the listed subscriptions leaks memory? I'm not really sure, but though that `unsubscribe` is called on such object destruction, fix me if I'm wrong

luckychess:
I built irohad target with leak sanitizer, than run and stop it. It clearly showed leaks without unsubscribe.

PR #1970, opened 2018-12-19
Add libc++ to docker build image

BulatSaif:

### Description of the Change
It is not trivial to install libc++6-dev on ubuntu 16.04,
The another solution was proposed:  Install  clang-7 (it has libc++7-dev in llvm.org )

### Benefits
One more compiler to test and build 

### Possible Drawbacks 
`clang-7` may update some library  that used in `clang-6.0`


l4l:
Why is it added only to x86_64 arch?

BulatSaif:
@l4l, there are no clang-7 armv7l deb packages in llvm repository

PR #1974, opened 2018-12-19
Refactored ITF fake peer initialization

MBoldyrev:
### Description of the Change

I stumbled upon a race condition when the ITF was run and started sending out votes but the Fake Peers were not yet assigned a proper behaviour to react to them. Overcoming that seems impossible without separating genesis block generation from ITF starting to different external interface functions. While doing that I could also rework the interface of Fake Peers addition to the ITF without `future`s, because we now can initialize ITF main peer's key without starting ITF and assigning the genesis block.

### Benefits

Easier and more granular Fake Peer initialization.

### Possible Drawbacks 

Two ITF functions got exposed. One of them `initPipeline` may be not very clear for the users that do not know the guts of the ITF, but I kept it as is, because some existing code relies on it.  

Please see `fake_peer_example_test`.

Providing an optional behaviour when requesting a fake peer, keeping the ugly design with `future`s.

### DIFF
--- a/test/framework/integration_framework/integration_test_framework.hpp
+++ b/test/framework/integration_framework/integration_test_framework.hpp
@@ -107,7 +107,7 @@ namespace integration_framework {
 
     ~IntegrationTestFramework();
 
-    std::future<std::shared_ptr<fake_peer::FakePeer>> addInitialPeer(
+    std::shared_ptr<fake_peer::FakePeer> addInitialPeer(
         const boost::optional<shared_model::crypto::Keypair> &key);
 
     /**
@@ -122,6 +122,13 @@ namespace integration_framework {
      */
     shared_model::proto::Block defaultBlock(
         const shared_model::crypto::Keypair &key) const;
+
+    /// Construct default genesis block using the my_key_ key.
+    shared_model::proto::Block defaultBlock() const;
+
+    /// Set the provided genesis block.
+    IntegrationTestFramework &setGenesisBlock(
+        const shared_model::interface::Block &block);
 
     /**
      * Initialize Iroha instance with default genesis block and provided signing
@@ -369,6 +376,12 @@ namespace integration_framework {
     /// Get the controlled Iroha instance.
     const std::shared_ptr<IrohaInstance> &getIrohaInstance() const;
 
+    /// Set the ITF peer keypair and initialize irohad pipeline.
+    void initPipeline(const shared_model::crypto::Keypair &keypair);
+
+    /// Start the ITF.
+    void subscribeQueuesAndRun();
+
    protected:
     using AsyncCall = iroha::network::AsyncGrpcClient<google::protobuf::Empty>;
 
@@ -406,9 +419,6 @@ namespace integration_framework {
 
     std::shared_ptr<AsyncCall> async_call_;
 
-    void initPipeline(const shared_model::crypto::Keypair &keypair);
-    void subscribeQueuesAndRun();
-
     // config area
 
     /// maximum time of waiting before appearing next proposal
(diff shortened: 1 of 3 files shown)


test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -40,30 +40,26 @@ class FakePeerExampleFixture : public AcceptanceFixture {
...
+    itf_->setGenesisBlock(itf_->defaultBlock()).subscribeQueuesAndRun();
 
     // inside prepareState we can use lambda for such assert, since
     // prepare transactions are not going to fail
     auto block_with_tx = [](auto &block) {
       ASSERT_EQ(block->transactions().size(), 1);
     };
 
+    auto permissions =
+        interface::RolePermissionSet({Role::kReceive, Role::kTransfer});
```
nickaleks:
```suggestion
        interface::RolePermissionSet{Role::kReceive, Role::kTransfer};
```

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -369,6 +376,12 @@ namespace integration_framework {
     /// Get the controlled Iroha instance.
     const std::shared_ptr<IrohaInstance> &getIrohaInstance() const;
 
+    /// Set the ITF peer keypair and initialize irohad pipeline.
+    void initPipeline(const shared_model::crypto::Keypair &keypair);
+
+    /// Start the ITF.
+    void subscribeQueuesAndRun();
```
muratovv: Does it seem that for each ITF test we will invoke this method? I suggest preserving old behavior as it is and to add additional API for fake peer case.

test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -40,30 +40,26 @@ class FakePeerExampleFixture : public AcceptanceFixture {
...
-
-    auto permissions =
-        interface::RolePermissionSet({Role::kReceive, Role::kTransfer});
-
-    // get the constructed fake peers & set honest behaviour
-    for (auto &fake_peer_future : fake_peers_futures) {
-      assert(fake_peer_future.valid() && "fake peer must be ready");
-      FakePeerPtr fake_peer = fake_peer_future.get();
+    // make the fake peers with honest behaviour
+    std::generate_n(std::back_inserter(fake_peers_), num_fake_peers, [this] {
```
muratovv: I've expected that network generation will be a part of the framework. Maybe move it to ITF?

test/framework/integration_framework/integration_test_framework.hpp
MBoldyrev: You could notice that CI fails, but if you look which tests fail, you see that it is only the fake peer example test, which I explained in a comment above. You could also notice that I did not do modifications to all present ITF usage. From that you cold conclude that the old interface still works as it is.

What really changed in the interface is the ways one can start the ITF. And these changes seem inevitable for the reasons stated in the PR description.

test/integration/acceptance/fake_peer_example_test.cpp
MBoldyrev: IMHO this will only complicate the ITF interfce. The code that initializes fake peers is only one loop with a single function to get the peer and another to set its behaviour. Current approach keeps things simple and flexible.

test/integration/acceptance/fake_peer_example_test.cpp
MBoldyrev: Decided to leave as is, because the inner braces make the initializer list easier to grasp.

test/integration/acceptance/fake_peer_example_test.cpp
muratovv: Yes, the point here is not to change the interface but leave it some shared place. Otherwise, developers will write the loop on each test case.

PR #1990, opened 2018-12-25
Fix GetAccountTransactions and GetAccountAssetTransactions behaviour

Akvinikym:
[IR-136](https://jira.hyperledger.org/browse/IR-136)

### Description of the Change

Those queries behaved incorrect in case invalid account_id or asset_id was supplied, leading to a success, but empty result in such a case. This PR adds a "fallback" check, which can deduce it and return a correct stateful error.

### Benefits

Correct behaviour of the queries.

### Possible Drawbacks 

Another DB check, which is though executed only in one specific case and does not touch success or another error ones.

### DIFF
--- a/irohad/ametsuchi/impl/postgres_query_executor.cpp
+++ b/irohad/ametsuchi/impl/postgres_query_executor.cpp
@@ -227,7 +227,7 @@ namespace iroha {
                   query_range, perms...);
             });
       } catch (const std::exception &e) {
-        return logAndReturnErrorResponse(
+        return this->logAndReturnErrorResponse(
             QueryErrorType::kStatefulFailed, e.what(), 1);
       }
     }
@@ -441,7 +441,7 @@ namespace iroha {
               // parameters were wrong
               if (auto query_incorrect =
                       std::forward<QueryChecker>(qry_checker)(q)) {
-                return logAndReturnErrorResponse(
+                return this->logAndReturnErrorResponse(
                     QueryErrorType::kStatefulFailed,
                     query_incorrect.error_message_,
                     query_incorrect.error_code_);
--- a/irohad/ametsuchi/impl/postgres_query_executor.cpp
+++ b/irohad/ametsuchi/impl/postgres_query_executor.cpp
@@ -443,8 +443,8 @@ namespace iroha {
                       std::forward<QueryChecker>(qry_checker)(q)) {
                 return this->logAndReturnErrorResponse(
                     QueryErrorType::kStatefulFailed,
-                    query_incorrect.error_message_,
-                    query_incorrect.error_code_);
+                    query_incorrect.error_message,
+                    query_incorrect.error_code);
               }
             }
 
@@ -599,7 +599,7 @@ namespace iroha {
       };
 
       auto check_query = [this](const auto &q) {
-        if (existsInDb<int>("account", "account_id", "quorum", q.accountId())) {
+        if (this->existsInDb<int>("account", "account_id", "quorum", q.accountId())) {
           return QueryFallbackCheckResult{};
         }
         return QueryFallbackCheckResult{
@@ -709,12 +709,12 @@ namespace iroha {
       };
 
       auto check_query = [this](const auto &q) {
-        if (not existsInDb<int>(
+        if (not this->existsInDb<int>(
                 "account", "account_id", "quorum", q.accountId())) {
           return QueryFallbackCheckResult{
               5, "no account with such id found: " + q.accountId()};
         }
-        if (not existsInDb<int>(
+        if (not this->existsInDb<int>(
                 "asset", "asset_id", "precision", q.assetId())) {
           return QueryFallbackCheckResult{
               6, "no asset with such id found: " + q.assetId()};
--- a/irohad/ametsuchi/impl/postgres_query_executor.cpp
+++ b/irohad/ametsuchi/impl/postgres_query_executor.cpp
@@ -340,9 +340,13 @@ namespace iroha {
           error_type, error, error_code, query_hash_);
     }
 
-    template <typename Query, typename QueryApplier, typename... Permissions>
+    template <typename Query,
+              typename QueryChecker,
+              typename QueryApplier,
+              typename... Permissions>
     QueryExecutorResult PostgresQueryExecutorVisitor::executeTransactionsQuery(
         const Query &q,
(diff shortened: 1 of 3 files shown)


irohad/ametsuchi/impl/postgres_query_executor.hpp
```diff
@@ -146,20 +146,60 @@ namespace iroha {
        * Execute query which returns list of transactions
        * uses pagination
        * @param query - query object
+       * @param qry_checker - fallback checker of the query, needed in some
```
lebdron: some unclear cases -> if paging hash is not specified and 0 transactions are returned

irohad/ametsuchi/impl/postgres_query_executor.hpp
```diff
@@ -146,20 +146,60 @@ namespace iroha {
...
+       * @tparam ReturnValueType - type of the value to be returned in the
+       * underlying query
+       * @param table_name - name of the table to be checked
+       * @param key_name - name of the table attritute, against which the search
+       * is performed
+       * @param selected_value_name - name of the value, which is to be returned
+       * from the search (attribute with such name is to exist)
+       * @param actual value of the key attribute
+       * @return true, if entry with such value of the key attribute exists,
+       * false otherwise
```
lebdron: A note should be added that the function throws an exception.

irohad/ametsuchi/impl/postgres_query_executor.hpp
```diff
@@ -146,20 +146,60 @@ namespace iroha {
...
           Permissions... perms);
 
+      /**
+       * Check if entry with such key exists in the database
+       * @tparam ReturnValueType - type of the value to be returned in the
+       * underlying query
+       * @param table_name - name of the table to be checked
+       * @param key_name - name of the table attritute, against which the search
+       * is performed
+       * @param selected_value_name - name of the value, which is to be returned
```
lebdron:
```suggestion
       * @param value_name - name of the value, which is to be returned
```

irohad/ametsuchi/impl/postgres_query_executor.hpp
```diff
@@ -146,20 +146,60 @@ namespace iroha {
...
+       * from the search (attribute with such name is to exist)
+       * @param actual value of the key attribute
```
lebdron:
```suggestion
       * @param value - actual value of the key attribute
```

irohad/ametsuchi/impl/postgres_query_executor.hpp
```diff
@@ -146,20 +146,60 @@ namespace iroha {
...
+        QueryFallbackCheckResult() = default;
+        QueryFallbackCheckResult(size_t error_code, std::string &&error_message)
+            : contains_error_{true},
+              error_code_{error_code},
+              error_message_{std::move(error_message)} {}
+
+        explicit operator bool() const {
+          return contains_error_;
+        }
+        bool contains_error_ = false;
```
lebdron:
```suggestion
        bool contains_error = false;
```

irohad/ametsuchi/impl/postgres_query_executor.hpp
```diff
@@ -146,20 +146,60 @@ namespace iroha {
...
+        size_t error_code_ = 0;
```
lebdron:
```suggestion
        shared_model::interface::ErrorQueryResponse::ErrorCodeType error_code = 0;
```

irohad/ametsuchi/impl/postgres_query_executor.hpp
```diff
@@ -146,20 +146,60 @@ namespace iroha {
...
+        std::string error_message_ = "";
```
lebdron:
```suggestion
        shared_model::interface::ErrorQueryResponse::ErrorMessageType error_message = "";
```
since these fields are public

irohad/ametsuchi/impl/postgres_query_executor.hpp
```diff
@@ -146,20 +146,60 @@ namespace iroha {
...
+       */
+      template <typename ReturnValueType>
+      bool existsInDb(const std::string &table_name,
+                      const std::string &key_name,
+                      const std::string &value_name,
+                      const std::string &value) const;
+
+      struct QueryFallbackCheckResult {
+        QueryFallbackCheckResult() = default;
+        QueryFallbackCheckResult(size_t error_code, std::string &&error_message)
```
lebdron:
```suggestion
        QueryFallbackCheckResult(shared_model::interface::ErrorQueryResponse::ErrorCodeType error_code, shared_model::interface::ErrorQueryResponse::ErrorMessageType &&error_message)
```

irohad/ametsuchi/impl/postgres_query_executor.cpp
```diff
@@ -929,5 +972,21 @@ namespace iroha {
           std::move(response_txs), query_hash_);
     }
 
+    template <typename ReturnValueType>
+    bool PostgresQueryExecutorVisitor::existsInDb(
+        const std::string &table_name,
+        const std::string &key_name,
+        const std::string &value_name,
+        const std::string &value) const {
+      auto cmd = (boost::format(R"(SELECT %s
```
lebdron: Please use data binding with `:value_name`, e.g. `SELECT :value_name` and `soci::use(value_name)` instead of raw format string.

irohad/ametsuchi/impl/postgres_query_executor.cpp
```diff
@@ -580,7 +598,16 @@ namespace iroha {
         };
       };
 
+      auto check_query = [this](const auto &q) {
+        if (existsInDb<int>("account", "account_id", "quorum", q.accountId())) {
+          return QueryFallbackCheckResult{};
+        }
+        return QueryFallbackCheckResult{
+            5, "no account with such id found: " + q.accountId()};
```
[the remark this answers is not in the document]
Akvinikym: Yes, when this task will be done, all magic number will be gone, and it is not a good idea, I think, to mark every such moment with a separate ToDo comment

irohad/ametsuchi/impl/postgres_query_executor.cpp
Akvinikym: As it seems, is is at least not possible to bind the name of the table to be selected from, and using both binding and boost::variant looks like a bad idea, so I decided to pick the boost

PR #1992, opened 2018-12-25
Move external dependencies dir

bakhtin:

### Description of the Change
Reopen #1692. It is not in `dev` somehow


lebdron:
Looks like #1692 is indeed in dev, but this line has to be also changed
https://github.com/hyperledger/iroha/blob/72fd90cc7562c907a7cddbb6b1992714b293dbd7/cmake/dependencies.cmake#L5

PR #1998, opened 2018-12-27
Hex format in client API

kamilsa:
### Description of the Change

This PR makes keys format in client api consistent. Not all keys are in hex format.

### Benefits

Consistent client API

### Possible Drawbacks 

None


test/module/irohad/model/converters/pb_commands_test.cpp
```diff
@@ -74,7 +74,7 @@ TEST(CommandTest, add_peer) {
   auto factory = iroha::model::converters::PbCommandFactory();
 
   auto proto_add_peer = factory.serializeAddPeer(orig_addPeer);
-  auto serial_addPeer = factory.deserializeAddPeer(proto_add_peer);
+  const auto& serial_addPeer = factory.deserializeAddPeer(proto_add_peer);
```
igor-egorov: Why the result of `deserialize` method is assigned to variable named `serial...`. It is opposite - it is `deserialized`.

shared_model/validators/protobuf/proto_transaction_validator.cpp
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+#include "validators/protobuf/proto_transaction_validator.hpp"
+
+#include "validators/validators_common.hpp"
+
+namespace shared_model {
+  namespace validation {
+
+    void validateHexField(const std::string &public_key,
```
MBoldyrev: The function name does not match the message of validation failure reason.
```suggestion
    void validatePublicKeyField(const std::string &public_key,
```

shared_model/validators/validators_common.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include "validators/validators_common.hpp"
+
+#include <regex>
+
+namespace shared_model {
+  namespace validation {
+
+    bool validateHexString(const std::string &str) {
+      static const std::regex hex_regex{R"([0-9a-fA-F]*)"};
```
MBoldyrev: So we do not check the length of the hex string in this function, although it is used directly to validate hashes and public keys. I would suggest to add specific functions that check our hash and public key formats to this file and use this function internally, like
```c++
bool validatePublicKeyString(const std::string &pk) {
    return pk.size() == crypto::kPublicKeyLength && validateHexString(pk);
}

bool validateHashString(const std::string &hash) {
    return hash.size() == crypto::kHashLength && validateHexString(hash);
}
```
and use them to validate both content and size.

MBoldyrev: PS: it seems now the length constants are defined in `shared_model/cryptography/ed25519_sha3_impl/crypto_provider.hpp` and initialized here:
https://github.com/hyperledger/iroha/blob/34bf8704d0dc8649a0a24ee0f75351d50e64094c/shared_model/cryptography/ed25519_sha3_impl/crypto_provider.cpp#L45-L49
So seems they can be accessed through `shared_model::crypto::DefaultCryptoAlgorithmType`.
Just do not forget that a byte in blob is two chars in hex string!

irohad/model/converters/impl/pb_command_factory.cpp
```diff
@@ -206,17 +217,15 @@ namespace iroha {
           const model::AddSignatory &add_signatory) {
         protocol::AddSignatory pb_add_signatory;
         pb_add_signatory.set_account_id(add_signatory.account_id);
-        pb_add_signatory.set_public_key(add_signatory.pubkey.data(),
-                                        add_signatory.pubkey.size());
+        pb_add_signatory.set_public_key(add_signatory.pubkey.to_hexstring());
```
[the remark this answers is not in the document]
kamilsa: https://jira.hyperledger.org/browse/IR-149

shared_model/validators/protobuf/proto_block_validator.cpp
```diff
@@ -18,7 +24,26 @@ namespace shared_model {
           == iroha::protocol::Block::BLOCK_VERSION_NOT_SET) {
         reason.second.emplace_back("Block version is not set");
         answer.addReason(std::move(reason));
-        return answer;
+      }
```
MBoldyrev: Seems we have to return the `answer` here if the block has no `block_version`, instead of examining non-existent data.

shared_model/validators/validators_common.cpp
kamilsa: This is checked in business logic. Particularly field_validator does that. In protobuf checker we only check if protobuf structure is okay. Size of the public key and hashes is checked later

deploy/ansible/roles/iroha-k8s/files/k8s-peer-keys.yaml
```diff
@@ -5,32 +5,32 @@ metadata:
   name: iroha-peer-key-0
 type: Opaque
 data:
-  node.priv: NWE5NTU5YjQ2ZGIwN2UwNzUwNGU5ZTFlOTM0NjFmODM3M2ExNzUyMzEyN2E1NmI4MDhkNjE0OWJmMTk1NDljZg==
-  node.pub: MWNmODlkMDgyOTgxZTRjOGFlNzJhNjI1MmIyMTc5NzgwNDZjMTZmZmY2ZjVmOGMxZmUwNTk0MGVhNTcwMmYwMA==
+  node.priv: YmI2MGE2ODNiZjQxYTc5ZWRhNzQxODhkYmUxYmM1M2RiOGY0ZmNlYTQwMWY2NDI3MTE1N2QwYWFkNzQxODdjYQ==
```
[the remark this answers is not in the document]
bakhtin: These are encrypted secrets of Kubernetes. Not related to Iroha keys

shared_model/validators/validators_common.cpp
MBoldyrev: I do not see a difference between checking the string content and string size: why we check one and dismiss the other?

PR #2003, opened 2019-01-09
Add CHECK_TXS macro to acceptance fixture. Reduce code duplication

igor-egorov:

### Description of the Change

The same macro for checking transactions quantity in response was defined in several places.
Now it is defined in acceptance fixture header.

### Benefits

Reduced code duplication.

### Possible Drawbacks 

Still a macro, but we have agreed to use it for the purpose.

### Usage Examples or Tests

```
add_signatory_test
create_account_test
get_account_test
remove_signatory_test
set_account_quorum_test
transfer_asset_test
```

### DIFF
--- a/test/integration/acceptance/acceptance_fixture.hpp
+++ b/test/integration/acceptance/acceptance_fixture.hpp
@@ -47,7 +47,7 @@ namespace {
 
 #define CHECK_MST_PENDING BASE_CHECK_RESPONSE(MstPendingResponse)
 
-#define CHECK_TXS(i) \
+#define CHECK_TXS_QUANTITY(i) \
   [](const auto &resp) { ASSERT_EQ(resp->transactions().size(), i); }
 }  // namespace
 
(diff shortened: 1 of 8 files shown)


test/integration/acceptance/acceptance_fixture.hpp
```diff
@@ -45,6 +46,9 @@ namespace {
 #define CHECK_COMMITTED BASE_CHECK_RESPONSE(CommittedTxResponse)
 
 #define CHECK_MST_PENDING BASE_CHECK_RESPONSE(MstPendingResponse)
+
+#define CHECK_TXS(i) \
```
Akvinikym: Maybe name it `CHECK_TXS_NUMBER` to be more specific?

igor-egorov: You are right. That was an option but on the other side everyone did it even shorter, just `check`.

muratovv: Some minor issue. Add a doc string for the macro.

PR #2009, opened 2019-01-09
Documentation change: develop branch changed to master

LiraLemur:
Signed-off-by: Sara <lira.lemur@gmail.com>

### Description of the Change

Changed develop branch in documentation (that one is terribly outdated) to master

### Benefits

Newer version will be used and no misunderstanding about develop-dev problem

### Possible Drawbacks 

Some old options (like cli) might not work with master. Also in some places I might have missed some changes. Though if any problems occur, it is easy to fix.

### DIFF
--- a/docs/source/getting_started/index.rst
+++ b/docs/source/getting_started/index.rst
@@ -84,7 +84,7 @@ In order to get those files, you need to clone the
 
 .. code-block:: shell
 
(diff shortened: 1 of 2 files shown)


l4l:
I would consider adding a note about active development branch (aka `dev`), master might be a bit stale after some time after release

PR #2010, opened 2019-01-09
New logger implementation part 1: main classes

MBoldyrev:
[IR-97](https://jira.hyperledger.org/browse/IR-97), [IR-99](https://jira.hyperledger.org/browse/IR-99), [IR-140](https://jira.hyperledger.org/browse/IR-140)

### Description of the Change

Main classes of a new logger implementation that ...

### Benefits

... supports tree config structure and hides spdlog library.

### Usage Examples

Loggers can be created with and without a tree config. Project components should pass children of their loggers to their sub-components. Child loggers get the settings from their parents, optionally overridden if a tree config was used.

Please note that this is the first step towards the working new logger, and the project is not yet using it, so it is not expected to compile successfully.

### DIFF
--- a/libs/logger/logger.cpp
+++ b/libs/logger/logger.cpp
@@ -11,7 +11,6 @@
 #include <spdlog/spdlog.h>
 
 static const std::string kDefaultPattern = R"([%Y-%m-%d %H:%M:%S.%F][%L] %n: %v)";
-static const std::string kTagHierarchySeparator = "/";
 
 namespace logger {
 
@@ -37,81 +36,17 @@ namespace logger {
     return kDefaultPattern;
   }
 
-  LoggerConfigTreeNode::LoggerConfigTreeNode(std::string tag,
-                                             LoggerConfig config)
-      : tag_(std::move(tag)), config_(std::move(config)){};
-
-  LoggerConfigTreeNodePtr LoggerConfigTreeNode::addChild(
-      std::string tag,
-      boost::optional<LogLevel> log_level,
-      boost::optional<LogPatterns> patterns) {
-    LoggerConfig child_config{
-        log_level.value_or(config_.log_level),
-        patterns ? LogPatterns{std::move(*patterns)} : config_.patterns};
-    auto child =
-        std::make_shared<LoggerConfigTreeNode>(tag, std::move(child_config));
-    auto map_elem = std::make_pair<const std::string, LoggerConfigTreeNodePtr>(
-        std::move(tag), std::move(child));
-    return children_.emplace(std::move(map_elem)).first->second;
-  }
-
-  const std::string &LoggerConfigTreeNode::getTag() const {
-    return tag_;
-  }
-
-  const LoggerConfig &LoggerConfigTreeNode::getConfig() const {
-    return config_;
-  }
-
-  boost::optional<LoggerConfigTreeNodePtr> LoggerConfigTreeNode::getChild(
-      const std::string &tag) {
-    const auto it = children_.find(tag);
-    if (it == children_.end()) {
-      return boost::none;
-    }
-    return it->second;
-  }
-
-  boost::optional<ConstLoggerConfigTreeNodePtr> LoggerConfigTreeNode::getChild(
-      const std::string &tag) const {
-    const auto it = children_.find(tag);
-    if (it == children_.end()) {
-      return boost::none;
-    }
-    return std::static_pointer_cast<const LoggerConfigTreeNode>(it->second);
-  }
-
   class Logger::Impl {
    public:
     virtual ~Impl() = default;
 
-    /// Constructor.
-    ///
-    /// @param tag - the tag of constructed logger
-    /// @param tree - the logger tree configuration
-    /// @param spdlog_logger - the spdlog logger to use
-    Impl(std::string tag, ConstLoggerConfigTreeNodePtr tree)
+    /// @param tag - the tag for logging (aka logger name)
+    /// @param config - logger configuration
+    Impl(std::string tag, ConstLoggerConfigPtr config)
         : tag_(std::move(tag)),
-          tree_(std::move(tree)),
+          config_(std::move(config)),
           logger_(spdlog::stdout_color_mt(tag)) {
       setupLogger();
-    }
-
-    LoggerPtr getChild(std::string tag) {
-      std::lock_guard<std::mutex> lock(children_mutex_);
-      const auto child_it = children_.find(tag);
-      if (child_it != children_.end()) {
-        return child_it->second;
-      }
-      // if a node for this child is not found in the tree config, create a
-      // new standalone logger using this logger's settings
-      auto child_config = tree_->getChild(tag).value_or(
-          std::make_shared<LoggerConfigTreeNode>(tag, tree_->getConfig()));
-      Logger new_child(tag_ + kTagHierarchySeparator + tag,
-                       std::move(child_config));
-      return children_
-          .emplace(std::make_pair(tag, std::make_shared<Logger>(new_child)))
-          .first->second;
     }
 
    private:
@@ -123,37 +58,23 @@ namespace logger {
                            {LogLevel::kWarn, spdlog::level::warn},
                            {LogLevel::kError, spdlog::level::err},
                            {LogLevel::kCritical, spdlog::level::critical}};
-      const auto it = kSpdLogLevels.find(tree_->getConfig().log_level);
+      const auto it = kSpdLogLevels.find(config_->log_level);
       BOOST_ASSERT_MSG(it != kSpdLogLevels.end(), "Unknown log level!");
       logger_->set_level(it == kSpdLogLevels.end()
                              ? kSpdLogLevels.at(kDefaultLogLevel)
                              : it->second);
-      logger_->set_pattern(
-          kDefaultLogPatterns.getPattern(tree_->getConfig().log_level));
+      logger_->set_pattern(kDefaultLogPatterns.getPattern(config_->log_level));
     }
 
     // --- Fields accessible by logger::Logger ---
    public:
     const std::string tag_;
-    const ConstLoggerConfigTreeNodePtr tree_;
+    const ConstLoggerConfigPtr config_;
     const std::shared_ptr<spdlog::logger> logger_;
-
-    // --- Internal use fields ---
-   private:
-    std::unordered_map<std::string, LoggerPtr> children_;
-    std::mutex children_mutex_;
   };
 
-  Logger::Logger(ConstLoggerConfigTreeNodePtr tree_config)
-      : impl_(std::make_shared<Impl>(tree_config->getTag(), tree_config)) {}
-
-  Logger::Logger(std::string tag, LoggerConfig config)
-      : impl_(std::make_shared<Impl>(
-            tag,
-            std::make_shared<LoggerConfigTreeNode>(tag, std::move(config)))) {}
-
-  Logger::Logger(std::string tag, ConstLoggerConfigTreeNodePtr tree_config)
-      : impl_(std::make_shared<Impl>(std::move(tag), std::move(tree_config))) {}
+  Logger::Logger(std::string tag, ConstLoggerConfigPtr config)
+      : impl_(std::make_shared<Impl>(tag, std::move(config))) {}
 
   Logger::Logger(const Logger &other) : impl_(other.impl_) {}
 
@@ -197,11 +118,7 @@ namespace logger {
   }
 
   bool Logger::shouldLog(Level level) const {
-    return impl_->tree_->getConfig().log_level <= level;
-  }
-
-  LoggerPtr Logger::getChild(std::string tag) {
-    return impl_->getChild(tag);
+    return impl_->config_->log_level <= level;
   }
 
   std::string boolRepr(bool value) {
--- a/libs/logger/logger.cpp
+++ b/libs/logger/logger.cpp
@@ -206,12 +206,14 @@ namespace logger {
 
   Logger::Logger(Logger &&other) : impl_(std::move(other.impl_)) {}
 
-  Logger &operator=(const Logger &other) {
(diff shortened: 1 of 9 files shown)


libs/logger/logger.hpp
```diff
@@ -18,35 +22,193 @@ auto operator<<(StreamType &os, const T &object)
...
+     template <typename... Args>
+     void critical(const std::string &format, const Args &... args) const {
+       log(LogLevel::kCritical, format, args...);
+     }
+
+     template <typename... Args>
+     void log(Level level,
+              const std::string &format,
+              const Args &... args) const {
+       if (shouldLog(level)) {
```
Warchant: You can get rid of conditional and make logger faster, if you rewrite this with 
```
1& logInternal(...)
```

Warchant:
This logger is synchronous, consider adding async logging for better performance

libs/logger/logger.hpp
l4l: Can be just constexpr and compiler will be able to optimize

libs/logger/logger.cpp
```diff
@@ -5,58 +5,262 @@
...
+    const std::string tag_;
+    const ConstLoggerConfigTreeNodePtr tree_;
+    const std::shared_ptr<spdlog::logger> logger_;
+
+    // --- Internal use fields ---
+   private:
+    std::unordered_map<std::string, std::map<LoggerThreadSafety, LoggerPtr>>
+        children_;
+  };
+
```
muratovv: Pls add a task with async logger

libs/logger/logger.cpp
```diff
@@ -5,58 +5,262 @@
...
+  class LoggerImplSingleThread : public Logger::Impl {
```
muratovv: Seems, that this type of logger is not required now for several reasons:
- now it can be introduced only in the initialization phase, where the profit of absence of locks is not significant
- other code transformations are not assumed where it can be helpful

We discussed those arguments with the author but didn't conclude for the changes here. I request @lebdron for facilitating and help with this issue.

libs/logger/logger.cpp
```diff
@@ -5,58 +5,262 @@
...
+    LoggerImplMultiThread(std::string tag, Types &&... args)
+        : Logger::Impl(
+              tag, std::forward<Types>(args)..., spdlog::stdout_color_mt(tag)) {
     }
-    return logger;
+
+    virtual ~LoggerImplMultiThread() = default;
+
+    LoggerPtr getChild(std::string tag, LoggerThreadSafety ts) {
+      std::lock_guard<std::mutex> lock(children_mutex);
```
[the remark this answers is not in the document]
MBoldyrev: The guard is needed to prevent concurrent access to children map.

libs/logger/logger.cpp
MBoldyrev: thread safety option removed in f82275d

libs/logger/logger.hpp
```diff
@@ -18,35 +22,199 @@ auto operator<<(StreamType &os, const T &object)
...
+    kTrace,
+    kDebug,
+    kInfo,
+    kWarn,
+    kError,
+    kCritical,
+  };
+
+  /// Logger thread safety
+  enum class LoggerThreadSafety {
```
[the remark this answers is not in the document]
MBoldyrev: Please see dbdaff3, in it I put the hierarchy management out of the `Logger` class.

libs/logger/logger.hpp
```diff
@@ -18,35 +22,199 @@ auto operator<<(StreamType &os, const T &object)
...
+    kSingleThread,
+    kMultiThread,
+  };
+
+  class Logger {
```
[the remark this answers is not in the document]
MBoldyrev: Please see dbdaff3, in it I put the hierarchy management out of the Logger class.

libs/logger/logger_manager.cpp
```diff
@@ -0,0 +1,80 @@
...
+  LoggerManagerTreePtr LoggerManagerTree::getChild(const std::string &tag) {
+    std::lock_guard<std::mutex> lock(children_mutex_);
+    const auto child_it = children_.find(tag);
+    if (child_it != children_.end()) {
+      return child_it->second;
+    }
+    // If a node for this child is not found in the tree config, create a
+    // new standalone logger using this logger's settings.
+    LoggerManagerTreePtr new_child(
+        new LoggerManagerTree(joinTags(full_tag_, tag), tag, config_));
```
muratovv: `new` triggered. I think it should be reworked with passing parameters (emplace instantiation) or pass unique_ptr inside instead.

libs/logger/logger_manager.hpp
```diff
@@ -0,0 +1,60 @@
...
+    LoggerManagerTree(ConstLoggerConfigPtr config);
+
+    /**
+     * Register a child configuration. The new child's cnofiguartion parameters
+     * are taken from the parent optionally overrided by the arguments.
+     *
+     * @param tag - the child's tag, without any parents' frefixes
+     * @param log_level - override the log level for the new child
+     * @param patterns - override the patterns
+     */
```
muratovv: Pls, mention thread-safe behavior of the method.

libs/logger/logger_manager.cpp
```diff
@@ -0,0 +1,80 @@
...
+                std::memory_order_acquire)) {
+          return new_logger;
+        }
+      }
+    }
+    return logger;
+  }
+
+  LoggerManagerTreePtr LoggerManagerTree::getChild(const std::string &tag) {
+    std::lock_guard<std::mutex> lock(children_mutex_);
```
muratovv: Why this lock is suitable? Even if the return value of LoggerManagerTreePtr will be consistent there is no guarantee about the safeness of usage of the pointer.

libs/logger/logger_manager.cpp
MBoldyrev: I explained the reason behind using new above:
https://github.com/hyperledger/iroha/blob/4f9913b7a7cea2f751fe40d8721c38e7d0bfb047/libs/logger/logger_manager.cpp#L36-L37
The stuff like `std::make_shared` does not work with private constructors.

I see a workaround with an internal method that creates unique or shared pointer in a derived class, but does it worth the problem? New result is obviously immediately passed to a shared pointer here - I think it is safe enough.

l4l:
Fix the build please

MBoldyrev:
> Fix the build please

I want to take the name of existing entity, `logger::Logger`, so unless I use it throughout the project or do some tricks with including and linking, or with renaming, it will not compile. And those tricks will have to be undone.
This change is for a trunk, so it will not break anything. And if you want to verify that it builds, you can build the `logger` lib, which does so.

libs/logger/logger_spdlog.hpp
```diff
@@ -0,0 +1,123 @@
...
+    std::map<LogLevel, std::string> patterns_;
+  };
+
+  // TODO mboldyrev 29.12.2018 IR-188 Add sink options (console, file, syslog)
+  struct LoggerConfig {
+    LogLevel log_level;
+    LogPatterns patterns;
+  };
+
+  class LoggerSpdlog : public Logger {
```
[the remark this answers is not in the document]
lebdron: I am pretty sure that `shared_ptr<Logger>` will be passed to components. In this case we can define `class MockLogger : public Logger {};` and use it in tests.

libs/logger/logger_spdlog.cpp
```diff
@@ -0,0 +1,92 @@
...
+    p.setPattern(LogLevel::kInfo, kDefaultPattern);
+    return p;
+  })();
+
+  void LogPatterns::setPattern(LogLevel level, std::string pattern) {
+    patterns_[level] = pattern;
+  }
+
+  std::string LogPatterns::getPattern(LogLevel level) const {
+    for (auto it = patterns_.rbegin(); it != patterns_.rend(); ++it) {
```
lebdron: Looks like a candidate for `std::find_if`.

libs/logger/logger_manager.hpp
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * A node of logger managers tree. It stores the configuration needed to
+   * create a logger corresponding to this node and its children. Thread safe.
+   */
+  class LoggerManagerTree {
+   public:
+    LoggerManagerTree(ConstLoggerConfigPtr config);
+
+    /**
+     * Register a child configuration. The new child's cnofiguartion parameters
```
lebdron:
```suggestion
     * Register a child configuration. The new child's configuration parameters
```

libs/logger/logger.hpp
```diff
@@ -6,10 +6,13 @@
 #ifndef IROHA_SPDLOG_LOGGER_LOGGER_HPP
```
lebdron:
```suggestion
 #ifndef IROHA_LOGGER_LOGGER_HPP
```

libs/logger/logger_manager.hpp
```diff
@@ -0,0 +1,65 @@
...
+     * are taken from the parent optionally overrided by the arguments. Thread
+     * safe.
+     *
+     * @param tag - the child's tag, without any parents' frefixes
```
lebdron:
```suggestion
     * @param tag - the child's tag, without any parents' prefixes
```

libs/logger/logger_manager.hpp
```diff
@@ -0,0 +1,65 @@
...
+     * @param patterns - override the patterns
+     */
+    LoggerManagerTreePtr registerChild(std::string tag,
+                                       boost::optional<LogLevel> log_level,
+                                       boost::optional<LogPatterns> patterns);
+
+    /// Get this node's logger. Thread safe.
+    LoggerPtr getLogger();
+
+    /// Get non-const child config by tag, if present. Thread safe.
```
lebdron:
```suggestion
    /// Get non-const child node by tag, if present. Thread safe.
```

libs/logger/logger_manager.cpp
```diff
@@ -0,0 +1,82 @@
...
+        full_tag_(std::move(full_tag)),
+        config_(std::move(config)){};
+
+  LoggerManagerTreePtr LoggerManagerTree::registerChild(
+      std::string tag,
+      boost::optional<LogLevel> log_level,
+      boost::optional<LogPatterns> patterns) {
+    LoggerConfig child_config{
+        log_level.value_or(config_->log_level),
+        patterns ? LogPatterns{std::move(*patterns)} : config_->patterns};
```
lebdron:
```suggestion
        patterns ? LogPatterns{*std::move(patterns)} : config_->patterns};
```

libs/logger/logger_spdlog.hpp
```diff
@@ -0,0 +1,123 @@
...
+#ifndef IROHA_LOGGER_SPDLOG_HPP
+#define IROHA_LOGGER_SPDLOG_HPP
+
+#include "logger/logger.hpp"
+
+#include <map>
+#include <memory>
+#include <string>
+
+#include <boost/optional.hpp>
```
lebdron: Looks like it can be removed too.

libs/logger/logger_spdlog.cpp
```diff
@@ -0,0 +1,92 @@
...
+    BOOST_ASSERT_MSG(it != kSpdLogLevels.end(), "Unknown log level!");
+    logger_->set_level(it == kSpdLogLevels.end()
+                           ? kSpdLogLevels.at(kDefaultLogLevel)
+                           : it->second);
+    logger_->set_pattern(config_->patterns.getPattern(config_->log_level));
+  }
+
+  void LoggerSpdlog::logInternal(Level level, const std::string &s) const {
+    switch (level) {
+      case LogLevel::kTrace:
```
lebdron: If there is an spdlog method which accepts `spdlog::level` as first argument, please use it instead of switch-case.

test/module/irohad/logger/logger_test.cpp
```diff
@@ -8,17 +8,14 @@
...
-  another_logger->info("temporal output {}, {}", 123, "string param");
-  another_logger->info(logger::red("color output"));
-  another_logger->info(
-      logger::yellow("color args output {} // note: require char *").c_str(),
-      "=^._.^=");
+TEST(LoggerTest, basicStandaloneLoggerTest) {
+  logger::LoggerConfig config;
+  config.log_level = logger::LogLevel::kInfo;
+  auto one_logger = logger::Logger(
+      "standalone_logger", config, logger::LoggerThreadSafety::kSingleThread);
```
lebdron: Does this compile? There is no mention of `LoggerThreadSafety`.

libs/logger/logger_spdlog.hpp
```diff
@@ -0,0 +1,123 @@
...
+     * @param config - logger configuration
+     */
+    LoggerSpdlog(std::string tag, ConstLoggerConfigPtr config);
+
+    virtual ~LoggerSpdlog();
+
+    // --- Logging functions ---
+
+    template <typename... Args>
+    void trace(const std::string &format, const Args &... args) const {
```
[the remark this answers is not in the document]
MBoldyrev: Sure, that was a careless copy-paste from the base class...

libs/logger/logger_spdlog.cpp
MBoldyrev: IMHO it complicates the code:
```c++
    auto it = std::find_if(patterns_.rbegin(),
                           patterns_.rend(),
                           [level](const auto &level_and_pattern) {
                             return level_and_pattern.first <= level;
                           });
    return it == patterns_.rend() ? kDefaultPattern : it->second;
```
Do you like this variant better?

libs/logger/logger_spdlog.cpp
MBoldyrev: Great! I overlooked that first time somehow..

libs/logger/logger_spdlog.cpp
lebdron: No 🙈

PR #2013, opened 2019-01-10
Irohad field reorder

luckychess:
### Description of the Change
Field declarations changed according their initialisation order and a bit of code cleaning.

### Benefits
A bit more consistent order of destructor calls.

### Possible Drawbacks 
None.

### DIFF
--- a/irohad/main/application.cpp
+++ b/irohad/main/application.cpp
@@ -387,8 +387,8 @@ void Irohad::initMstProcessor() {
     mst_propagation = std::make_shared<GossipPropagationStrategy>(
         storage, rxcpp::observe_on_new_thread(), *opt_mst_gossip_params_);
   } else {
-    mst_propagation = std::make_shared<iroha::PropagationStrategyStub>();
     mst_transport = std::make_shared<iroha::network::MstTransportStub>();
+    mst_propagation = std::make_shared<iroha::PropagationStrategyStub>();
   }
 
   auto mst_time = std::make_shared<MstTimeProviderImpl>();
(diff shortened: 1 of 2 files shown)


irohad/main/application.cpp
```diff
@@ -514,9 +514,3 @@ Irohad::RunResult Irohad::run() {
             return e;
           });
 }
-
-Irohad::~Irohad() {
```
Akvinikym: Why is `freeConnections()` function call no longer needed?

luckychess: It's called in Storage destructor, no need to do it manually from here.

irohad/main/application.hpp
```diff
@@ -166,6 +166,24 @@ class Irohad {
...
+ protected:
+  logger::Logger log_;
+
+  // initialization objects
+  iroha::network::OnDemandOrderingInit ordering_init;
+  iroha::consensus::yac::YacInit yac_init;
+  iroha::network::BlockLoaderInit loader_init;
+
+  // common objects factory
+  std::shared_ptr<shared_model::interface::CommonObjectsFactory>
```
lebdron: Technically `common_objects_factory_` should be above `std::shared_ptr<iroha::ametsuchi::Storage> storage`
https://github.com/hyperledger/iroha/blob/d617982653b0cf8b03e349f937fafac37bce37e2/irohad/main/application.cpp#L122-L136

luckychess: I agree but I don't really want to do a protected-public-protected mix of fields. I think it could be done later.

PR #2016, opened 2019-01-11
Move some WsvQuer-ies to test namespace

Akvinikym:
### Description of the Change

WsvQuery class contains a lot of functions, which are either not used anywhere but in tests or are only tested and not used at all. To be precise, only two queries are actually needed in the main code, so all other ones are moved to specially created test class.

### Benefits

Less code.

### Possible Drawbacks 

Functions `execute` and `fromResult` are repeated in both `WsvQuery` and `SqlQuery` classes, as I found no beautiful way to avoid it. Suggestions are accepted.


test/framework/sql_query.cpp
```diff
@@ -0,0 +1,297 @@
...
+ */
+
+#include "framework/sql_query.hpp"
+
+#include <soci/boost-tuple.h>
+#include "ametsuchi/impl/soci_utils.hpp"
+#include "backend/protobuf/permissions.hpp"
+#include "common/bind.hpp"
+#include "common/result.hpp"
+#include "interfaces/common_objects/common_objects_factory.hpp"
```
lebdron: Already included in "framework/sql_query.hpp", so it is better to forward declare it there.

PR #2024, opened 2019-01-15
Feature/iroha tests coredump

Stayer:
### Description of the Change

This PR allows to archive and upload all core dumps to nexus server
All files will be stored at [/artifacts/iroha/coredumps/](https://nexus.iroha.tech/#browse/browse:artifacts:iroha%2Fcoredumps)
[Example file](https://nexus.iroha.tech/repository/artifacts/iroha/coredumps/coredumps-959fb10e.tar)
Task: [DOPS-198](https://soramitsu.atlassian.net/browse/DOPS-198)
### Benefits

Easy debug for failed tests

### Possible Drawbacks 

`ulimit -c unlimited` settings doesn't restrict size of core dump. In theory, it might be too large

### DIFF
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -25,7 +25,7 @@ def doDebugBuild(coverageEnabled=false) {
 
   // for saving coredumps
   sh "sudo echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
-  sh "sudo ulimit -c unlimited"
+  sh "sudo bash -c 'ulimit -c unlimited'"
 
   sh "docker network create ${env.IROHA_NETWORK}"
   def iC = dPullOrBuild.dockerPullOrUpdate("${platform}-develop-build",
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -24,8 +24,8 @@ def doDebugBuild(coverageEnabled=false) {
   }
 
   // for saving coredumps
-  sh "sudo echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
-  sh "sudo bash -c 'ulimit -c unlimited'"
+  sh "echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
+  sh "bash -c 'ulimit -c unlimited'"
 
   sh "docker network create ${env.IROHA_NETWORK}"
   def iC = dPullOrBuild.dockerPullOrUpdate("${platform}-develop-build",
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -22,6 +22,8 @@ def doDebugBuild(coverageEnabled=false) {
   if (env.NODE_NAME.contains('arm7')) {
     parallelism = 1
   }
+  sh "sudo echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
+  sh "sudo ulimit -c unlimited"
 
   sh "docker network create ${env.IROHA_NETWORK}"
   def iC = dPullOrBuild.dockerPullOrUpdate("${platform}-develop-build",
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -22,6 +22,8 @@ def doDebugBuild(coverageEnabled=false) {
   if (env.NODE_NAME.contains('arm7')) {
     parallelism = 1
   }
+
+  // for saving coredumps
   sh "sudo echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
   sh "sudo ulimit -c unlimited"
 
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -26,7 +26,6 @@ def doDebugBuild(coverageEnabled=false) {
   // enable coredumps collecting
   if (params.coredump) {
     sh "echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
-    sh "echo \$0"
     sh "ulimit -c unlimited"
   }
 
--- a/.jenkinsci/debug-build.groovy
+++ b/.jenkinsci/debug-build.groovy
@@ -23,9 +23,12 @@ def doDebugBuild(coverageEnabled=false) {
     parallelism = 1
   }
 
-  // for saving coredumps
-  sh "echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
-  sh "bash -c 'ulimit -c unlimited'"
+  // enable coredumps collecting
+  if (params.coredump) {
+    sh "echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
+    sh "echo \$0"
+    sh "ulimit -c unlimited"
+  }
 
   sh "docker network create ${env.IROHA_NETWORK}"
   def iC = dPullOrBuild.dockerPullOrUpdate("${platform}-develop-build",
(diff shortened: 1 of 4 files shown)


l4l:
> ulimit -c unlimited settings doesn't restrict size of core dump. In theory, it might be too large

Looks dangerous, might be consider using something like logrotate (but for coredumps)?

Stayer:
> > ulimit -c unlimited settings doesn't restrict size of core dump. In theory, it might be too large
> 
> Looks dangerous, might be consider using something like logrotate (but for coredumps)?

Our build servers works like that:
1. Creates VM on AWS
2. Run docker inside for build & tests
3. Upload artifacts, binaries, coredumps, etc. to artifacts server
4. Remove this VM

Typically, AWS instances are using only for build of one task, and very rarely for two or three jobs. At the end of working day everything is deleted.

.jenkinsci/linux-post-step.groovy
```diff
@@ -1,6 +1,23 @@
...
     try {
+      // stop write core dumps
+      sh "bash -c 'ulimit -c 0'"
+      if (currentBuild.currentResult != "SUCCESS") {
+        // handling coredumps (if tests crashed)
+        def currentPath = sh(script: "pwd", returnStdout: true).trim()
+        def dumpsFileName = sprintf('coredumps-%1$s.tar',
+          [GIT_COMMIT.substring(0,8)])
+
+        // sh(script: "find ${currentPath} -type f -name '*.coredump' | zip -j ${dumpsFileName} -@")
```
bakhtin: Is that comment any useful?

.jenkinsci/linux-post-step.groovy
```diff
@@ -1,6 +1,23 @@
...
+        sh(script: "find ${currentPath} -type f -name '*.coredump' -exec tar -rvfj ${dumpsFileName} {} \\;")
```
bakhtin: Please, flatten directory structure in an archive. So that it only contains files, not directory trees.

.jenkinsci/linux-post-step.groovy
```diff
@@ -1,6 +1,23 @@
 def linuxPostStep() {
   timeout(time: 600, unit: "SECONDS") {
     try {
+      // stop write core dumps
+      sh "bash -c 'ulimit -c 0'"
```
bakhtin: Default Jenkins interpreter is Bash. Why do you need to explicitly use it to invoke the command?

.jenkinsci/linux-post-step.groovy
```diff
@@ -1,6 +1,23 @@
...
+      if (currentBuild.currentResult != "SUCCESS") {
+        // handling coredumps (if tests crashed)
+        def currentPath = sh(script: "pwd", returnStdout: true).trim()
```
bakhtin: There is a `$WORKSPACE` variable exists

.jenkinsci/debug-build.groovy
```diff
@@ -23,6 +23,10 @@ def doDebugBuild(coverageEnabled=false) {
     parallelism = 1
   }
 
+  // for saving coredumps
+  sh "echo %e.%p.coredump > /proc/sys/kernel/core_pattern"
```
bakhtin: Can we make coredumps collection option configurable? That can be disabled by request (using Jeknins GUI). It should be enabled by default

.jenkinsci/linux-post-step.groovy
MBoldyrev: Why should we flatten the directories? I will only complicate the code (which is now very readable) and cause possible name conflicts resulting in dumps being overwritten or skipped. I do not think it is worth a couple of times pressing \<Tab\> to auto-complete the path to core dump in a shell. Also the original directory tree will make it easier to understand which executable it corresponds to.

bakhtin: If the way coredumps are stored in an archive is OK with other developers then we can keep it

.jenkinsci/linux-post-step.groovy
```diff
@@ -2,18 +2,16 @@ def linuxPostStep() {
...
+      if (currentBuild.currentResult != "SUCCESS" && params.coredump) {
         // handling coredumps (if tests crashed)
-        def currentPath = sh(script: "pwd", returnStdout: true).trim()
-        def dumpsFileName = sprintf('coredumps-%1$s.tar',
+        def dumpsFileName = sprintf('coredumps-%1$s.bzip2',
           [GIT_COMMIT.substring(0,8)])
 
-        // sh(script: "find ${currentPath} -type f -name '*.coredump' | zip -j ${dumpsFileName} -@")
-        sh(script: "find ${currentPath} -type f -name '*.coredump' -exec tar -rvfj ${dumpsFileName} {} \\;")
+        sh(script: "find ${$WORKSPACE} -type f -name '*.coredump' -exec tar -rvfj ${dumpsFileName} {} \\;")
```
bakhtin: Please verify this command actually works. I'm getting an error while verifying it locally:
```
$ find /tmp/test-tar/ -type f -name '123*' -exec tar -rvfj coredumps.bzip2 {} \;
tar: coredumps.bzip2: Cannot stat: No such file or directory
tar: Removing leading `/' from member names
/tmp/test-tar/1/2/1234.txt
tar: Exiting with failure status due to previous errors
tar: coredumps.bzip2: Cannot stat: No such file or directory
tar: Removing leading `/' from member names
/tmp/test-tar/1/2/123.txt
tar: Exiting with failure status due to previous errors
tar: coredumps.bzip2: Cannot stat: No such file or directory
tar: Removing leading `/' from member names
/tmp/test-tar/2/3/123.txt
tar: Exiting with failure status due to previous errors
```

MBoldyrev: I would suggest `-exec tar cjvf ${dumpsFileName} {} \\+ \\;`, but please check the escape backslashes.

.jenkinsci/linux-post-step.groovy
```diff
@@ -3,12 +3,12 @@ def linuxPostStep() {
...
       // stop write core dumps
       sh "ulimit -c 0"
+      // handling coredumps (if tests crashed)
       if (currentBuild.currentResult != "SUCCESS" && params.coredump) {
-        // handling coredumps (if tests crashed)
         def dumpsFileName = sprintf('coredumps-%1$s.bzip2',
           [GIT_COMMIT.substring(0,8)])
 
-        sh(script: "find ${$WORKSPACE} -type f -name '*.coredump' -exec tar -rvfj ${dumpsFileName} {} \\;")
+        sh(script: "find . -type f -name '*.coredump'-exec tar -cjvf ${dumpsFileName} {} \\;")
```
bakhtin: This will create an archive, containing just the first file found. The rest will be left out. The working solution is to use `r` option to create an uncompressed archive. And only then it can be compressed.

bakhtin:
```suggestion
        sh(script: "find . -type f -name '*.coredump' -exec tar -cjvf ${dumpsFileName} {} \\;")
```

test/integration/consensus/consensus_sunny_day.cpp
```diff
@@ -23,7 +23,7 @@ using namespace framework::test_subscriber;
 static size_t num_peers = 1, my_num = 0;
 
 auto mk_local_peer(uint64_t num) {
-  auto address = "0.0.0.0:" + std::to_string(num);
+  auto address = "0.0.0.0:--" + std::to_string(num);
```
bakhtin: Do not forget to undo changes to the test

.jenkinsci/linux-post-step.groovy
MBoldyrev: @bakhtin are you sure? I do not see any problem with file aggregation except (most likely) a typo - a missing whitespace before `-exec`.

MBoldyrev: ow, I am wrong, there is also a `+` missing after `{}`

MBoldyrev: that should do:
```suggestion
        sh(script: "find . -type f -name '*.coredump' -exec tar -cjvf ${dumpsFileName} {} \\+;")
```

PR #2025, opened 2019-01-15
IR-156: Remove node.js library

Pobepto:
### Description of the Change
NodeJS library is obsolete and is not compatible with current version of Iroha. The task is to remove it from Iroha including all corresponding documentation 

### Benefits
🐌Preparing for new library for JS 🐌

- [x] Removed old node.js library
- [x] Removed old documentation for library
- [x] Added documentation for new library

### Possible Drawbacks 
None


neewy:
Thank you for your contribution!
If you delete this library — where can we find new one?

Please fix your pull request: it has issues with DCO. You can do it automatically using git commit -s.

Pobepto:
> If you delete this library — where can we find new?

@neewy, the new version of javascript library will be available soon. I will create new PR for including it to the iroha repository 🙂

l4l:
I'm quite afraid that docs separated from the code might cause inconsistency quite fast

docs/source/guides/libraries/javascript.rst
```diff
@@ -0,0 +1,134 @@
...
+    $ yarn add iroha-helpers
+
+Commands
+^^^^^^^^
+For usage of any command you need to provide ``commandOptions`` as a first argument.
+
+.. code-block:: javascript
+
+  const commandOptions = {
+    privateKeys: [''], // Array of private keys in hex format
```
igor-egorov: I suggest putting an example of a key here. It may not be clear for the reader what kind of string is expected here.

An example will implicitly show the alphabet, the length.

PR #2027, opened 2019-01-15
DOPS-202: Jenkins GUI For Build with Parameters

BulatSaif:
### Description of the Change
1) Changed to human readable options
screenshot from 2019-01-15 17-43-34
2) Fix sonar token
3) Always run release step if we in dev or master branch 
### Benefits
Simple looking, and very flexible. 

### Possible Drawbacks 
None


BulatSaif:
Now README file inside build page:
screenshot from 2019-01-17 11-45-10
``` bash

```
screenshot from 2019-01-17 11-46-10

bakhtin:
Please, verify that this section actually works: https://github.com/hyperledger/iroha/blob/51853c898a4687d7aa9affaab88b9c586f34cd28/Jenkinsfile#L40
AFAIK, `UNSTABLE` result does not throw an exception (probably, `ABORTED` also not).

bakhtin:
`cpusAvailable` should be configurable (see current pipeline). Default is 8 for Linux and 4 for Mac
https://github.com/hyperledger/iroha/blob/51853c898a4687d7aa9affaab88b9c586f34cd28/Jenkinsfile#L226-L227

Jenkinsfile
```diff
@@ -136,6 +136,7 @@ node ('master') {
   pushDockerTag = 'not-supposed-to-be-pushed'
   packagePush = false
   specialBranch = false
+  parallelism = 0
```
bakhtin: Did you test it with `0`? Is that even a valid option for `make` command? I suggest making `1` a default value

BulatSaif: `0`  is converted to default: 8 for Linux and 4 for Mac [here]( https://github.com/hyperledger/iroha/pull/2027/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R251) ,  you can choose `1` and build in one thread both Linux and Mac. 
`0` not valid option,  and it not be used. 
 `make: the '-j' option requires a positive integer argument`,

PR #2029, opened 2019-01-16
Change Simulator functions to return objects instead of emitting events

nickaleks:
### Description of the Change

This PR changes simulator functions processProposal and processVerifiedProposal, so that they return resulting objects, instead of broadcasting events. Event broadcasting still happens, but in a subscription lambda. This means that we can now access all field of incoming events when forming new event.

This is useful, if, for example we want to pass some information from one event to the other. This will be useful in the upcoming changes which store peer's list as some sort of "Ledger state" in all events.

### Benefits

Event broadcasting happens in one place. Simulator functions are now purer and easier to test.

### Possible Drawbacks 

None

### DIFF
--- a/test/module/irohad/simulator/simulator_test.cpp
+++ b/test/module/irohad/simulator/simulator_test.cpp
@@ -253,7 +253,7 @@ TEST_F(SimulatorTest, FailWhenSameAsProposalHeight) {
  * @then verified proposal consists of txs we did not fail, and the failed
  * transactions are provided as well
  */
- TEST_F(SimulatorTest, SomeFailingTxs) {
+TEST_F(SimulatorTest, SomeFailingTxs) {
   // create a 3-height proposal, but validator returns only a 2-height
   // verified proposal
   const int kNumTransactions = 3;
@@ -269,8 +269,8 @@ TEST_F(SimulatorTest, FailWhenSameAsProposalHeight) {
           .build());
   auto verified_proposal_and_errors =
       std::make_unique<VerifiedProposalAndErrors>();
-  const shared_model::interface::types::HeightType verified_proposal_height =
-  2; const std::vector<shared_model::proto::Transaction>
+  const shared_model::interface::types::HeightType verified_proposal_height = 2;
+  const std::vector<shared_model::proto::Transaction>
       verified_proposal_transactions{txs[0]};
   verified_proposal_and_errors->verified_proposal =
       std::make_unique<shared_model::proto::Proposal>(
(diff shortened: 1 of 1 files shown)


test/module/irohad/simulator/simulator_test.cpp
```diff
@@ -168,30 +168,18 @@ TEST_F(SimulatorTest, ValidWhenPreviousBlock) {
 
   init();
 
-  auto proposal_wrapper =
-      make_test_subscriber<CallExact>(simulator->onVerifiedProposal(), 1);
```
lebdron: There are no tests which check the successful branch in the subscription lambda created in constructor, so the coverage is reduced.
I would suggest to either: 
- add a separate test case with `  auto proposal_wrapper = make_test_subscriber<CallExact>(simulator->onVerifiedProposal(), 1);`, so that successful branch is covered.
- or rewrite the test with `subject<OrderingEvent>`, create `Simulator` in setup and subscribe it to the subject, remove init method, leave the wrappers in modified tests, and trigger the processing by calling `subject.get_subscriber().on_next()`. This approach is used in `synchronizer_test`

test/module/irohad/simulator/simulator_test.cpp
```diff
@@ -270,22 +253,13 @@ TEST_F(SimulatorTest, FailWhenSameAsProposalHeight) {
  * @then verified proposal consists of txs we did not fail, and the failed
  * transactions are provided as well
  */
-TEST_F(SimulatorTest, SomeFailingTxs) {
-  // create a 3-height proposal, but validator returns only a 2-height verified
-  // proposal
+ TEST_F(SimulatorTest, SomeFailingTxs) {
```
lebdron:
```suggestion
TEST_F(SimulatorTest, SomeFailingTxs) {
```

test/module/irohad/simulator/simulator_test.cpp
```diff
@@ -295,8 +269,8 @@ TEST_F(SimulatorTest, SomeFailingTxs) {
           .build());
   auto verified_proposal_and_errors =
       std::make_unique<VerifiedProposalAndErrors>();
-  const shared_model::interface::types::HeightType verified_proposal_height = 2;
-  const std::vector<shared_model::proto::Transaction>
+  const shared_model::interface::types::HeightType verified_proposal_height =
+  2; const std::vector<shared_model::proto::Transaction>
```
lebdron: clang-format please

PR #2030, opened 2019-01-16
DOPS-243 Custom github notifications

BulatSaif:

### Description of the Change

Add custom notifications in github.
screenshot from 2019-01-16 17-39-25

`  * Details`  direct you to the run build page

### Benefits
It will help us to migrate to new CI 

In new multi-branch pipeline we need disable a default notifications,otherwise it will  rewrite old CI notifications.
screenshot from 2019-01-16 16-32-11

### Possible Drawbacks 

None


Jenkinsfile
```diff
@@ -83,6 +86,10 @@ def cmd_sanitize(String cmd){
   return true
 }
 
+def gitNotify (context, description, status, targetUrl='' ){
```
bakhtin: What do you about implementing `Notification` as an abstract class? In the future we would need email, IM notifications (maybe some others). Each notification provider would've done its own class implementation. This would allow for a more uniform behaviour across different notifiers.

Jenkinsfile
```diff
@@ -32,27 +32,32 @@ def build(Build build) {
     node(build.worker.label) {
       try {
         echo "Worker: ${env.NODE_NAME}"
+        gitNotify ("New CI: " + build.name, "Started...", 'PENDING')
```
l4l: Why is that "New"?

Jenkinsfile
```diff
@@ -277,6 +290,9 @@ node ('master') {
   cleanWs()
   parallel tasks
 
+  if (build_scenario == 'Before merge to trunk')
+    gitNotify ("New CI: Merge to trunk", "Finish", 'SUCCESS')
+
   // if (currentBuild.currentResult == 'SUCCESS') {
```
l4l: Quite unrelated to the pr, but still, why is there commented out code?

BulatSaif: It was for testing purposes, I will delete it.

Jenkinsfile
BulatSaif: We decided for some period use 2 CI: this CI(new) and current one(old) 
Old CI generate: `continuous-integration/jenkins/pr-head` check name.
New CI generate `New CI: <info>` check names.
'New CI' prefix needs to not scare developers if build marked as failed with new ci, because "new" can be not stable. 
After Old CI will be delete, the 'New CI' prefix will be rename into  'Jenkins'

Jenkinsfile
```diff
@@ -34,34 +36,35 @@ def build(Build build) {
...
+      } catch(Exception e) {
+        print "Error was detected: " + e
+        currentBuild.result = 'FAILURE'
+        build.builder.postSteps.failure.each {
+          it()
         }
       }
       // ALWAYS
       finally {
+        if (currentBuild.currentResult == 'SUCCESS')
```
bakhtin: What about reporting other statuses? What I was suggesting is to just execute:
`gitNotify ('JenkinsCI', build.name, currentBuild.currentResult)`

BulatSaif: The github status is different from Jenkins status

```
      Jenkins  : Github
               : "PENDING"
     "SUCCESS" : "SUCCESS",	
     "UNSTABLE":  	
     "ABORTED" :  
     "FAILURE" : "FAILURE",
               : "ERROR"
```
gitNotify (Name , Jenkins status or massage , Github Status )

PR #2031, opened 2019-01-17
Selective block streaming

Akvinikym:


shared_model/validators/field_validator.cpp
```diff
@@ -365,6 +365,17 @@ namespace shared_model {
       }
     }
 
+    void FieldValidator::validateNonNegativeHeight(
```
nickaleks: I don't think this method is necessary, since height cannot be negative due to its type.

test/module/irohad/torii/processor/query_processor_test.cpp
```diff
@@ -158,16 +168,74 @@ TEST_F(QueryProcessorTest, GetBlocksQuery) {
   ASSERT_TRUE(wrapper.validate());
 }
 
+/**
+ * @given account, ametsuchi queries @and mocked ledger with some blocks
+ * @when valid block query with specified height is sent @and this height is
+ * less than the ledger's one
+ * @then Query Processor should emit blocks starting from the specified height
+ * @and finish by the newest ones @and handle the case when new block is
+ * comitted in the process of retreiving old ones
```
nickaleks:
```suggestion
 * committed in the process of retrieving old ones
```

neewy:
We need to discuss responsibilities first. There is no problem with this interface solely, but having individual buffers for each client means taking too much on shoulders of Iroha peer. We won't be able to provide an adequate level of fault tolerance in case of huge loads. Although, it might be enough just to warn users about their responsibilities in this case.

shared_model/schema/queries.proto
```diff
@@ -103,4 +103,7 @@ message Query {
 message BlocksQuery {
   QueryPayloadMeta meta = 1;
   Signature signature = 2;
+  oneof opt_height {
```
lebdron: It does not seem intuitive that a single parameter changes the logic of the query - from a continuous stream of new blocks to returning a single block. Why not to create a non-streaming query that will return a single block?

shared_model/interfaces/query_responses/get_block_response.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "interfaces/base/model_primitive.hpp"
+#include "interfaces/iroha_internal/block.hpp"
+
+namespace shared_model {
+  namespace interface {
+    /**
+     * Provide response with the block
+     */
+    class GetBlockResponse : public ModelPrimitive<GetBlockResponse> {
```
lebdron: It looks identical to BlockResponse, why not reuse it?
https://github.com/hyperledger/iroha/blob/5671e2d1300b344714aa883f2dbf6f241ec7bb50/shared_model/interfaces/query_responses/block_response.hpp#L20

shared_model/backend/protobuf/query_responses/impl/proto_get_block_response.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "backend/protobuf/query_responses/proto_get_block_response.hpp"
+
+namespace shared_model {
+  namespace proto {
+
+    template <typename QueryResponseType>
+    GetBlockResponse::GetBlockResponse(QueryResponseType &&queryResponse)
+        : CopyableProto(std::forward<QueryResponseType>(queryResponse)),
+          block_response_{proto_->block_response()},
```
lebdron: https://github.com/hyperledger/iroha/blob/5671e2d1300b344714aa883f2dbf6f241ec7bb50/shared_model/backend/protobuf/query_responses/impl/proto_block_response.cpp#L14
🙈

irohad/ametsuchi/impl/postgres_query_executor.cpp
```diff
@@ -312,19 +313,34 @@ namespace iroha {
...
       try {
         soci::rowset<T> st =
-            (sql_->prepare
-                 << (cmd % checkAccountRolePermission(Role::kGetBlocks)).str(),
-             soci::use(query.creatorAccountId(), "role_account_id"));
+            (sql_.prepare
+                 << (cmd % getAccountRolePermissionCheckSql(permission)).str(),
+             soci::use(account_id, "role_account_id"));
 
-        return st.begin()->get<0>();
```
lebdron: `return st.begin()->get<0>();` seems to work well here, so the change does not seem necessary.

test/module/irohad/ametsuchi/postgres_query_executor_test.cpp
```diff
@@ -774,6 +777,115 @@ namespace iroha {
...
 
+    class GetBlockExecutorTest : public QueryExecutorTest {
+      class StorageImpl {
+        friend class GetBlockExecutorTest;
+      };
+
+     public:
+      friend class StorageImpl;
+      void SetUp() override {
+        QueryExecutorTest::SetUp();
```
lebdron: Please remove SetUp, friend class, and StorageImpl since they have no effect.

test/module/irohad/ametsuchi/postgres_query_executor_test.cpp
```diff
@@ -774,6 +777,115 @@ namespace iroha {
...
+      }
+
+      // TODO [IR-257] Akvinikym 30.01.19: remove the method and use mocks
+      /**
+       * Commit some number of blocks to the storage
+       * @param blocks_amount - number of blocks to be committed
+       */
+      void commitBlocks(shared_model::interface::types::HeightType
+                            blocks_amount = kLedgerHeight) {
```
lebdron:
```suggestion
                            number_of_blocks = kLedgerHeight) {
```

PR #2032, opened 2019-01-17
ITF Fake Peers: behaviours & internal communication capabilities

MBoldyrev:
### Description of the Change

This makes ITF Fake Peers apt for testing basic internal network communication between Iroha peers. Fake Peers get all needed network stubs and handles to simulate network environment in ITF tests. Also Fake Peers can be assigned standard or custom behaviours to reuse some common automation patterns.

### Benefits

Extended network testing facilities in ITF.

### Possible Drawbacks 

A this point we need to be extra careful with synchronization when using Fake Peers in tests, as they run asynchronously. Careless usage may lead to randomly failing tests.

Please see `fake_peer_example_test`.


test/framework/integration_framework/fake_peer/behaviour/behaviour.hpp
```diff
@@ -0,0 +1,80 @@
...
+      virtual ~Behaviour();
+
+      /// Enable the behaviour for the given peer
+      void adopt(const std::shared_ptr<FakePeer> &fake_peer);
+
+      /// Disable the behaviour
+      void absolve();
+
+      /// This method gets subscribed on Fake Peer's MST messages.
+      virtual void processMstMessage(MstMessagePtr message) = 0;
```
l4l: probably default implementation as "do nothing" for these abstract methods might be ok

test/framework/integration_framework/fake_peer/block_storage.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include <boost/algorithm/string/join.hpp>
+#include <boost/range/adaptor/map.hpp>
+#include <boost/range/algorithm/max_element.hpp>
+#include "backend/protobuf/block.hpp"
+#include "framework/integration_framework/fake_peer/fake_peer.hpp"
+
+/// Emplace to a map.
+/// @return true if overwritten.
+template <typename MapType, typename KeyType, typename ValueType>
+static bool emplaceCheckingOverwrite(MapType &map,
```
l4l, line 20: wondering for moving to c++17 :( https://en.cppreference.com/w/cpp/container/map/insert_or_assign

test/framework/integration_framework/fake_peer/block_storage.cpp
```diff
@@ -0,0 +1,125 @@
...
+      return found->second;
+    }
+
+    BlockStorage::BlockPtr BlockStorage::getTopBlock() const {
+      if (blocks_by_height_.empty()) {
+        getLogger()->info(
+            "Requested top block, but the block storage is empty.");
+        return {};
+      }
+      return blocks_by_height_.at(*boost::range::max_element(
```
l4l, line 85: can be operator[]

test/framework/integration_framework/fake_peer/block_storage.cpp
```diff
@@ -0,0 +1,125 @@
...
+        getLogger()->warn(
+            "Requested removal of a using fake peer "
+            "without previously registering it.");
+        return;
+      }
+      using_peers_.erase(found);
+    }
+
+    logger::Logger BlockStorage::getLogger() const {
+      std::vector<std::string> using_peers_addresses;
```
l4l: Не очень имя, в первый раз перепутал с `using_peers_` и долго не мог понять у кого вызывается lock. Мб лучше `addresses_str`?

test/framework/integration_framework/fake_peer/block_storage.hpp
```diff
@@ -0,0 +1,60 @@
...
+      void claimNotUsingPeer(const std::shared_ptr<FakePeer> &peer);
+
+     private:
+      logger::Logger getLogger() const;
+
+      std::unordered_map<HeightType, BlockPtr> blocks_by_height_;
+      std::unordered_map<HashType, BlockPtr, HashType::Hasher> blocks_by_hash_;
+
+      /// The collection of peers claiming to use this block storage.
+      // I am using vector (although a kind of set would fit better) because
```
l4l: I am using vector -> Vector is used

test/framework/integration_framework/fake_peer/block_storage.hpp
```diff
@@ -0,0 +1,60 @@
...
+      // weak pointers seem unsuitable for any comparison, as the shared pointer
+      // they point to may change.
```
l4l: as the <s>shared pointer</s> shared_ptr <s>they point to</s> reference may change

test/framework/integration_framework/fake_peer/fake_peer.hpp
```diff
@@ -10,169 +10,229 @@
...
+       * @param listen_ip - IP on which this fake peer should listen
+       * @param internal_port - the port for internal commulications
+       * @param key - the keypair of this peer
+       * @param real_peer - the main tested peer managed by ITF
+       * @param common_objects_factory - common_objects_factory
+       * @param transaction_factory - transaction_factory
+       * @param batch_parser - batch_parser
+       * @param transaction_batch_factory - transaction_batch_factory
+       * @param gree_all_proposals - whether this peer should agree all
+       * proposals
```
l4l: Sweet yeti of the Serengeti!
That a tremendous number of ctor args. Probably not an issue of the PR, but that tech debt should be considered at least in future

test/framework/integration_framework/fake_peer/proposal_storage.cpp
```diff
@@ -0,0 +1,45 @@
...
+  namespace fake_peer {
+
+    ProposalStorage::ProposalStorage()
+        : default_provider_([](auto &) { return boost::none; }) {}
+
+    OrderingProposalRequestResult ProposalStorage::getProposal(
+        const Round &round) const {
+      auto it = proposals_map_.find(round);
+      if (it != proposals_map_.end()) {
+        if (it->second) {
```
l4l: might be a bit clearer:
`return it->second ? OrderingProposalRequestResult{*it->second}: boost::none;`

test/framework/integration_framework/fake_peer/types.hpp
```diff
@@ -0,0 +1,99 @@
...
+  }    // namespace consensus
+  namespace ordering {
+    namespace transport {
+      class OnDemandOsServerGrpc;
+    }
+    class OrderingGateTransportGrpc;
+    class OrderingServiceTransportGrpc;
+  }  // namespace ordering
+  class MstState;
+}  // namespace iroha
```
l4l, line 63: I guess these fds should be in core, rather in tests

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -634,4 +675,9 @@ namespace integration_framework {
       boost::filesystem::remove_all(iroha_instance_->block_store_dir_);
     }
   }
+
+  const std::shared_ptr<IrohaInstance>
+      &IntegrationTestFramework::getIrohaInstance() const {
```
l4l: `const shared_ptr<..>`? :\ Maybe just `const IrohaInstance&`?

Also I'm not really sure that this component should be publicly available

test/integration/acceptance/basic_mst_state_propagation.cpp
```diff
@@ -1,111 +0,0 @@
-/**
```
l4l, line 1: Why is that removed?

test/framework/integration_framework/fake_peer/block_storage.cpp
MBoldyrev: Comrade l4l, by using `at` I strengthen the immutability of the map!

test/framework/integration_framework/fake_peer/proposal_storage.cpp
MBoldyrev: the ternary operator wants both alternatives of  the same type, and accomplishing this makes the code way uglier than the current version, which is at least easily readable.

test/integration/acceptance/basic_mst_state_propagation.cpp
MBoldyrev: it is replaced with `test/integration/acceptance/fake_peer_example_test.cpp`

shared_model/interfaces/iroha_internal/transaction_batch.hpp
```diff
@@ -53,6 +53,9 @@ namespace shared_model {
           size_t number_of_tx,
           const shared_model::crypto::Signed &signed_blob,
           const shared_model::crypto::PublicKey &public_key) = 0;
+
+      /// Pretty print the batch contents.
```
igor-egorov, line 57: minor
Seems that just "batch contents" (without "the") will be ok.

test/framework/integration_framework/fake_peer/behaviour/honest.cpp
```diff
@@ -0,0 +1,166 @@
...
+      getFakePeer().voteForTheSame(message);
+    }
+
+    std::string HonestBehaviour::getName() {
+      return "honest behaviour";
+    }
+
+    LoaderBlockRequestResult HonestBehaviour::processLoaderBlockRequest(
+        LoaderBlockRequest request) {
+      const auto block_storage = getFakePeer().getBlockStorage();
```
Akvinikym: Why not `const auto &` here and below?

test/framework/integration_framework/integration_test_framework.hpp
```diff
@@ -251,6 +277,32 @@ namespace integration_framework {
      */
     IntegrationTestFramework &sendQuery(const shared_model::proto::Query &qry);
 
+    /// Send proposal to this peer' s ordering service.
```
Akvinikym:
```suggestion
    /// Send proposal to this peer's ordering service.
```

test/framework/integration_framework/fake_peer/network/mst_network_notifier.hpp
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef FAKE_PEER_MST_NETWORK_NOTIFIER_HPP_
+#define FAKE_PEER_MST_NETWORK_NOTIFIER_HPP_
+
+#include <rxcpp/rx.hpp>
+
```
igor-egorov: redundant line break

test/framework/integration_framework/fake_peer/network/ordering_service_network_notifier.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef FAKE_PEER_OS_NETWORK_NOTIFIER_HPP_
+#define FAKE_PEER_OS_NETWORK_NOTIFIER_HPP_
+
+#include "network/ordering_service_transport.hpp"
+
+#include <rxcpp/rx.hpp>
+
```
igor-egorov: redundant line break

test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace integration_framework {
+  namespace fake_peer {
+
+    Behaviour::~Behaviour() {
+      absolve();
+    }
+
+    void Behaviour::adopt(const std::shared_ptr<FakePeer> &fake_peer) {
+      // This code feels like part of constructor, but the use of `this'
```
igor-egorov, line 19: please use the same type of enclosing quotes around `this`

test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
```diff
@@ -0,0 +1,53 @@
...
+      // Hint: such calls would precede the derived class construction.
+      fake_peer_wptr_ = fake_peer;
+      log_ = logger::log(getName() + " of fake peer at "
+                         + fake_peer->getAddress());
+      // subscribe for all messages
+      subscriptions_.emplace_back(
+          getFakePeer().getMstStatesObservable().subscribe(
+              [this](const auto &message) {
+                this->processMstMessage(message);
+              }));
```
igor-egorov, line 53: can we avoid intermediary lambdas and pass class methods like `processMstMethods` directly to `subscribe` call?

something like
```
getFakePeer().getMstStatesObservable().subscribe(processMstMessage);
```

test/framework/integration_framework/fake_peer/behaviour/honest.cpp
MBoldyrev: here because the method signature is
```c++
      boost::optional<const BlockStorage &> getBlockStorage() const;
```
neither below I can see where to make this change.

test/framework/integration_framework/fake_peer/behaviour/behaviour.hpp
igor-egorov: logging method calls also is an option. Please consider

test/framework/integration_framework/fake_peer/behaviour/honest.cpp
igor-egorov: Probably Pavel meant the line 42
```
const auto block_storage = getFakePeer().getBlockStorage();
```

test/framework/integration_framework/fake_peer/block_storage.cpp
igor-egorov: why not just `operator[]` ?


```
result = end() == find(key)
map[key] = val
return result
```

test/framework/integration_framework/fake_peer/fake_peer.cpp
```diff
@@ -119,6 +130,7 @@ namespace integration_framework {
 
     FakePeer &FakePeer::setBlockStorage(
         const std::shared_ptr<BlockStorage> &block_storage) {
+      ensureInitialized();
```
igor-egorov, line 173: Why we can't use `initialize()` here instead `ensureInitialized()`?

Can we put such check into `initialize()`?
```diff
    FakePeer &FakePeer::initialize() {
-      BOOST_VERIFY_MSG(!initialized_, "Already initialized!"); 
+      if (not initialized_) {
        // here comes the initialization of members requiring shared_from_this()
        synchronizer_transport_ =
            std::make_shared<LoaderGrpc>(shared_from_this());

        initialized_ = true;
+      }
      return *this;
    }
```

test/framework/integration_framework/fake_peer/fake_peer.cpp
```diff
@@ -106,8 +106,19 @@ namespace integration_framework {
           + getAddress() + ")");
     }
 
+    FakePeer &FakePeer::initialize() {
+      BOOST_VERIFY_MSG(!initialized_, "Already initialized!");
```
igor-egorov: our practice is using of `not,and,or` instead of `!,&&,||`

test/framework/integration_framework/integration_test_framework.cpp
```diff
@@ -534,11 +535,43 @@ namespace integration_framework {
...
 
+  IntegrationTestFramework &IntegrationTestFramework::sendBatches(
+      const iroha::consensus::Round &round,
+      const std::vector<TransactionBatchSPtr> &batches) {
+    auto on_demand_os_transport =
+        iroha::ordering::transport::OnDemandOsClientGrpcFactory(
+            async_call_,
+            [] { return std::chrono::system_clock::now(); },
+            std::chrono::milliseconds(0)  // the proposal waiting
+                                          // timeout does not apply here
```
igor-egorov: please change the comment.

unfortunately, the comment brings more confusion than clarification.
Please add the reason, **why** timeout does not apply here.

without the clarification, the comment just states the fact without explanation

test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -274,14 +282,17 @@ TEST_F(FakePeerExampleFixture,
...
+ *
  * \attention this code is nothing more but an example of Fake Peer usage
  *
  * @given a network of two iroha peers
  * @when a valid command is sent to one
  * @then it must propagate either a proposal or a batch
  */
 TEST_F(FakePeerExampleFixture,
-       OrderingMessagePropagationAfterValidCommandReceived) {
+       DISABLED_OrderingMessagePropagationAfterValidCommandReceived) {
```
igor-egorov: Please create a task with explanation why the test is disabled and add a todo here.

test/framework/integration_framework/fake_peer/behaviour/honest.cpp
MBoldyrev: Yes, now I got it, we can assign an rvalue to a const lvalue reference.

MBoldyrev: Yes, now I got it, we can assign an rvalue to a const lvalue reference.

test/framework/integration_framework/fake_peer/fake_peer.cpp
MBoldyrev: I would like to force initialization at creation time before the object starts being used. That is why there is an assertion.

irohad/ordering/impl/ordering_gate_transport_grpc.cpp
```diff
@@ -17,7 +17,7 @@ grpc::Status OrderingGateTransportGrpc::onProposal(
     ::grpc::ServerContext *context,
     const iroha::protocol::Proposal *request,
     ::google::protobuf::Empty *response) {
-  async_call_->log_->info("receive proposal");
+  async_call_->log_->info("received a proposal");
```
igor-egorov: test single comment

test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
MBoldyrev: Seems we cannot as the `this` pointer has to be bound somehow to the callable, and a lambda is a cheap way to do that.

test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
MBoldyrev: This type of quotes is widely used in program documentation and I personally like it.

test/integration/acceptance/fake_peer_example_test.cpp
MBoldyrev: I found no task about removing the old ordering service code, so we decided the comment above the test code is enough.

test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
igor-egorov: D:

MBoldyrev:
Need to fix some multithreading concurrency issues prior to merging this.

PR #2033, opened 2019-01-17
Fix tx status streaming

igor-egorov:

### Description of the Change

1. Consensus events now are delivered to all subscribers (TransportCommandService and Synchronizer)
2. Fixed non-graceful shutdown of tx status stream.

The first issue may lead to hang of client application in case when no events were emitted in the network.

### Benefits

Correctness and stability improved.

### Possible Drawbacks 

?

### Usage Examples or Tests

```torii_transport_command_test```

Thanks to @lebdron !

### DIFF
--- a/irohad/torii/impl/command_service_transport_grpc.cpp
+++ b/irohad/torii/impl/command_service_transport_grpc.cpp
@@ -23,237 +23,235 @@
 
 namespace torii {
 
-    CommandServiceTransportGrpc::CommandServiceTransportGrpc(
-        std::shared_ptr<CommandService> command_service,
-        std::shared_ptr<iroha::torii::StatusBus> status_bus,
-        std::shared_ptr<shared_model::interface::TxStatusFactory>
-            status_factory,
-        std::shared_ptr<TransportFactoryType> transaction_factory,
-        std::shared_ptr<shared_model::interface::TransactionBatchParser>
-            batch_parser,
-        std::shared_ptr<shared_model::interface::TransactionBatchFactory>
-            transaction_batch_factory,
-        rxcpp::observable<ConsensusGateEvent> consensus_gate_objects,
-        int maximum_rounds_without_update,
-        logger::Logger log)
-        : command_service_(std::move(command_service)),
-          status_bus_(std::move(status_bus)),
-          status_factory_(std::move(status_factory)),
-          transaction_factory_(std::move(transaction_factory)),
-          batch_parser_(std::move(batch_parser)),
-          batch_factory_(std::move(transaction_batch_factory)),
-          log_(std::move(log)),
-          consensus_gate_objects_(std::move(consensus_gate_objects)),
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by neewy** [comment]
We might want to release a hotfix for this bug. Is this change atomic? Can we rebase it onto master? @lebdron

**Comment by lebdron** [comment]
Yes, it should work fine with master branch.

File: irohad/torii/impl/command_service_transport_grpc.cpp [hunk 1]
```diff
@@ -14,240 +14,246 @@
...
+      command_service_
+          ->getStatusStream(hash)
+          // convert to transport objects
+          .map([&](auto response) {
+            log_->info("mapped {}, {}", *response, client_id);
+            return std::static_pointer_cast<
+                       shared_model::proto::TransactionResponse>(response)
+                ->getTransport();
+          })
+          .combine_latest(consensus_gate_observable)
```
MBoldyrev: Maybe we should throw away the second tuple member?
```c++
.combine_latest(consensus_gate_observable)
.map([] (auto &tuple) { return std::move(std::get<0>(tuple)); })
```

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -185,20 +185,14 @@ namespace torii {
...
-          }
-        });
+    auto consensus_gate_observable =
+        consensus_gate_objects_
+            // a dummy start_with lets us don't wait for the consensus event
+            // on further combine_latest
+            .start_with(ConsensusGateEvent{});
 
+    boost::optional<iroha::protocol::TxStatus> last_tx_status;
+    auto rounds_counter{0};
```
Akvinikym: It is used only in `take_while` lambda. Can't it be made static there and not be declared here?

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -185,20 +185,14 @@ namespace torii {
...
-            round_counter++;
-          }
-        });
+    auto consensus_gate_observable =
+        consensus_gate_objects_
+            // a dummy start_with lets us don't wait for the consensus event
+            // on further combine_latest
+            .start_with(ConsensusGateEvent{});
 
+    boost::optional<iroha::protocol::TxStatus> last_tx_status;
```
Akvinikym: I am strongly not sure about thread safety of this variable - it is captured by reference in two lambdas below, which can cause race conditions. Either thread-safe variant of optional should be used (if exists) or operations must be under mutex.
Also there's another way - atomic shared_ptr

l4l: Can be just std::atomic, since TxStatus is just a enum. But maybe it's possible just merge `take_while` and `filter` combinators?

Akvinikym:
Please also pay attention you are merging into master

PR #2034, opened 2019-01-17
Comments on reworking integration tests

MBoldyrev:
### Description of the Change

This change adds comments to some integration tests that suggest rewriting or removing them.

### Benefits

It will help us to reach consensus and rework these tests in the future.

### Possible Drawbacks 

If we misinterpret some test scopes, we may unintentionally cause lowering the test coverage.

Read the tests, read the comments, agree silently or argue noticeably.

Design new test coverage from scratch.

### DIFF
--- a/test/integration/acceptance/add_asset_qty_test.cpp
+++ b/test/integration/acceptance/add_asset_qty_test.cpp
@@ -22,7 +22,7 @@ class AddAssetQuantity : public AcceptanceFixture {
 };
 
 /**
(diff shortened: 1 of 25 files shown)


l4l:
I think it's better remove all redundant tests here, rather than just marking. As an example - #1047, where all these markings was useless

Though it might be nice to leave TODOs "as is" for review and remove code pieces in separate commit

MBoldyrev:
> Though it might be nice to leave TODOs "as is" for review and remove code pieces in separate commit

That is exactly what we intend here - reviewing change proposals and creating separate tasks for them.

PR #2036, opened 2019-01-21
New logger forward headers

MBoldyrev:
### Description of the Change

Forward declaration headers for the new logger.

### Benefits

The new logger related objects will be used through pointers. So the code that does not dereference those pointers can omit the inclusion of full headers. The major reason why we want it is the `fmt` header-only library used in the `logger.hpp` header, which one can keep away by using the `logger_fwd.hpp` header.

### Possible Drawbacks 

The logger-related includes appear in both headers (`logger_fwd.hpp`) and cpp-files (`logger.hpp`) - more includes overall.


libs/logger/logger_fwd.hpp
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+#ifndef IROHA_SPDLOG_LOGGER_LOGGER_FWD_HPP
+#define IROHA_SPDLOG_LOGGER_LOGGER_FWD_HPP
+
+#include <memory>
+
+namespace logger {
+
+  class Logger;
```
muratovv: Please, add a brief description why f_decls are required.

PR #2037, opened 2019-01-21
New libs for the new logger

MBoldyrev:
### Description of the Change

This change

- updates used `spdlog` library to 1.3.1
- adds external `fmt` library and its usage to the new logger
- adds version checking of the mentioned libs to the cmake `find_package` scripts

### Benefits

### Possible Drawbacks 

The cmake code needs a thorough review as I am not experienced in it.

### DIFF
--- a/cmake/Modules/Findfmt.cmake
+++ b/cmake/Modules/Findfmt.cmake
@@ -22,23 +22,23 @@ if (NOT DEFINED fmt_VERSION OR fmt_VERSION VERSION_LESS fmt_FIND_VERSION)
   message(STATUS "Package 'fmt' of version ${fmt_FIND_VERSION} not found. "
           "Will download it from git repo.")
 
-  set(git_url https://github.com/fmtlib/fmt.git)
+  set(GIT_URL https://github.com/fmtlib/fmt.git)
   set(fmt_VERSION ${fmt_FIND_VERSION})
-  set(git_tag "refs/tags/${fmt_VERSION}")
+  set(GIT_TAG "refs/tags/${fmt_VERSION}")
   set_package_properties(fmt
       PROPERTIES
-      URL ${git_url}
+      URL ${GIT_URL}
       DESCRIPTION "String formatting library"
       )
 
   externalproject_add(fmtlib_fmt
-      GIT_REPOSITORY  ${git_url}
-      GIT_TAG         ${git_tag}
+      GIT_REPOSITORY  ${GIT_URL}
+      GIT_TAG         ${GIT_TAG}
+      GIT_SHALLOW     1
       CONFIGURE_COMMAND "" # remove configure step
       BUILD_COMMAND "" # remove build step
       INSTALL_COMMAND "" # remove install step
       TEST_COMMAND "" # remove test step
-      UPDATE_COMMAND "" # remove update step
       )
   externalproject_get_property(fmtlib_fmt source_dir)
   set(fmt_INCLUDE_DIR ${source_dir}/include)
--- a/cmake/Modules/Findfmt.cmake
+++ b/cmake/Modules/Findfmt.cmake
@@ -1,9 +1,7 @@
 add_library(fmt INTERFACE IMPORTED)
 
-set(fmt_external_project_name fmtlib_fmt)
-
 find_path(fmt_INCLUDE_DIR fmt/format.h
-  PATHS ${EP_PREFIX}/src/${fmt_external_project_name}/include)
+  PATHS ${EP_PREFIX}/src/fmtlib_fmt/include)
 mark_as_advanced(fmt_INCLUDE_DIR)
 
 # read the fmt version stored in fmt/core.h
@@ -33,7 +31,7 @@ if (NOT DEFINED fmt_VERSION OR fmt_VERSION VERSION_LESS fmt_FIND_VERSION)
       DESCRIPTION "String formatting library"
       )
 
-  externalproject_add(${fmt_external_project_name}
+  externalproject_add(fmtlib_fmt
       GIT_REPOSITORY  ${git_url}
       GIT_TAG         ${git_tag}
       CONFIGURE_COMMAND "" # remove configure step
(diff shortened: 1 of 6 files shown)


cmake/Modules/Findfmt.cmake
```diff
@@ -0,0 +1,64 @@
+add_library(fmt INTERFACE IMPORTED)
+
+set(fmt_external_project_name fmtlib_fmt)
```
l4l: Quite misleading, it's fine to duplicate project name, since it a common practice.

cmake/Modules/Findfmt.cmake
```diff
@@ -0,0 +1,64 @@
...
+
+find_path(fmt_INCLUDE_DIR fmt/format.h
+  PATHS ${EP_PREFIX}/src/${fmt_external_project_name}/include)
```
l4l: Not really a good idea, since the user might want to use system-wide installed library

cmake/Modules/Findfmt.cmake
```diff
@@ -0,0 +1,64 @@
+add_library(fmt INTERFACE IMPORTED)
+
+set(fmt_external_project_name fmtlib_fmt)
+
+find_path(fmt_INCLUDE_DIR fmt/format.h
```
l4l: Where does it come from, if the library haven't been downloaded yet?

libs/logger/logger.hpp
```diff
@@ -79,7 +83,7 @@ namespace logger {
               const std::string &format,
               const Args &... args) const {
        if (shouldLog(level)) {
-         logInternal(level, format); // TODO perform the actual formatting here
+         logInternal(level, fmt::format(format, args...));
```
l4l: Additional dependency for just one line :( probably ok, but still sad

cmake/Modules/Findfmt.cmake
```diff
@@ -0,0 +1,64 @@
...
+    math(EXPR ver_patch ${CMAKE_MATCH_3})
+    set(fmt_VERSION "${ver_major}.${ver_minor}.${ver_patch}")
+  endif ()
+endif ()
+
+if (NOT DEFINED fmt_VERSION OR fmt_VERSION VERSION_LESS fmt_FIND_VERSION)
+  message(STATUS "Package 'fmt' of version ${fmt_FIND_VERSION} not found. "
+          "Will download it from git repo.")
+
+  set(git_url https://github.com/fmtlib/fmt.git)
```
l4l: It's a common practice to use UPPER_CASE for cmake vars, except names (e.g lib name)

cmake/Modules/Findfmt.cmake
```diff
@@ -0,0 +1,64 @@
...
+      )
+
+  externalproject_add(${fmt_external_project_name}
+      GIT_REPOSITORY  ${git_url}
+      GIT_TAG         ${git_tag}
+      CONFIGURE_COMMAND "" # remove configure step
+      BUILD_COMMAND "" # remove build step
+      INSTALL_COMMAND "" # remove install step
+      TEST_COMMAND "" # remove test step
+      UPDATE_COMMAND "" # remove update step
```
l4l: Most of it empty by default, so might be removed

cmake/Modules/Findfmt.cmake
```diff
@@ -0,0 +1,64 @@
...
+
+set(fmt_external_project_name fmtlib_fmt)
+
+find_path(fmt_INCLUDE_DIR fmt/format.h
+  PATHS ${EP_PREFIX}/src/${fmt_external_project_name}/include)
+mark_as_advanced(fmt_INCLUDE_DIR)
+
+# read the fmt version stored in fmt/core.h
+unset(fmt_VERSION)
+set(fmt_version_file_path ${fmt_INCLUDE_DIR}/fmt/core.h)
```
l4l: That makes me cry :( Maybe move to conan? or at least to hunter

cmake/Modules/Findspdlog.cmake
```diff
@@ -1,35 +1,63 @@
...
+      )
 
-if (NOT SPDLOG_FOUND)
-  externalproject_add(gabime_spdlog
-      GIT_REPOSITORY  ${URL}
-      GIT_TAG         ${VERSION}
+  externalproject_add(${spdlog_external_project_name}
+      GIT_REPOSITORY  ${git_url}
+      GIT_TAG         ${git_tag}
+      GIT_SHALLOW     1
```
l4l: useful for fmtlib as well

cmake/Modules/Findfmt.cmake
MBoldyrev: According to [the docs](https://cmake.org/cmake/help/latest/command/find_path.html), the system standard paths are searched as well.

cmake/Modules/Findfmt.cmake
MBoldyrev: It is searched in the system too.

libs/logger/logger.hpp
MBoldyrev: That is for the formatting we use throughout the project, and it is reasonable to keep it.

cmake/Modules/Findfmt.cmake
MBoldyrev: In a separate PR, please!

cmake/Modules/Findfmt.cmake
MBoldyrev: In [the docs](https://cmake.org/cmake/help/latest/module/ExternalProject.html) I can see that the default is to use the external project's CMake configuration. Both added libs use CMake, and both are used as header-only, so we have to explicitly disable the steps we do not need.

cmake/Modules/Findspdlog.cmake
```diff
@@ -1,27 +1,45 @@
 add_library(spdlog INTERFACE IMPORTED)
 
-find_path(spdlog_INCLUDE_DIR spdlog/spdlog.h)
+unset(spdlog_INCLUDE_DIR CACHE)
```
lebdron: Why is is required for spdlog, and not for fmt?
>If the file in a directory is found the result is stored in the variable and the search will not be repeated unless the variable is cleared. If nothing is found, the result will be <VAR>-NOTFOUND, and the search will be attempted again the next time find_path is invoked with the same variable.

Please explain why is it required.

libs/logger/logger.hpp
```diff
@@ -20,6 +20,10 @@ auto operator<<(StreamType &os, const T &object)
   return os << object.toString();
 }
 
+#include <fmt/format.h>
+#include <fmt/ostream.h>
+#include <boost/optional.hpp>
```
lebdron: `optional` include can be removed.

cmake/Modules/Findspdlog.cmake
MBoldyrev: It was left unintentionally after debugging.

PR #2043, opened 2019-01-21
Fix saving rejected txs hashes for failed atomic batches

igor-egorov:

This pr will be a part of RC2 hotfix release.
The #2033 should be merged before this pr. Then this pr has to be rebased to master.

### Description of the Change

When an atomic batch arrived with one stateful valid and one stateful invalid transaction, only one tx's hash was saved to block as rejected (stateful invalid).
The second tx (stateful valid) was not saved as rejected too.

### Benefits

Correct operating.

### Possible Drawbacks

The area for improvements is initialization of fail reason for valid transactions which are within the same atomic batch as a tranasction that did not pass stateful validation. 

### Usage Examples or Tests

```
stateful_validator_test
```

### DIFF
--- a/irohad/validation/impl/stateful_validator_impl.cpp
+++ b/irohad/validation/impl/stateful_validator_impl.cpp
@@ -82,10 +82,12 @@ namespace iroha {
                 transactions_errors_log.emplace_back(
                     validation::TransactionError{
                         tx.hash(),
+                        // TODO igor-egorov 22.01.2019 IR-245 add a separate
+                        // error code for failed batch case
                         validation::CommandError{
                             "",
                             1,  // internal error code
-                            "The other transaction failed the batch",
+                            "Another transaction failed the batch",
                             true,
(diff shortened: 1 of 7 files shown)


irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -75,6 +75,21 @@ namespace iroha {
...
+          } else {
+            auto failed_tx_hash = transactions_errors_log.back().tx_hash;
+            for (const auto &tx : batch) {
+              if (tx.hash() != failed_tx_hash) {
+                transactions_errors_log.emplace_back(
+                    validation::TransactionError{
+                        tx.hash(),
+                        validation::CommandError{
+                            "",
+                            1,  // internal error code
```
Akvinikym: I think, there must be a separate error code for such a reason, as it is hard for client to understand, what happened. Maybe a ToDo for adding new code?

PR #2046, opened 2019-01-22
Make MST expiration time configurable

luckychess:
### Description of the Change
Pull request makes MST expiration time configurable. 

### Benefits
Now Iroha administrators can use `mst_expiration_time` config parameter as they want. Before that this limit was 24 hours and came indirectly from field validator.

### Possible Drawbacks 
Configs became incompatible with old Iroha versions.


irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -28,9 +31,15 @@ namespace iroha {
                        });
   }
 
-  bool DefaultCompleter::operator()(const DataType &tx,
+  bool DefaultCompleter::operator()(const DataType &batch,
```
muratovv: nice catch

irohad/multi_sig_transactions/state/mst_state.hpp
```diff
@@ -54,13 +54,24 @@ namespace iroha {
   };
 
   /**
-   * Class provides the default behavior for the batch completer:
-   * complete, if all transactions have at least quorum number of signatures
+   * Class provides the default behavior for the batch completer.
+   * Complete, if all transactions have at least quorum number of signatures,
+   * complete, if all transactions are not expired.
```
muratovv: Maybe relate "complete, if all" with "and"?

test/framework/integration_framework/iroha_instance.cpp
```diff
@@ -33,6 +33,8 @@ namespace integration_framework {
         proposal_delay_(1h),
         // not required due to solo consensus
         vote_delay_(0ms),
+        // let's use the same value as in the default config
```
muratovv: It may be changed in the default config. Better to describe what is `1440`

test/module/irohad/multi_sig_transactions/storage_test.cpp
```diff
@@ -9,27 +9,17 @@
...
+auto log_ = logger::log("MstStorageTest");
 
 class StorageTest : public testing::Test {
  public:
   StorageTest() : absent_peer_key("absent") {}
 
   void SetUp() override {
-    storage = std::make_shared<MstStorageStateImpl>(
-        std::make_shared<StorageTestCompleter>());
+    completer_ = std::make_shared<DefaultCompleter>(std::chrono::minutes(0));
```
muratovv: I think better to provide test compliter for testing business-logic and make a test for default compliter.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -28,9 +31,15 @@ namespace iroha {
...
                                     const TimeType &time) const {
-    return false;
+    return std::any_of(batch->transactions().begin(),
+                       batch->transactions().end(),
+                       [&](const auto &tx) {
+                         return tx->createdTime()
+                             + expiration_time_ / std::chrono::milliseconds(1)
+                             < time;
```
MBoldyrev: That is not a problem of your PR, but it is not evident, which time scale is used in the `TimeType`. @muratovv, should we place a TODO to make such conversions in a standard way?

irohad/multi_sig_transactions/transport/mst_transport_grpc.hpp
```diff
@@ -41,6 +42,7 @@ namespace iroha {
           std::shared_ptr<shared_model::interface::TransactionBatchFactory>
               transaction_batch_factory,
           std::shared_ptr<iroha::ametsuchi::TxPresenceCache> tx_presence_cache,
+          std::shared_ptr<DefaultCompleter> mst_completer,
```
MBoldyrev: why not
```suggestion
          std::shared_ptr<Completer> mst_completer,
```

test/module/irohad/multi_sig_transactions/CMakeLists.txt
```diff
@@ -19,6 +19,13 @@ target_link_libraries(storage_test
     shared_model_interfaces_factories
     )
 
+AddTest(completer_test completer_test.cpp)
+target_link_libraries(completer_test
+    mst_state
+    shared_model_proto_builders
```
muratovv: Seems we have to use mocks instead.

test/module/irohad/multi_sig_transactions/storage_test.cpp
```diff
@@ -13,12 +13,23 @@ using namespace iroha;
 
 auto log_ = logger::log("MstStorageTest");
 
+class StorageTestCompleter : public DefaultCompleter {
```
muratovv: Looks like code duplication. I suggest to refactor it with one class in a shared header file.

irohad/main/iroha_conf_loader.hpp
```diff
@@ -103,6 +104,8 @@ inline rapidjson::Document parse_iroha_config(const std::string &conf_path) {
                    ac::no_member_error(mbr::MstSupport));
   ac::assert_fatal(doc[mbr::MstSupport].IsBool(),
                    ac::type_error(mbr::MstSupport, kBoolType));
+  ac::assert_fatal(doc[mbr::MstExpirationTime].IsUint(),
```
l4l: Doesn't have to be fatal

MBoldyrev: Not agreed - the type check has to be fatal, but the issue here is that the presence of this key - value pair is not checked and hence reported correctly. We should either treat it as optional and then check the type only when it is set, or demand its presence explicitly and fail with a corresponding message when it is not set.

Warchant: IMO should not be fatal, instead - some default value should be used

PR #2049, opened 2019-01-23
Fix/itf fake peer block storage safety

MBoldyrev:
### Description of the Change

This change makes the Block Storage of ITF Fake Peers thread-safe. It is necessary due to the shared usage of these objects.

Also, the logging of using peers is removed from block storage, as it introduced too much complexity and was not thread-safe.

### Benefits

Bug fixed.

### Possible Drawbacks 

Synchronization slows down execution.

### DIFF
--- a/test/framework/integration_framework/fake_peer/block_storage.cpp
+++ b/test/framework/integration_framework/fake_peer/block_storage.cpp
@@ -5,11 +5,11 @@
 
 #include "framework/integration_framework/fake_peer/block_storage.hpp"
 
+#include <mutex>
+
 #include <boost/algorithm/string/join.hpp>
 #include <boost/range/adaptor/map.hpp>
 #include <boost/range/algorithm/max_element.hpp>
-#include <boost/thread/lock_guard.hpp>
-#include <boost/thread/shared_lock_guard.hpp>
 #include "backend/protobuf/block.hpp"
 #include "framework/integration_framework/fake_peer/fake_peer.hpp"
 
@@ -41,7 +41,7 @@ namespace integration_framework {
           log_(logger::log("Fake peer block storage")) {}
 
     void BlockStorage::storeBlock(const BlockPtr &block) {
-      boost::lock_guard<boost::shared_mutex> lock(block_maps_mutex_);
+      std::lock_guard<std::shared_timed_mutex> lock(block_maps_mutex_);
       if (emplaceCheckingOverwrite(blocks_by_height_, block->height(), block)) {
         log_->warn("Overwriting block with height {}.", block->height());
       }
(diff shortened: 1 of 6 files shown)


test/framework/integration_framework/fake_peer/block_storage.hpp
```diff
@@ -25,33 +26,28 @@ namespace integration_framework {
...
-
-      // claim that a fake peer is not using this storage any more.
-      void claimNotUsingPeer(const std::shared_ptr<FakePeer> &peer);
-
      private:
-      logger::Logger getLogger() const;
-
       std::unordered_map<HeightType, BlockPtr> blocks_by_height_;
       std::unordered_map<HashType, BlockPtr, HashType::Hasher> blocks_by_hash_;
+      mutable boost::shared_mutex block_maps_mutex_;
```
lebdron: Same for `std::shared_timed_mutex` and locks.

test/framework/integration_framework/fake_peer/block_storage.cpp
```diff
@@ -8,6 +8,8 @@
 #include <boost/algorithm/string/join.hpp>
 #include <boost/range/adaptor/map.hpp>
 #include <boost/range/algorithm/max_element.hpp>
+#include <boost/thread/lock_guard.hpp>
+#include <boost/thread/shared_lock_guard.hpp>
```
igor-egorov: Probably boost is not needed here and std primitives can be used here.

test/framework/integration_framework/fake_peer/block_storage.hpp
```diff
@@ -25,33 +26,28 @@ namespace integration_framework {
 
     class BlockStorage final {
      public:
-      using BlockPtr = std::shared_ptr<shared_model::proto::Block>;
+      using BlockPtr = std::shared_ptr<const shared_model::proto::Block>;
```
lebdron: https://www.fluentcpp.com/2019/01/22/pitfalls-of-aliasing-a-pointer-in-modern-cpp/
Please do not use generic `Ptr` suffix, as it can mean raw pointer, `shared_ptr`, and `unique_ptr`.

igor-egorov:
`pr-head` has to pass

test/framework/integration_framework/fake_peer/block_storage.hpp
```diff
@@ -25,33 +26,28 @@ namespace integration_framework {
...
       using HeightType = shared_model::interface::types::HeightType;
       using HashType = shared_model::crypto::Hash;
 
+      BlockStorage();
```
lebdron: Maybe pass `log` as a dependency, similar to `irohad` components?

test/framework/integration_framework/fake_peer/block_storage.hpp
```diff
@@ -25,33 +26,28 @@ namespace integration_framework {
...
+      BlockStorage(const BlockStorage &);
```
lebdron: Are these constructors required to be defined explicitly?

MBoldyrev: That is because the mutex class member, which cannot be copied.

test/framework/integration_framework/fake_peer/block_storage.hpp
MBoldyrev: I would leave it for the new hierarchical logger, which is now in development.

PR #2050, opened 2019-01-23
New logger integration to irohad: part 1

MBoldyrev:

### Description of the Change

This is a part of new logger integration to irohad. As the complete integration patch is too big for a single PR (+743 -462), it is split in three parts (#2050, #2051, #2052) according to the source code directory structure. This PR includes changes made to `irohad/{ametsuchi,consensus}`. You can use the unified patch 496b2e9d0 (which was exactly equal to these three PRs put together when they were opened) to build Irohad with the new logger.

### Benefits

### DIFF
--- a/irohad/ametsuchi/impl/postgres_query_executor.cpp
+++ b/irohad/ametsuchi/impl/postgres_query_executor.cpp
@@ -33,6 +33,8 @@
 #include "interfaces/queries/get_transactions.hpp"
 #include "interfaces/queries/query.hpp"
 #include "interfaces/queries/tx_pagination_meta.hpp"
+#include "logger/logger.hpp"
+#include "logger/logger_manager.hpp"
 
 using namespace shared_model::interface::permissions;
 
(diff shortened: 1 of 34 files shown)


irohad/ametsuchi/impl/postgres_query_executor.cpp
```diff
@@ -251,9 +253,10 @@ namespace iroha {
                    pending_txs_storage_,
                    std::move(converter),
                    response_factory,
-                   perm_converter),
+                   perm_converter,
+                   log_manager->getChild("Visitor")->getLogger()),
```
muratovv: I guess we should pass parameters for objects which we create via ctor. So better to pass a factory which returns parameters required for child logger instantiation.

Also, I am ok with leaving only todo now.

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -104,7 +107,10 @@ namespace iroha {
 
       return expected::makeValue<std::unique_ptr<TemporaryWsv>>(
           std::make_unique<TemporaryWsvImpl>(
-              std::move(sql), factory_, perm_converter_));
+              std::move(sql),
+              factory_,
+              perm_converter_,
+              log_manager_->getChild("TemporaryWorldStateView")));
```
muratovv: Seems, that all `getChild` invocations should be reworked with another approach.

PR #2053, opened 2019-01-23
Fix segfault in StatusStream

Akvinikym:
### Description of the Change

Because of the misunderstanding of gRPC API, it was assumed that method Write(..) was thread-safe, but it isn't, and it caused a segfault when several threads called it at once. But, several threads should also not be in one lambda of the rxcpp observable, which is also a bug in our code. This is going to be fixed later, so now a mutex was added to ensure Write(..) is called only from one thread at the time.

Another thing fixed in this PR is that unsuccessful Write(..) was not taken into account, and the code tried to call it even after failure (for example, channel to client did not exist anymore), which is the reason for another potential segfault. This was fixed by checking Write(..)'s result.

### Benefits

Fix of the critical bug.

### Possible Drawbacks 

Code becomes slower because of the mutex - it must be removed after the mentioned issue is fixed.

### DIFF
--- a/irohad/torii/command_service.hpp
+++ b/irohad/torii/command_service.hpp
@@ -52,7 +52,7 @@ namespace torii {
      */
     virtual rxcpp::observable<
         std::shared_ptr<shared_model::interface::TransactionResponse>>
-    getStatusStream(shared_model::crypto::Hash hash) = 0;
+    getStatusStream(const shared_model::crypto::Hash &hash) = 0;
   };
 
 }  // namespace torii
--- a/irohad/torii/command_service.hpp
+++ b/irohad/torii/command_service.hpp
@@ -52,7 +52,7 @@ namespace torii {
      */
     virtual rxcpp::observable<
         std::shared_ptr<shared_model::interface::TransactionResponse>>
-    getStatusStream(const shared_model::crypto::Hash &hash) = 0;
+    getStatusStream(shared_model::crypto::Hash hash) = 0;
   };
 
 }  // namespace torii
--- a/irohad/torii/impl/command_service_impl.cpp
+++ b/irohad/torii/impl/command_service_impl.cpp
@@ -100,7 +100,7 @@ namespace torii {
 
   rxcpp::observable<
       std::shared_ptr<shared_model::interface::TransactionResponse>>
-  CommandServiceImpl::getStatusStream(const shared_model::crypto::Hash &hash) {
+  CommandServiceImpl::getStatusStream(shared_model::crypto::Hash hash) {
     using ResponsePtrType =
         std::shared_ptr<shared_model::interface::TransactionResponse>;
     auto initial_status = cache_->findItem(hash).value_or([&] {
@@ -113,7 +113,7 @@ namespace torii {
         .start_with(initial_status)
         // select statuses with requested hash
         .filter(
-            [&](auto response) { return response->transactionHash() == hash; })
+            [=](auto response) { return response->transactionHash() == hash; })
         // successfully complete the observable if final status is received.
         // final status is included in the observable
         .template lift<ResponsePtrType>([](rxcpp::subscriber<ResponsePtrType>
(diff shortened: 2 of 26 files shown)


igor-egorov:
The test is missing btw

neewy:
@igor-egorov it is hard to reproduce — it takes from 2 to 10 minutes of variable load and I trust if Andrei approved the PR, as they were trying to verify this yesterday

irohad/torii/command_service.hpp
```diff
@@ -52,7 +52,7 @@ namespace torii {
      */
     virtual rxcpp::observable<
         std::shared_ptr<shared_model::interface::TransactionResponse>>
-    getStatusStream(const shared_model::crypto::Hash &hash) = 0;
```
lebdron: Please revert this change, as it can be fixed without modifications in the interface.

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -113,7 +113,7 @@ namespace torii {
         .start_with(initial_status)
         // select statuses with requested hash
         .filter(
-            [&](auto response) { return response->transactionHash() == hash; })
+            [=](auto response) { return response->transactionHash() == hash; })
```
lebdron: After reverting the argument type, capture can be changed to `[hash]`.

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -161,8 +161,10 @@ namespace torii {
 
   namespace {
     void handleEvents(rxcpp::composite_subscription &subscription,
-                      rxcpp::schedulers::run_loop &run_loop) {
-      while (subscription.is_subscribed() or not run_loop.empty()) {
+                      rxcpp::schedulers::run_loop &run_loop,
+                      bool &stream_failed) {
```
lebdron: Please refactor without passing additional variables to this loop.

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -234,21 +241,32 @@ namespace torii {
...
-                   },
-                   [&] { log_->debug("stream done, {}", client_id); });
+        .subscribe(
+            subscription,
+            [this, &subscription, &response_writer, &client_id, &stream_failed](
+                const auto &response) {
+              // TODO [IR-249] akvinikym 23.01.19: remove the mutex after
+              // ensuring only one thread can be here
+              std::lock_guard<std::mutex> lg{stream_write_mutex_};
+              if (not stream_failed and response_writer->Write(response)) {
```
lebdron: Looks like a candidate for `take_while`.

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -193,6 +195,7 @@ namespace torii {
 
     boost::optional<iroha::protocol::TxStatus> last_tx_status;
     auto rounds_counter{0};
+    auto stream_failed = false;
```
lebdron: It should be possible to refactor without using this variable.

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -234,21 +241,32 @@ namespace torii {
...
+                log_->error("write to stream has failed to client {}",
+                            client_id);
+                stream_failed = true;
+                subscription.unsubscribe();
+              }
+            },
+            [&](std::exception_ptr ep) {
+              log_->error("something bad happened, client_id {}", client_id);
+              stream_failed = true;
+              subscription.unsubscribe();
```
lebdron: This is not required, since the observable will not emit any events after an error.

irohad/torii/impl/command_service_transport_grpc.hpp
```diff
@@ -134,6 +134,7 @@ namespace torii {
 
     rxcpp::observable<ConsensusGateEvent> consensus_gate_objects_;
     const int maximum_rounds_without_update_;
+    std::mutex stream_write_mutex_;
```
lebdron: This should be a local variable in `StatusStream`, since writes to each client should be synchronized.

PR #2064, opened 2019-01-30
Remove single peer ordering service

lebdron:
### Description of the Change
Remove single peer ordering service, since on demand ordering service replaced it.

### Benefits
Less legacy code to support.

### Possible Drawbacks 
None?


lebdron:
CI error was noted by @BulatSaif

PR #2067, opened 2019-01-31
Fix itf fake peers on demand ordering test

MBoldyrev:
### Description of the Change

Rewrote the test...

### Benefits

...to fix an unidentified bug and to improve the test logic.

### DIFF
--- a/test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
+++ b/test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
@@ -25,25 +25,37 @@ namespace integration_framework {
       subscriptions_.emplace_back(
           getFakePeer().getMstStatesObservable().subscribe(
               [this](const auto &message) {
-                this->processMstMessage(message);
+                if (auto alive = fake_peer_wptr_.lock()) {
+                  this->processMstMessage(message);
+                }
               }));
       subscriptions_.emplace_back(
           getFakePeer().getYacStatesObservable().subscribe(
               [this](const auto &message) {
-                this->processYacMessage(message);
+                if (auto alive = fake_peer_wptr_.lock()) {
+                  this->processYacMessage(message);
+                }
               }));
       subscriptions_.emplace_back(
           getFakePeer().getOsBatchesObservable().subscribe(
-              [this](const auto &batch) { this->processOsBatch(batch); }));
+              [this](const auto &batch) {
+                if (auto alive = fake_peer_wptr_.lock()) {
+                  this->processOsBatch(batch);
+                }
+              }));
       subscriptions_.emplace_back(
           getFakePeer().getOgProposalsObservable().subscribe(
               [this](const auto &proposal) {
-                this->processOgProposal(proposal);
+                if (auto alive = fake_peer_wptr_.lock()) {
+                  this->processOgProposal(proposal);
+                }
               }));
       subscriptions_.emplace_back(
           getFakePeer().getBatchesObservable().subscribe(
               [this](const auto &batches) {
-                this->processOrderingBatches(*batches);
+                if (auto alive = fake_peer_wptr_.lock()) {
+                  this->processOrderingBatches(*batches);
+                }
               }));
     }
 
--- a/test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
+++ b/test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
@@ -19,30 +19,32 @@ namespace integration_framework {
       // This code feels like part of constructor, but the use of `this'
       // to call virtual functions from base class constructor seems wrong.
       // Hint: such calls would precede the derived class construction.
-      fake_peer_wptr_ = fake_peer;
+      fake_peer_ = fake_peer;
       log_ = std::move(log);
       // subscribe for all messages
       subscriptions_.emplace_back(
           getFakePeer().getMstStatesObservable().subscribe(
-              [this](const auto &message) {
+              [this, alive = fake_peer](const auto &message) {
                 this->processMstMessage(message);
               }));
       subscriptions_.emplace_back(
           getFakePeer().getYacStatesObservable().subscribe(
-              [this](const auto &message) {
+              [this, alive = fake_peer](const auto &message) {
                 this->processYacMessage(message);
               }));
       subscriptions_.emplace_back(
           getFakePeer().getOsBatchesObservable().subscribe(
-              [this](const auto &batch) { this->processOsBatch(batch); }));
+              [this, alive = fake_peer](const auto &batch) {
+                this->processOsBatch(batch);
+              }));
       subscriptions_.emplace_back(
           getFakePeer().getOgProposalsObservable().subscribe(
-              [this](const auto &proposal) {
+              [this, alive = fake_peer](const auto &proposal) {
                 this->processOgProposal(proposal);
               }));
       subscriptions_.emplace_back(
           getFakePeer().getBatchesObservable().subscribe(
-              [this](const auto &batches) {
+              [this, alive = fake_peer](const auto &batches) {
                 this->processOrderingBatches(*batches);
               }));
     }
@@ -51,14 +53,11 @@ namespace integration_framework {
       for (auto &subscription : subscriptions_) {
         subscription.unsubscribe();
       }
-      fake_peer_wptr_.reset();
+      fake_peer_.reset();
     }
 
     FakePeer &Behaviour::getFakePeer() {
-      auto fake_peer = fake_peer_wptr_.lock();
-      assert(fake_peer && "Fake peer shared pointer is not set!"
-        " Probably the fake peer has gone before the associated behaviour.");
-      return *fake_peer;
+      return *fake_peer_;
     }
 
     logger::LoggerPtr &Behaviour::getLogger() {
--- a/test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
+++ b/test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
@@ -21,40 +21,55 @@ namespace integration_framework {
       // Hint: such calls would precede the derived class construction.
       fake_peer_wptr_ = fake_peer;
       log_ = std::move(log);
+      std::weak_ptr<Behaviour> weak_this = shared_from_this();
       // subscribe for all messages
       subscriptions_.emplace_back(
           getFakePeer().getMstStatesObservable().subscribe(
-              [this](const auto &message) {
-                if (auto alive = fake_peer_wptr_.lock()) {
-                  this->processMstMessage(message);
+              [weak_this,
+               weak_fake_peer = fake_peer_wptr_](const auto &message) {
+                auto me_alive = weak_this.lock();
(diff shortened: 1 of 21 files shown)


test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -318,80 +330,98 @@ TEST_F(FakePeerExampleFixture,
...
-  };
-
   // Create the tx:
   const auto tx = complete(
       baseTx(kAdminId).transferAsset(kAdminId, kUserId, kAssetId, "tx1", "1.0"),
       kAdminKeypair);
-  const auto hash = tx.reducedHash();
 
-  std::atomic_flag got_proposal_from_main_peer = ATOMIC_FLAG_INIT;
+  auto state_ready = std::make_shared<std::atomic_bool>(false);
```
l4l: I guess it's possible just passing a reference, w/o shared_ptr,

test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -318,80 +330,98 @@ TEST_F(FakePeerExampleFixture,
...
+        std::shared_ptr<const shared_model::proto::Proposal> proposal;
+        valid_factory
+            .createProposal(round.block_round,
+                            iroha::time::now(),
+                            std::vector<shared_model::proto::Transaction>({tx}))
+            .match(
+                [&proposal](iroha::expected::Value<std::unique_ptr<
+                                shared_model::interface::Proposal>> &v) {
+                  auto p1 =
+                      std::static_pointer_cast<shared_model::proto::Proposal>(
```
l4l: D: may you a bit elaborate what is happening here?

test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -318,80 +330,98 @@ TEST_F(FakePeerExampleFixture,
...
+  auto &itf = prepareState();
+  state_ready->store(true, std::memory_order_relaxed);
   // watch the proposal requests to fake peer
-  bool got_proposal_from_fake_peer = false;
-  fake_peers_.front()->getProposalRequestsObservable().subscribe(
-      [&got_proposal_from_fake_peer](const auto &round) {
-        got_proposal_from_fake_peer = true;
+  itf.getPcsOnCommitObservable()
+      /*
+      .start_with(iroha::synchronizer::SynchronizationEvent{
```
l4l: Why isn't it deleted?

test/integration/acceptance/fake_peer_example_test.cpp
MBoldyrev: Then it would have lifetime issues when the variable gets wiped from stack but still used by rx streams in fake peer on heap.

test/integration/acceptance/fake_peer_example_test.cpp
MBoldyrev: That was a debugging code I forgot to remove, sorry.

test/integration/acceptance/fake_peer_example_test.cpp
MBoldyrev: That was a debugging code I forgot to remove, sorry.

test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -85,26 +88,29 @@ class FakePeerExampleFixture : public AcceptanceFixture {
...
       kAdminKeypair));
-  std::unique_lock<std::mutex> mst_lock(mst_mutex);
-  mst_cv.wait_for(mst_lock, kMstStateWaitingTime, [&got_state_notification] {
-    return got_state_notification.load();
-  });
-  EXPECT_TRUE(got_state_notification.load())
-      << "Reached timeout waiting for MST State.";
+
+  mst_states_observable.timeout(kMstStateWaitingTime)
+      .take(1)
```
lebdron, line 105: Please add a TODO to remove unnecessary waiting time after `on_completed`.

test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -246,115 +253,123 @@ TEST_F(FakePeerExampleFixture, SynchronizeTheRightVersionOfForkedLedger) {
...
+               const std::shared_ptr<shared_model::interface::Block>
+                   &committed_block) {
         const auto valid_hash =
             valid_block_storage->getBlockByHeight(committed_block->height())
                 ->hash()
                 .hex();
         const auto commited_hash = committed_block->hash().hex();
-        if (commited_hash != valid_hash) {
+        if (commited_hash == valid_hash) {
           ASSERT_EQ(commited_hash, valid_hash) << "Wrong block got committed!";
```
lebdron: Looks like conditional in previous line should be removed, otherwise `ASSERT_EQ` will always pass.

test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -246,115 +253,123 @@ TEST_F(FakePeerExampleFixture, SynchronizeTheRightVersionOfForkedLedger) {
...
+        valid_factory
+            .createProposal(round.block_round,
+                            iroha::time::now(),
+                            std::vector<shared_model::proto::Transaction>({tx}))
+            .match(
+                [&proposal](iroha::expected::Value<std::unique_ptr<
+                                shared_model::interface::Proposal>> &v) {
+                  // casting std::shared_ptr<shared_model::interface::Proposal>
+                  // to std::shared_ptr<const shared_model::proto::Proposal>
+                  proposal = std::const_pointer_cast<
```
lebdron:
```suggestion
                  proposal = std::static_pointer_cast<
```
Because static cast allows to add `const`.

test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
```diff
@@ -25,25 +25,37 @@ namespace integration_framework {
       subscriptions_.emplace_back(
           getFakePeer().getMstStatesObservable().subscribe(
               [this](const auto &message) {
-                this->processMstMessage(message);
+                if (auto alive = fake_peer_wptr_.lock()) {
```
lebdron: Will `this` object outlive the observable? If there are no such guarantees, probably `fake_peer_wptr_` has to be captured by value.
Same below.

MBoldyrev: great catch, thanks!

test/framework/integration_framework/fake_peer/fake_peer.cpp
```diff
@@ -124,6 +124,12 @@ namespace integration_framework {
       yac_transport_->subscribe(yac_network_notifier_);
     }
 
+    FakePeer::~FakePeer() {
+      if (behaviour_) {
+        behaviour_->absolve();
```
luckychess, line 129: As I can see `Behaviour` class calls `absolve` by itself in its destructor. Could you explain why we should do it manually here?

test/framework/integration_framework/fake_peer/proposal_storage.cpp
```diff
@@ -114,5 +116,23 @@ namespace integration_framework {
 
     }
 
+    ProposalStorage &ProposalStorage::setDefaultProvider(DefaultProvider provider) {
+      std::atomic_store_explicit(
+          &default_provider_,
+          std::make_shared<DefaultProvider>(std::move(provider)),
+          std::memory_order_release);
```
luckychess: I would suggest to add some comments about why memory order here and below is important.

test/integration/acceptance/fake_peer_example_test.cpp
```diff
@@ -23,32 +25,33 @@ using namespace shared_model::interface::permissions;
 using ::testing::_;
 using ::testing::Invoke;
 
-static constexpr std::chrono::seconds kMstStateWaitingTime(10);
-static constexpr std::chrono::seconds kSynchronizerWaitingTime(10);
-static constexpr std::chrono::seconds kProposalWaitingTime(1);
+static constexpr std::chrono::seconds kMstStateWaitingTime(20);
```
luckychess, line 27: Pretty suspicious change.

test/framework/integration_framework/fake_peer/fake_peer.cpp
MBoldyrev: We do it to ensure it will stop its operation on this peer. The destructor may not be called if the behaviour is managed by a shared pointer stored elsewhere.

test/framework/integration_framework/fake_peer/proposal_storage.cpp
MBoldyrev: It is a general case of release-acquire model with nothing specific.

test/framework/integration_framework/fake_peer/behaviour/behaviour.cpp
```diff
@@ -21,29 +21,56 @@ namespace integration_framework {
...
       log_ = std::move(log);
+      std::weak_ptr<Behaviour> weak_this = shared_from_this();
       // subscribe for all messages
       subscriptions_.emplace_back(
           getFakePeer().getMstStatesObservable().subscribe(
-              [this](const auto &message) {
-                this->processMstMessage(message);
+              [weak_this,
+               weak_fake_peer = fake_peer_wptr_](const auto &message) {
+                auto me_alive = weak_this.lock();
```
muratovv: Code of locking and checking is duplicated 5 times. I think better to refactor it.

test/framework/integration_framework/iroha_instance.cpp
```diff
@@ -16,6 +16,8 @@
 
 using namespace std::chrono_literals;
 
+static constexpr std::chrono::milliseconds kMstEmissionPeriod = 100ms;
```
muratovv, line 19: I guess better to move default constants from itf to a separate header.

test/framework/integration_framework/fake_peer/fake_peer.cpp
luckychess: But it will be called later anyway, what's the problem?

test/framework/integration_framework/fake_peer/proposal_storage.cpp
luckychess: Ok, what will happen if I omit memory_order arguments?

test/integration/acceptance/fake_peer_example_test.cpp
MBoldyrev: Since rxcpp `timeout` works well now, we can loosen the time limits, that were previously lowered to reduce the test execution time.

test/framework/integration_framework/fake_peer/proposal_storage.cpp
MBoldyrev: You are right that this optimization is excessive in tests. I believe no noticeable overhead will be added if the default memory order is restored.

test/framework/integration_framework/fake_peer/fake_peer.cpp
MBoldyrev: If we do not call `absolve()` here, a behaviour (if it overlives the peer) will try to lock its weak pointer and will not succeed. This line absolves its duties on fake peer exit.

PR #2069, opened 2019-01-31
Feature/iroha build time graph

Stayer:
### Description of the Change

[[DOPS-116]](https://soramitsu.atlassian.net/browse/DOPS-116)
This PR allows to view compilation time per target (g++ run time) per each build for all branches, and compare this time to `develop` tree. After that we have csv file with time difference in seconds and %

### Benefits

Developers can detail view and compare time to improve their code.

### Possible Drawbacks 

1. For now it is not possible to print `cmake` output to jenkins console (using tee gives some duplicates in the result file)
2. When "fuzzingEnabled" is true, there is no results

See attachment [diff.csv.zip](https://github.com/hyperledger/iroha/files/2817751/diff.csv.zip)


.jenkinsci/linux-post-step.groovy
```diff
@@ -16,6 +17,17 @@ def linuxPostStep() {
           echo "Build is not SUCCESS! See core dumps at: https://nexus.iroha.tech/repository/artifacts/iroha/coredumps/${dumpsFileName}"
         }
       }
+
+      // handling build time results
+      sh(".jenkinsci/helpers/exportBuildTime.py buildTimeResult.txt")
+      zip archive: true, dir: '', glob: 'buildTimeResult.csv', zipFile: 'buildTimeMeasurement.zip'
+      archiveArtifacts artifacts: 'buildTimeMeasurement.zip'
+
+      copyArtifacts(projectName: 'feature%2Firoha_build_time_graph', filter: 'buildTimeMeasurement.zip', target: 'buildTimeMeasurement-develop');
```
Stayer: this will be changed to develop

Jenkinsfile
```diff
@@ -17,10 +17,11 @@ properties([parameters([
   choice(choices: '26\n25\n24\n23\n22\n21\n20\n19\n18\n17\n16\n15\n14', description: 'Android Bindings ABI Version', name: 'ABABIVersion'),
   choice(choices: 'Release\nDebug', description: 'Android bindings build type', name: 'ABBuildType'),
   choice(choices: 'arm64-v8a\narmeabi-v7a\narmeabi\nx86_64\nx86', description: 'Android bindings platform', name: 'ABPlatform'),
-  booleanParam(defaultValue: true, description: 'Build docs', name: 'Doxygen'),
-  booleanParam(defaultValue: true, description: 'Sanitize address;leak', name: 'sanitize'),
+  booleanParam(defaultValue: false, description: 'Build docs', name: 'Doxygen'),
```
Stayer: it will be switched back after review

Jenkinsfile
```diff
@@ -4,7 +4,7 @@ properties([parameters([
   booleanParam(defaultValue: true, description: '', name: 'x86_64_linux'),
   booleanParam(defaultValue: false, description: '', name: 'armv7_linux'),
   booleanParam(defaultValue: false, description: '', name: 'armv8_linux'),
-  booleanParam(defaultValue: true, description: '', name: 'x86_64_macos'),
+  booleanParam(defaultValue: false, description: '', name: 'x86_64_macos'),
```
Stayer: it will be switched back after review

.jenkinsci/linux-post-step.groovy
```diff
@@ -25,8 +37,8 @@ def linuxPostStep() {
       }
     }
     finally {
-      def cleanup = load ".jenkinsci/docker-cleanup.groovy"
-      cleanup.doDockerCleanup()
+      // def cleanup = load ".jenkinsci/docker-cleanup.groovy"
```
bakhtin: Why?

bakhtin:
Why the build has failed (pr-head)?

.jenkinsci/linux-post-step.groovy
bakhtin: Will this step fail if there is no reports in `develop` yet?

BulatSaif:
@Stayer , can you add this feature to new CI also?

.jenkinsci/debug-build.groovy
```diff
@@ -66,7 +66,7 @@ def doDebugBuild(coverageEnabled=false) {
     + " -v /tmp/${GIT_COMMIT}-${BUILD_NUMBER}:/tmp/${GIT_COMMIT}") {
 
     def scmVars = checkout scm
-    def cmakeOptions = ""
+    def cmakeOptions = "-DCMAKE_CXX_COMPILER=${WORKSPACE}/.jenkinsci/helpers/build.sh"
```
BulatSaif, line 69: are you sure we do not need `DCMAKE_C_COMPILER` also ?

BulatSaif: how it will work if `fuzzingEnabled` options is set to true?
because it adds
`cmakeOptions += " -DCMAKE_C_COMPILER=clang-6.0 -DCMAKE_CXX_COMPILER=clang++-6.0 -DFUZZING=ON "`

.jenkinsci/linux-post-step.groovy
```diff
@@ -16,6 +17,20 @@ def linuxPostStep() {
           echo "Build is not SUCCESS! See core dumps at: https://nexus.iroha.tech/repository/artifacts/iroha/coredumps/${dumpsFileName}"
         }
       }
+
+      // handling build time results
+      if (currentBuild.currentResult == "SUCCESS" && !fuzzingEnabled) {
```
BulatSaif: Add this :
`def fuzzingEnabled = params.fuzzing`
or just use `params.fuzzing`
Keep in mind it can be `null`, in some case
>  // params are always null unless job is started
>  // this is the case for the FIRST build only.	 
 // So just set this to same value as default.	
  // This is a known bug. See https://issues.jenkins-ci.org/browse/JENKINS-41929

.jenkinsci/debug-build.groovy
Stayer: I think yes, because compiler tracks only "g++" operations

.jenkinsci/linux-post-step.groovy
Stayer: Ok, will fix this

Stayer:
> @Stayer , can you add this feature to new CI also?

Ok

.jenkinsci-new/builders/x64-linux-build-steps.groovy
```diff
@@ -54,7 +54,7 @@ def buildSteps(int parallelism, List compilerVersions, String build_type, boolea
     cmakeBooleanOption = [ (true): 'ON', (false): 'OFF' ]
     platform = sh(script: 'uname -m', returnStdout: true).trim()
     cmakeBuildOptions = ""
-    cmakeOptions = ""
+    cmakeOptions = " -DCMAKE_CXX_COMPILER=${WORKSPACE}/.jenkinsci-new/helpers/build.sh "
```
BulatSaif: New ci can build with multiple compiler is already define DCMAKE_CXX_COMPILER below:
https://github.com/hyperledger/iroha/blob/4edfdf7ccaa0449f9b96dbfa84a9586ee1c63398/.jenkinsci-new/builders/x64-linux-build-steps.groovy#L95

You can set compiler binary name in this file, (and please add comment to it):
https://github.com/hyperledger/iroha/blob/4edfdf7ccaa0449f9b96dbfa84a9586ee1c63398/.jenkinsci-new/utils/vars.groovy#L12

BulatSaif:
General comment form Nikoly to last PR in new CI: 
>Per each module or file we might need to include license header as described here:
>https://wiki.hyperledger.org/community/copyright-and-license-policy
https://www.hyperledger.org/about/charter №13 
>you can reuse code from here
hyperledger/iroha#1944

>Also, per each file it is desired to see what does it do or store. Can we please add just a line or something so that others can quickly understand what a purpose some file serves?

BulatSaif:
as I understand it will no work with clang?

.jenkinsci-new/build.groovy
```diff
@@ -13,7 +13,7 @@ def cmakeConfigure(String buildDir, String cmakeOptions, String sourceTreeDir=".
 }
 
 def cmakeBuild(String buildDir, String cmakeOptions, int parallelism) {
-  sh "cmake --build ${buildDir} ${cmakeOptions} -- -j${parallelism}"
+  sh "cmake --build ${buildDir} ${cmakeOptions} -- -j${parallelism} | sponge buildTimeResult.txt"
```
BulatSaif, line 16: This fails on mac: `sponge: command not found`
you can change it to 

```
def cmakeBuild(String buildDir, String cmakeOptions, int parallelism, String pipe=' ') {
sh "cmake --build ${buildDir} ${cmakeOptions} -- -j${parallelism}  ${pipe}"
```

Where can be empty string or:
```
  def pipe = '  | sponge buildTimeResult.txt '
```

Stayer:
> as I understand it will no work with clang?

yes, you are

.jenkinsci-new/utils/vars.groovy
```diff
@@ -10,7 +10,7 @@
 
 def compilerMapping () {
   return ['gcc5': ['cxx_compiler':'g++-5', 'cc_compiler':'gcc-5'],
-          'gcc7' : ['cxx_compiler':'g++-7', 'cc_compiler':'gcc-7'],
+          'gcc7' : ['cxx_compiler':"${WORKSPACE}/.jenkinsci-new/helpers/build.sh", 'cc_compiler':'gcc-7',],
```
BulatSaif: Default compiler is gcc5, in gcc7 it builds just last build before merge to trunk(starts by hand)

.jenkinsci/helpers/build_cmake.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
BulatSaif, line 1: There `build_cmake.sh` is used? I find referents  only for `build.sh`

.jenkinsci-new/builders/x64-linux-build-steps.groovy
```diff
@@ -132,6 +132,20 @@ def buildSteps(int parallelism, List compilerVersions, String build_type, boolea
 def successPostSteps(scmVars, boolean packagePush, String dockerTag, List environment) {
   stage('Linux success PostSteps') {
     withEnv(environment) {
+
+      // handling build time results
+      if (currentBuild.currentResult == "SUCCESS" && !params.fuzzing) {
+        sh(".jenkinsci-new/helpers/exportBuildTime.py buildTimeResult.txt")
+        zip archive: true, dir: '', glob: 'buildTimeResult.csv', zipFile: 'buildTimeMeasurement.zip'
+        archiveArtifacts artifacts: 'buildTimeMeasurement.zip'
```
BulatSaif, line 142: New CI can run different sets of compiler and if it runs only with `clang` it will fail in `successPostSteps` because `buildTimeMeasurement.zip`  do not exist. 
so you need to check which compiler is used or just:
```suggestion
        archiveArtifacts artifacts: 'buildTimeMeasurement.zip', allowEmptyArchive: true
```

.jenkinsci-new/builders/x64-linux-build-steps.groovy
```diff
@@ -132,6 +132,20 @@ def buildSteps(int parallelism, List compilerVersions, String build_type, boolea
...
+
+        copyArtifacts(projectName: 'develop', filter: 'buildTimeMeasurement.zip', target: 'buildTimeMeasurement-develop');
+        unzip zipFile: 'buildTimeMeasurement-develop/buildTimeMeasurement.zip', dir: 'buildTimeMeasurement-develop'
+        sh ".jenkinsci-new/helpers/analyzeBuildTime.py buildTimeMeasurement-develop/buildTimeResult.csv buildTimeResult.csv"
+        zip archive: true, dir: '', glob: 'diff.csv', zipFile: 'diff.zip'
+        archiveArtifacts artifacts: 'diff.zip'
```
BulatSaif, line 148:
```suggestion
        archiveArtifacts artifacts: 'diff.zip', allowEmptyArchive: true
```

PR #2070, opened 2019-02-01
ITF fake peer: applying new aliasing rules

MBoldyrev:
### Description of the Change

During the review of the fake peers trunk, there arose an issue regarding using aliases. Together with @lebdron, @muratovv and @luckychess we came to a consensus of not hiding top-level type modifiers like `const`, `&`, `*` and smart pointers behind aliases, and also not to alias a type with no templates. The only exception is using an alias as a business logic type.

### Benefits

Standardized approach to using aliases.

### Possible Drawbacks 

As for me, the code has got less readable due to a lot of long and nested types being exposed distract from the higher-level picture.

### DIFF
--- a/test/framework/integration_framework/fake_peer/fake_peer.cpp
+++ b/test/framework/integration_framework/fake_peer/fake_peer.cpp
@@ -183,7 +183,7 @@ namespace integration_framework {
 
     boost::optional<ProposalStorage &> FakePeer::getProposalStorage() const {
       if (proposal_storage_) {
-       return *proposal_storage_;
+        return *proposal_storage_;
       }
       return boost::none;
     }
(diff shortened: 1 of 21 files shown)


test/framework/integration_framework/fake_peer/fake_peer.cpp
```diff
@@ -350,8 +356,10 @@ namespace integration_framework {
                                                         request);
     }
 
-    void FakePeer::sendBatchesForRound(iroha::consensus::Round round,
-                                       std::vector<OsBatchPtr> batches) {
+    void FakePeer::sendBatchesForRound(
+        iroha::consensus::Round round,
+        std::vector<std::shared_ptr<shared_model::interface::TransactionBatch>>
```
muratovv: Seems `using` is appropriate here.

test/framework/integration_framework/fake_peer/fake_peer.hpp
```diff
@@ -157,15 +162,20 @@ namespace integration_framework {
...
       bool sendBlockRequest(const LoaderBlockRequest &request);
 
       size_t sendBlocksRequest(const LoaderBlocksRequest &request);
 
       /// Send the real peer the provided batches for the provided round.
-      void sendBatchesForRound(iroha::consensus::Round round,
-                               std::vector<OsBatchPtr> batches);
+      void sendBatchesForRound(
+          iroha::consensus::Round round,
+          std::vector<
```
muratovv: Same as before

PR #2072, opened 2019-02-01
Feature/logger new impl integration irohad test

MBoldyrev:
### Description of the Change

The change integrates the new logger into the `irohad_test`.

### Benefits


test/system/irohad_test.cpp
```diff
@@ -36,29 +38,51 @@ using namespace std::chrono_literals;
...
+            getIrohadTestLoggerManager()->getChild("KeysManager")->getLogger()),
+        log_(getIrohadTestLoggerManager()->getLogger()) {}
 
   void SetUp() override {
     setPaths();
-    auto config = parse_iroha_config(path_config_.string());
+
+    rapidjson::Document doc;
+    std::ifstream ifs_iroha(path_config_.string());
+    rapidjson::IStreamWrapper isw(ifs_iroha);
```
igor-egorov: This name is OK. Since now I would not expect rejects for that kind of names for "very" local variables.

test/system/irohad_test.cpp
```diff
@@ -36,29 +38,51 @@ using namespace std::chrono_literals;
...
+  }
+  return irohad_test_logger_manager->getChild("IrohadTest");
+}
+
 class IrohadTest : public AcceptanceFixture {
  public:
-  IrohadTest() : kAddress("127.0.0.1"), kPort(50051) {}
+  IrohadTest()
+      : kAddress("127.0.0.1"),
+        kPort(50051),
```
igor-egorov: Can we use our port manager? Can we avoid hard coding the port?

MBoldyrev: Maybe we can, but it's not related to the new logger integration, and actually we do not need it here now.

PR #2073, opened 2019-02-01
Fix Status Stream Performance Degradation

igor-egorov:

### Description of the Change

In some cases, the transaction status stream may not end in 10 seconds even if a final status (e.g. "committed") was already reported.

#### Details

That might happen when there are no transactions committed in the network and the delay between consensus rounds was increased up to 5 seconds.
The status stream gets forced to end in the case when tx status is not updated during two consensus rounds.

### Benefits

Clients that expect immediate stream end after a final tx status receiving would not suffer from redundant delay.

### Possible Drawbacks 

The solution is not elegant enough, because we do a kind of synchronization on a local variable. 
There is no data race because we do the processing in the same thread.

### Usage Examples or Tests

The gist is a repro script - https://gist.github.com/igor-egorov/e07d803b3ef9f17eee48804713140c04

In normal mode, the script will end less than in 20 seconds.

In degraded mode, the script would not finish even in a minute.

### Alternate Designs

Probably, `take_until(status_bus.last())` can be applied somehow.
That approach will avoid the usage of custom synchronization variables.
Unfortunately, take_until prevents the execution of subscribed lambdas when the source stream has ended by the moment of intended processing.

### DIFF
--- a/irohad/torii/impl/command_service_transport_grpc.cpp
+++ b/irohad/torii/impl/command_service_transport_grpc.cpp
@@ -175,7 +175,7 @@ namespace torii {
     rxcpp::schedulers::run_loop rl;
 
     auto current_thread =
-        rxcpp::observe_on_one_worker(rxcpp::schedulers::make_run_loop(rl));
+        rxcpp::synchronize_in_one_worker(rxcpp::schedulers::make_run_loop(rl));
 
     rxcpp::composite_subscription subscription;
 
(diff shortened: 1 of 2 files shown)


Warchant:
You should add that script to continuous integration

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -239,6 +253,12 @@ namespace torii {
             return false;
           }
 
+          if (last_tx_status_received) {
+            // force stream to end because no more tx statuses will arrive.
+            // it is thread safe because of synchronization on current_thread
+            return false;
+          }
+
           log_->debug("status written, {}", client_id);
```
lebdron: I get the following order in a minimal example:
```
a next
a complete
a finally
b next
take while
complete
```
where `a` is status bus, and `b` is consensus events observable.
It means that it will execute `take_while` after `finally`, and this log `status written` should be moved before `last_tx_status_received` conditional.

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -184,18 +184,27 @@ namespace torii {
     auto client_id_format = boost::format("Peer: '%s', %s");
     std::string client_id =
         (client_id_format % context->peer() % hash.toString()).str();
-
+    bool last_tx_status_received = false;
+    auto status_bus = command_service_->getStatusStream(hash)
+                          .subscribe_on(current_thread)
```
lebdron:
```suggestion
```
It might be better to use `.combine_latest(current_thread, consensus_gate_observable)` below instead.

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -184,18 +184,27 @@ namespace torii {
...
+                          })
+                          .publish()
+                          .ref_count();
     auto consensus_gate_observable =
         consensus_gate_objects_
             // a dummy start_with lets us don't wait for the consensus event
             // on further combine_latest
-            .start_with(ConsensusGateEvent{});
+            .start_with(ConsensusGateEvent{})
+            .subscribe_on(current_thread)
```
lebdron:
```suggestion
```
Same as above.

PR #2081, opened 2019-02-07
OnDemandOrderingServiceImpl: reworked emitProposal

MBoldyrev:

### Description of the Change

Replaced a thread-unsafe class method with a standalone local function.

### Benefits

Thread-unsafe logic concentrated in a single place. (Hopefully) better readability.

### DIFF
--- a/irohad/ordering/impl/on_demand_ordering_service_impl.cpp
+++ b/irohad/ordering/impl/on_demand_ordering_service_impl.cpp
@@ -22,6 +22,7 @@
 
 using namespace iroha;
 using namespace iroha::ordering;
+using TransactionBatchType = transport::OdOsNotification::TransactionBatchType;
 
 OnDemandOrderingServiceImpl::OnDemandOrderingServiceImpl(
     size_t transaction_limit,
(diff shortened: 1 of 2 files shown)


irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -106,6 +107,35 @@ OnDemandOrderingServiceImpl::onRequestProposal(consensus::Round round) {
...
+ * @param requested_tx_amount - amount of transactions to get
+ * @param tx_batches_queue - the queue to get transactions from
+ * @return transactions
+ */
+static std::vector<std::shared_ptr<shared_model::interface::Transaction>>
+getTransactions(size_t requested_tx_amount,
+                tbb::concurrent_queue<TransactionBatchType> &tx_batches_queue) {
+  TransactionBatchType batch;
+  std::vector<std::shared_ptr<shared_model::interface::Transaction>> collection;
+  std::unordered_set<std::string> inserted;
```
Akvinikym: Why do you need this set?

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -115,9 +145,18 @@ void OnDemandOrderingServiceImpl::packNextProposals(
     if (it != current_proposals_.end()) {
       log_->debug("proposal found");
       if (not it->second.empty()) {
-        proposal_map_.emplace(round, emitProposal(round));
-        log_->debug("packNextProposal: data has been fetched for {}", round);
-        round_queue_.push(round);
+        log_->debug("Mutable proposal generation for round {}", round);
+        auto txs = getTransactions(transaction_limit_, it->second);
+        if (not txs.empty()) {
+          log_->debug("Number of transactions in proposal = {}", txs.size());
```
Akvinikym: I think, result of `getTransactions` should be logged independently of transactions amount

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
MBoldyrev: It is there to avoid hash duplicates. It was there originally: https://github.com/hyperledger/iroha/blob/0c373792a7a21afca6f5644013457238af01b43f/irohad/ordering/impl/on_demand_ordering_service_impl.cpp#L187

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
MBoldyrev: But if there are no transactions we do not create a proposal, so this log line would be misleading.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -106,6 +107,35 @@ OnDemandOrderingServiceImpl::onRequestProposal(consensus::Round round) {
...
+ */
+static std::vector<std::shared_ptr<shared_model::interface::Transaction>>
+getTransactions(size_t requested_tx_amount,
+                tbb::concurrent_queue<TransactionBatchType> &tx_batches_queue) {
+  TransactionBatchType batch;
+  std::vector<std::shared_ptr<shared_model::interface::Transaction>> collection;
+  std::unordered_set<std::string> inserted;
+
+  size_t collection_size = 0;
+  while (collection_size < requested_tx_amount
```
lebdron:
```suggestion
  while (collection.size() < requested_tx_amount
```

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -115,9 +145,18 @@ void OnDemandOrderingServiceImpl::packNextProposals(
...
+          auto proposal = proposal_factory_->unsafeCreateProposal(
+              round.block_round,
+              iroha::time::now(),
+              std::move(txs) | boost::adaptors::indirected);
+          proposal_map_.emplace(round, std::move(proposal));
+          log_->debug("packNextProposal: data has been fetched for {}", round);
+          round_queue_.push(round);
+        }
       }
       current_proposals_.erase(it);
```
lebdron: Information about lost transactions should be logged, probably by checking `it->second.unsafe_size()`.

PR #2085, opened 2019-02-07
Reworked proposal sharing

MBoldyrev:

### Description of the Change

This change avoids copying of proposal where possible and makes the shared proposal constant. Also it applies the new aliasing policy to `ProposalType`.

### Benefits

Less copying, less chances of unintended changes to shared proposals.

### Possible Drawbacks 

Longer type names due to the new aliasing policy.

### DIFF
--- a/irohad/validation/impl/stateful_validator_impl.cpp
+++ b/irohad/validation/impl/stateful_validator_impl.cpp
@@ -138,10 +138,13 @@ namespace iroha {
                                *batch_parser_);
 
       // Since proposal came from ordering gate it was already validated.
-      // All transactions has been validated as well
+      // All transactions are validated as well
       // This allows for unsafe construction of proposal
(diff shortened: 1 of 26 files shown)


irohad/validation/impl/stateful_validator_impl.cpp
```diff
@@ -138,10 +138,13 @@ namespace iroha {
                                *batch_parser_);
 
       // Since proposal came from ordering gate it was already validated.
-      // All transactions has been validated as well
+      // All transactions are validated as well
       // This allows for unsafe construction of proposal
-      validation_result->verified_proposal = factory_->unsafeCreateProposal(
-          proposal.height(), proposal.createdTime(), valid_txs);
+      validation_result->verified_proposal =
+          std::const_pointer_cast<const shared_model::interface::Proposal>(
```
lebdron: Maybe add a TODO here to refactor proposal factory with const pointer and remove const cast here?

PR #2091, opened 2019-02-12
Pluggable consensus consistency model

MBoldyrev:
[IR-130](https://jira.hyperledger.org/browse/IR-310)

### Description of the Change

`iroha::consensus::yac::SupermajorityChecker` interface now has two implementations: a BFT and a CFT checker. Besides that all the signature-related logic was put away from that interface, and it now operates solely on numbers.

### Benefits

Now a user can choose the consensus consistency model in compile time. Startup-time choosing is coming in a following PR (#1962).

### Usage Example

In `irohad/main/application.cpp` one can set the constant `kConsensusConsistencyModel` to any of the values defined in `irohad/consensus/yac/consistency_model.hpp` and recompile irohad to apply a different consistency model (now set to BFT).


irohad/consensus/yac/CMakeLists.txt
```diff
@@ -1,22 +1,12 @@
...
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
+# Copyright Soramitsu Co., Ltd. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
 #
 
 add_library(supermajority_check
-    impl/supermajority_checker_impl.cpp
+    impl/supermajority_checker_cft.cpp
```
igor-egorov: I suggest renaming of the library to `supermajority_checker`

MBoldyrev:
> Cannot find K const definition for the BFT case :(

It's here:
https://github.com/hyperledger/iroha/blob/90a7eef46c5da4f2fae2baea989dd4e85f90815e/irohad/consensus/yac/impl/supermajority_checker_bft.hpp#L16

irohad/consensus/yac/impl/supermajority_checker_bft.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define IROHA_SUPERMAJORITY_CHECKER_BFT_HPP
+
+#include "consensus/yac/supermajority_checker.hpp"
+
+namespace iroha {
+  namespace consensus {
+    namespace yac {
+
+      namespace detail {
+        constexpr unsigned int kSupermajorityCheckerKfPlus1Bft = 3;
```
muratovv: Pls, add a description to the constant.

irohad/consensus/yac/impl/supermajority_checker_cft.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define IROHA_SUPERMAJORITY_CHECKER_CFT_HPP
+
+#include "consensus/yac/supermajority_checker.hpp"
+
+namespace iroha {
+  namespace consensus {
+    namespace yac {
+
+      namespace detail {
+        constexpr unsigned int kSupermajorityCheckerKfPlus1Cft = 2;
```
muratovv: Maybe put this constant in the base class?

irohad/consensus/yac/impl/supermajority_checker_kf1.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "consensus/yac/supermajority_checker.hpp"
+
+namespace iroha {
+  namespace consensus {
+    namespace yac {
+
+      /**
+       * A generic implementation of N = K * f + 1 supermajority checker.
+       * N is the amount of peers in the network, f is the number of tolerated
+       * faulty peers, and K is a free parameter. Supermajority is achieved when
```
muratovv: Maybe better to write "K is a consistency model parameter".

test/integration/consensus/consensus_sunny_day.cpp
```diff
@@ -27,6 +27,10 @@ using ::testing::Return;
 using namespace iroha::consensus::yac;
 using namespace framework::test_subscriber;
 
+// TODO mboldyrev 13.12.2018 IR- Parametrize the tests with consistency models
```
muratovv: There are two issues: first, missed IR number. Second, I am not sure that the task is required. Because in these tests we want to check how it works in general. So, for checking another consistency model we have to check such things - yac invokes supermajority checker, verify cft supermajority checker and ideally, test application's method which invokes different models.
Another argument to not do it is the following: cft model is one round consensus, but new yac will have two phases. In this perspective, I think we will have two different consensus classes.

test/integration/validation/chain_validator_storage_test.cpp
```diff
@@ -3,25 +3,29 @@
...
 
 #include "ametsuchi/mutable_storage.hpp"
 #include "builders/protobuf/transaction.hpp"
+#include "consensus/yac/supermajority_checker.hpp"
 #include "cryptography/crypto_provider/crypto_defaults.hpp"
 #include "cryptography/default_hash_provider.hpp"
 #include "cryptography/keypair.hpp"
 #include "module/shared_model/builders/protobuf/block.hpp"
 
+// TODO mboldyrev 13.12.2018 IR- Parametrize the tests with consistency models
```
muratovv: sab

test/integration/validation/chain_validator_storage_test.cpp
```diff
@@ -142,8 +151,8 @@ namespace iroha {
   /**
    * @given initialized storage with 4 peers
    * block 1 - initial block with 4 peers
-   * block 2 - signed by supermajority of ledger peers
-   * block 3 - signed by supermajority of ledger peers
+   * block 2 - signed by all ledger peers
```
muratovv: What are the reasons for these changes?

test/module/irohad/consensus/yac/yac_block_storage_test.cpp
```diff
@@ -7,50 +7,63 @@
...
 
+#include "consensus/yac/impl/supermajority_checker_bft.hpp"
 #include "consensus/yac/storage/yac_proposal_storage.hpp"
 #include "logger/logger.hpp"
 
 #include "module/irohad/consensus/yac/yac_test_util.hpp"
 
 using namespace iroha::consensus::yac;
 
+// TODO mboldyrev 13.12.2018 IR- Parametrize the tests with consistency models
```
muratovv: Same issue as for yac tests

test/module/irohad/consensus/yac/yac_fixture.hpp
```diff
@@ -16,6 +16,10 @@
 #include "module/irohad/consensus/yac/mock_yac_timer.hpp"
 #include "module/irohad/consensus/yac/yac_test_util.hpp"
 
+// TODO mboldyrev 13.12.2018 IR- Parametrize the tests with consistency models
```
muratovv: sab

test/integration/validation/chain_validator_storage_test.cpp
MBoldyrev: I want to remove the supermajority logic related assumptions from this test. So previously we assumed that some amount of peers are supermajority, and now we remove that assumption because all peers voting unanimously are supermajority (almost) regardless of its model.

PR #2092, opened 2019-02-12
Add -DUSE_BTF=ON cmake configuration parameter on CI

BulatSaif:
[DOPS-292](https://soramitsu.atlassian.net/browse/DOPS-292)

### Description of the Change
add -DUSE_BTF=ON for cmake configuration on CI

### Benefits

It always` false`,  `true` if  build type "Before merge to trunk"

### DIFF
--- a/Jenkinsfile-new
+++ b/Jenkinsfile-new
@@ -231,7 +231,7 @@ node ('master') {
   }
 
   echo "specialBranch=${specialBranch}, packageBuild=${packageBuild}, pushDockerTag=${pushDockerTag}, packagePush=${packagePush} "
-  echo "testing=${testing}, testList=${testList}, parallelism=${parallelism}"
+  echo "testing=${testing}, testList=${testList}, parallelism=${parallelism}, useBTF=${useBTF}"
   echo "x64linux_compiler_list=${x64linux_compiler_list}"
   echo "mac_compiler_list=${mac_compiler_list}"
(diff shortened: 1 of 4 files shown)


Jenkinsfile-new
```diff
@@ -152,6 +152,7 @@ node ('master') {
   packagePush = false
   specialBranch = false
   parallelism = 0
+  useBTF = false
 
   if (scmVars.GIT_LOCAL_BRANCH in ["master","develop","dev"] || scmVars.CHANGE_BRANCH_LOCAL in ["develop","dev"])
```
bakhtin: Shall we get rid of 'dev' branch mentions since we moved back to good old `develop` ?

BulatSaif:
Currently  build fails due to error :
```
CMake Error at /usr/local/Cellar/cmake/3.11.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):

  Could NOT find PythonInterp: Found unsuitable version "1.4", but required
  is at least "3.5" (found /Users/jenkins/.pyenv/shims/python3.5)
```
But looks like it work fine on `mac-mcst-1`,
@bakhtin, can you please add pysha3 to rest of mac

BulatSaif:
@bakhtin `mac-mcst-1 `and `mac-mcst-2` works fine, but problem with `mac-mcst-3`

BulatSaif:
mac-mcst-3 now works

PR #2096, opened 2019-02-13
Fix nested object copy in shared_model proto backend

lebdron:
### Description of the Change
Fix nested object shared_model construction in protobuf backend. Now nested objects are not copied into shared_model objects, reducing the memory consumption twice.
Remove redundant vector copy in `yac/transport/impl/network_impl.cpp`.

### Benefits
Reduce memory consumption.

### Possible Drawbacks 
None.

### Usage Examples or Tests
Add static object counters to `shared_model::proto` and `iroha::protocol` object and compare the output.

### DIFF
--- a/irohad/consensus/yac/transport/impl/network_impl.cpp
+++ b/irohad/consensus/yac/transport/impl/network_impl.cpp
@@ -67,7 +67,7 @@ namespace iroha {
         async_call_->log_->info(
             "Receive votes[size={}] from {}", state.size(), context->peer());
 
-        handler_.lock()->onState(state);
+        handler_.lock()->onState(std::move(state));
         return grpc::Status::OK;
       }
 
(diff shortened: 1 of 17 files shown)


irohad/consensus/yac/transport/impl/network_impl.cpp
```diff
@@ -67,7 +67,7 @@ namespace iroha {
         async_call_->log_->info(
             "Receive votes[size={}] from {}", state.size(), context->peer());
 
-        handler_.lock()->onState(state);
+        handler_.lock()->onState(std::move(state));
```
igor-egorov: Unconditional lock of the weak pointer will cause undefined behavior when a pointer type does not have a default constructor.

Please do something like
```
if (auto notifications = handler_.lock()) {
  notifications->onState(std::move(state));
} else {
  // some error handling if applicable
}
```

test/module/irohad/simulator/simulator_test.cpp
```diff
@@ -131,7 +131,9 @@ auto makeTx() {
 
 TEST_F(SimulatorTest, ValidWhenPreviousBlock) {
   // proposal with height 2 => height 1 block present => new block generated
-  std::vector<shared_model::proto::Transaction> txs = {makeTx(), makeTx()};
```
MBoldyrev: Could you please also address the problem of temporary protocol transactions captured by reference in these cases? It seems that this problem may arise somewhere without being noticed.

PR #2087, opened 2019-02-09
Add performance test based on locust.io

lebdron:
### Description of the Change
Performance test that sends a transaction using asynchronous API and waits until the final status - committed or rejected - is received from the status stream.

### How to run
[README.md](https://github.com/lebdron/iroha/blob/7426dc869654076b2d8c37cd74d4add6d9ffde9f/test/load/README.md)
>Specify locustfile in `LOCUSTFILE_PATH` in Compose file

PR #2101, opened 2019-02-14
Fix busy waiting in tx status stream loop

igor-egorov:

### Description of the Change

The change contains modified copies of run_loop_state and run_loop objects from rxcpp.

The method `dispatch` of `run_loop` would not immediately return with the applied patch.
The method will pause execution on an additional cond var till some task will be scheduled for execution.

That fix solves the problem of busy waiting when we call dispatch inside a loop.

### Benefits

Decreased CPU usage

### Usage Examples or Tests

Locust #2087 settings are 1/1

#### Before the fix
image

#### After the fix
image


MBoldyrev:
Please see an alternative approach in #2104. While implementing the same idea, it avoids copying the external source code (less maintenance troubles) and also causes less false wakeups (when an event is scheduled after next scheduled wakeup).

igor-egorov:
#2104 will be applied with an additional fix for rx combine_latest from @lebdron

PR #2102, opened 2019-02-14
Fix command service infinite replay

igor-egorov:

### Description of the Change

Persistent cache for some reason was not integrated into command service.
That was resulting in infinite resending of a replayed transaction between Iroha components.

The fix makes CommandServiceImpl aware of persistent cache and the infinite loop is broken now.

Caches lookup strategy during batch processing in command service (even single incoming transaction will be represented as a batch internally):
* check command service tx status cache
* check persistent storage cache for replays

### Benefits

Potentially improved CPU and memory consumption.
Reduced network load.
An infinite resending of a replayed transaction between Iroha components is avoided.

### Possible Drawbacks 

Weird cache dependency for CommandServiceImpl appeared.

### Usage Examples or Tests

```
command_service_replay_test
```

compile and run irohad

#### Check that there is no infinite resending
run repro script https://gist.github.com/igor-egorov/ff56db9ca1de50215d6830f5980155d6

Iroha output should not contain inifinitely repeated message about duplicated transaction from `OnDemandOrderingServiceImpl` like "Duplicate transaction: ..."

#### Check that correct statuses are delivered in tx status stream

repro script https://gist.github.com/igor-egorov/152b821faa335193569dec2246e578fb

You have to compile two versions of irohad: one from `develop` branch, another - from that pr.

Run the script with both irohad's and capture the output to files.
The statuses (their order) should not differ.


test/integration/acceptance/replay_test.cpp
```diff
@@ -40,7 +40,13 @@ class ReplayFixture : public AcceptanceFixture {
  * @when the transaction is sent to ITF twice
  * @then the second submission should be rejected
  */
-TEST_F(ReplayFixture, OrderingGateReplay) {
+TEST_F(ReplayFixture, DISABLE_OrderingGateReplay) {
+  /*
```
kamilsa: Maybe leave a todo here?

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -150,6 +152,27 @@ namespace iroha {
 
     void CommandServiceImpl::processBatch(
         std::shared_ptr<shared_model::interface::TransactionBatch> batch) {
+      auto cache_presence = tx_presence_cache_->check(*batch);
```
lebdron: I suggest to make a look up in `cache_` for response first, then access `tx_presence_cache_`. If there was no entry in `cache_` for the transaction, but `tx_presence_cache_` returns `Committed` or `Rejected`, it can be pushed into status bus. In this case there will be no problem testing the replay in ITF.

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -150,6 +152,27 @@ namespace iroha {
...
+      if (not cache_presence) {
+        // TODO andrei 30.11.18 IR-51 Handle database error
+        log_->warn("Check tx presence database error. {}", *batch);
+        return;
+      }
+      auto is_replay = std::any_of(
```
MBoldyrev: It would be nice to place a TODO to make a method to `TxPresenceCache` that generalizes this check, because there are at least 5 analogous places in the project.

igor-egorov: Caches lookup order was changed and the code was modified. 
Now it could not be generalized with the rest part of replay checks.

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -150,30 +153,91 @@ namespace iroha {
...
+                      },
+                      [this, &found, &status_issuer](const auto &status) {
+                        this->pushStatus(status_issuer, *found);
+                      });
+                  lookup_result.first = true;
+                }
+                return lookup_result;
+              });
+
+      if (cache_has_at_least_one_tx and not batch_has_mst_pending_tx) {
```
muratovv: add an explanation for the reason for the `if` statement.

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -150,30 +153,91 @@ namespace iroha {
...
+        return;
+      }
+
+      if (not batch_has_mst_pending_tx) {
```
muratovv: Is it possible to pass batch with pending txes which was already processed?

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -150,30 +153,91 @@ namespace iroha {
...
+                          const shared_model::interface::MstPendingResponse &) {
+                        this->pushStatus(status_issuer, *found);
+                        lookup_result.second = true;
+                      },
+                      [this, &tx_hash, &status_issuer](
+                          const shared_model::interface::NotReceivedTxResponse
+                              &) {
+                        this->pushStatus(
+                            status_issuer,
+                            status_factory_->makeStatelessValid(tx_hash));
```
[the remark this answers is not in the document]
muratovv: Add comment

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -150,30 +153,91 @@ namespace iroha {
...
+      std::tie(cache_has_at_least_one_tx, batch_has_mst_pending_tx) =
+          std::accumulate(
+              txs.begin(),
+              txs.end(),
+              std::make_pair<bool, bool>(false, false),
+              [this, &status_issuer](std::pair<bool, bool> lookup_result,
+                                     const auto &tx) {
+                const auto &tx_hash = tx->hash();
+                if (auto found = cache_->findItem(tx_hash)) {
+                  iroha::visit_in_place(
```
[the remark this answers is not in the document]
igor-egorov: As we discussed - can be done sometimes later during refactoring. I have left a comment in the code.

irohad/torii/impl/command_service_impl.cpp
igor-egorov: I have added a comment.

irohad/torii/impl/command_service_impl.cpp
igor-egorov: fixed

igor-egorov:
Please review the two last commits.

test/fuzzing/status_fuzz.cpp
```diff
@@ -16,6 +16,7 @@
 #include "interfaces/iroha_internal/transaction_batch_factory_impl.hpp"
 #include "interfaces/iroha_internal/transaction_batch_parser_impl.hpp"
 #include "module/irohad/ametsuchi/ametsuchi_mocks.hpp"
+#include "module/irohad/ametsuchi/mock_tx_presence_cache.hpp"
```
lebdron: I would suggest to revert the changes in this file since it does not compile, and fix it in another pull request.

test/fuzzing/torii_fuzz.cpp
```diff
@@ -15,6 +15,7 @@
 #include "interfaces/iroha_internal/transaction_batch_factory_impl.hpp"
 #include "interfaces/iroha_internal/transaction_batch_parser_impl.hpp"
 #include "module/irohad/ametsuchi/ametsuchi_mocks.hpp"
+#include "module/irohad/ametsuchi/mock_tx_presence_cache.hpp"
```
lebdron: Same as in status_fuzz.cpp

test/module/irohad/torii/command_service_replay_test.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * @given a CommandService with internal cache that contains MstPending status
+ * for the transaction
+ * @when an already rejected transaction arrives (it is a replay and may have
+ * pending status in internal cache, for example, after cache overflow)
+ * @then transaction is not passed to transaction processor
+ */
+TEST_F(CommandServiceReplayTest, OrderingGateReplayWhenPendingTest) {
```
lebdron:
```suggestion
TEST_F(CommandServiceReplayTest, ReplayWhenPendingAndRejectedTest) {
```

test/module/irohad/torii/command_service_replay_test.cpp
```diff
@@ -0,0 +1,139 @@
...
+  command_service->handleTransactionBatch(batch);
+}
+
+/**
+ * @given a CommandService with internal cache that contains MstPending status
+ * for the transaction
+ * @when the transaction arrives
+ * @then the transaction is passed to transaction processor
+ */
+TEST_F(CommandServiceReplayTest, OrderingGateReplayWhenPending2Test) {
```
lebdron:
```suggestion
TEST_F(CommandServiceReplayTest, ReplayWhenPendingAndMissingTest) {
```

PR #2105, opened 2019-02-18
Add rxcpp patch for run loop data race

lebdron:
### Description of the Change
Introduce fixes from Rx repository https://github.com/ReactiveX/RxCpp/pull/486

### Benefits
No data race in status streaming

### Possible Drawbacks 
None

### DIFF
--- a/cmake/Modules/Findrxcpp.cmake
+++ b/cmake/Modules/Findrxcpp.cmake
@@ -15,11 +15,13 @@ set_target_description(rxcpp "Library for reactive programming" ${URL} ${VERSION
 
 
 if (NOT rxcpp_FOUND)
+  find_package(Git REQUIRED)
   externalproject_add(reactive_extensions_rxcpp
       GIT_REPOSITORY ${URL}
       GIT_TAG        ${VERSION}
       CONFIGURE_COMMAND ""
       BUILD_COMMAND ""
+      PATCH_COMMAND ${GIT_EXECUTABLE} am ${PROJECT_SOURCE_DIR}/patch/0001-Fix-data-race-in-run-loop-usage.patch
       INSTALL_COMMAND "" # remove install step
       UPDATE_COMMAND "" # remove update step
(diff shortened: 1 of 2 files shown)


neewy:
@MBoldyrev, unfortunately, this is an issue we have to live with for some time until RxCpp is fixed

MBoldyrev:
> @MBoldyrev, unfortunately, this is an issue we have to live with for some time until RxCpp is fixed

If we are going to use Rx from elsewhere, we need to make sure it has the patch. I do not see a reasonable way to check it even after the Andrew's pull request to Rx repo is merged - users still may have the old version (and most probably they will for a long time). I think we must use a fixed version of the library applying the patch with CMake until it gets into the Rx repo, and after that switch to a fixed version from the repo, but not blindly use any Rx lib found in the user environment.

cmake/Modules/Findrxcpp.cmake
```diff
@@ -13,13 +13,15 @@ set(URL https://github.com/Reactive-Extensions/rxcpp.git)
...
+#if (NOT rxcpp_FOUND)
+  find_package(Git REQUIRED)
   externalproject_add(reactive_extensions_rxcpp
       GIT_REPOSITORY ${URL}
       GIT_TAG        ${VERSION}
       CONFIGURE_COMMAND ""
       BUILD_COMMAND ""
+      PATCH_COMMAND ${GIT_EXECUTABLE} apply ${PROJECT_SOURCE_DIR}/patch/0001-Fix-data-race-in-run-loop-usage.patch
       INSTALL_COMMAND "" # remove install step
       UPDATE_COMMAND "" # remove update step
```
MBoldyrev: I am not sure, but seems that this line can prevent the existing cloned repo from updating to the version we need.

lebdron:
Since https://github.com/ReactiveX/RxCpp/pull/486 was merged, it might be better to bump the version instead.

PR #2106, opened 2019-02-18
Less Transaction Final Statuses

igor-egorov:

### Description of the Change

Non final txs statuses were removed from the list of final statuses.


neewy:
Draft? wow

PR #2108, opened 2019-02-19
New logger to iroha-cli

MBoldyrev:
### Description of the Change

Integrate the new logger to iroha-cli.

### Benefits

Another step towards new logger merge.

### Possible Drawbacks 

More code.

Avoid tree structure in iroha-cli for simplicity.


iroha-cli/client.cpp
```diff
@@ -14,8 +14,12 @@
 
 namespace iroha_cli {
 
-  CliClient::CliClient(std::string target_ip, int port)
-      : command_client_(target_ip, port), query_client_(target_ip, port) {}
+  CliClient::CliClient(std::string target_ip,
+                       int port,
+                       logger::LoggerPtr pb_qry_factory_log)
+      : command_client_(target_ip, port, pb_qry_factory_log),
```
igor-egorov: Can we use & references or moves here?

MBoldyrev: moved

iroha-cli/interactive/interactive_query_cli.hpp
```diff
@@ -150,11 +158,20 @@ namespace iroha_cli {
...
       // Crypto provider
       std::shared_ptr<iroha::model::ModelCryptoProvider> provider_;
+
+      /// Logger manager for GrpcResponseHandler
+      logger::LoggerManagerTreePtr responce_handler_log_manager_;
+
+      /// Logger for PbQueryFactory
+      logger::LoggerPtr pb_qry_factory_log_;
+
+      /// Logger for JsonQueryFactory
```
lebdron: Are you sure that these comments are useful? It looks like they can be derived from type and variable name.

irohad/model/converters/impl/json_query_factory.cpp
```diff
@@ -18,7 +19,7 @@ using namespace rapidjson;
 namespace iroha {
   namespace model {
     namespace converters {
-      JsonQueryFactory::JsonQueryFactory(logger::Logger log)
+      JsonQueryFactory::JsonQueryFactory(logger::LoggerPtr log)
```
lebdron: `log_` is not used in this file. However, since this code is legacy, it is probably better not to refactor it.

PR #2116, opened 2019-02-20
Fixes to the new logger

MBoldyrev:
### Description of the Change

Here is a set of changes (each in a separate commit):
- moved log levels translation to a separate function which initializes a static map as soon as it is needed (fixes static initialization races)
- fixed formatting due to libfmt upgrade (seems that any `{..}` means a formatted argument now)
- added `tryGetValByKey` specialization for optionals to config file parser. Now optional config members can be made seamlessly
- made the logger configuration section of the config file optional (to avoid an unwanted breaking change)

### Benefits

Fixes & improvements.

### Possible Drawbacks 

Undefined.
b

### DIFF
--- a/libs/logger/logger_spdlog.cpp
+++ b/libs/logger/logger_spdlog.cpp
@@ -13,16 +13,27 @@
 #include <spdlog/spdlog.h>
 #include <boost/assert.hpp>
 
-static const std::string kDefaultPattern =
-    R"([%Y-%m-%d %H:%M:%S.%F] [%L] [%n]: %v)";
+namespace {
 
-static const std::map<logger::LogLevel, const spdlog::level::level_enum>
-    kSpdLogLevels = {{logger::LogLevel::kTrace, spdlog::level::trace},
-                     {logger::LogLevel::kDebug, spdlog::level::debug},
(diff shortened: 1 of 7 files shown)


libs/logger/logger_spdlog.cpp
```diff
@@ -13,16 +13,27 @@
...
+    static const std::map<logger::LogLevel, const spdlog::level::level_enum>
+        kSpdLogLevels = {
+            {logger::LogLevel::kTrace, spdlog::level::trace},
+            {logger::LogLevel::kDebug, spdlog::level::debug},
+            {logger::LogLevel::kInfo, spdlog::level::info},
+            {logger::LogLevel::kWarn, spdlog::level::warn},
+            {logger::LogLevel::kError, spdlog::level::err},
+            {logger::LogLevel::kCritical, spdlog::level::critical}};
+    const auto it = kSpdLogLevels.find(level);
+    BOOST_ASSERT_MSG(it != kSpdLogLevels.end(), "Unknown log level!");
```
igor-egorov: Not sure we need to use asserts in logger that is going to be a part of everything.

MBoldyrev: It will not affect the release build.

PR #2120, opened 2019-02-21
rx subscription manager

muratovv:
### Description of the Change
Add naive but useful implementation of subscription manager for rx subscriptions.

### Benefits

Don't miss your subscriptions in dtor

### Possible Drawbacks 
Additional inheritance level. But in fact, the class is a mixin.

```addSubscription(observer.subcribe(...))```

* addSubscription could return a reference for subscription for further usage
* thread safety

### DIFF
--- a/libs/common/subscription_manager.hpp
+++ b/libs/common/subscription_manager.hpp
@@ -19,10 +19,10 @@ namespace iroha {
      * Note: class is NOT thread-safe
      */
     class SubscriptionManager {
-    public:
+     public:
       virtual ~SubscriptionManager();
 
-    protected:
+     protected:
       /**
        * Add new subscription for the management
        */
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: libs/common/subscription_manager.hpp [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+      virtual ~SubscriptionManager();
+
+     protected:
+      /**
+       * Add new subscription for the management
+       */
+      void addSubscription(rxcpp::composite_subscription &&subscription);
+
+      /// list of all subscriptions
+      std::vector<rxcpp::composite_subscription> subscriptions;
```
l4l: `std::forward_list` might be used as a type hint for a reader, that subscription list is only need to be traversed

muratovv:
This PR was introduced as sugar for managing subscriptions but it seems that the maintainers want to find a more elegant solution which includes both: automatic management of subscriptions and subject/observer lifetime.

PR #2123, opened 2019-02-24
Ansible Rewrite

sudomann:
### Description of the Change
Rewritten for peers to be able to discover each other within the cluster and communicate within a given namespace. A service is also defined to expose `torii-port` to another specified namespace (has been preset to `namespace: default`).

### Benefits

Ansible configuration for Kubernetes that works and is flexible to customize via variables without breaking functionality.
Configuration follows good semantics, and allow easy deployment with minimal effort.
### Possible Drawbacks 

None


sudomann:
Would be great if someone could help out in setting appropriate resource limits. I have no idea what iroha needs and lack expertise to measure/analyze.

Baydarich:
@sudomann You were asking me about improving security handling. Deleting files with private keys from a deployment host after the process is finished sounds like a good idea, but be sure to backup them somewhere in a safe place.

I don't understand why there are private and public keys embedded directly in configuration files, can you clarify a bit?

sudomann:
@Baydarich When Ansible generates keys for each peer, they are first recorded into a [csv file](https://github.com/hyperledger/iroha/pull/2123/files#diff-ddc699e3ae6686442dc6b6ec37d8b027). 

1.  For lack of a better alternative, I improved upon the existing kubernetes setup, which [preserves each peer as its own unique entity](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) with fixed persistent storage. The csv file is handy for subsequent tasks such as registering all peers in the `genesis.block` file, since the peer's hostname and public key are extracted from it.

2. Later on in the process, the public and private keys of each peer are extracted from that csv file to generate [a `yml` resource file for cluster secrets](https://github.com/hyperledger/iroha/pull/2123/files#diff-4a95290b9cf2e56f70a1c73d80792872). The keys are encoded in base64 (requirement of Kubernetes) but can be very easily reverted to readable hex. This file is used to create all the secrets at once with a single command.

3. Before each pod hosting a peer is started, it these keys are [mounted](https://github.com/hyperledger/iroha/pull/2123/files#diff-595af1be216cc2b8cac1f232038f5211L41) [from secrets](https://github.com/hyperledger/iroha/pull/2123/files#diff-595af1be216cc2b8cac1f232038f5211R100) and [copied](https://github.com/hyperledger/iroha/pull/2123/files#diff-595af1be216cc2b8cac1f232038f5211R28) to appropriate pods.

As I understand, there is no further use in knowing/keeping the private keys beyond that since the peers all have theirs mounted. For potential debugging purposes however, I thought it made sense to leave the files to show that the role was working as intended. e.g verifying using the csv file that the peers were assigned the right combination of hostname and public key when added in the `genesis.block` (in the same order), along with the mounted secrets (in the same order).

I've thought about changing the Ansible steps to instead add a peer to the `genesis.block` file with its hostname, and create its secret as they are generated, but this is much more annoying for a developer to debug, with nothing but a trail of scattered stdout text to piece together.

bakhtin:
Thanks for the contribution. I skimmed through the code and overall it looks good. But we need some time to actually test the new playbook. Based on the test results I'll be able to give a proper feedback.

PR #2125, opened 2019-02-25
Pluggable consensus

MBoldyrev:
### Description of the Change

This change brings back the pluggable consensus initially added in #2091, but reverted (#2109) due to a bug, which is fixed in this PR.

### Benefits

Pluggable consensus is back.

### Possible Drawbacks 

I do not fully understand why the fix works O_o.

Run iroha in release build with several peers in the genesis block (they do not need to be reachable). It does not crash anymore, and valgrind does not report unitialized memory usage.


irohad/consensus/yac/impl/supermajority_checker_bft.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <boost/range/algorithm/max_element.hpp>
+#include <boost/range/numeric.hpp>
+#include "consensus/yac/impl/supermajority_checker_kf1.hpp"
+
+namespace iroha {
+  namespace consensus {
+    namespace yac {
+
+      bool SupermajorityCheckerBft::hasSupermajority(
```
lebdron: This file is almost the same as `supermajority_checker_cft.cpp`, the only change is `k` parameter in `hasSupermajority`. Maybe it is better to replace these two classes with class template, and make `k` a template parameter?

irohad/consensus/yac/storage/yac_vote_storage.hpp
```diff
@@ -81,6 +83,9 @@ namespace iroha {
        public:
         // --------| public api |--------
 
+        /// @param consistency_model - consensus consistency model (CFT, BFT).
+        YacVoteStorage(ConsistencyModel consistency_model);
```
lebdron: Why not parametrize it with `std::shared_ptr<SupermajorityChecker>`? This will allow simpler testing with `MockSupermajorityChecker`

test/module/irohad/consensus/yac/supermajority_checker_test.cpp
```diff
@@ -18,96 +22,165 @@ using ::testing::ReturnRefOfCopy;
...
-TEST_F(SupermajorityCheckerTest, SuperMajorityCheckWithSize2) {
-  log_->info("-----------| F(x, 2), x in {0..3} -----------");
-
-  int N = 2;
-  ASSERT_FALSE(checkSize(0, N));
-  ASSERT_FALSE(checkSize(1, N));
-  ASSERT_TRUE(checkSize(2, N));
-  ASSERT_FALSE(checkSize(3, N));
+TEST_P(CftAndBftSupermajorityCheckerTest, SuperMajorityCheckWithSize2) {
+  log_->info("-----------| F(x, 2), x in [0..3] |-----------");
```
lebdron: Probably these logs can be removed, as they were used as test descriptions before given-when-then was introduced.

PR #2127, opened 2019-02-26
Provide git version info

MBoldyrev:
### Description of the Change

Make iroha version available through `--version` argument and in log pattern.

The log pattern can now include git commit hash formatted like `%h` or `%6h`, where the first will be substituted with the whole hash, and the second with the first 6 chars.

### Benefits

When someone sends error logs, exact source version is available (as by default).

### Possible Drawbacks 

Longer log lines. Requires Git to work.


MBoldyrev:
In a discussion on commit hash inclusion to log lines format we decided it is an excessive info, so I reverted that change and added a single log line containing the version to be printed during startup.

MBoldyrev:
That is how it looks compiled on my machine:
```
[2019-02-28 09:32:51.729015315] [I] [Init]: Irohad version: v0.75-4927-g50675ce8d
```

irohad/main/irohad.cpp
```diff
@@ -95,6 +96,8 @@ static bool validateVerbosity(const char *flagname, const std::string &val) {
 DEFINE_string(verbosity, kLogSettingsFromConfigFile, "Log verbosity");
 DEFINE_validator(verbosity, &validateVerbosity);
 
+
+
```
l4l: why?

libs/common/irohad_version.hpp
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef LIBS_COMMON_IROHAD_VERSION_HPP
+#define LIBS_COMMON_IROHAD_VERSION_HPP
+
+namespace iroha {
+
+  extern const char *kGitPrettyVersion;
```
l4l: Why a separate file?

MBoldyrev: When these definitions change, only this small file has to be recompiled.

irohad/main/irohad.cpp
MBoldyrev: Surely these empty lines are odd.

MBoldyrev:
As the logger formatting with git hash was considered unneeded, I rebased the branch from the new logger to develop.

libs/common/CMakeLists.txt
```diff
@@ -33,3 +33,36 @@ target_link_libraries(libs_timeout INTERFACE
...
   rxcpp
   )
+
+add_library(irohad_version irohad_version.cpp)
+
+# Get the git repo data
+set(GIT_REPO_PRETTY_VER "version info unavailable")
+if (EXISTS "${PROJECT_SOURCE_DIR}/.git")
+  if(NOT GIT_FOUND)
+    find_package(Git QUIET)
```
muratovv: Seems, find_package will install git. I think better assume that git is already installed, but, ofc, there should be a fallback strategy and appropriate message when git is unavailable.

libs/common/irohad_version.hpp
```diff
@@ -0,0 +1,16 @@
...
+  extern const char *kGitHeadCommitHash;
```
muratovv: Is this variable set up in cmake? I think it should be removed.

libs/common/CMakeLists.txt
MBoldyrev: We found out that the package will not be installed if missing.

libs/common/CMakeLists.txt
```diff
@@ -33,3 +33,36 @@ target_link_libraries(libs_timeout INTERFACE
...
+        WORKING_DIRECTORY
+        "${PROJECT_SOURCE_DIR}"
+        RESULT_VARIABLE
+        res
+        OUTPUT_VARIABLE
+        GIT_REPO_PRETTY_VER
+        ERROR_QUIET
+        OUTPUT_STRIP_TRAILING_WHITESPACE)
+    if(NOT res EQUAL 0)
+      message(STATUS "Error running `git describe': ${res}")
```
muratovv: I suggest in error cases initialize `irohad_version` with "unknown version" or something similar.

MBoldyrev: it is actually done here:
https://github.com/hyperledger/iroha/blob/404c95243536034642b8ac9a36146be8ce7426af/libs/common/CMakeLists.txt#L40

PR #2131, opened 2019-02-28
Cmake sanitizers

luckychess:
### Description of the Change

Add different sanitizers flags into cmake.

### Benefits
- Convenient way to sanitize Iroha
- Different sanitizers can be turned on simultaneously
- Dependencies are also under sanitizer checks
- Make sanitizing in CI possible

### Possible Drawbacks 
- Have no idea about how does it work in Windows
- Standard library still remains uninstrumented
- Incompatible sanitizing options can be used
- Ubsan doesn't work with clang on Linux (linking issue, probably could be solved)

### Usage Examples or Tests 
Four new cmake flags were introduced: _SANITIZE_THREAD_, _SANITIZE_ADDRESS_, _SANITIZE_MEMORY_, and _SANITIZE_UNDEFINED._ To use run:
`cmake -DSANITIZE_ADDRESS=ON ..`
Different flags can be set simultaneously but if they are incompatible you will get a build error.
For better results I would recommend to remove all dependencies from the system and let Iroha's cmake pull them - that way they will be also sanitized.
For better results standard library should be sanitized as well.
Leak detection is on by default on Linux when address sanitizer is on. For Mac you should use next command:
```
ASAN_OPTIONS=detect_leaks=1 bin/irohad ...
```


CMakeLists.txt
```diff
@@ -67,21 +67,42 @@ if(COVERAGE)
...
+option(TESTING               "Build tests"                              ON )
+option(USE_BTF               "Build Binary Testing Framework"           OFF)
+option(COVERAGE              "Enable coverage"                          OFF)
+option(FUZZING               "Build fuzzing binaries"                   OFF)
+option(PACKAGE_ZIP           "Create ZIP package"                       OFF)
+option(PACKAGE_TGZ           "Create TGZ package"                       OFF)
+option(PACKAGE_RPM           "Create RPM package"                       OFF)
+option(PACKAGE_DEB           "Create DEB package"                       OFF)
+option(ENABLE_LIBS_PACKAGING "Enable libs packaging"                    ON)
+option(SWIG_JAVA             "Generate Swig Java bindings"              OFF)
```
neewy: Pls pull develop, these options should be missing

cmake/Modules/Findpq.cmake
```diff
@@ -40,9 +40,14 @@ if (NOT pq_FOUND)
   externalproject_add(postgres_postgres
       GIT_REPOSITORY  ${URL}
       GIT_TAG         ${VERSION}
-      CONFIGURE_COMMAND ./configure --without-readline
+      CONFIGURE_COMMAND
+                      ./configure --without-readline
+                      CC=${CMAKE_C_COMPILER}
+                      CXX=${CMAKE_CXX_COMPILER}
```
lebdron: `CXX` is probably unused https://www.postgresql.org/docs/9.5/install-procedure.html

cmake/Modules/Findgflags.cmake
```diff
@@ -20,6 +20,11 @@ if (NOT gflags_FOUND)
   externalproject_add(gflags_gflags
       GIT_REPOSITORY ${URL}
       GIT_TAG        ${VERSION}
+      CMAKE_ARGS
+        -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
```
lebdron: Maybe it is also possible to define a variable like this and reuse it in different modules
```
set(DEPS_CMAKE_ARGS -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
        -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
        -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
        -DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS})
```

PR #2136, opened 2019-03-02
Fuzzing fix

MBoldyrev:
### Description of the Change

As our CI does not build fuzzing targets, these get outdated over time. This change fixes broken things because of recent changes in dependees.

### Benefits

Hopefully fuzzing is back.

### Possible Drawbacks 

This does not prevent the cause of such problems - missing CI checks for fuzzing.


test/fuzzing/status_fuzz.cpp
```diff
@@ -98,24 +102,34 @@ struct CommandFixture {
...
         status_bus,
         status_factory,
-        transaction_factory,
-        batch_parser,
-        transaction_batch_factory,
-        consensus_gate_,
-        2);
+        cache_,
+        tx_presence_cache_, );
+    service_transport_ =
```
lebdron: Please check line 117, it is the same. The file does not compile.

PR #2138, opened 2019-03-04
Add service endpoint for remote daemon shutdown

igor-egorov:


neewy:
Btw we can merge these changes in case there is a feature toggle (if you explicitly allow perform remote shutdown in config)

PR #2139, opened 2019-03-04
Add fix for commit direct propagation

muratovv:
### Description of the Change
Add fix for direct propagation in Yac. If a peer sends a vote for the old round we will share last finalized state.

### Benefits
Fix of synchronization in some cases.

### Possible Drawbacks 
No so elegant solution: YAC checks stateless invariants inside functions, vote storage responsibility is increased.

### DIFF
--- a/irohad/consensus/yac/impl/yac.cpp
+++ b/irohad/consensus/yac/impl/yac.cpp
@@ -18,6 +18,12 @@
 #include "interfaces/common_objects/peer.hpp"
 #include "logger/logger.hpp"
 
+// TODO: 2019-03-04 @muratovv refactor std::vector<VoteMessage> with a
+// separate class IR-374
+auto &getRound(const std::vector<iroha::consensus::yac::VoteMessage> &state) {
+  return state.at(0).hash.vote_round;
+}
+
 namespace iroha {
   namespace consensus {
     namespace yac {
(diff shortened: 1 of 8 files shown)


irohad/consensus/yac/impl/yac.cpp
```diff
@@ -136,55 +142,77 @@ namespace iroha {
...
+                  break;
               }
-              break;
+            },
+            // sent a state which didn't match with current one
+            [&]() { tryPropagateBack(state); });
+      }
+
+      void Yac::tryPropagateBack(const std::vector<VoteMessage> &state) {
+        // yac back propagation will work  only if another peer is in
```
MBoldyrev:
```suggestion
        // yac back propagation will work only if another peer is in
```

irohad/consensus/yac/storage/impl/yac_vote_storage.cpp
```diff
@@ -69,6 +69,11 @@ namespace iroha {
               return storage->insert(state) |
                          [this, &round](
                              auto &&insert_outcome) -> boost::optional<Answer> {
+
+                if (round > last_round_) {
+                  last_round_ = round;
+                }
```
MBoldyrev: `last_round_` is `boost::optional`:
```
                last_round_ = std::max(last_round_.value_or(round), round);
```

irohad/consensus/yac/storage/impl/yac_vote_storage.cpp
```diff
@@ -107,6 +112,25 @@ namespace iroha {
...
+      }
+
+      boost::optional<Answer> YacVoteStorage::getState(
+          const Round &round) const {
+        auto proposal_storage =
+            std::find_if(proposal_storages_.begin(),
+                         proposal_storages_.end(),
+                         [&round](const auto &proposal_state) {
+                           return proposal_state.getStorageKey() == round;
+                         });
```
MBoldyrev: Could `YacVoteStorage::getProposalStorage(const Round &round)` be reused here?
https://github.com/hyperledger/iroha/blob/d0a9e924c4498b89c70f8560587d5fa95429254f/irohad/consensus/yac/storage/impl/yac_vote_storage.cpp#L20-L26

irohad/consensus/yac/storage/yac_vote_storage.hpp
```diff
@@ -128,6 +128,19 @@ namespace iroha {
...
         void nextProcessingState(const Round &round);
 
+        /**
+         * Get last by order finalized round
+         * @return round if it exists
+         */
+        boost::optional<Round> getLastFinalizedRound() const;
+
+        /**
+         * Get state attached to passed round
```
MBoldyrev:
```suggestion
         * Get the state of a past round
```

irohad/consensus/yac/storage/yac_vote_storage.hpp
```diff
@@ -128,6 +128,19 @@ namespace iroha {
...
+         * @param round - target round for finding
```
MBoldyrev:
```suggestion
         * @param round - queried round
```

test/module/irohad/consensus/yac/yac_synchronization_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  auto createHash(const iroha::consensus::Round &r,
+                  const std::string &block_hash = "default_block",
+                  const std::string &proposal_hash = "default_proposal") const {
+    return YacHash(r, proposal_hash, block_hash);
+  }
+
+  auto createVote(size_t from, const YacHash &yac_hash) const {
+    return iroha::consensus::yac::createVote(
+        yac_hash,
+        *iroha::hexstringToBytestring(peers_.at(from)->pubkey().hex()));
```
MBoldyrev: Accessing vector elements by index looks scary, but can be tolerated in tests. Could you please add an assertion to catch indices outside the vector boundaries?

test/module/irohad/consensus/yac/yac_synchronization_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  }
+  /// create votes of peers by their number
+  auto createVotes(
+      const std::vector<size_t> &peers,
+      const iroha::consensus::Round &r,
+      const std::string &block_hash = "default_block",
+      const std::string &proposal_hash = "default_proposal") const {
```
MBoldyrev: If you rely on these literals to be equal, it is better to store them in a single place, like a global const.
https://github.com/hyperledger/iroha/blob/d0a9e924c4498b89c70f8560587d5fa95429254f/test/module/irohad/consensus/yac/yac_synchronization_test.cpp#L29-L30
https://github.com/hyperledger/iroha/blob/d0a9e924c4498b89c70f8560587d5fa95429254f/test/module/irohad/consensus/yac/yac_synchronization_test.cpp#L43-L44

irohad/consensus/yac/impl/yac.cpp
```diff
@@ -136,55 +142,77 @@ namespace iroha {
...
+        // propagation stage because if peer sends list of votes this means that
+        // state is already committed
+        if (state.size() != 1) {
+          return;
+        }
+
+        vote_storage_.getLastFinalizedRound() | [&](const auto &last_round) {
+          if (getRound(state) <= last_round) {
```
MBoldyrev: It would be nice to add some logging to both branches of this `if`.

test/module/irohad/consensus/yac/yac_synchronization_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  }
+};
+
+/**
+ * @given Yac which stores commit
+ * @when  Vote from known peer from old round which was presented in the cache
+ * @then  Yac sends commit for the last round
+ */
+TEST_F(YacSynchronizationTest, SynchronizationOncommitInTheCahe) {
+  NetworkUtil network_util(7);
```
luckychess: Looks like this line and line just below are the same in every test. Could we move them into `SetUp()`?

irohad/consensus/yac/storage/impl/yac_vote_storage.cpp
muratovv: Optional is required because in the first round there is no committed state yet.

irohad/consensus/yac/storage/impl/yac_vote_storage.cpp
muratovv: Nice catch, thanks

test/module/irohad/consensus/yac/yac_synchronization_test.cpp
muratovv: Values by itself don't matter. If a user wants to check some specific value it will be done explicitly by a test.

irohad/consensus/yac/impl/yac.cpp
muratovv: What is the point to do that?

MBoldyrev: keysManagerLogger

MBoldyrev: It may be very helpful for synchronization debugging. Much more helpful than other log messages that we currently have.

irohad/consensus/yac/storage/impl/yac_vote_storage.cpp
muratovv: But naive reusing impossible here:
`getProposalStorage` method is not const. We want to use it in a const context. Fix declaration of `getProposalStorage` with const we will fail because auto has deduced const iterator which uses in non-const context.

I add ``getProposalStorage` for const version.

PR #2142, opened 2019-03-05
synchronizer applies individual blocks

MBoldyrev:
### Description of the Change

Synchronizer applies each block it gets to a mutable storage individually. In case of failure it asks another peer. Blocks request obtains a timeout.

### Benefits

One step towards block observables optimization.

### Possible Drawbacks 

This PR complicates things, but it is a necessary step.

### DIFF
--- a/irohad/synchronizer/impl/synchronizer_impl.cpp
+++ b/irohad/synchronizer/impl/synchronizer_impl.cpp
@@ -74,20 +74,21 @@ namespace iroha {
     boost::optional<SynchronizationEvent>
     SynchronizerImpl::downloadMissingBlocks(
         const consensus::VoteOther &msg,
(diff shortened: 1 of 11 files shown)


irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -74,46 +75,75 @@ namespace iroha {
...
+                            block->height());
+                blocks.push_back(block);
+              }
+            },
+            [&, this]() {
+              log_->info("Applied {} blocks from peer {}",
+                         my_height() - peer_start_height,
+                         public_key);
+            });
+        network_chain.as_blocking().subscribe(subscriber);
```
lebdron: It might be possible to subscribe in `MutableStorage` and pass network chain directly to `ChainValidator`, and check its result.

PR #2149, opened 2019-03-06
Fixes for OS/OG

muratovv:
### Description of the Change
* remove setPcs method from OG
* minor issues with clang-format
* add max transaction size to OG. Not it shares only a limited number of transactions.
* OS queue replaced with an unordered set for checking not-unique transactions

### Benefits

Performance improvements and DDoS prevention

### Possible Drawbacks 
The absence of tests. I think we have to discuss it during the review


irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -118,34 +118,26 @@ OnDemandOrderingServiceImpl::onRequestProposal(consensus::Round round) {
...
     collection.insert(
         std::end(collection),
-        std::make_move_iterator(std::begin(batch->transactions())),
-        std::make_move_iterator(std::end(batch->transactions())));
-  }
-
-  if (discarded_txs_amount) {
-    *discarded_txs_amount = 0;
-    while (tx_batches_queue.try_pop(batch)) {
-      *discarded_txs_amount += boost::size(batch->transactions());
```
MBoldyrev: why did you remove this? now the `discarded_txs_amount` it not set.

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -120,6 +101,35 @@ OnDemandOrderingGate::processProposalRequest(
...
+                   // no blocks committed, no transactions to remove
+                 });
+
+  auto batches = cache_->pop();
+  cache_->addToBack(batches);
+
+  // get only transactions which fit to next proposal
+  auto end_iterator = batches.begin();
+  if (batches.size() > transaction_limit_) {
+    std::advance(end_iterator, transaction_limit_);
```
MBoldyrev: You count batches as if they were transactions. Is it the intended behaviour?

irohad/ordering/impl/on_demand_ordering_gate.hpp
```diff
@@ -86,13 +84,17 @@ namespace iroha {
               std::shared_ptr<const OnDemandOrderingService::ProposalType>>
               proposal) const;
 
+      void sendTransactionsOS(const BlockRoundEventType &event);
```
lebdron:
```suggestion
      void sendCachedTransactions(const BlockRoundEventType &event);
```

irohad/main/impl/on_demand_ordering_init.hpp
```diff
@@ -75,6 +75,7 @@ namespace iroha {
           consensus::Round initial_round,
           std::function<std::chrono::milliseconds(
               const synchronizer::SynchronizationEvent &)> delay_func,
+          size_t max_number_of_transactions,
```
lebdron: Please make this variable name consistent with `max_size` in `initOrderingGate`, maybe by renaming `max_size` to `max_number_of_transactions`.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
lebdron: It was expected to be replaced with atomic counter in `onBatches` method, so that there are no transactions to discard. However, since it is not introduced yet, I would suggest to make the following changes:
```C++
  auto it = tx_batches_queue.begin();
  for (;
       it != tx_batches_queue.end() and collection.size() < requested_tx_amount;
       ++it) {
    collection.insert(std::end(collection),
                      std::begin((*it)->transactions()),
                      std::end((*it)->transactions()));
  }

  if (discarded_txs_amount) {
    *discarded_txs_amount = 0;
    for (; it != tx_batches_queue.end(); ++it) {
      *discarded_txs_amount += boost::size((*it)->transactions());
    }
  }
```

libs/common/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ if (EXISTS "${PROJECT_SOURCE_DIR}/.git")
     # Get pretty version
     execute_process(COMMAND
         "${GIT_EXECUTABLE}"
-        describe --always
+        describe --tags
```
MBoldyrev:
```suggestion
        describe --tags --always
```

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
MBoldyrev: whoops. please either remove the discarded txs counter at all, or set it in `getTransactions`

PR #2150, opened 2019-03-06
Refactor ordering service queues

lebdron:
### Description of the Change
Use two queues in ordering service for transactions, one for current reject round, and another for future reject and commit rounds.

### Benefits
Less memory usage in case the ordering service is used for multiple rounds.

### Possible Drawbacks 
Two proposals are stored for next rounds instead of one.

### DIFF
--- a/irohad/ordering/impl/on_demand_ordering_gate.cpp
+++ b/irohad/ordering/impl/on_demand_ordering_gate.cpp
@@ -18,6 +18,20 @@
 using namespace iroha;
 using namespace iroha::ordering;
 
+std::string OnDemandOrderingGate::BlockEvent::toString() const {
+  return shared_model::detail::PrettyStringBuilder()
+      .init("BlockEvent")
+      .append(round.toString())
+      .finalize();
+}
+
+std::string OnDemandOrderingGate::EmptyEvent::toString() const {
+  return shared_model::detail::PrettyStringBuilder()
+      .init("EmptyEvent")
+      .append(round.toString())
+      .finalize();
+}
+
 OnDemandOrderingGate::OnDemandOrderingGate(
     std::shared_ptr<OnDemandOrderingService> ordering_service,
     std::shared_ptr<transport::OdOsNotification> network_client,
@@ -30,20 +44,11 @@ OnDemandOrderingGate::OnDemandOrderingGate(
       ordering_service_(std::move(ordering_service)),
       network_client_(std::move(network_client)),
       events_subscription_(events.subscribe([this](auto event) {
-        // exclusive lock
-        consensus::Round current_round_;
-        visit_in_place(event,
-                       [this, &current_round_](const BlockEvent &block_event) {
-                         // block committed, increment block round
-                         log_->debug("BlockEvent. {}", block_event.round);
-                         current_round_ = block_event.round;
-                       },
-                       [this, &current_round_](const EmptyEvent &empty_event) {
-                         // no blocks committed, increment reject round
-                         log_->debug("EmptyEvent");
-                         current_round_ = empty_event.round;
-                       });
-        log_->debug("Current: {}", current_round_);
+        consensus::Round current_round =
+            visit_in_place(event, [this, &current_round](const auto &event) {
+              log_->debug("{}", event);
+              return event.round;
+            });
 
         visit_in_place(event,
                        [this](const BlockEvent &block_event) {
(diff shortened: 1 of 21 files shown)


irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -217,23 +169,48 @@ void OnDemandOrderingServiceImpl::packNextProposals(
...
 
-  // new reject round
-  open_round(
-      {round.block_round, currentRejectRoundConsumer(round.reject_round)});
+  if (not next_round_batches_.empty()
+      and round.reject_round == kFirstRejectRound) {
+    auto txs = get_transactions(next_round_batches_, [](auto &) {});
+
+    if (not txs.empty()) {
+      generate_proposal({round.block_round, round.reject_round + 1}, txs);
```
MBoldyrev:
```suggestion
      generate_proposal({round.block_round, kNextRejectRoundConsumer}, txs);
```

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -217,23 +169,48 @@ void OnDemandOrderingServiceImpl::packNextProposals(
...
+      generate_proposal({round.block_round + 1, round.reject_round}, txs);
```
MBoldyrev:
```suggestion
      generate_proposal({round.block_round + 1, kNextCommitRoundConsumer}, txs);
```

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -123,21 +106,24 @@ OnDemandOrderingServiceImpl::onRequestProposal(consensus::Round round) {
...
-    collection.insert(
-        std::end(collection),
-        std::make_move_iterator(std::begin(batch->transactions())),
-        std::make_move_iterator(std::end(batch->transactions())));
+         and tx_batches_queue.try_pop(batch)) {
+    if (inserted.insert(batch->reducedHash().hex()).second) {
+      collection.insert(std::end(collection),
+                        std::begin(batch->transactions()),
+                        std::end(batch->transactions()));
+      batch_operation(batch);
```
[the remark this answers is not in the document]
lebdron: - https://github.com/hyperledger/iroha/blob/de84114275dba5c821715ad99945173b25e50a2d/shared_model/interfaces/iroha_internal/transaction_batch.hpp#L28
So move iterator did not have an effect.
- https://github.com/hyperledger/iroha/blob/de84114275dba5c821715ad99945173b25e50a2d/irohad/ordering/impl/on_demand_ordering_service_impl.cpp#L196
Batch pointer is moved to the next queue, so the transactions cannot be invalidated.

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -25,27 +25,25 @@ OnDemandOrderingGate::OnDemandOrderingGate(
     std::shared_ptr<cache::OrderingGateCache> cache,
     std::shared_ptr<shared_model::interface::UnsafeProposalFactory> factory,
     std::shared_ptr<ametsuchi::TxPresenceCache> tx_cache,
-    consensus::Round initial_round,
     logger::LoggerPtr log)
     : log_(std::move(log)),
       ordering_service_(std::move(ordering_service)),
       network_client_(std::move(network_client)),
       events_subscription_(events.subscribe([this](auto event) {
         // exclusive lock
```
muratovv: Seems line is redundant

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -25,27 +25,25 @@ OnDemandOrderingGate::OnDemandOrderingGate(
...
-        std::unique_lock<std::shared_timed_mutex> lock(mutex_);
+        consensus::Round current_round_;
         visit_in_place(event,
-                       [this](const BlockEvent &block_event) {
+                       [this, &current_round_](const BlockEvent &block_event) {
```
muratovv: Maybe leave todo here about generalization:
```suggestion
                       [this, &current_round_](const auto &event) {
                         log_->debug("{}",  event);
                         current_round_ = event.round;
                       }
```
It could be possible when we generalize block empty events with toString (operator<<) method.

irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -102,10 +102,8 @@ namespace iroha {
       /**
        * Proposals for current rounds
```
muratovv: I think better to update doc here.

PR #2151, opened 2019-03-06
Change managing way for round queue

luckychess:
### Description of the Change
Pull request fixes a race with proposal retrieval

### Benefits
No race ^^

### Possible Drawbacks 
Tests code could be better for erase case.


irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -234,6 +234,14 @@ void OnDemandOrderingServiceImpl::packNextProposals(
   // new reject round
   open_round(
       {round.block_round, currentRejectRoundConsumer(round.reject_round)});
+
+  std::queue<consensus::Round>{}.swap(round_queue_);
```
muratovv: I guess better to move this piece of code to a separate function for better readability.

test/module/irohad/ordering/on_demand_os_test.cpp
```diff
@@ -197,40 +197,54 @@ TEST_F(OnDemandOsTest, DISABLED_ConcurrentInsert) {
...
+
+  consensus::Round last_round = {commit_round.block_round + 2 * proposal_limit,
+                                 commit_round.reject_round};
+  for (consensus::BlockRoundType i = commit_round.block_round;
+       i < commit_round.block_round + 2 * proposal_limit;
+       ++i) {
+    consensus::Round current_round = {i + 1, commit_round.reject_round};
+    auto proposal = os->onRequestProposal({i + 1, commit_round.reject_round});
+
+    if (current_round.block_round >= last_round.block_round) {
```
muratovv: Seems, if else ladder could be refactored with inline function and used twice for this and `EraseReject` tests.

test/module/irohad/ordering/on_demand_os_test.cpp
```diff
@@ -334,14 +334,16 @@ TEST_F(OnDemandOsTest, PassMissingTransaction) {
   auto proposal = os->onRequestProposal(target_round);
 
   // since we only sent one transaction,
-  // if the proposal is present, there is no need to check for that specific tx
+  // if the proposal is present, there is no need to check for that specific
```
lebdron: Changes in this file are not really required by clang-format and can be reverted.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -237,11 +237,9 @@ void OnDemandOrderingServiceImpl::packNextProposals(
 }
 
 void OnDemandOrderingServiceImpl::tryErase() {
-  while (round_queue_.size() > number_of_proposals_) {
-    auto &round = round_queue_.front();
-    proposal_map_.erase(round);
-    log_->info("tryErase: erased {}", round);
-    round_queue_.pop();
+  while (proposal_map_.size() > number_of_proposals_) {
```
lebdron: I would rather use the previous logic, where this loop operated only on subset of `proposal_map_` with `map.round < packNextProposals.round`.
I would suggest to first `it = proposal_map_.find(round)`, and then operate on `range(proposal_map_.begin(), it)`.

MBoldyrev: this makes sense. It could be also done by passing current round to `tryErase` and adding condition
```suggestion
  while (proposal_map_.size() > number_of_proposals_ and proposal_map_.begin()->first < current_round) {
```

lebdron: I think that the first condition should only consider the subset, e.g. `size(range(proposal_map_.begin(), it)) > number_of_proposals_` instead of `proposal_map_.size() > number_of_proposals_`.

test/module/irohad/ordering/on_demand_os_test.cpp
luckychess: Well, it's possible but these 2 code fragments are actually a bit different (using `.block_round` in the first one instead of `.reject_round` in the second) so I guess the resulting code will be too difficult to read.

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -236,12 +236,16 @@ void OnDemandOrderingServiceImpl::packNextProposals(
...
-    proposal_map_.erase(round);
-    log_->info("tryErase: erased {}", round);
-    round_queue_.pop();
+void OnDemandOrderingServiceImpl::tryErase(
+    const consensus::Round &current_round) {
+  auto current_proposal = proposal_map_.find(current_round);
+  auto proposal_range_size = boost::size(
+      boost::make_iterator_range(proposal_map_.begin(), current_proposal));
+
+  while (proposal_range_size > number_of_proposals_
```
MBoldyrev: quite a strange condition, because `proposal_range_size` is unchanged in the loop. maybe
```suggestion
  while (proposal_range_size-- > number_of_proposals_
```

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -236,12 +236,16 @@ void OnDemandOrderingServiceImpl::packNextProposals(
...
+         and proposal_map_.begin()->first < current_round) {
```
MBoldyrev: and if I get the code right, this can be converted into an assertion

PR #2152, opened 2019-03-07
Llogger: fixed patterns

MBoldyrev:
### Description of the Change

Fixes:
- child config overrides only the patterns specified in it, while the unspecified are taken from parent config.
- log message contains the verbosity of message, not the logger.
- the default patterns are used in root logger only if there is no patterns specification for it.
- verbosity padding increased to fit `warning` in default pattern.

### Benefits

Logger behaves as expected.

### DIFF
--- a/docs/source/guides/configuration.rst
+++ b/docs/source/guides/configuration.rst
@@ -60,7 +60,11 @@ Environment-specific parameters
 Logging
 -------
 
-In Iroha logging can be adjusted as granularly as you want. Each component has its own logging configuration with properties inherited from its parent, able to be overriden through config file. This means all the component loggers are organized in a tree with a single root. The relevant section of the  configuration file contains the overriden values:
+In Iroha logging can be adjusted as granularly as you want. Each component
+has its own logging configuration with properties inherited from its parent,
+able to be overriden through config file. This means all the component loggers
+are organized in a tree with a single root. The relevant section of the
+configuration file contains the overriden values:
 
 .. code-block:: json
   :linenos:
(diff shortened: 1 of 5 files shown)


docs/source/guides/configuration.rst
```diff
@@ -60,7 +60,11 @@ Environment-specific parameters
 Logging
 -------
 
-In Iroha logging can be adjusted as granularly as you want. Each component has its own logging configuration with properties inherited from its parent, able to be overriden through config file. This means all the component loggers are organized in a tree with a single root. The relevant section of the  configuration file contains the overriden values:
+In Iroha logging can be adjusted as granularly as you want. Each component
```
igor-egorov: Please split sentences with newline separator.

docs/source/guides/configuration.rst
```diff
@@ -60,7 +60,11 @@ Environment-specific parameters
...
+has its own logging configuration with properties inherited from its parent,
+able to be overriden through config file. This means all the component loggers
+are organized in a tree with a single root. The relevant section of the
+configuration file contains the overriden values:
```
igor-egorov: The last sentence is not clear to me. Maybe it can be improved.

docs/source/guides/configuration.rst
```diff
@@ -89,6 +93,14 @@ In Iroha logging can be adjusted as granularly as you want. Each component has i
...
 
 Every part of this config section is optional.
-- ``level`` sets the verbosity. Available valies are (in decreasing verbosity order): trace, debug, info, warning, error, critical.
-- ``patterns`` controls the formatting of each log string for different verbosity levels. Each value overrides the less verbose levels too. So in the example above, the "don't panic" pattern also applies to info and warning levels, and the trace level pattern is the only one that is not initialized in the config (it will be set to default hardcoded value).
-- ``children`` describes the overrides of child nodes. The keys are the names of the components, and the values have the same syntax and semantics as the root log configuration.
+
+- ``level`` sets the verbosity. Available valies are (in decreasing verbosity
+  order): trace, debug, info, warning, error, critical.
+- ``patterns`` controls the formatting of each log string for different
+  verbosity levels. Each value overrides the less verbose levels too. So in the
```
igor-egorov: Please put each sentence on its own line.

igor-egorov: Each value overrides the less verbose levels too.
Probably can be rephrased. It looks like "values" and "levels" are the objects of the same kind. Not clear.

docs/source/guides/configuration.rst
```diff
@@ -89,6 +93,14 @@ In Iroha logging can be adjusted as granularly as you want. Each component has i
...
+  example above, the "don't panic" pattern also applies to info and warning
```
igor-egorov: looks like "don't panic" is not a pattern, but a prefix of a debug message when the name of the pattern is "debug"

libs/logger/logger_spdlog.cpp
```diff
@@ -52,9 +52,9 @@ namespace logger {
     static LogPatterns default_patterns;
     if (not is_initialized.test_and_set()) {
       default_patterns.setPattern(
-          LogLevel::kTrace, R"([%Y-%m-%d %H:%M:%S.%F] [th:%t] [%5l] [%n]: %v)");
+          LogLevel::kTrace, R"([%Y-%m-%d %H:%M:%S.%F][th:%t][%=7l][%n]: %v)");
```
lebdron:
```suggestion
          LogLevel::kTrace, R"([%Y-%m-%d %H:%M:%S.%F][th:%t][%=8l][%n]: %v)");
```
Because `critical` is 8 characters.

PR #2158, opened 2019-03-12
Update config template for docker deployment

lebdron:
### Description of the Change

- Update config template to include new optional parameters such as `mst_expiration_time`
- Make `docker-compose pull` an explicit step - https://github.com/ansible/ansible/issues/47978
- Add missing prepared transaction setting for postgres in compose template
- Stop execution of `config_gen.sh` on error

### Benefits
New parameters support, possible to see if something gone wrong during the execution of `config_gen.sh`

### Possible Drawbacks 
None?


deploy/ansible/roles/iroha-docker/templates/config.docker.j2
```diff
@@ -7,4 +7,13 @@
   "proposal_delay" : {{ iroha_proposal_delay }},
   "vote_delay" : {{ iroha_vote_delay }},
   "mst_enable" : {{ iroha_mst_enable }}
+  {% if iroha_mst_expiration_time is defined %}
```
bakhtin: These new variables should also be defined in `defaults/main.yml`

deploy/ansible/roles/iroha-docker/templates/docker-compose.yml.j2
```diff
@@ -39,6 +39,7 @@ services:
       - psql_storage-{{ node.human_hostname }}:/var/lib/postgresql/data
     networks:
       - iroha-db-net
+    command: -c max_prepared_transactions=100
```
bakhtin: Shall we make the parameter configurable?

deploy/ansible/roles/iroha-docker/templates/config.docker.j2
lebdron: Since they are optional and have defaults values hardcoded, maybe I can put commented out examples in `main.yml`?

bakhtin: Yes, this would work

PR #2159, opened 2019-03-12
Add info about vulnerabilities

neewy:
### Description of the Change

This pull request contains additional contribution docs on how to submit discovered vulnerabilities. 

### Benefits

More docs.

### Possible Drawbacks 

None

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,8 @@ The following is a short set of guidelines for contributing to Iroha.
 
 ##### [How Can I Contribute?](#how-can-i-contribute-1)
 
-- [Reporting bugs](#reporting-bugs)
+- [Reporting Bugs](#reporting-bugs)
+- [Reporting Vulnerabilities](#reporting-vulnerabilities)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -42,6 +43,17 @@ To submit a bug, [create new issue](https://jira.hyperledger.org/secure/CreateIs
 | Environment             | Your OS, device's specs, Virtual Environment if you use one, version of Iroha etc. |
 
 
+### Reporting Vulnerabilities
+
+While we try to be proactive in preventing security problems, we do not assume they’ll never come up.
```
LiraLemur: Double negative seems confusing. Maybe put it like "Although we try to be proactive in preventing security problems, we cannot ignore the possibility" or smth like that

neewy: TBH, I reused text from Wordpress website here: https://make.wordpress.org/core/handbook/testing/reporting-security-vulnerabilities/

LiraLemur: Ok ) I've just got confused myself when I read it.

PR #2162, opened 2019-03-13
Documentation/fix build docs

vadimreutskiy:
### Description of the Change

Changed a few small things in the documentation:
- Added note about no need to compile Iroha to start using it.
- Fixed list format at the "Building" documentation page
- Added example for key generation procedure

### Benefits

Users will struggle less with their frequent questions.

### Possible Drawbacks 

Not in this case


neewy:
Just a hint for the future: if you update the docs include [ci skip] to your commits so that CI is not triggered for your build

PR #2163, opened 2019-03-13
Synchronizer: observable passthrough to ChainValidator

MBoldyrev:
### Description of the Change

[IR-361](https://jira.hyperledger.org/browse/IR-361)

This change makes the synced blocks observable provided by block loader transport be passed to ChainValidator directly (hot observable), without using the stored collection. The collection is still kept in this PR, but it will be removed when all the necessary changes (like this one) are made so that no component needs it.

This is part of closed #2142.

### Benefits

One step towards removing in memory blocks accumulation during synchronization. Together with  #2160 provides early cancellation of failing block sequence loading.


irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -74,49 +74,48 @@ namespace iroha {
...
-              [&blocks](auto block) { blocks.push_back(block); });
-          if (blocks.empty()) {
-            log_->info("Downloaded an empty chain");
-            continue;
-          } else {
-            log_->info("Successfully downloaded {} blocks", blocks.size());
-          }
+      // TODO andrei 17.10.18 IR-1763 Add delay strategy for loading blocks
+      for (const auto &public_key : msg.public_keys) {
+        auto storage = getStorage().value_or(nullptr);
```
luckychess: Do we need `value_or()` call?

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -74,49 +74,48 @@ namespace iroha {
...
+        if (not storage) {
+          return boost::none;
+        }
 
+        std::vector<std::shared_ptr<shared_model::interface::Block>> blocks;
+        auto network_chain =
+            block_loader_->retrieveBlocks(height, public_key)
+                .tap(
```
luckychess: Why `tap` and not `subscribe` as before?

irohad/synchronizer/impl/synchronizer_impl.hpp
```diff
@@ -50,7 +50,6 @@ namespace iroha {
        */
       boost::optional<SynchronizationEvent> downloadMissingBlocks(
           const consensus::VoteOther &msg,
-          std::unique_ptr<ametsuchi::MutableStorage> storage,
```
luckychess, line 53: Please fix method docs as well.

irohad/synchronizer/impl/synchronizer_impl.cpp
MBoldyrev: I use it to get the pointer. If you have a better idea - please provide it.

irohad/synchronizer/impl/synchronizer_impl.cpp
MBoldyrev: The blocks are now accumulated on their way to ChainValidator, so `tap` seems the right operator.

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -74,49 +74,48 @@ namespace iroha {
     boost::optional<SynchronizationEvent>
     SynchronizerImpl::downloadMissingBlocks(
         const consensus::VoteOther &msg,
-        std::unique_ptr<ametsuchi::MutableStorage> storage,
         const shared_model::interface::types::HeightType height) {
       auto expected_height = msg.round.block_round;
 
-      // while blocks are not loaded and not committed
-      while (true) {
```
lebdron: This loop has to be restored since it can affect consensus.

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -55,6 +58,14 @@ class SynchronizerTest : public ::testing::Test {
     consensus_gate = std::make_shared<MockConsensusGate>();
     block_query = std::make_shared<::testing::NiceMock<MockBlockQuery>>();
 
+    ledger_peers = std::make_shared<PeerList>();
+    for (int i = 0; i < 10; ++i) {
+      ledger_peer_keys.emplace_back(
+          shared_model::crypto::DefaultCryptoAlgorithmType::generateKeypair());
```
lebdron, line 66: It would be great to avoid specific implementation here.

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -79,23 +90,18 @@ class SynchronizerTest : public ::testing::Test {
...
-                     .createdTime(time)
-                     .build()
-                     .signAndAddSignature(
-                         shared_model::crypto::DefaultCryptoAlgorithmType::
-                             generateKeypair())
-                     .finish();
-    return std::make_shared<shared_model::proto::Block>(std::move(block));
+    shared_model::proto::UnsignedWrapper<shared_model::proto::Block> block{
+        TestUnsignedBlockBuilder().height(height).createdTime(time).build()};
+    for (const auto &key  : ledger_peer_keys) {
```
lebdron:
```suggestion
    for (const auto &key : ledger_peer_keys) {
```

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -302,19 +378,26 @@ TEST_F(SynchronizerTest, ExactlyThreeRetrievals) {
 TEST_F(SynchronizerTest, RetrieveBlockTwoFailures) {
   DefaultValue<expected::Result<std::unique_ptr<MutableStorage>, std::string>>::
       SetFactory(&createMockMutableStorage);
-  EXPECT_CALL(*mutable_factory, createMutableStorage()).Times(1);
+  EXPECT_CALL(*mutable_factory, createMutableStorage()).Times(3);
```
lebdron: As I remeber, this test was made for `while(true)` case, so with the new change
https://github.com/hyperledger/iroha/blob/9618614d78c5d13c39786c1002082a6d7b0770c7/test/module/irohad/synchronizer/synchronizer_test.cpp#L62
it should try to retrieve the chain at least 11 times in this test.

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -110,14 +116,87 @@ class SynchronizerTest : public ::testing::Test {
...
+inline ::testing::Matcher<Chain> ChainEq(
+    std::vector<std::shared_ptr<shared_model::interface::Block>>
+        expected_chain) {
+  return ::testing::MakeMatcher(new ChainMatcher(expected_chain));
+}
+
+void mutableStorageExpectChain(
+    iroha::ametsuchi::MockMutableFactory &mutable_factory,
+    std::vector<std::shared_ptr<shared_model::interface::Block>> chain) {
+  const bool must_create_storage = not chain.empty();
```
lebdron, line 177: `must_create_storage` sounds misleading, because the code below uses `WillRepeatedly` if `must_create_storage` is false. Should it be `must_create_once`?

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -110,14 +116,87 @@ class SynchronizerTest : public ::testing::Test {
...
+    }
+    return expected::Value<std::unique_ptr<MutableStorage>>{
+        std::move(mutable_storage)};
+  };
+  if (must_create_storage) {
+    EXPECT_CALL(mutable_factory, createMutableStorage())
+        .WillOnce(::testing::Invoke(create_mutable_storage));
+  } else {
+    EXPECT_CALL(mutable_factory, createMutableStorage())
+        .WillRepeatedly(::testing::Invoke(create_mutable_storage));
```
lebdron, line 202: Otherwise it should be `Times(0)` here.

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -117,6 +117,7 @@ namespace iroha {
           }
         }
       }
+      return boost::none;
```
lebdron: Code should be removed, as it is placed after `while(true)`.

test/module/irohad/synchronizer/synchronizer_test.cpp
```diff
@@ -300,21 +379,28 @@ TEST_F(SynchronizerTest, ExactlyThreeRetrievals) {
  * @then it will try until success
  */
 TEST_F(SynchronizerTest, RetrieveBlockTwoFailures) {
```
lebdron: Maybe `TwoFailures` can be updated to reflect the new logic.

PR #2167, opened 2019-03-13
Fix race bug in command service

muratovv:

### Description of the Change
Fix bug when not committed/rejected transaction will not pass to Iroha's pipeline because it is a "replay".

### Benefits
Add fix for the problem. The idea of the fix to check only final statuses in the ram cache in command service. If there is a final status we will decline further checking of the transaction. Also, thanks @lebdron for cooperation in founding the bug.

### Possible Drawbacks 
Lack of tests. I hope we will discuss with reviewers about how to write the correct test for the situation.

### DIFF
--- a/irohad/torii/impl/command_service_impl.cpp
+++ b/irohad/torii/impl/command_service_impl.cpp
@@ -205,10 +205,12 @@ namespace iroha {
               (*found)->get(),
               [](const auto &final_responses)
                   -> std::enable_if_t<
-                      FinalStatusValue<decltype(final_responses), bool>> {
-                return true;
-              },
-              [](const auto &rest_responses) { return false; });
+                      FinalStatusValue<decltype(final_responses)>,
+                      bool> { return true; },
+              [](const auto &rest_responses)
+                  -> std::enable_if_t<
+                      not FinalStatusValue<decltype(rest_responses)>,
+                      bool> { return false; });
         }
 
         if (has_final_status) {
--- a/irohad/torii/impl/command_service_impl.cpp
+++ b/irohad/torii/impl/command_service_impl.cpp
@@ -195,62 +195,38 @@ namespace iroha {
       const auto status_issuer = "ToriiBatchProcessor";
       const auto &txs = batch->transactions();
 
-      bool cache_has_at_least_one_tx{false};
-      bool batch_has_mst_pending_tx{false};
-      std::tie(cache_has_at_least_one_tx, batch_has_mst_pending_tx) =
-          // this accumulate can be split on two parts to perform visit_in_place
-          // two times - one for cache lookup with booleans initialization and
-          // another for statuses pushing. That can allow to move a part of code
-          // to a separate method for simplification
-          std::accumulate(
-              txs.begin(),
-              txs.end(),
-              std::make_pair<bool, bool>(false, false),
-              [this, &status_issuer](std::pair<bool, bool> lookup_result,
-                                     const auto &tx) {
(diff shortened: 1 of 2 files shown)


irohad/torii/impl/command_service_impl.cpp
```diff
@@ -174,62 +174,38 @@ namespace iroha {
...
+            if (res) {
+              return res;
+            }
 
-      if (cache_has_at_least_one_tx and not batch_has_mst_pending_tx) {
-        // If a non-persistent cache says that a transaction has pending status
-        // that means we have to check persistent cache too.
-        // Non-persistent cache might be overflowed and mst replay become
-        // possible without checking persistent cache.
+            auto on_final_status = [] { return true; };
```
MBoldyrev: unused?

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -174,62 +174,38 @@ namespace iroha {
...
-                      },
-                      [this, &found, &status_issuer](const auto &status) {
-                        this->pushStatus(status_issuer, *found);
-                      });
-                  lookup_result.first = true;
-                }
-                return lookup_result;
-              });
+      bool has_final_status{false};
+      has_final_status = std::accumulate(
```
MBoldyrev: Could you rewrite it with `for` or `std::any_of`? `std::accumulate` with early-return lambda looks somewhat ill-formed.

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -174,62 +174,34 @@ namespace iroha {
...
-        // from further propagation.
+      bool has_final_status{false};
+
+      for (auto tx : txs) {
+        const auto &tx_hash = tx->hash();
+        if (auto found = cache_->findItem(tx_hash)) {
+          log_->debug("Found in cache: {}", **found);
+          has_final_status = iroha::visit_in_place(
+              (*found)->get(),
+              [](const shared_model::interface::CommittedTxResponse &) {
```
lebdron: Another option is to reuse `FinalStatusValue`:
https://github.com/hyperledger/iroha/blob/204a61960b772e85f6a38fa47ddefdec5bf99d61/irohad/torii/impl/command_service_impl.cpp#L152-L159

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -174,62 +174,34 @@ namespace iroha {
...
+          break;
+        }
+      }
 
-        // If non-persistent cache does not contain any info about a
-        // transaction, then we just check persistent cache.
-        log_->warn("Replayed batch would not be served. {}", *batch);
+      if (has_final_status) {
+        // if the transaction or batch has appeared in the cache with final
+        // status this guarantees the transaction was passed to a consensus
```
lebdron:
```suggestion
        // status guarantees that the transaction was passed to consensus
```

irohad/torii/impl/command_service_impl.cpp
```diff
@@ -174,62 +174,34 @@ namespace iroha {
...
+        if (has_final_status) {
+          break;
+        }
+      }
 
-        // If non-persistent cache does not contain any info about a
-        // transaction, then we just check persistent cache.
-        log_->warn("Replayed batch would not be served. {}", *batch);
+      if (has_final_status) {
+        // if the transaction or batch has appeared in the cache with final
```
lebdron:
```suggestion
        // presence of the transaction or batch in the cache with final
```

PR #2170, opened 2019-03-14
Fix hash set usage in OS. Fix block leaks in YacGate and Simulator

lebdron:
### Description of the Change
- Fix duplicate transactions in proposals due to missing `BatchHashEquality` comparator
- Fix block leaks in Simulator and YacGate

### Benefits
- No duplicate transactions in proposals
- Less memory usage

### Possible Drawbacks 
None

### Usage Examples or Tests
YacGateTest.CacheReleased, OnDemandOsTest.DuplicateTxTest

### DIFF
--- a/irohad/consensus/yac/impl/yac_gate_impl.cpp
+++ b/irohad/consensus/yac/impl/yac_gate_impl.cpp
@@ -55,6 +55,9 @@ namespace iroha {
 
         if (not event.round_data) {
           current_block_ = boost::none;
+          // previous block is committed to block storage, it is safe to clear
+          // the cache
+          consensus_result_cache_->release();
           log_->debug("Agreed on nothing to commit");
         } else {
           current_block_ = event.round_data->block;
--- a/irohad/consensus/yac/impl/yac_gate_impl.cpp
+++ b/irohad/consensus/yac/impl/yac_gate_impl.cpp
@@ -57,6 +57,8 @@ namespace iroha {
           current_block_ = boost::none;
           // previous block is committed to block storage, it is safe to clear
           // the cache
+          // TODO 2019-03-15 andrei: IR-405 Subscribe BlockLoaderService to
+          // BlockCreator::onBlock
           consensus_result_cache_->release();
           log_->debug("Agreed on nothing to commit");
         } else {
--- a/irohad/ordering/impl/on_demand_ordering_service_impl.hpp
+++ b/irohad/ordering/impl/on_demand_ordering_service_impl.hpp
@@ -9,14 +9,13 @@
 #include "ordering/on_demand_ordering_service.hpp"
 
 #include <map>
-#include <queue>
 #include <shared_mutex>
-#include <unordered_map>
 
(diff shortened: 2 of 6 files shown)


irohad/ordering/impl/on_demand_ordering_service_impl.hpp
```diff
@@ -9,14 +9,13 @@
...
 #include <map>
-#include <queue>
 #include <shared_mutex>
-#include <unordered_map>
 
 #include <tbb/concurrent_unordered_set.h>
 #include "interfaces/iroha_internal/unsafe_proposal_factory.hpp"
 #include "logger/logger_fwd.hpp"
 #include "multi_sig_transactions/hash.hpp"
+#include "multi_sig_transactions/state/mst_state.hpp"
```
muratovv: I think better to leave a todo about changing a place of PointerBatchHasher and BatchHashEquality

test/module/irohad/ordering/on_demand_os_test.cpp
```diff
@@ -343,3 +343,21 @@ TEST_F(OnDemandOsTest, SeveralTransactionsOneCommited) {
...
+  auto txs1 = generateTransactions({1, 2}, now);
+  os->onBatches(txs1);
+
+  auto txs2 = generateTransactions({1, 2}, now);
+  os->onBatches(txs2);
+  os->onCollaborationOutcome(commit_round);
+  auto proposal = os->onRequestProposal(target_round);
+
+  ASSERT_EQ(1, boost::size((*proposal)->transactions()));
+}
```
muratovv: I think better to add/updated existing test for simulator and check block invalidation.

lebdron: It it a private field in Simulator, which is used internally and not accessible with API of Simulator. Could you suggest how to make the test?

irohad/simulator/impl/simulator.cpp
```diff
@@ -149,6 +149,8 @@ namespace iroha {
                                             rejected_hashes);
       crypto_signer_->sign(*block);
 
+      last_block.reset();
```
MBoldyrev: please add a TODO to refactor `processVerifiedProposal` in such a way that it accepts both verified proposal and last block, and is called from `processProposal` which provides it the last block as an argument. This will let us remove the ugly `last_block` which is now sometimes nulled and never tested for null...

irohad/consensus/yac/impl/yac_gate_impl.cpp
```diff
@@ -55,6 +55,9 @@ namespace iroha {
 
         if (not event.round_data) {
           current_block_ = boost::none;
+          // previous block is committed to block storage, it is safe to clear
+          // the cache
+          consensus_result_cache_->release();
```
MBoldyrev: please add a TODO to use `onBlock()` subscription in `BlockLoaderService` instead of unneeded complications with `consensus_result_cache_`

PR #2178, opened 2019-03-18
Storage: avoid block copying

MBoldyrev:
### Description of the Change

This change makes most block manipulation functions of storage take const blocks by shared pointer.

### Benefits

Blocks can be reused - no need to copy.

### Possible Drawbacks 

Negligible additional cost of shared pointers creation compared to just references.


irohad/validation/impl/chain_validator_impl.cpp
```diff
@@ -31,7 +31,7 @@ namespace iroha {
 
       return storage.apply(
           blocks,
-          [this](const auto &block, auto &queries, const auto &top_hash) {
+          [this](auto block, auto &queries, const auto &top_hash) {
```
igor-egorov: let's try 
```suggestion
          [this](auto &block, auto &queries, const auto &top_hash) {
```

MBoldyrev: sadly, gcc didn't like your idea..

PR #2179, opened 2019-03-18
Storage: notify block commit after actual commit

MBoldyrev:
### Description of the Change

[IR-406](https://jira.hyperledger.org/browse/IR-406)

Make storage notify subscribers of newly committed blocks only after they have been actually applied to the WSV. Also, send these notifications only if the commit to WSV succeeded.

### Benefits

Hopefully fixes the unexpected old state reported through query after a block that updates it is reported committed.


test/module/irohad/ametsuchi/ametsuchi_test.cpp
```diff
@@ -511,6 +511,84 @@ TEST_F(AmetsuchiTest, TestRestoreWSV) {
...
+          .transactions(
+              std::vector<shared_model::proto::Transaction>({genesis_tx}))
+          .height(1)
+          .prevHash(shared_model::crypto::Sha3_256::makeHash(
+              shared_model::crypto::Blob("")))
+          .createdTime(iroha::time::now())
+          .build();
+  apply(storage, genesis_block);
+
+  // std::vector<shared_model::proto::Transaction> txs;
```
luckychess: Remove plz.

test/module/irohad/ametsuchi/ametsuchi_test.cpp
```diff
@@ -511,6 +511,84 @@ TEST_F(AmetsuchiTest, TestRestoreWSV) {
...
+
+  auto expected_block =
+      TestBlockBuilder()
+          .transactions(
+              std::vector<shared_model::proto::Transaction>({add_ast_tx}))
+          .height(1)
+          .prevHash(genesis_block.hash())
+          .createdTime(iroha::time::now())
+          .build();
+  // auto expected_block =
```
luckychess: And here as well.

PR #2182, opened 2019-03-19
Redirect postgres log messages to spdlog

lebdron:
### Description of the Change
Print notices from postgres to spdlog instead of std::cerr

### Benefits
Logging is controlled consistently with other logs

### Possible Drawbacks 
`reinterpret_cast`...

### DIFF
--- a/irohad/ametsuchi/impl/storage_impl.cpp
+++ b/irohad/ametsuchi/impl/storage_impl.cpp
@@ -6,8 +6,9 @@
 #include "ametsuchi/impl/storage_impl.hpp"
 
 #include <soci/postgresql/soci-postgresql.h>
-#include <boost/algorithm/string/replace.hpp>
+#include <boost/algorithm/string.hpp>
 #include <boost/format.hpp>
+#include <boost/range/algorithm/replace_if.hpp>
 #include "ametsuchi/impl/flat_file/flat_file.hpp"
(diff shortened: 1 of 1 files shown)


irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -44,6 +45,17 @@ namespace {
     }
   }
 
+  std::string formatPostgresMessage(const char *message) {
+    std::string formatted_message(message);
+    boost::replace_all(formatted_message, "\n", " ");
```
MBoldyrev: maybe replace "\r" too?

irohad/ametsuchi/impl/storage_impl.cpp
```diff
@@ -44,6 +45,17 @@ namespace {
...
+    return formatted_message;
+  }
+
+  void processPqNotice(void *arg, const char *message) {
+    auto *log = reinterpret_cast<logger::Logger *>(arg);
```
igor-egorov, line 56: Please check if dynamic_cast can be employed here instead of reinterpret_cast.

MBoldyrev: no, it cannot be employed with `void*`

lebdron:
Linux CI has passed, Mac CI has a known issue.

PR #2188, opened 2019-03-20
Replace TransportBuilder with ProtoBlocksQueryFactory in QueryService

igor-egorov:

### Description of the Change

Remove legacy code.

### Benefits

A step towards bug resolution with maximum batch size.

### Possible Drawbacks 

Unknown

### Usage Examples or Tests 

query_service_test
find_fuzzing
torii_queries_test
torii_service_query_test


igor-egorov:
>do we need honest validators in all these module tests?
Not sure. I left it as is because the main focus was not to break tests, which were already using the real instance of ProtoQueryFactory (just one more dependency was added - ProtoBlocksQueryFactory).

PR #2195, opened 2019-03-25
CommandServiceImpl: fix observer lifetime problem

MBoldyrev:
### Description of the Change

This is a quite dirty fix of the issue discovered in failing `command_service_replay_test`. The problem is that calling `unsubscribe()` on a subscription from `rxcpp::subjects::synchronize<T, rxcpp::observe_on_one_worker>` does not block until a currently running subscribed observer finishes. This lead to `CommandServiceImpl` destruction while there was still running an observer on statuses that was using  `CommandServiceImpl::cache_` member, which got invalidated.

There is a task to find a better way to handle this type of problems: [IR-426](https://jira.hyperledger.org/browse/IR-426).

### Benefits

Fix.

### Possible Drawbacks 

Need to capture everything needed for a subscriber by value.

`command_service_replay_test`

Wait till the task finishes. Calling `unsubscribe()` allows it if we have a dedicated thread for our subscriber, but not when we reuse one.

### DIFF
--- a/irohad/torii/impl/command_service_impl.cpp
+++ b/irohad/torii/impl/command_service_impl.cpp
@@ -36,7 +36,8 @@ namespace iroha {
           log_(std::move(log)) {
       // Notifier for all clients
       status_subscription_ = status_bus_->statuses().subscribe(
-          // TODO mboldyrev IR-426 capture this, wait termination in destructor
+          // TODO mboldyrev IR-426 research approaches to the problem of member
+          // observer lifetime.
           [cache = cache_](auto response) {
             // find response for this tx in cache; if status of received
             // response isn't "greater" than cached one, dismiss received one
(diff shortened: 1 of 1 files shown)


irohad/torii/impl/command_service_impl.cpp
```diff
@@ -35,19 +35,20 @@ namespace iroha {
           tx_presence_cache_(std::move(tx_presence_cache)),
           log_(std::move(log)) {
       // Notifier for all clients
-      status_subscription_ =
-          status_bus_->statuses().subscribe([this](auto response) {
+      status_subscription_ = status_bus_->statuses().subscribe(
+          // TODO mboldyrev IR-426 capture this, wait termination in destructor
```
lebdron: >capture this, wait termination in destructor

Not sure if blocking is the rx way to solve this. I would rather rephrase it as "research and list the approaches to this problem".

lebdron (approved):

igor-egorov (approved):

PR #2197, opened 2019-03-26
Ordering: batches propagation for (reject, commit) case

MBoldyrev:

### Description of the Change

Recently @luckychess discovered a round transition case not handled by our ordering service - when there is a reject followed by a commit. This change brings this transition equal to others.

### Benefits

Correct batches propagation for (reject, commit) case.

### Possible Drawbacks 

More network communications.

Define a single batches receiver based on current state (using linear scheme without commit,reject pair).

### DIFF
--- a/irohad/main/impl/on_demand_ordering_init.cpp
+++ b/irohad/main/impl/on_demand_ordering_init.cpp
@@ -158,9 +158,9 @@ namespace iroha {
          * irohad/ordering/impl/on_demand_connection_manager.cpp
          *
          *    0 1 2         0 1 2         0 1 2         0 1 2
-         *  0 o x v 0     0 o . .       0 o x .       0 o . .
-         *  1 . . . 1     1 x v .       1 v . .       1 x . .
-         *  2 . . . 2     2 . . .       2 . . .       2 v . .
+         *  0 o x v       0 o . .       0 o x .       0 o . .
+         *  1 . . .       1 x v .       1 v . .       1 x . .
+         *  2 . . .       2 . . .       2 . . .       2 v . .
          * RejectReject  CommitReject  RejectCommit  CommitCommit
          *
          * o - current round, x - next round, v - target round
--- a/irohad/main/impl/on_demand_ordering_init.cpp
+++ b/irohad/main/impl/on_demand_ordering_init.cpp
@@ -157,14 +157,17 @@ namespace iroha {
          * See detailed description in
          * irohad/ordering/impl/on_demand_connection_manager.cpp
          *
-         *   0 1 2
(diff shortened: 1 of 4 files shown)


irohad/main/impl/on_demand_ordering_init.cpp
```diff
@@ -159,21 +159,24 @@ namespace iroha {
          *
          *   0 1 2
          * 0 o x v
-         * 1 x v .
+         * 1 v v .
```
lebdron: 1,0 should be something like `x v`, because it gets closed when the round starts, but it will be updated with new transactions if reject happens (i.e. transition to 1,1).

irohad/main/impl/on_demand_ordering_init.cpp
```diff
@@ -159,21 +159,24 @@ namespace iroha {
...
          * 2 v . .
          *
-         * v, round 0 - kCurrentRoundRejectConsumer
-         * v, round 1 - kNextRoundRejectConsumer
-         * v, round 2 - kNextRoundCommitConsumer
-         * o, round 0 - kIssuer
+         * v, round 0,2 - kRejectRejectConsumer
+         * v, round 1,1 - kRejectCommitConsumer
```
lebdron:
```suggestion
         * v, round 1,1 - kCommitRejectConsumer
```
because 0,0 ->commit 1,0 ->reject 1,1

irohad/main/impl/on_demand_ordering_init.cpp
```diff
@@ -159,21 +159,24 @@ namespace iroha {
...
+         * v, round 1,0 - kCommitRejectConsumer
```
lebdron:
```suggestion
         * v, round 1,0 - kRejectCommitConsumer
```
0,0 ->reject 0,1 ->commit 1,0

irohad/ordering/impl/on_demand_connection_manager.cpp
```diff
@@ -44,20 +44,20 @@ void OnDemandConnectionManager::onBatches(CollectionType batches) {
...
-   * This can be visualised as a diagram, where:
-   * o - current round, x - next round, v - target round
+   * There are 4 possibilities - all combinations of commits and rejects in the
+   * following two rounds. This can be visualised as a diagram, where: o -
+   * current round, x - next round, v - target round
    *
    *   0 1 2
    * 0 o x v
-   * 1 x v .
+   * 1 v v .
```
lebdron: Same question with 1,0 and `x v`

lebdron (approved):

luckychess (approved):

PR #2198, opened 2019-03-26
Ordering Service max proposal size

igor-egorov:

### Description of the Change

Ordering Service would not produce proposal bigger than max_proposal_size.

The issue was:
max_proposal_size = 3
Two batches of two transactions arrive.
Both will be processed.
The resulting proposal will have 4 transactions inside.

### Benefits

Correct behavior.

### Usage Examples or Tests 

All the tests

### DIFF
--- a/irohad/ordering/impl/on_demand_ordering_service_impl.cpp
+++ b/irohad/ordering/impl/on_demand_ordering_service_impl.cpp
@@ -116,8 +116,9 @@ getTransactions(size_t requested_tx_amount,
   std::vector<std::shared_ptr<shared_model::interface::Transaction>> collection;
 
   auto it = batch_collection.begin();
-  for (;
-       it != batch_collection.end() and collection.size() < requested_tx_amount;
+  for (; it != batch_collection.end()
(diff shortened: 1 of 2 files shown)


MBoldyrev (approved):

irohad/ordering/impl/on_demand_ordering_service_impl.cpp
```diff
@@ -116,8 +116,9 @@ getTransactions(size_t requested_tx_amount,
   std::vector<std::shared_ptr<shared_model::interface::Transaction>> collection;
 
   auto it = batch_collection.begin();
-  for (;
-       it != batch_collection.end() and collection.size() < requested_tx_amount;
+  for (; it != batch_collection.end()
+       and collection.size() + boost::size((*it)->transactions())
```
muratovv, line 112: I suggest using brackets for plus:
```suggestion
       and (collection.size() + boost::size((*it)->transactions()))
```

muratovv (approved):

igor-egorov:
CI failed due to known issue of 9.0 apple clang.

PR #2200, opened 2019-03-27
Resending votes in consensus

muratovv:
### Description of the Change
The PR introduce changes for resending votes. The feature is required due to we want to simulate the asynchronous environment among YAC processors.

### Design
* Fix voting interface of consensus: now it requires value types for propagation.
* Add new interface for proto transport level with has feedback about the status of sending
* Add wrapper which can resend votes if it is required

### Possible Drawbacks 
* Lack of new proto transport test with different statuses.
* Probably there are missing two things: the decline of existed calls and attempt limitation.

### DIFF
--- a/irohad/consensus/yac/cluster_order.hpp
+++ b/irohad/consensus/yac/cluster_order.hpp
@@ -39,7 +39,7 @@ namespace iroha {
         /**
          * Provide current leader peer
          */
-        const shared_model::interface::Peer &currentLeader();
+        const std::shared_ptr<shared_model::interface::Peer> &currentLeader();
 
         /**
          * Switch to next peer as leader
--- a/irohad/consensus/yac/transport/impl/network_impl.cpp
+++ b/irohad/consensus/yac/transport/impl/network_impl.cpp
@@ -32,8 +32,9 @@ namespace iroha {
         handler_ = handler;
       }
 
-      void NetworkImpl::sendState(const shared_model::interface::Peer &to,
-                                  const std::vector<VoteMessage> &state) {
+      YacNetworkWithFeedBack::SendStateReturnType
+      NetworkImpl::sendState(const shared_model::interface::Peer &to,
+                             const std::vector<VoteMessage> &state) {
         createPeerConnection(to);
 
         proto::State request;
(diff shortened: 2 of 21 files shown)


irohad/consensus/yac/cluster_order.hpp
```diff
@@ -39,7 +39,7 @@ namespace iroha {
         /**
          * Provide current leader peer
          */
-        const shared_model::interface::Peer &currentLeader();
+        const std::shared_ptr<shared_model::interface::Peer> &currentLeader();
```
MBoldyrev, line 42: I would make
```suggestion
        std::shared_ptr<shared_model::interface::Peer> currentLeader();
```
as returning a reference to a local object is scary, and then
```suggestion
        std::shared_ptr<const shared_model::interface::Peer> currentLeader();
```
as we do not want anyone to change the object (no matter it is an interface)

irohad/consensus/yac/transport/impl/network_impl.cpp
```diff
@@ -42,12 +43,60 @@ namespace iroha {
...
+          using namespace grpc;
+          std::vector<int> codes = {StatusCode::CANCELLED,
+                                    StatusCode::INVALID_ARGUMENT,
+                                    StatusCode::UNAUTHENTICATED,
+                                    StatusCode::RESOURCE_EXHAUSTED,
+                                    StatusCode::ABORTED,
+                                    StatusCode::UNIMPLEMENTED,
+                                    StatusCode::UNAVAILABLE,
+                                    StatusCode::DATA_LOSS};
+          return std::any_of(codes.begin(), codes.end(), [code](auto val) {
```
MBoldyrev, line 82:
```suggestion
          return std::find(codes.begin(), codes.end(), code) != codes.end();
```
or, if using set,
```suggestion
          return codes.find(code) != codes.end();
```
which is faster

irohad/consensus/yac/transport/impl/network_impl.cpp
```diff
@@ -42,12 +43,60 @@ namespace iroha {
...
+
+      YacNetworkWithFeedBack::ValueStateReturnType
+      NetworkImpl::makeSendStateStatus(const grpc::Status &status) {
+        auto is_ok = [](const auto &code) {
+          return code == grpc::StatusCode::OK;
+        };
+
+        auto is_troubles_with_recipient = [](const auto &code) {
+          using namespace grpc;
+          std::vector<int> codes = {StatusCode::CANCELLED,
```
MBoldyrev, line 74:
```suggestion
          std::set<StatusCode> codes = {StatusCode::CANCELLED,
```
performs better lookup

test/integration/acceptance/basic_mst_state_propagation.cpp
```diff
@@ -82,8 +82,9 @@ class BasicMstPropagationFixture : public AcceptanceFixture {
  * @when such transaction is sent to one of two iroha peers in the network
  * @then that peer propagates MST state to another peer
  */
+// TODO: 2019-03-27 @muratovv the test is blocked by IR-276
 TEST_F(BasicMstPropagationFixture,
-       MstStateOfTransactionWithoutAllSignaturesPropagtesToOtherPeer) {
+       DISABLED_MstStateOfTransactionWithoutAllSignaturesPropagtesToOtherPeer) {
```
MBoldyrev, line 87: Why is the test disabled? Please provide a reason. Taking into account another failing test related to MST (see CI), I suggest possible investigating MST regression - it could cause this test failing.

MBoldyrev (changes requested):
Please re-enable the test or provide a reason of its disabling. Also, there might be a problem as MST regression, as there is another MST test showing spontaneous failures - it is worth further investigation.

irohad/consensus/yac/transport/impl/network_impl.cpp
muratovv: Seems that set is more appropriate because it is a more reliable way to stay unique constants.

test/integration/acceptance/basic_mst_state_propagation.cpp
muratovv: Seems, the branch was outdated and now basic_mst_state_propagation test absent.

PR #2202, opened 2019-03-29
MST: Limit stored transactions

MBoldyrev:
### Description of the Change

[IR-431](https://jira.hyperledger.org/browse/IR-431)

`MstState` gets a limit of transactions it can store. The value is configured through iroha config file.

### Benefits

Prevent unlimited growth of MST storage.

### Possible Drawbacks 

Some batches may get dropped. Malicious peer can keep other peers' MST states filled with fake batches, so that the legitimate batches will be dropped.

Use something like circular buffer to allow newer batches push out the oldest ones.


igor-egorov:
A comment before the code review regarding an **alternative design**:
> Use something like circular buffer to allow newer batches push out the oldest ones.

I don't think that applying circular buffer here is a good solution. That case can cause a situation when a user puts half of the signatures to a transaction and then the transaction will disappear at some point in `GetPendingTransactions` query response. It will be neither passed to stateful validation nor "mst"-expired in case of circular buffer substitution. 
Probably for that case, we will have to introduce some additional status (what will look very weird for me).

irohad/main/iroha_conf_literals.cpp
```diff
@@ -16,6 +16,7 @@ namespace config_members {
   const char *VoteDelay = "vote_delay";
   const char *MstSupport = "mst_enable";
   const char *MstExpirationTime = "mst_expiration_time";
+  const char *MstStateTxLimit = "mst_state_transactions_limit";
```
igor-egorov:
```suggestion
  const char *MstStateTxsLimit = "mst_state_transactions_limit";
```

irohad/main/irohad.cpp
```diff
@@ -23,6 +23,7 @@
 static const std::string kListenIp = "0.0.0.0";
 static const std::string kLogSettingsFromConfigFile = "config_file";
 static const uint32_t kMstExpirationTimeDefault = 1440;
+static const uint32_t kMstStateTxLimitDefault = 1000;
```
igor-egorov:
```suggestion
static const uint32_t kMstStateTxsLimitDefault = 1000;
```

igor-egorov: Should not this setting have a value equal to `max_proposal_size`?

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -14,6 +14,15 @@
 #include "interfaces/transaction.hpp"
 #include "logger/logger.hpp"
 
+template <class T>
+static size_t countTxsInBatches(const T &batches) {
```
igor-egorov, line 18: I did the same for `OnDemandCache`.

Could you please pick the best and move it to some common place?

```
namespace {
  uint64_t countTransactions(const OrderingGateCache::BatchesSetType &batches) {
    return std::accumulate(batches.begin(),
                           batches.end(),
                           0u,
                           [](const uint64_t txs_quantity, const auto &batch) {
                             return txs_quantity
                                 + boost::size(batch->transactions());
                           });
  }
}  // namespace
```

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -46,22 +55,27 @@ namespace iroha {
...
+    return MstState(completer, transaction_limit, std::move(log));
   }
 
   StateUpdateResult MstState::operator+=(const DataType &rhs) {
-    auto state_update = StateUpdateResult{
-        std::make_shared<MstState>(MstState::empty(log_, completer_)),
-        std::make_shared<MstState>(MstState::empty(log_, completer_))};
+    auto state_update =
+        StateUpdateResult{std::make_shared<MstState>(
+                              MstState::empty(log_, completer_, txs_limit_)),
```
igor-egorov: `empty` method takes the same set of arguments as `MstState` constructor.

Could you please refactor the order of its arguments to be the same in constructor and `empty` method in addition to the main changes?

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -93,17 +108,28 @@ namespace iroha {
   }
 
   MstState MstState::eraseByTime(const TimeType &time) {
-    MstState out = MstState::empty(log_, completer_);
+    MstState out = MstState::empty(log_, completer_, txs_limit_);
     while (not index_.empty() and (*completer_)(index_.top(), time)) {
       auto iter = internal_state_.find(index_.top());
 
       out += *iter;
+      assert(txs_amount_ >= boost::size((*iter)->transactions()));
```
igor-egorov: And what should happen when an assertion fails at runtime?

irohad/multi_sig_transactions/state/mst_state.hpp
```diff
@@ -148,6 +150,12 @@ namespace iroha {
      */
     bool contains(const DataType &element) const;
 
+    /// Get the amount of stored transactions.
```
igor-egorov:
```suggestion
    /// Get the number of stored transactions over batches.
```
1. amount -> number or quantity
2. \+ "over batches"
https://www.bkacontent.com/commonly-confused-words-amount-number-and-quantity/

irohad/multi_sig_transactions/state/mst_state.hpp
```diff
@@ -148,6 +150,12 @@ namespace iroha {
...
+    size_t transactionsAmount() const;
+
+    /// Get the amount of stored batches.
```
igor-egorov:
```suggestion
    /// Get the number of stored batches.
```

igor-egorov (commented):
Please review the comments. I cannot approve now.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
MBoldyrev: a crash that forces one to investigate the problem. this chack should never fail - it is guaranteed by the code for now. the assertion is needed for later changes, when somebody possibly breaks this contract.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -154,8 +187,20 @@ namespace iroha {
...
-      rawInsert(rhs_batch);
-      state_update.updated_state_->rawInsert(rhs_batch);
+      if (transactionsQuantity() + boost::size(rhs_batch->transactions())
+          <= txs_limit_) {
+        // there is enough room for the new batch
+        rawInsert(rhs_batch);
+        state_update.updated_state_->rawInsert(rhs_batch);
+      } else {
+        // there is not enough room for the new batch
+        log_->info(
```
luckychess, line 193: I would suggest to use warning level for logger.

test/module/irohad/multi_sig_transactions/state_test.cpp
```diff
@@ -129,17 +125,20 @@ TEST(StateTest, DifferentSignaturesUnionTest) {
...
-  auto state2 = MstState::empty(mst_state_log_, completer_);
+  auto state2 = MstState::empty(completer_, kMstStateTxLimit, mst_state_log_);
   state2 +=
       addSignatures(makeTestBatch(txBuilder(4)), 0, makeSignature("4", "4"));
   state2 +=
       addSignatures(makeTestBatch(txBuilder(5)), 0, makeSignature("5", "5"));
-  ASSERT_EQ(2, state2.getBatches().size());
+  ASSERT_EQ(2, state2.batchesQuantity());
 
+  ASSERT_LE(state1.transactionsQuantity() + state2.transactionsQuantity(),
```
luckychess, line 137: I guess this check is unnecessary. Test will fail at the very last assertion anyway. Also testing tests looks a bit strange :)

luckychess (approved):
I guess you first should merge https://github.com/hyperledger/iroha/pull/2208 and then update this one.

test/module/irohad/multi_sig_transactions/state_test.cpp
MBoldyrev: I cannot agree because here we check against the transaction limit, and in the last line against a number not related to it. Suppose someone changes `kMstStateTxLimit`. Will he be responsible for checking every magic number in all related tests?

lebdron (commented):
Please merge this into trunk, since we cannot have code which silently drops transactions. Handles for insertion feedback will be implemented in another task. The same trunk should be used for https://github.com/hyperledger/iroha/pull/2206

PR #2203, opened 2019-03-29
MstState: iteration methods for batches & txs

MBoldyrev:

### Description of the Change

New interface to access all batches and transactions of MST State.

### Benefits

Copying can be avoided. Code simplified.


irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -104,6 +104,21 @@ namespace iroha {
     return out;
   }
 
+  void MstState::iterateBatches(std::function<void(const DataType &)> visitor) const {
+    std::for_each(internal_state_.cbegin(), internal_state_.cend(), visitor);
+  }
+
+  void MstState::iterateTransactions(
+      std::function<
```
igor-egorov: std::function is not considered as a good solution for non-tests code due to its performance drawbacks.

igor-egorov (commented):
The new interface looks sexy but I cannot approve the code with `std::function` used in non-tests area. 

A possible solution can be `template <typename Lambda>`.
We need more opinions regarding that pull request.

luckychess (approved):

igor-egorov (approved):

PR #2168, opened 2019-03-13
Fix/coredumps add binaries

Stayer:
### Description of the Change

[DOPS-378](https://soramitsu.atlassian.net/browse/DOPS-378)
This PR allows to archive binaries (located in build/bin folder) and upload it to Nexus artifact server with coredumps (see #2024 ). Also, coredumps handling added to new pipeline.

Clean-up script: https://github.com/hlavki/nexus-scripts

### Benefits

Binaries can be useful to work with coredumps

### Possible Drawbacks 

There is no possible drawbacks

### Usage Examples or Tests *[optional]*

Use coredumps flag in Jenkins build to switch on/off this feature

PR #2205, opened 2019-04-01
Update the CI

BulatSaif:
### Description of the Change

1. fix: docker `develop-build` containers is build from iroha dir which is can be up to 25 Gb

2. For mac kill any postures at the build end

3. build and push git tags to docker (not tested need to create tag in )

4. add core dump ( already  done in: https://github.com/hyperledger/iroha/pull/2168 )

5. Fix: **Merge to trunk** always finish green, (now we check the build status)

6. For **on open pr** job  run all tests

### Benefits

1. develop branch will stop fail with `no space left` error

2.  Mac agent will clean itself, if one job fails and do not clean second job will clean agent.

3. Git tags are pushed in docker 

4. 

5. no confusing status 

6. more info in open pr 

### Possible Drawbacks 

6.  **on open pr** job will take ~3 mins longer


bakhtin (approved):

kogeler (approved):
OK

PR #2206, opened 2019-04-01
Add transactions limit to OnDemandCache

igor-egorov:

### Description of the Change

Iroha will not accept new transactions via Torii or ListTorii if its internals is not ready for new incoming transactions.

If Iroha is full then Torii or ListTorii will return `grpc::StatusCode::RESOURCE_EXHAUSTED` with textual description that includes "reduced hashes" of batches that were not processed.

### Benefits

Improved robustness. Still the same API.

### Possible Drawbacks 

Unknown

### Usage Examples or Tests 

All the tests.


irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -150,7 +163,20 @@ namespace iroha {
             });
       }
 
-      return grpc::Status::OK;
+      auto status = grpc::Status::OK;
+      if (not unprocessed_batches.empty()) {
+        std::stringstream message;
+        message << "The batches with the following reduced hashes were not "
+                   "processed, please resend them: [";
```
MBoldyrev: parsing such a string on clients to get the unprocessed batches is not quite a good idea. I suggest extending the API to add a `ListTorii` call with status response.

irohad/ordering/impl/ordering_gate_cache/on_demand_cache.cpp
```diff
@@ -13,24 +15,47 @@ using namespace iroha::ordering::cache;
...
+OnDemandCache::OnDemandCache(uint64_t max_cache_size)
+    : max_cache_size_(max_cache_size) {}
+
+bool OnDemandCache::addToBack(
     const OrderingGateCache::BatchesSetType &batches) {
   std::unique_lock<std::shared_timed_mutex> lock(mutex_);
-  circ_buffer.back().insert(batches.begin(), batches.end());
+  const uint64_t input_size = countTransactions(batches);
+
+  if (circ_buffer.back().first + input_size > max_cache_size_) {
```
MBoldyrev, line 38: as you assume there is only one batch in the collection, maybe we should make the interface to explicitly accept a single batch?

MBoldyrev (approved):

test/module/shared_model/interface_mocks.hpp
```diff
@@ -160,6 +160,28 @@ auto createMockBatchWithTransactions(
   return res;
 }
 
+/**
+ * Creates mock batch with provided transactions
+ * @param txs -- list of transactions in the batch
+ * @param hash -- const ref to hash to be returned by the batch
+ * @return unique_ptr for batch
+ */
+auto createUniqueMockBatchWithTransactions(
```
lebdron: `unique_ptr` can be converted to `shared_ptr`. The method above should be removed, and this method can be renamed to `createMockBatchWithTransactions`.

test/module/irohad/torii/torii_transport_command_test.cpp
```diff
@@ -160,7 +161,18 @@ TEST_F(CommandServiceTransportGrpcTest, ListTorii) {
       *batch_factory,
       createTransactionBatch(
           A<const shared_model::interface::types::SharedTxsCollectionType &>()))
-      .Times(kTimes);
+      .Times(kTimes)
+      .WillRepeatedly(Invoke(
+          [](const shared_model::interface::types::SharedTxsCollectionType
```
lebdron: Please make a separate function, so that the code is not repeated below.

test/module/irohad/ordering/on_demand_cache_test.cpp
```diff
@@ -18,19 +18,27 @@ using ::testing::Return;
 using ::testing::ReturnRef;
 using ::testing::UnorderedElementsAre;
 
+const uint64_t kMaxCacheSize = 10000;
```
lebdron, line 21: False case for `addToBack` should be covered.

irohad/torii/processor/impl/transaction_processor_impl.cpp
```diff
@@ -121,19 +128,26 @@ namespace iroha {
...
         std::shared_ptr<shared_model::interface::TransactionBatch>
             transaction_batch) const {
       log_->info("handle batch");
+      bool batch_passed_for_further_processing(true);
       if (transaction_batch->hasAllSignatures()
           and not mst_processor_->batchInStorage(transaction_batch)) {
         log_->info("propagating batch to PCS");
         this->publishEnoughSignaturesStatus(transaction_batch->transactions());
-        pcs_->propagate_batch(transaction_batch);
+        log_->critical("BATCH {}", transaction_batch);
```
lebdron: Looks like this log call was used for debug and can be removed.

irohad/torii/processor/impl/transaction_processor_impl.cpp
```diff
@@ -111,7 +111,14 @@ namespace iroha {
       mst_processor_->onPreparedBatches().subscribe([this](auto &&batch) {
         log_->info("MST batch prepared");
         this->publishEnoughSignaturesStatus(batch->transactions());
-        this->pcs_->propagate_batch(batch);
+        log_->critical("BATCH {}", batch);
```
lebdron: Same for removal here.

irohad/torii/processor/impl/transaction_processor_impl.cpp
```diff
@@ -121,19 +128,26 @@ namespace iroha {
...
+      bool batch_passed_for_further_processing(true);
       if (transaction_batch->hasAllSignatures()
           and not mst_processor_->batchInStorage(transaction_batch)) {
         log_->info("propagating batch to PCS");
         this->publishEnoughSignaturesStatus(transaction_batch->transactions());
-        pcs_->propagate_batch(transaction_batch);
+        log_->critical("BATCH {}", transaction_batch);
+        if (not(batch_passed_for_further_processing =
+                    pcs_->propagate_batch(transaction_batch))) {
+          log_->info("PCS was unable to serve the batch {}",
```
lebdron: Is it required to log here? Maybe the log in CommandServiceTransport is enough. In this case the code here can be simplified to the following, without using the addidional `batch_passed_for_further_processing` variable.
```C++
return pcs_->propagate_batch(transaction_batch);
```

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -150,7 +162,20 @@ namespace iroha {
             });
       }
 
-      return grpc::Status::OK;
+      auto status = grpc::Status::OK;
+      if (not unprocessed_batches.empty()) {
+        std::stringstream message;
+        message << "The batches with the following reduced hashes were not "
```
lebdron: In an offline discussion we decided to make a generic error message without hashes.

irohad/torii/impl/command_service_transport_grpc.cpp
```diff
@@ -122,13 +122,25 @@ namespace iroha {
...
+              const shared_model::interface::types::HashType hash =
+                  value.value->reducedHash();
+              bool batch_is_processed =
+                  this->command_service_->handleTransactionBatch(
+                      std::move(value).value);
+              if (not batch_is_processed) {
+                log_->info(
+                    "Transaction batch with reduced hash {} is not processed "
+                    "and has "
+                    "to be resent",
```
lebdron: "has to be resent" is misleading, since `false` is also returned when it is a replay. Please rephrase.

irohad/ordering/impl/ordering_gate_cache/on_demand_cache.hpp
```diff
@@ -18,7 +18,14 @@ namespace iroha {
 
       class OnDemandCache : public OrderingGateCache {
        public:
-        void addToBack(const BatchesSetType &batches) override;
+        /**
+         * Create OnDemandCache instance
+         * @param max_cache_size - maximum amount of transactions contained in
+         * cache (calculated as a sum over all stored batches)
```
lebdron, line 24: in cache -> in a single cache entry, since there are 3 entries, and the restriction is for a single entry.

irohad/ordering/impl/ordering_gate_cache/on_demand_cache.hpp
```diff
@@ -29,9 +36,11 @@ namespace iroha {
         virtual const BatchesSetType &tail() const override;
 
        private:
+        const uint64_t max_cache_size_;
         mutable std::shared_timed_mutex mutex_;
-        using BatchesQueueType = boost::circular_buffer<BatchesSetType>;
-        BatchesQueueType circ_buffer{3, BatchesSetType{}};
+        using CacheElementType = std::pair<uint64_t, BatchesSetType>;
```
lebdron: Please add a comment that pair.first is the cumulative number of transactions in stored batches.

irohad/ordering/impl/ordering_gate_cache/on_demand_cache.cpp
```diff
@@ -13,24 +15,47 @@ using namespace iroha::ordering::cache;
...
+    for (auto it = batches.second.begin(); it != batches.second.end();) {
       if (std::any_of(it->get()->transactions().begin(),
                       it->get()->transactions().end(),
                       [&hashes](const auto &tx) {
                         return hashes.find(tx->hash()) != hashes.end();
                       })) {
         // returns iterator following the last removed element
         // hence there is no increment in loop iteration_expression
-        it = batches.erase(it);
+        batches.first -= countTransactions(batches.second);
```
lebdron: It subtracts the count of transaction in all the batches, but only a single batch `it` is removed below. Should it be the following?
```suggestion
        batches.first -= boost::size((*it)->transactions());
```

lebdron (changes requested):
1. Pull request should be made into trunk, since it introduces an issue with lost mst batches
2. The pop-addToBack combination in OnDemandOrderingGate::sendCachedTransactions should be carefully handled, since a following race can happen at the moment:
 - front cache entry contains max_number_of_txs
 - pop is called, moving the batches from back cache entry
 - OnDemandOrderingGate::propagateBatch is called, adding a batch to back entry
 - addToBack in sendCachedTransactions fails, since there is no place for all the batches in popped collection

Consider adding `rotate` method instead, which will move the elements from front to back

test/module/irohad/ordering/on_demand_cache_test.cpp
igor-egorov: need to write a test

irohad/ordering/impl/ordering_gate_cache/on_demand_cache.hpp
igor-egorov: discussed offline: not a subject of change

irohad/ordering/impl/ordering_gate_cache/on_demand_cache.cpp
igor-egorov: need to write a test

irohad/ordering/impl/on_demand_ordering_gate.cpp
```diff
@@ -100,8 +104,9 @@ OnDemandOrderingGate::processProposalRequest(
 void OnDemandOrderingGate::sendCachedTransactions() {
   // TODO mboldyrev 22.03.2019 IR-425
   // make cache_->getBatchesForRound(current_round) that respects sync
-  auto batches = cache_->pop();
-  cache_->addToBack(batches);
+
+  cache_->rotate();
+  auto batches = cache_->back();
```
lebdron: It will be better to call `auto batches = front(); cache_->rotate();` to prevent new batches being accidentally resent due to concurrency.

lebdron (approved):

PR #2207, opened 2019-04-01
OG batch resend strategy

luckychess:
### Description of the Change
Pull request implements strategy for resending batches from cache.

### Benefits
We don't resend batches to the rounds we already sent before.

### Possible Drawbacks 
- Code complexity
- Strategy partially intersects with OG cache logic
- remove function complexity :(

### Usage Examples or Tests
Check `on_demand_resend_strategy_test`


irohad/ordering/impl/on_demand_connection_manager.cpp
```diff
@@ -55,10 +62,41 @@ void OnDemandConnectionManager::onBatches(CollectionType batches) {
...
+    }
+    if (rounds.find(nextRejectRound(nextCommitRound(current_round)))
+        != rounds.end()) {
+      commit_reject_batches.push_back(batch);
+    }
+    if (rounds.find(nextCommitRound(nextCommitRound(current_round)))
+        != rounds.end()) {
+      commit_commit_batches.push_back(batch);
+    }
+  }
```
MBoldyrev: I think the strategy interface could be made easier to use if it provided
```
bool shouldSend(const BatchPtr &, const Round &) const
```
instead of these two:
```
RoundSetType extract(BatchPtr)
Round getCurrentRound() const
```

irohad/ordering/impl/ordering_gate_cache/on_demand_resend_strategy.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+      void setCurrentRound(const consensus::Round &current_round) override;
+
+      consensus::Round getCurrentRound() const override;
+
+     private:
+      std::unordered_map<
+          std::shared_ptr<shared_model::interface::TransactionBatch>,
+          std::pair<consensus::Round, bool>>
+          sent_batches_;
```
MBoldyrev, line 51: I would suggest specifying the batch pointer hasher and equality checker classes for this map. This way we can correctly handle same batch after copying (pointer changes, but reduced hash does not).

irohad/ordering/impl/ordering_gate_cache/on_demand_resend_strategy.cpp
```diff
@@ -0,0 +1,106 @@
...
+      sent_batches_.emplace(batch, std::make_pair(current_round_, false));
+  return res.second;
+}
+
+bool OnDemandResendStrategy::readyToUse(
+    std::shared_ptr<shared_model::interface::TransactionBatch> batch) {
+  std::shared_lock<std::shared_timed_mutex> lock(access_mutex_);
+  auto batch_found = std::find_if(
+      sent_batches_.begin(), sent_batches_.end(), [&batch](auto saved_batch) {
+        return *(saved_batch.first) == *batch;
```
MBoldyrev: nooo! please use `sent_batches_.find(batch)` to benefit from the map structure, otherwise you get a worse performance than on a vector.

irohad/ordering/impl/ordering_gate_cache/on_demand_resend_strategy.cpp
```diff
@@ -0,0 +1,106 @@
...
+OnDemandResendStrategy::RoundSetType OnDemandResendStrategy::extract(
+    std::shared_ptr<shared_model::interface::TransactionBatch> batch) {
+  std::shared_lock<std::shared_timed_mutex> lock(access_mutex_);
+  std::set<consensus::Round> valid_rounds{
+      nextCommitRound(nextCommitRound(current_round_)),
+      nextCommitRound(nextRejectRound(current_round_)),
+      nextRejectRound(nextCommitRound(current_round_)),
+      nextRejectRound(nextRejectRound(current_round_))};
+
+  auto saved_round_iterator = std::find_if(
```
MBoldyrev: could you use `sent_batches_.find` here?

irohad/ordering/impl/ordering_gate_cache/on_demand_resend_strategy.cpp
```diff
@@ -0,0 +1,106 @@
...
+      });
+  if (saved_round_iterator == sent_batches_.end()) {
+    return valid_rounds;
+  }
+  if (not saved_round_iterator->second.second) {
+    return valid_rounds;
+  }
+  auto saved_round = saved_round_iterator->second.first;
+
+  if (nextCommitRound(nextCommitRound(saved_round)) == current_round_) {
```
MBoldyrev: Could this whole block
https://github.com/hyperledger/iroha/blob/e0dcced4843337d48dac7f269aaac21830ca9515/irohad/ordering/impl/ordering_gate_cache/on_demand_resend_strategy.cpp#L59-L73
be substituted with something like
```c++
(pseudo-code)
std::set<consensus::Round> target_rounds;
copy_if(valid_rounds,
    inserter(target_rounds),
    [current_round_](target_round){ return target_round.block_round <= current_round_.block_round; });
```

irohad/ordering/impl/ordering_gate_cache/on_demand_resend_strategy.cpp
```diff
@@ -0,0 +1,106 @@
...
+  } else if (nextCommitRound(nextRejectRound(saved_round)) == current_round_) {
+    valid_rounds.erase(nextCommitRound(nextRejectRound(current_round_)));
+  } else if (nextRejectRound(nextRejectRound(saved_round)) == current_round_) {
+    valid_rounds.erase(nextCommitRound(nextRejectRound(current_round_)));
+    valid_rounds.erase(nextRejectRound(nextCommitRound(current_round_)));
+    valid_rounds.erase(nextCommitRound(nextCommitRound(current_round_)));
+  } else {
+    BOOST_ASSERT_MSG(
+        false,
+        "Current round can't be reached after 2 rounds from the saved one");
```
MBoldyrev: As we discussed, this assertion should be removed, because such condition could happen after a synchronization on a delayed node.

irohad/ordering/impl/ordering_gate_cache/on_demand_resend_strategy.cpp
luckychess: I afraid it will work only for (reject, reject) case.

irohad/ordering/impl/on_demand_connection_manager.cpp
MBoldyrev: sorry, it was meant to be
```
void sendBatch(const BatchPtr &, const CurrentConnections &)
```
instead of the above comment

irohad/ordering/impl/ordering_gate_cache/on_demand_resend_strategy.cpp
MBoldyrev: yes, I understood your code wrong way. as we discussed, could you apply something like
```
target_rounds = setDifference(
    reachableIn2Steps(current_round_),
    reachableIn2Steps(saved_round));
```

PR #2208, opened 2019-04-02
MstState: fixed issues.

MBoldyrev:
### Description of the Change

The priority queue `index_` in `MstState` is expected to be sorted in the order of batches expiration, but that does not hold in general case when the queue gets sorted by the first transaction timestamp. This change sorts the queue by the oldest transaction timestamp to fix the issue.

#### UPDATE ####

- `internal_state_` and `index_` were both used to store the batches, which complicated the logic and lead to a missing cleanup operation. Now there is only one storage - `batches_`.
- `MstState::operator==(const MstState &rhs)` previously checked whether `this->internal_state_` is a subset of `rhs.internal_state_`. Implementing it like a honest equality would be ineffective with batches stored in an unordered set. As this operator is used only once in a single test, I decided better to remove it than to keep an ineffective implementation.

The index type was changed to 

### Benefits

Fixed `MstState::eraseByTime` and storage cleanup.

### Possible Drawbacks 

Additional boost includes present in the `mst_state.hpp` header.
Removed `MstState::operator==(const MstState &rhs)`.

Rework `MstState` with pimpl to hide the includes.


irohad/multi_sig_transactions/state/mst_state.hpp
```diff
@@ -164,8 +167,12 @@ namespace iroha {
                            iroha::model::PointerBatchHasher,
                            BatchHashEquality>;
 
-    using IndexType =
-        std::priority_queue<DataType, std::vector<DataType>, Less>;
```
luckychess, line 176: Looks like we don't need class `Less` anymore.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -5,15 +5,34 @@
...
+  shared_model::interface::types::TimestampType oldestTimestamp(
+      const iroha::BatchPtr &batch) {
+    auto timestamps =
+        batch->transactions()
+        | boost::adaptors::transformed(
+              +[](const std::shared_ptr<shared_model::interface::Transaction>
+                      &tx) { return tx->createdTime(); });
+    const auto min_it = std::min_element(timestamps.begin(), timestamps.end());
+    assert(min_it != timestamps.end());
+    return min_it == timestamps.end() ? 0 : *min_it;
```
luckychess, line 37: It looks like assertion will check only case when batch is empty so maybe better check this explicitly?
Also if empty batches are invalid here may be `optional<TimestampType>` will be a better choice? Technically 0 is a valid value (even though pretty unlikely).

test/module/irohad/multi_sig_transactions/completer_test.cpp
```diff
@@ -99,11 +99,11 @@ TEST(CompleterTest, BatchExpirationTestExpired) {
   auto completer = std::make_shared<DefaultCompleter>(std::chrono::minutes(1));
   auto time = iroha::time::now();
   auto tx1 = std::make_shared<MockTransaction>();
-  EXPECT_CALL(*tx1, createdTime()).WillOnce(Return(time));
+  EXPECT_CALL(*tx1, createdTime()).WillRepeatedly(Return(time));
```
luckychess, line 102: This and below changes look unclear, could you explain them please?

luckychess (approved):

test/module/irohad/multi_sig_transactions/completer_test.cpp
MBoldyrev: Here we impose no assertions on the number of times the function must be called - it does not matter.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
MBoldyrev: A batch cannot have empty transactions by external contract. I added a more explicit assertion to ensure that at the beginning of the function, as you suggested. But I want to keep this one too because I am dereferencing an iterator, which might cause a crash.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -5,15 +5,39 @@
...
+      const iroha::BatchPtr &batch) {
+    const bool batch_is_empty = boost::empty(batch->transactions());
+    assert(not batch_is_empty);
+    if (batch_is_empty) {
+      return 0;
+    }
+    auto timestamps =
+        batch->transactions()
+        | boost::adaptors::transformed(
+              +[](const std::shared_ptr<shared_model::interface::Transaction>
```
lebdron, line 33: It should be possible to pass lambda to `transformed` without casting it to function pointer.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -177,26 +192,23 @@ namespace iroha {
   }
 
   void MstState::rawInsert(const DataType &rhs_batch) {
-    internal_state_.insert(rhs_batch);
-    index_.push(rhs_batch);
+    batches_.insert(
+        BatchesBimap::value_type(oldestTimestamp(rhs_batch), rhs_batch));
```
lebdron:
```suggestion
        {oldestTimestamp(rhs_batch), rhs_batch});
```

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -136,35 +150,36 @@ namespace iroha {
...
-                     const InternalStateType &transactions,
+                     const BatchesForwardCollectionType &batches,
                      logger::LoggerPtr log)
-      : completer_(completer),
-        internal_state_(transactions.begin(), transactions.end()),
-        index_(transactions.begin(), transactions.end()),
-        log_(std::move(log)) {}
+      : completer_(completer), log_(std::move(log)) {
+    for (const auto &batch : batches) {
+      batches_.insert(BatchesBimap::value_type(oldestTimestamp(batch), batch));
```
lebdron:
```suggestion
      batches_.insert({oldestTimestamp(batch), batch});
```

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -62,34 +82,34 @@ namespace iroha {
...
-  }
-
-  bool MstState::operator==(const MstState &rhs) const {
-    return std::all_of(
-        internal_state_.begin(), internal_state_.end(), [&rhs](auto &i) {
-          return rhs.internal_state_.find(i) != rhs.internal_state_.end();
-        });
+    const auto &my_batches = batches_.right | boost::adaptors::map_keys;
+    std::vector<DataType> difference;
+    difference.reserve(boost::size(batches_));
```
lebdron, line 94: Maybe call `shrink_to_fit` afterwards then?

lebdron (approved):

irohad/multi_sig_transactions/state/impl/mst_state.cpp
MBoldyrev: It would be worth shrinking if we were to keep the vector. But it is discarded with the function scope. Please see the return statement and the new constructor of `MstState`:
https://github.com/hyperledger/iroha/blob/19af4575d9047925e34a29e5effa9afcb4ab377a/irohad/multi_sig_transactions/state/mst_state.hpp#L171-L173
https://github.com/hyperledger/iroha/blob/19af4575d9047925e34a29e5effa9afcb4ab377a/irohad/multi_sig_transactions/state/mst_state.hpp#L90-L91

PR #2209, opened 2019-04-02
Event loop for pipeline processing

lebdron:
### Description of the Change

Create a `synchronize` subject in consensus to queue the events from the network without locking the operation of consensus
- Patch rxcpp to remove data race and allow using `current_thread` scheduler with `synchronize`
- Add subscriptions for subjects in irohad to manage their lifetime
- Add round parameter to Yac to close rounds that are less than or equal to current round
- Add `publish().ref_count()` to YacGateImpl observable to eliminate redundant calls to `flat_map` if multiple subscribers are present
- Fix data race in `getSpdlogLogLevel`
- Fix nondeterminism in `consensus_sunny_day` test 🎉 
- Allow usage of `blocking_observable` in TestSubscriber 

### Benefits
Consensus can process messages from the network while the pipeline is running

### Possible Drawbacks 
A separate thread is allocated to the pipeline


cmake/Modules/Findrxcpp.cmake
```diff
@@ -1,6 +1,7 @@
 add_library(rxcpp INTERFACE IMPORTED)
 
-find_path(rxcpp_INCLUDE_DIR rxcpp/rx.hpp)
+# TODO 2019-04-02 lebdron: IR-346 Update Rx commit
+#find_path(rxcpp_INCLUDE_DIR rxcpp/rx.hpp)
```
luckychess: I guess we don't need this line anymore.

cmake/Modules/Findrxcpp.cmake
```diff
@@ -9,27 +10,26 @@ find_package_handle_standard_args(rxcpp DEFAULT_MSG
...
 
 set(URL https://github.com/Reactive-Extensions/rxcpp.git)
-# this version is chosen, because it fixes 100% node overload [IR-1736] bug
-set(VERSION a7d5856385f126e874db6010d9dbfd37290c61de)
+set(VERSION 7c79c18ab252eb37bcf122c75070f8d557868c08)
 set_target_description(rxcpp "Library for reactive programming" ${URL} ${VERSION})
 
 
-if (NOT rxcpp_FOUND)
+#if (NOT rxcpp_FOUND)
```
luckychess: Maybe just remove?

cmake/Modules/Findrxcpp.cmake
```diff
@@ -9,27 +10,26 @@ find_package_handle_standard_args(rxcpp DEFAULT_MSG
...
-      UPDATE_COMMAND "" # remove update step
       TEST_COMMAND "" # remove test step
       )
   externalproject_get_property(reactive_extensions_rxcpp source_dir)
   set(rxcpp_INCLUDE_DIR ${source_dir}/Rx/v2/src)
   file(MAKE_DIRECTORY ${rxcpp_INCLUDE_DIR})
 
   add_dependencies(rxcpp reactive_extensions_rxcpp)
-endif ()
+#endif ()
```
luckychess: And here as well.

luckychess (approved):

luckychess:
CI confuses me a bit - looks like all tests are passed but it shows failure status.

irohad/consensus/yac/impl/yac.cpp
```diff
@@ -150,13 +172,17 @@ namespace iroha {
 
         network_->sendState(current_leader, {vote});
         cluster_order_.switchToNext();
-        if (cluster_order_.hasNext()) {
+        auto has_next = cluster_order_.hasNext();
+        lock.unlock();
```
MBoldyrev, line 176: should not we synchronize the other accesses to `cluster_order_` for example with shared locks?
https://github.com/hyperledger/iroha/blob/6b97c3e467ca628d3a1e7f44a22b7d3420498813/irohad/consensus/yac/impl/yac.cpp#L190
https://github.com/hyperledger/iroha/blob/6b97c3e467ca628d3a1e7f44a22b7d3420498813/irohad/consensus/yac/impl/yac.cpp#L203
https://github.com/hyperledger/iroha/blob/6b97c3e467ca628d3a1e7f44a22b7d3420498813/irohad/consensus/yac/impl/yac.cpp#L286

MBoldyrev: and `round_` btw

irohad/consensus/yac/impl/yac.cpp
```diff
@@ -150,13 +172,17 @@ namespace iroha {
...
+        if (has_next) {
           timer_->invokeAfterDelay([this, vote] { this->votingStep(vote); });
         }
       }
 
-      void Yac::closeRound() {
-        timer_->deny();
+      void Yac::closeRound(Round round) {
```
MBoldyrev: as for me, the `closeRound` function should better take no `round` argument, because we do not maintain more than one round at a time. Better would be to call it only if the `proposal_round` matches the condition here:
https://github.com/hyperledger/iroha/blob/6b97c3e467ca628d3a1e7f44a22b7d3420498813/irohad/consensus/yac/impl/yac.cpp#L247
```c++
                  if (proposal_round >= round_) {
                    this->closeRound();
                  }
```

irohad/main/impl/on_demand_ordering_init.hpp
```diff
@@ -98,7 +100,8 @@ namespace iroha {
       /**
        * Initializes on-demand ordering gate and ordering sevice components
        *
-       * @param max_number_of_transactions maximum number of transaction in a proposal
+       * @param max_number_of_transactions maximum number of transaction in a
```
MBoldyrev:
```suggestion
       * @param max_number_of_transactions maximum number of transactions in a
```

libs/logger/logger_spdlog.cpp
```diff
@@ -17,19 +17,23 @@
...
+        return spdlog::level::info;
+      case logger::LogLevel::kWarn:
+        return spdlog::level::warn;
+      case logger::LogLevel::kError:
+        return spdlog::level::err;
+      case logger::LogLevel::kCritical:
+        return spdlog::level::critical;
+      default:
+        BOOST_ASSERT_MSG(false, "Unknown log level!");
+        return getSpdlogLogLevel(logger::kDefaultLogLevel);
```
MBoldyrev: looks like a place for stack overflow! if something changes and `logger::kDefaultLogLevel` will hit `default` too.. better something like
```suggestion
        return getSpdlogLogLevel(spdlog::level::info);
```

MBoldyrev (commented):
I will finish the review later.

MBoldyrev (approved):

MBoldyrev:
also it would be nice to give the patch `patch/synchronize-lock-recursion-thread-safety.patch` a more meaningful and descriptive name. I would suggest either renaming it with some `rxcpp` prefix or moving to a directory like `patches/external/rxcpp` (it other patches to rxcpp are probable).

irohad/consensus/yac/impl/yac.cpp
lebdron: They can be synchronized with shared lock as optimization, but it will introduce more complexity to code. I am not sure that it is required at the moment.
The accesses you mentioned are covered by lock in `onState` now.

MBoldyrev (approved):

PR #2210, opened 2019-04-03
Add Synchronization on Rejects

igor-egorov:

### Description of the Change

Synchronizer now is able to download missing blocks even when reject is happened in consensus.

### Benefits

Performance of the network can be increased in some cases.

**Explanation**
Some peers at the network can be out of sync (the peer is in round 3, when the rest in 5).
Such a peer previously could sync even when a commit of 6th block happened.
Now the peer can sync early and be in a synchronized state when a commit of the 6th block is happening. 
Thus there is a chance to do consensus step faster.

### Possible Drawbacks 

Still some changes has to be introduced by #2066 to the structure of `SynchronizationEvent`.

### Usage Examples or Tests 

synchronizer_test

### DIFF
--- a/irohad/consensus/gate_object.hpp
+++ b/irohad/consensus/gate_object.hpp
@@ -27,26 +27,39 @@ namespace iroha {
       consensus::Round round;
     };
 
-    /// Network votes for another pair and round
-    struct VoteOther {
+    struct Synchronizable {
       shared_model::interface::types::PublicKeyCollectionType public_keys;
-      shared_model::interface::types::HashType hash;
       consensus::Round round;
+      Synchronizable(
+          shared_model::interface::types::PublicKeyCollectionType _public_keys,
+          consensus::Round _round)
+          : public_keys(std::move(_public_keys)), round(std::move(_round)) {}
+    };
+
+    /// Network votes for another pair and round
+    struct VoteOther : public Synchronizable {
+      shared_model::interface::types::HashType hash;
+      VoteOther(
+          shared_model::interface::types::PublicKeyCollectionType _public_keys,
+          consensus::Round _round,
+          shared_model::interface::types::HashType _hash)
+          : Synchronizable(std::move(_public_keys), std::move(_round)),
+            hash(std::move(_hash)) {}
     };
 
     /// Reject on proposal
-    struct ProposalReject {
-      consensus::Round round;
+    struct ProposalReject : public Synchronizable {
+      using Synchronizable::Synchronizable;
     };
 
     /// Reject on block
-    struct BlockReject {
-      consensus::Round round;
+    struct BlockReject : public Synchronizable {
+      using Synchronizable::Synchronizable;
     };
 
     /// Agreement on <None, None>
-    struct AgreementOnNone {
-      consensus::Round round;
+    struct AgreementOnNone : public Synchronizable {
+      using Synchronizable::Synchronizable;
     };
 
     using GateObject = boost::variant<PairValid,
--- a/irohad/synchronizer/impl/synchronizer_impl.cpp
+++ b/irohad/synchronizer/impl/synchronizer_impl.cpp
@@ -41,28 +41,16 @@ namespace iroha {
           object,
           [this](const consensus::PairValid &msg) { this->processNext(msg); },
           [this](const consensus::VoteOther &msg) {
-            this->processDifferent(msg,
-                                   false,
-                                   msg.round.block_round,
-                                   SynchronizationOutcomeType::kCommit);
+            this->processDifferent(msg, SynchronizationOutcomeType::kCommit);
           },
           [this](const consensus::ProposalReject &msg) {
-            this->processDifferent(msg,
-                                   true,
-                                   msg.round.block_round - 1,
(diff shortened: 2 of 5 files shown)


irohad/consensus/gate_object.hpp
```diff
@@ -27,26 +27,39 @@ namespace iroha {
...
     };
 
-    /// Network votes for another pair and round
-    struct VoteOther {
+    struct Synchronizable {
       shared_model::interface::types::PublicKeyCollectionType public_keys;
-      shared_model::interface::types::HashType hash;
       consensus::Round round;
+      Synchronizable(
+          shared_model::interface::types::PublicKeyCollectionType _public_keys,
```
luckychess, line 34: Is this such a codestyle for struct constructor params? (names starting from  underscore)

irohad/synchronizer/impl/synchronizer_impl.hpp
```diff
@@ -43,15 +43,40 @@ namespace iroha {
        * Iterate through the peers which signed the commit_message, load and
        * apply the missing blocks
        * @param commit_message - the commit that triggered synchronization
-       * @param height - the top block height of a peer that needs to be
-       * synchronized
+       * @param top_block_height - the top block height of a peer that needs to
```
luckychess, line 46: Please fix also first param name according to name in docs (or vice versa).

luckychess (approved):

irohad/consensus/gate_object.hpp
MBoldyrev: you can safely use the same names in constructor arguments and class members, here and below

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -167,26 +164,55 @@ namespace iroha {
...
+      if (height_diff < 0) {
         log_->info(
             "Storage is already in synchronized state. Top block height is {}",
-            top_block_height);
+            *top_block_height);
+        return;
+      }
+
+      if (process_small_height_difference
+          and (0 == height_diff or 1 == height_diff)) {
```
MBoldyrev, line 190:
```suggestion
          and (height_diff < 2)) {
```

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -167,26 +164,55 @@ namespace iroha {
...
+      int64_t height_diff = msg.round.block_round - *top_block_height;
+      if (height_diff < 0) {
         log_->info(
             "Storage is already in synchronized state. Top block height is {}",
-            top_block_height);
+            *top_block_height);
+        return;
+      }
+
+      if (process_small_height_difference
```
MBoldyrev: maybe
```suggestion
      if (alternative_outcome != SynchronizationOutcomeType::kCommit
```

irohad/synchronizer/impl/synchronizer_impl.cpp
```diff
@@ -167,26 +164,55 @@ namespace iroha {
...
+                    std::shared_ptr<shared_model::interface::Block>>(),
+                alternative_outcome,
+                msg.round,
+                nullptr /* this is going to be fixed in PR-2066 (IR-147) */});
         return;
       }
 
-      auto result = downloadMissingBlocks(msg, top_block_height);
+      auto result = downloadMissingBlocks(
+          msg, *top_block_height, expected_height, alternative_outcome);
```
MBoldyrev: if we apply the suggestion above and this one, we can remove the `process_small_height_difference` and `expected_height` arguments of `processDifferent` and deduce them from `alternative_outcome`
```suggestion
          msg,
          *top_block_height,
          (alternative_outcome == SynchronizationOutcomeType::kCommit
               ? msg.round.block_round
               : msg.round.block_round - 1),
          alternative_outcome);
```

irohad/synchronizer/impl/synchronizer_impl.hpp
```diff
@@ -43,15 +43,40 @@ namespace iroha {
...
+          const SynchronizationOutcomeType alternative_outcome);
 
       void processNext(const consensus::PairValid &msg);
-      void processDifferent(const consensus::VoteOther &msg);
+
+      /**
+       * Performs synchronization on rejects
+       * @param msg - consensus gate message with a list of peers and a round
+       * @param process_small_height_difference - turn on or off small height
+       * difference handling (should be off for VoteOther type of messages)
```
MBoldyrev: I suggest removing this parameter as non-obvious an deduced from `alternative_outcome` anyway

MBoldyrev (approved):

irohad/consensus/gate_object.hpp
igor-egorov: 1. There is no any special style guideline for that case.
2. According to fast googling results, that is true - we can but that may cause false positive gcc warnings about unused variables.

irohad/synchronizer/impl/synchronizer_impl.cpp
igor-egorov: The suggested transformation is valid, but for clarity purposes, it was written in a bit excessive way.

PR #2211, opened 2019-04-04
MST: fixed expired incoming state logic

MBoldyrev:
### Description of the Change

This change removes a very strange line the rationale behind which was not found:
https://github.com/hyperledger/iroha/blob/df5272c243db9eadaf737fa510af1e5566c823d0/irohad/multi_sig_transactions/impl/mst_processor_impl.cpp#L110

...and adds early removal of expired batches from incoming `MstState`. Also move semantics get used.

### Benefits

see above

### Possible Drawbacks 

There remains a possibility that the removed line of code was helpful...

Regarding the removed line, it broke no tests when it was commented out.
Regarding the outdated batches, please see mst_processor_test `MstProcessorTest.receivedOutdatedState` test.


igor-egorov (approved):

irohad/multi_sig_transactions/impl/mst_processor_impl.cpp
```diff
@@ -92,10 +92,11 @@ namespace iroha {
   // -------------------| MstTransportNotification override |-------------------
 
   void FairMstProcessor::onNewState(const shared_model::crypto::PublicKey &from,
-                                    ConstRefState new_state) {
+                                    MstState new_state) {
     log_->info("Applying new state");
     auto current_time = time_provider_->getCurrentTime();
 
+    new_state.eraseExpired(current_time);
```
muratovv, line 94: There is could be an explanation of why we throw batches from new_state

irohad/multi_sig_transactions/impl/mst_processor_impl.cpp
```diff
@@ -92,10 +92,11 @@ namespace iroha {
...
     auto state_update = storage_->apply(from, new_state);
```
muratovv, line 96: Seems, there is missing `expiredBatchesNotify` invocation.

test/module/irohad/multi_sig_transactions/mst_processor_test.cpp
```diff
@@ -319,3 +319,30 @@ TEST_F(MstProcessorTest, emptyStatePropagation) {
...
+  auto transported_state = MstState::empty(getTestLogger("MstState"),
+                                           std::make_shared<TestCompleter>());
+  const auto expired_batch = makeTestBatch(txBuilder(1, time_before, 3));
+  transported_state += addSignaturesFromKeyPairs(expired_batch, 0, makeKey());
+  mst_processor->onNewState(another_peer_key, transported_state);
+
+  // ---------------------------------| then |----------------------------------
+  EXPECT_FALSE(storage->batchInStorage(expired_batch));
+  check(observers);
+}
```
muratovv, line 348: Also, there could be another test which checks that expired transactions from new_state will be not propagated by the expired notifier.

muratovv (changes requested):

PR #2218, opened 2019-04-09
Fix server dependency in FakePeer; fix timer thread in YAC

lebdron:
### Description of the Change
- Fix cyclic dependency in `Behavior` and `FakePeer` when it was possible for gRPC server and services to deadlock on destruction
- Fix thread management in Yac `TimerImpl`. Now a single thread is used for all invocations, and thread is destroyed together with `TimerImpl` class.

### Benefits
No random segfaults and locks in `fake_peer_example_test`

### Possible Drawbacks 
None

### Usage Examples or Tests
`fake_peer_example_test`, `yac_timer_test`

### DIFF
--- a/irohad/main/impl/consensus_init.cpp
+++ b/irohad/main/impl/consensus_init.cpp
@@ -74,8 +74,10 @@ namespace iroha {
       }
 
       auto YacInit::createTimer(std::chrono::milliseconds delay_milliseconds) {
-        return std::make_shared<TimerImpl>(delay_milliseconds,
-                                           rxcpp::observe_on_new_thread());
+        return std::make_shared<TimerImpl>(
+            delay_milliseconds,
+            // TODO 2019-04-10 andrei: IR-441 Share a thread between MST and YAC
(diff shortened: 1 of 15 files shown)


muratovv (approved):

irohad/main/impl/consensus_init.cpp
```diff
@@ -74,20 +74,8 @@ namespace iroha {
...
-          //
-          // scheduler owns a timeline that is exposed by the now() method.
-          // scheduler is also a factory for workers in that timeline.
-          //
-          // coordination is a factory for coordinators and has a scheduler.
-          return rxcpp::observable<>::timer(
-              std::chrono::milliseconds(delay_milliseconds), coordination_);
-        });
+        return std::make_shared<TimerImpl>(delay_milliseconds,
+                                           rxcpp::observe_on_new_thread());
```
muratovv: I think here we can consider an opportunity to reuse thread from MST-propagator. I am okay with both fix or todo.

igor-egorov (approved):

PR #2222, opened 2019-04-11
shared storage limit

MBoldyrev:
### Description of the Change

A set of classes providing a shared limit between several storages.

please refer to `storage_shared_limit_test`: https://github.com/MBoldyrev/iroha/blob/wip/mst-shared-limit/test/module/libs/storage_shared_limit/storage_shared_limit_test.cpp


irohad/multi_sig_transactions/impl/mst_processor_impl.cpp
```diff
@@ -57,11 +57,13 @@ namespace iroha {
...
 
   // TODO [IR-1687] Akvinikym 10.09.18: three methods below should be one
-  void FairMstProcessor::completedBatchesNotify(ConstRefState state) const {
-    if (not state.isEmpty()) {
-      state.iterateBatches([this](const auto &batch) {
-        batches_subject_.get_subscriber().on_next(batch);
-      });
+  void FairMstProcessor::completedBatchesNotify(
+      std::vector<std::shared_ptr<MovedBatchPtr>> completed) const {
+    if (not completed.empty()) {
```
muratovv: We can omit checking of empty here because `batches_subject_.get_subscriber().on_next(batch);` will be not invoked.

irohad/multi_sig_transactions/impl/propagation_to_pcs.cpp
```diff
@@ -0,0 +1,69 @@
...
+using namespace iroha;
+
+MstToPcsPropagation::MstToPcsPropagation(
+    std::shared_ptr<iroha::network::PeerCommunicationService> pcs,
+    std::shared_ptr<StorageLimit> storage_limit,
+    rxcpp::observable<size_t> propagation_available,
+    logger::LoggerPtr log)
+    : log_(std::move(log)),
+      pcs_(pcs),
+      pending_batches_(storage_limit->create<InternalStorage>(
```
muratovv: I guess better to create storage in the outer scope of the ctor and pass it here.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -172,92 +178,102 @@ namespace iroha {
...
-            *rhs_batch);
+    auto completed_batches = batches_.move([this, &state_update, &rhs_batch](
+                                               auto &storage)
+                                               -> std::vector<BatchPtr> {
+      auto corresponding = storage.batches.right.find(rhs_batch);
+      if (corresponding == storage.batches.right.end()) {
+        // when state does not contain transaction
+        if (this->rawInsert(rhs_batch)) {
+          // there is enough room for the new batch
+          BOOST_VERIFY_MSG(state_update.updated_state_->rawInsert(rhs_batch),
```
muratovv: What is the point to check the insertion twice?

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -172,92 +178,102 @@ namespace iroha {
...
+                           "Could not insert new MST batch to state update.");
+        } else {
+          // there is not enough room for the new batch
+          log_->info("Dropped a batch because it did not fit into storage: {}",
+                     *rhs_batch);
+        }
+        return {};
```
muratovv: The method should return void as I see. What is the default value is assumed here?

irohad/multi_sig_transactions/storage/storage_limit.hpp
```diff
@@ -0,0 +1,219 @@
...
+#include <type_traits>
+
+#include <boost/core/noncopyable.hpp>
+#include <boost/optional.hpp>
+#include "interfaces/iroha_internal/transaction_batch.hpp"
+#include "multi_sig_transactions/mst_types.hpp"
+
+namespace iroha {
+
+  struct LimitableStorage {
```
muratovv: Add documentation for this and further classes/methods in this file.

irohad/multi_sig_transactions/storage/storage_limit.hpp
```diff
@@ -0,0 +1,219 @@
...
+    virtual ~LimitableStorage() = default;
+    virtual bool insert(BatchPtr batch) = 0;
+  };
+
+  template <class StorageImpl>
+  class LimitedStorage;
+
+  struct StorageLimit : public std::enable_shared_from_this<StorageLimit> {
```
muratovv: Add explanation of why shared_from_this is required here.

irohad/multi_sig_transactions/storage/storage_limit.hpp
```diff
@@ -0,0 +1,219 @@
...
+      return LimitedStorage<StorageImpl>(shared_from_this(),
+                                         std::move(storage));
+    }
+
+    virtual bool addIfAllowed(const BatchPtr &batch) = 0;
+
+    virtual void remove(const BatchPtr &batch) = 0;
+  };
+
+  struct StorageLimitDummy : public StorageLimit {
```
muratovv: Dummy sounds like the class is using in tests. Maybe better to write something like `UnlimitedStorage`?

muratovv: This class and `StorageLimitByTxs` classes should be on separate files. Also, I think better to separate classes for header and implementation files.

irohad/multi_sig_transactions/storage/storage_limit.hpp
muratovv: I think this class is required only methods operated on size. For example:
```
bool AddIfAllwed(size_t new_collection_size)
```
`StorageLimit` class by itself shouldn't know about the collection or collection items.

irohad/multi_sig_transactions/storage/storage_limit.hpp
muratovv: It looks like with this PR we want to solve two tasks:
1. Limit some storage with an upper bound size
2. Make "defer" remove from the collection.

The first task can be solved by `StorageLimit` strategy and this strategy should be passed into a collection as the dependency.

The second task should be solved by inheritance at the collection from an interface with such semantics:

```
template<typename Item>
class DeferRemove {
Proxy deferRemove(Key); // Key is some tag for removing objects from the collection here it is doesn't matter finalized interface. Only proxy is important here.
}

template<typename Item>
class Proxy {

Item release(); // method removes item from initial collection
private:
Iter<Item> obj; // there could be a pointer, iterator or identifier in the collection.
}
```
With this approach, all synchronization stuff will hold in the collection and will be consistent with the rest of the collection class and code will look cleaner because we separate two responsibilities - there is no intersection between add functionality in the collection and add which bad case of removing. Also, there is no need util classes such as `MstToPcsPropagation` because all stuff is doing in one subscribe on PCS.

muratovv (changes requested):
I have reviewed only the part of the PR. It seems we have to discuss the interface first and move further with the review.

irohad/multi_sig_transactions/state/impl/mst_state.cpp
MBoldyrev: it is a lambda that extracts batches to be moved from MstState's internal storage. The return type is any iterable, in this case it is a vector: https://github.com/hyperledger/iroha/blob/a3e6f62460abc4c086a94ca32c4c0c19ca7b98a8/irohad/multi_sig_transactions/state/impl/mst_state.cpp#L206

irohad/multi_sig_transactions/state/impl/mst_state.cpp
MBoldyrev: here we check the insertion to state update, which is expected to accept the batch unconditionally (it has no limits). and above we check the insertion to local storage, which is limited and may deny new batch, which is OK.

irohad/multi_sig_transactions/storage/storage_limit.hpp
MBoldyrev: it got removed

irohad/multi_sig_transactions/storage/storage_limit.hpp
MBoldyrev: renamed Dummy limit to None

MBoldyrev:
The PR got too big and could be split to two: this one brings a more general shared storage limit, and #2230 employs it to limit MST storages.

PR #2230, opened 2019-04-16
Feature/mst shared limit

MBoldyrev:
### Description of the Change

Reworked completed MST batches propagation to PCS:
- completed batches are sent to an instance of `MstToPcsPropagation`
- `MstToPcsPropagation` tries to pass them to PCS
- if PCS rejects them, `MstToPcsPropagation` waits till a notification with available room for transactions and resends the fitting batches

`MstToPcsPropagation` uses the same shared limit with the rest MST batches storages, which prevents new batches from being accepted until previously accepted either expire or come to PCS.

### Benefits

Solves the memory growth problem.

Please see `mst_to_psc_propagation_test`.

### DIFF
--- a/irohad/multi_sig_transactions/impl/propagation_to_pcs.cpp
+++ b/irohad/multi_sig_transactions/impl/propagation_to_pcs.cpp
@@ -1,0 +1,43 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "multi_sig_transactions/propagation_to_pcs.hpp"
+
+#include <utility>
+
+#include "interfaces/iroha_internal/transaction_batch.hpp"
+#include "logger/logger.hpp"
+
+using namespace iroha;
+
+MstToPcsPropagation::MstToPcsPropagation(
+    std::shared_ptr<iroha::network::PeerCommunicationService> pcs,
+    rxcpp::observable<size_t> propagation_available,
+    logger::LoggerPtr log)
+    : log_(std::move(log)),
+      pcs_(pcs),
+      propagation_available_subscription_(
+          propagation_available.subscribe([this, pcs](size_t available_txs) {
+            auto it = pending_batches_.begin();
+            while (it != pending_batches_.end()) {
+              const auto txs_in_batch = (*it)->transactions().size();
+              if (txs_in_batch <= available_txs and pcs->propagate_batch(*it)) {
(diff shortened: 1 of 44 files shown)


irohad/multi_sig_transactions/impl/propagation_to_pcs.cpp
```diff
@@ -0,0 +1,69 @@
...
+MstToPcsPropagation::~MstToPcsPropagation() {
+  propagation_available_subscription_.unsubscribe();
+}
+
+void MstToPcsPropagation::notifyCompletedBatch(
+    std::shared_ptr<MovedBatch> moved_batch) {
+  if (not pcs_->propagate_batch(moved_batch->get())) {
+    if (not pending_batches_.insert(std::move(moved_batch))) {
+      log_->critical(
+          "Dropped a completed MST batch because no place left in storage: {}",
```
igor-egorov, line 55: Usually, space is used in that context (even if place is more technically correct because it is internals that are not an object of interest for the external interface user).
```suggestion
          "Dropped a completed MST batch because no space left in storage: {}",
```

irohad/multi_sig_transactions/impl/propagation_to_pcs.cpp
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+size_t MstToPcsPropagation::pendingBatchesQuantity() const {
+  return pending_batches_.itemsQuantity();
+}
+
+bool MstToPcsPropagation::InternalStorage::insert(BatchPtr batch) {
+  pending_batches.emplace_back(std::move(batch));
+  return true;
```
igor-egorov, line 68: Unconditional return type 🤔

irohad/multi_sig_transactions/state/impl/mst_state.cpp
```diff
@@ -172,92 +187,110 @@ namespace iroha {
...
-        log_->info(
-            "Dropped a batch because it would exceed the transaction limit "
-            "(currently have {} out of {} transactions in state): {}",
-            transactionsQuantity(),
-            txs_limit_,
-            *rhs_batch);
+    auto completed_batches = batches_.move([this, &state_update, &rhs_batch](
+                                               auto &storage)
+                                               -> std::vector<BatchPtr> {
+      auto corresponding = storage.batches.right.find(rhs_batch);
```
igor-egorov, line 217: I would be very glad if you will use any other variable name instead of "corresponding". Hope this is possible.

igor-egorov (approved):
Thanks! Cool

